PR #168, opened 2026-03-02
feat(store): user-level aggregation layer (~/.edda/)

### LINKED ISSUE #151: feat(store): user-level aggregation layer (~/.edda/)
## Context

edda 目前是 repo-scoped（每個專案獨立 `.edda/`）。Chronicle 需要跨 repo 的用戶級聚合。

**決策背景：** `chronicle.subject=collaboration-relationship` — 主體是 user × tool 的協作關係，不是單一 repo。

## 目標結構

```
~/.edda/
├── collabs/
│   └── claude-code/
│       ├── chronicle.db      # 跨 repo 聚合數據
│       └── writings/         # 用戶文章（未來）
├── projects/
│   ├── <hash-a> → /path/to/project-a/.edda/
│   └── <hash-b> → /path/to/project-b/.edda/
└── config.json               # 用戶級設定
```

## 需要實作

1. **Project registry** — 在 `~/.edda/projects/` 登記所有已知專案
2. **Aggregation queries** — 跨 repo 查詢 sessions, decisions, commits
3. **Daily/weekly/monthly rollup** — 預計算的聚合統計
4. **edda-store 擴展** — user-level store API

## 注意

- `edda init` 時自動登記到 user-level registry
- 聚合是 lazy 的（查詢時才讀各 repo ledger），不是 eager sync
- 保持零 LLM、確定性

## Phase

Phase 1 — edda 數據基礎（chronicle 前提）

fagemx:
## Summary

Implements the user-level aggregation layer at `~/.edda/` for cross-repo queries and rollup statistics, providing the data foundation for Chronicle (#152).

**Architecture: Hybrid approach** — registry stays in `edda-store` (no new deps besides `time`), aggregation + rollup move to new `edda-aggregate` crate (avoids circular dependency with `edda-ledger`).

### What's included

- **`edda-store::registry`** — project registry at `~/.edda/registry.json` with register, unregister, list, validate, touch operations and file locking
- **`edda-store::user_config`** — user-level config at `~/.edda/config.json` with get/set/load/save
- **`edda-aggregate` crate** (new) — cross-repo aggregation queries (`aggregate_overview`, `aggregate_commits`, `aggregate_decisions`, `events_by_date`, `commits_by_date`) and rollup statistics (`compute_rollup`, `compute_rollup_incremental`, `merge_rollups`) with daily/weekly/monthly granularity
- **`edda user` CLI subcommand group** — `projects`, `overview`, `commits`, `decisions`, `rollup`, `config get/set/list` (all with `--json` support)
- **`edda init` integration** — auto-registers project in user-level registry
- **`edda gc --global` extension** — prunes stale registry entries
- **`edda-serve` dependency wiring** — ready for #152 API endpoints

### File changes

| File | Action | Description |
|------|--------|-------------|
| `crates/edda-store/src/lib.rs` | Edit | Add `pub mod registry; pub mod user_config;` |
| `crates/edda-store/Cargo.toml` | Edit | Add `time.workspace = true` |
| `crates/edda-store/src/registry.rs` | New | Project registry (~/.edda/registry.json) |
| `crates/edda-store/src/user_config.rs` | New | User-level config (~/.edda/config.json) |
| `crates/edda-aggregate/` | New crate | Cross-repo aggregation + rollup |
| `Cargo.toml` | Edit | Add `edda-aggregate` to workspace members |
| `crates/edda-cli/Cargo.toml` | Edit | Add `edda-aggregate` dependency |
| `crates/edda-cli/src/main.rs` | Edit | Add `mod cmd_user;` + `User` command variant |
| `crates/edda-cli/src/cmd_user.rs` | New | User-level CLI subcommands |
| `crates/edda-cli/src/cmd_init.rs` | Edit | Add `register_project()` call |
| `crates/edda-cli/src/cmd_gc.rs` | Edit | Add registry cleanup to `--global` |
| `crates/edda-serve/Cargo.toml` | Edit | Add `edda-aggregate`, `edda-store` deps |

### Design decisions

- **Lazy aggregation** — queries read each repo's ledger on demand, no persistent cross-repo DB
- **Incremental rollup** — `compute_rollup_incremental` merges new data into existing cache
- **No circular deps** — `edda-aggregate` depends on both `edda-store` and `edda-ledger`; `edda-ledger` depends on `edda-store` only
- **Best-effort registration** — `register_project()` failure in `edda init` prints warning, doesn't block init

## Test plan

- [x] `cargo check --workspace` passes (zero errors, zero warnings)
- [x] `cargo test --workspace` passes (965+ tests, 0 failures)
- [x] `cargo test -p edda-store` — 16 tests (registry: 4, user_config: 2, existing: 10)
- [x] `cargo test -p edda-aggregate` — 11 tests (aggregate: 6, rollup: 5)
- [ ] Manual: `edda init` registers project in `~/.edda/registry.json`
- [ ] Manual: `edda user projects` shows registered projects
- [ ] Manual: `edda user overview` shows cross-repo summary
- [ ] Manual: `edda gc --global` cleans stale entries

Closes #151

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -642,6 +642,7 @@ dependencies = [
  "clap",
  "crossterm",
  "ctrlc",
+ "edda-aggregate",
  "edda-ask",
  "edda-bridge-claude",
  "edda-bridge-openclaw",
@@ -668,6 +669,20 @@ dependencies = [
  "tokio",
  "tokio-util",
  "ulid",
+]
+
+[[package]]
+name = "edda-aggregate"
+version = "0.1.1"
+dependencies = [
+ "anyhow",
+ "edda-core",
+ "edda-ledger",
+ "edda-store",
+ "serde",
+ "serde_json",
+ "tempfile",
+ "time",
 ]
 
 [[package]]
@@ -859,10 +874,12 @@ version = "0.1.1"
 dependencies = [
  "anyhow",
  "axum",
+ "edda-aggregate",
  "edda-ask",
  "edda-core",
  "edda-derive",
  "edda-ledger",
+ "edda-store",
  "serde",
  "serde_json",
  "tempfile",
@@ -883,6 +900,7 @@ dependencies = [
  "serde_json",
  "tempfile",
  "thiserror",
+ "time",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ members = [
     "crates/edda-conductor",
     "crates/edda-serve",
     "crates/edda-notify",
+    "crates/edda-aggregate",
     "crates/edda-cli",  # crate name: "edda"
 ]
 
--- a/crates/edda-aggregate/Cargo.toml
+++ b/crates/edda-aggregate/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "edda-aggregate"
+description = "Cross-repo aggregation queries and rollup statistics for Edda"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+homepage.workspace = true
+categories.workspace = true
+keywords.workspace = true
+
+[dependencies]
+edda-core = { path = "../edda-core", version = "0.1.1" }
+edda-ledger = { path = "../edda-ledger", version = "0.1.1" }
+edda-store = { path = "../edda-store", version = "0.1.1" }
+anyhow.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+time.workspace = true
+
+[dev-dependencies]
+tempfile.workspace = true
--- a/crates/edda-aggregate/src/aggregate.rs
+++ b/crates/edda-aggregate/src/aggregate.rs
@@ -1,0 +1,368 @@
+//! Cross-repo aggregation queries.
+//!
+//! Lazy aggregation: reads each project's ledger on demand, no persistent DB.
+
+use edda_core::Event;
+use edda_ledger::Ledger;
+use edda_store::registry::ProjectEntry;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+use std::path::Path;
+
+/// Inclusive date range filter using ISO 8601 prefix comparison.
+#[derive(Debug, Clone, Default)]
+pub struct DateRange {
+    pub after: Option<String>,
+    pub before: Option<String>,
+}
+
+impl DateRange {
+    pub fn matches(&self, ts: &str) -> bool {
+        if let Some(ref after) = self.after {
+            if ts < after.as_str() {
+                return false;
+            }
+        }
+        if let Some(ref before) = self.before {
+            if ts > before.as_str() {
+                return false;
+            }
+        }
+        true
+    }
+}
+
+/// Per-project summary within an aggregation result.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ProjectSummary {
+    pub project_id: String,
+    pub name: String,
+    pub path: String,
+    pub event_count: usize,
+    pub commit_count: usize,
+    pub decision_count: usize,
+    pub session_count: usize,
+}
+
+/// A commit record from any project.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CommitRecord {
+    pub project_id: String,
+    pub project_name: String,
+    pub event_id: String,
+    pub ts: String,
+    pub title: String,
+    pub branch: String,
+}
+
+/// A decision record from any project.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DecisionRecord {
+    pub project_id: String,
+    pub project_name: String,
+    pub event_id: String,
+    pub key: String,
+    pub value: String,
+    pub reason: String,
+    pub domain: String,
+    pub branch: String,
+    pub ts: Option<String>,
+}
+
+/// Full cross-repo aggregation result.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AggregateResult {
+    pub projects: Vec<ProjectSummary>,
+    pub total_events: usize,
+    pub total_commits: usize,
+    pub total_decisions: usize,
+}
+
+/// Aggregate overview across all given projects.
+pub fn aggregate_overview(projects: &[ProjectEntry], range: &DateRange) -> AggregateResult {
+    let mut summaries = Vec::new();
+    let mut total_events = 0usize;
+    let mut total_commits = 0usize;
+    let mut total_decisions = 0usize;
+
+    for entry in projects {
+        let root = Path::new(&entry.path);
+        let ledger = match Ledger::open(root) {
+            Ok(l) => l,
+            Err(_) => continue,
+        };
+
+        let events = match ledger.iter_events() {
+            Ok(e) => e,
+            Err(_) => continue,
+        };
+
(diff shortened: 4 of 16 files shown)


fagemx:
## Code Review: PR #168

Thanks for the thorough implementation.

### Summary

Adds user-level aggregation layer with project registry, cross-repo queries, rollup statistics, and `edda user` CLI subcommand group. Well-structured crate split. A few issues to address before merge.

### Findings

#### Blockers

1. **O(N^2) registry reads in `execute_projects`** — `cmd_user.rs:91`: `validate_projects()` is called inside the `for p in &projects` loop, re-reading `~/.edda/registry.json` from disk once per project. With 50 registered projects this makes 50 redundant file reads. Call `validate_projects()` once before the loop and look up status from the result.

2. **Registry tests mutate real `~/.edda/registry.json`** — `registry.rs` tests `register_and_list_roundtrip`, `register_is_idempotent`, `unregister_removes_entry`, `validate_detects_stale` all write to the user's actual registry file. If tests run in parallel or the user has real data, this corrupts state. Tests should either override `store_root` or use a scoped approach that doesn't touch the global registry.

3. **User config tests mutate real `~/.edda/config.json`** — `user_config.rs:set_get_and_overwrite_roundtrip` writes to the actual config file. Same concern as above. The PID-keyed cleanup helps but doesn't protect against test failures leaving stale keys.

#### Suggestions

1. **`edda-serve` deps added but unused** — `edda-serve/Cargo.toml` adds `edda-aggregate` and `edda-store` dependencies but no source files in `edda-serve/src/` reference them. Consider deferring to #152 when the API endpoints are actually built, to keep the dependency graph minimal.

2. **`ensure_collabs_dir` creates `writings/` preemptively** — `rollup.rs:43` creates `collabs/<tool>/writings/` but nothing in this PR uses it. YAGNI — let the feature that needs it create it.

3. **`DayStat.sessions` always 0** — `rollup.rs:64` hardcodes `sessions: 0` with a comment "sessions don't have a natural daily granularity." If it's never populated, consider removing the field from `DayStat` to avoid confusion. Or populate it.

4. **`DateRange` uses lexicographic string comparison** — `aggregate.rs:20-32`: this works for ISO 8601 dates but could be surprising for callers passing timestamps with timezone offsets (e.g., `+08:00` vs `Z`). Worth a doc comment clarifying the assumption.

### Four-Point Check

| Check | Status | Notes |
|-------|--------|-------|
| Scope | ✅ | Matches #151 spec; `edda-serve` deps are premature but minor |
| Reality | ✅ | All referenced APIs (`store_root`, `write_atomic`, `lock_file`, `Ledger::open`, `active_decisions`, `iter_events`, `DecisionRow` fields) verified against codebase |
| Testing | ❌ | Tests write to real `~/.edda/` files — side-effecting on user's actual system |
| YAGNI | ✅ | Minor: `writings/` dir, `sessions` field, unused serve deps |

### Verdict

Changes Requested — fix the test isolation (blocker 2-3) and the O(N^2) validate loop (blocker 1).

---
*Reviewed by AI Tech Lead (e7h4n pattern)*

fagemx (commented):
## Code Review: feat(store): user-level aggregation layer (~/.edda/)

Overall this is a well-structured PR that cleanly separates the registry (in `edda-store`) from aggregation logic (in `edda-aggregate`), avoiding circular dependencies. The lazy aggregation approach is pragmatic for an MVP. Tests cover the core logic paths. However, there are several issues to address before merging.

---

### CI Failures (blocking)

**1. `cargo fmt` failures** — multiple formatting issues in `aggregate.rs` and `rollup.rs`:
- `events_by_date` and `commits_by_date` signatures should be single-line per `rustfmt`
- `compute_rollup_incremental` signature same issue
- `iso_week_start` has an unnecessary line break in the `let format = ...` binding
- Long `new_note_event` call in the test should be wrapped

**2. Clippy error** — `needless-question-mark` in `rollup.rs` line 239:
```rust
// Current (fails clippy):
Some(monday.format(&format).ok()?)
// Should be:
monday.format(&format).ok()
```

---

### Correctness Issues

**3. `execute_projects` calls `validate_projects()` inside a loop (O(n^2) I/O)**
In `cmd_user.rs` `execute_projects`, `validate_projects()` is called once per project inside the `for p in &projects` loop (~line 991). This re-reads the registry and checks every project's `.edda/` directory on every iteration. Should be called once before the loop and reused:
```rust
let projects = registry::list_projects();
let (valid, _) = registry::validate_projects(); // call once
// ...
for p in &projects {
    let status = if valid.iter().any(|v| v.project_id == p.project_id) { "ok" } else { "stale" };
    // ...
}
```

**4. `aggregate_overview` decision count ignores date range**
`decision_count` is computed via `ledger.active_decisions(None, None)` which returns all active decisions regardless of the `DateRange` filter. The event counts and commit counts are filtered by range, but decisions are not. This is semantically inconsistent — either document that decisions are always "current state" (not time-filtered), or filter by the decision's timestamp field.

**5. `compute_rollup_incremental` uses last daily date, not `last_updated`**
The incremental range uses `existing.daily.last().map(|d| d.date.clone())` as the `after` boundary. This means if a rollup was computed but had no daily entries (e.g., empty projects), the `after` will be empty and trigger a full re-scan anyway. Consider using `last_updated` (truncated to date) as the boundary instead, which is always set.

**6. `DateRange::matches` uses lexicographic string comparison**
The `after`/`before` fields are documented as "ISO 8601 prefix" but compared as raw strings. This works for ISO 8601 dates (e.g., `"2026-03-01" < "2026-03-01T00:00:00Z"`) but could produce surprising results if someone passes a non-date prefix. The comparison `ts > before` means `before` is **exclusive** for full timestamps but **inclusive** for dates matching the prefix. This semantic should be documented clearly, especially since the CLI arg names (`--after`/`--before`) suggest the filter is inclusive on both ends.

---

### Test Coverage Gaps

**7. `user_config` tests write to the real `~/.edda/config.json`**
The `set_get_and_overwrite_roundtrip` test writes to the actual user's global config file. While it uses a PID-based key name and cleans up, this is fragile — if the test panics, the key remains. Consider using a test-specific `store_root` override or `tempdir`-based approach. The `registry` tests have the same issue (they write to the real `~/.edda/registry.json`).

**8. No integration test for `edda init` → registry registration**
The `cmd_init.rs` change adds `register_project()` but there's no test verifying the integration. A test that runs `edda init` in a tempdir and checks the registry would catch regressions.

**9. No test for `aggregate_commits` or `aggregate_decisions`**
Only `aggregate_overview` has a test with a real ledger. The commit/decision aggregation functions have no real-ledger tests, only the `DateRange` unit tests.

---

### Minor / Style

**10. Phase 4d placement in `cmd_gc.rs` is after Phase 4c but before Phase 5 "Execute or dry-run"**
The registry cleanup runs unconditionally (not gated behind `total_items == 0` early return at line 293), which is correct. However, the `total_items` count on line 292 doesn't include stale registry entries, so the "Nothing to clean up" message could be misleading if only registry entries were cleaned but no file candidates existed. Consider either including registry cleanup in the count or printing the message before the registry phase.

**11. `DayStat.sessions` is always 0**
`build_daily_stats` sets `sessions: 0` with the comment "sessions don't have a natural daily granularity." If this field will never be populated at the daily level, consider removing it from `DayStat` or documenting the plan for it.

**12. `collabs` directory structure**
`ensure_collabs_dir` creates `collabs/<tool>/writings/` but this directory structure isn't documented or used anywhere in the PR. Presumably this is for Chronicle (#152), but it's dead code currently.

---

### Summary

The architecture is sound (hybrid store + aggregate split, lazy reads, incremental rollup). The main blockers are the CI failures (format + clippy) and the O(n^2) `validate_projects` loop. The date range semantics for decisions and the test isolation issues are worth fixing before this becomes load-bearing infrastructure.

Issue #157, opened 2026-03-02
feat(L3): post-mortem analysis → learned rules with TTL decay


fagemx:
## Context

L3 系統：從 session 異常中自動學習規則，用免疫系統模型管理規則生命週期。

**設計哲學：** 不告訴代理該做什麼，讓它做不了錯的事。規則不是提示（低遵從），是 hook（100% 遵從）。

### 1. Post-mortem 分析（LLM, session 後）

觸發條件（確定性，不需要 LLM）：
- Session 有失敗（PR blocked, test 壞）
- 異常長（> 20 turns）
- 同一檔案改 3+ 次
- 決策被推翻（supersede）
- 多代理衝突

分析流程：
- 收集 transcript delta + edda events + git diff
- LLM (Sonnet) 分析：提取教訓，判斷能否提煉為規則
- 輸出：Lesson（描述性）或 Rule proposal（處方性）

### 2. 規則生命週期（免疫系統模型）

```
誕生（pattern 重複 2x）→ 活躍 → 休眠（30天沒觸發）→ 沉降 → 消亡
```

三個衰減機制：
- **時間衰減**：TTL 30 天，每次觸發 reset
- **錨定衰減**：規則錨定到檔案路徑，檔案大改則 stale
- **矛盾檢測**：同 trigger 不同 action → 新 supersede 舊

活躍窗口上限 ~15 條。

規則存儲：
```
~/.edda/rules.json
欄位：id, trigger, action, anchor_file, anchor_hash,
      created, last_hit, hits, ttl_days, superseded_by, status
```

### 3. 規則執行（Hook, 確定性）

規則不注入 context（遵從性 50-70%），而是變成 hook：
- PreCommit hook 讀 rules store → 動態執行檢查
- 觸發時 block 或 warn（不是「建議」）
- 100% 遵從

## 產出物層級

| 產出物 | 機制 | 遵從性 |
|--------|------|--------|
| 規則 | Hook (block/auto-run) | 100% |
| 教訓 | CLAUDE.md 自動維護段落 | 90% |
| 觀察 | `edda ask` 按需查詢 | on-demand |

## Phase

Phase 3 — chronicle 成熟 / L3 系統

fagemx:
# Research Phase: Issue #157 — Post-mortem Analysis → Learned Rules with TTL Decay

## Executive Summary

[...]

PR #169, opened 2026-03-02
feat(L3): post-mortem analysis with learned rules and TTL decay

fagemx:
## Summary

Implements #157: L3 post-mortem analysis system with immune-system lifecycle for learned rules.

- **New crate `edda-postmortem`** (6 modules, 38 tests): deterministic trigger detection, heuristic analysis, rules store with TTL decay (time/anchor/contradiction), lessons store with CLAUDE.md sync
- **Bridge integration**: `run_postmortem()` wired into SessionEnd (best-effort); `evaluate_learned_rules()` wired into PreToolUse (warn-only via additionalContext)
- **CLI integration**: `edda rules list/show/decay/stats/gc` subcommands
- **Rate-limited decay**: only runs once per calendar day

## Architecture

Rule lifecycle: `Proposed → Active → Dormant → Settled → Dead` (+ `Superseded` for contradictions)

Three decay mechanisms:
1. **Time decay**: TTL (30 days default), reset on each trigger hit
2. **Anchor decay**: rule anchored to file hash; file changes → dormant
3. **Contradiction detection**: same trigger, different action → supersede old rule

Five trigger reasons: session failures, abnormally long sessions, excessive file edits, decision superseded, multi-agent conflict.

## Test plan

- [x] `cargo test -p edda-postmortem` — 38 tests pass (triggers, analyzer, rules lifecycle, lessons store, hooks evaluation)
- [x] `cargo test --workspace` — all 967 tests pass, 0 failures
- [x] Workspace compiles cleanly (`cargo check --workspace`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -653,6 +653,7 @@ dependencies = [
  "edda-mcp",
  "edda-notify",
  "edda-pack",
+ "edda-postmortem",
  "edda-search-fts",
  "edda-serve",
  "edda-store",
@@ -695,6 +696,7 @@ dependencies = [
  "edda-ledger",
  "edda-notify",
  "edda-pack",
+ "edda-postmortem",
  "edda-store",
  "edda-transcript",
  "globset",
@@ -835,6 +837,22 @@ dependencies = [
  "serde_json",
  "tempfile",
  "thiserror",
+]
+
+[[package]]
+name = "edda-postmortem"
+version = "0.1.1"
+dependencies = [
+ "anyhow",
+ "edda-core",
+ "edda-store",
+ "hex",
+ "serde",
+ "serde_json",
+ "sha2",
+ "tempfile",
+ "time",
+ "ulid",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ members = [
     "crates/edda-conductor",
     "crates/edda-serve",
     "crates/edda-notify",
+    "crates/edda-postmortem",
     "crates/edda-cli",  # crate name: "edda"
 ]
 
--- a/crates/edda-bridge-claude/Cargo.toml
+++ b/crates/edda-bridge-claude/Cargo.toml
@@ -17,6 +17,7 @@ edda-index = { path = "../edda-index", version = "0.1.1" }
 edda-pack = { path = "../edda-pack", version = "0.1.1" }
 edda-ledger = { path = "../edda-ledger", version = "0.1.1" }
 edda-notify = { path = "../edda-notify", version = "0.1.1" }
+edda-postmortem = { path = "../edda-postmortem", version = "0.1.1" }
 edda-derive = { path = "../edda-derive", version = "0.1.1" }
 anyhow.workspace = true
 thiserror.workspace = true
--- a/crates/edda-bridge-claude/src/digest.rs
+++ b/crates/edda-bridge-claude/src/digest.rs
@@ -4,7 +4,7 @@
 //! `edda_core::Event` milestone summarizing the session — without LLM,
 //! without touching the workspace ledger.
 
-use std::collections::BTreeSet;
+use std::collections::BTreeMap;
 use std::io::BufRead;
 use std::path::Path;
 
@@ -77,12 +77,29 @@ pub struct SessionStats {
     pub signal_count: u64,
     /// Dependency packages added during this session (for passive harvest).
     pub deps_added: Vec<String>,
+    /// Per-file edit counts: (file_path, count).
+    /// Used by L3 post-mortem trigger to detect file churn.
+    pub file_edit_counts: Vec<(String, u64)>,
+    /// Per-tool call counts (e.g. "Read" -> 15, "Edit" -> 8).
+    pub tool_call_breakdown: BTreeMap<String, u64>,
+    /// Model name used in this session.
+    pub model: String,
+    /// Total input tokens consumed.
+    pub input_tokens: u64,
+    /// Total output tokens consumed.
+    pub output_tokens: u64,
+    /// Total cache-read input tokens.
+    pub cache_read_tokens: u64,
+    /// Total cache-creation input tokens.
+    pub cache_creation_tokens: u64,
+    /// Estimated cost in USD.
+    pub estimated_cost_usd: f64,
 }
 
 /// Extract statistics from a session ledger file.
 pub fn extract_stats(session_ledger_path: &Path) -> anyhow::Result<SessionStats> {
     let mut stats = SessionStats::default();
-    let mut files_set: BTreeSet<String> = BTreeSet::new();
+    let mut file_edit_map: BTreeMap<String, u64> = BTreeMap::new();
 
     // Track session outcome: last event type + trailing failure count
     let mut last_event_name = String::new();
@@ -142,10 +159,13 @@ pub fn extract_stats(session_ledger_path: &Path) -> anyhow::Result<SessionStats>
                     })
                     .and_then(|v| v.as_str())
                     .unwrap_or("");
+                if !tool_name.is_empty() {
+                    *stats.tool_call_breakdown.entry(tool_name.to_string()).or_insert(0) += 1;
+                }
                 if tool_name == "Edit" || tool_name == "Write" {
                     if let Some(fp) = extract_file_path(&envelope) {
                         if !crate::signals::is_noise_file(&fp) {
-                            files_set.insert(fp);
+                            *file_edit_map.entry(fp).or_insert(0) += 1;
                         }
                     }
                 }
@@ -197,7 +217,8 @@ pub fn extract_stats(session_ledger_path: &Path) -> anyhow::Result<SessionStats>
         }
     }
 
-    stats.files_modified = files_set.into_iter().collect();
+    stats.file_edit_counts = file_edit_map.iter().map(|(k, v)| (k.clone(), *v)).collect();
+    stats.files_modified = file_edit_map.into_keys().collect();
     stats.duration_minutes = compute_duration_minutes(&stats.first_ts, &stats.last_ts);
 
     // Determine session outcome
@@ -254,6 +275,18 @@ pub fn render_digest_text(session_id: &str, stats: &SessionStats) -> String {
         stats.duration_minutes,
     ));
 
+    if !stats.tool_call_breakdown.is_empty() {
+        let breakdown: Vec<String> = stats
+            .tool_call_breakdown
+            .iter()
+            .map(|(name, count)| format!("{name}:{count}"))
+            .collect();
+        lines.push(format!("Tools: {}", breakdown.join(" ")));
+        let edit_ratio = compute_edit_ratio(&stats.tool_call_breakdown, stats.tool_calls);
+        let search_ratio = compute_search_ratio(&stats.tool_call_breakdown, stats.tool_calls);
+        lines.push(format!("edit_ratio: {edit_ratio:.2}, search_ratio: {search_ratio:.2}"));
+    }
+
     if !stats.files_modified.is_empty() {
         lines.push(format!(
             "Files modified: {}",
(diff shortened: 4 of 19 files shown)


fagemx (commented):
## PR Review: feat(L3): post-mortem analysis with learned rules and TTL decay

### Overall Assessment

This is a substantial, well-structured PR that implements the L3 post-mortem analysis system described in issue #157. The architecture follows the immune-system lifecycle model faithfully, and the new `edda-postmortem` crate has excellent test coverage (38 tests). The bridge and CLI integration points are clean and non-invasive.

However, **all CI checks are currently failing** -- the issues are minor (clippy/format/dead code) but must be fixed before merge. Beyond CI, I have several design-level observations.

---

### CI Failures (Must Fix)

1. **Clippy: "this `impl` can be derived"** -- In `edda-postmortem`, the `Default` impl for `RulesStore` can use `#[derive(Default)]` instead of a manual impl. The manual impl just does `Vec::new()` and `None` which is exactly what derive generates.

2. **Clippy: "this `if` statement can be collapsed"** -- In `rules.rs`, the nested `if` in anchor decay can be collapsed:
   ```rust
   // Current (two separate if-blocks with same body):
   if rule.status == RuleStatus::Active {
       rule.status = RuleStatus::Dormant;
   }
   ```
   The two `if` statements with the same body should be combined with the outer condition.

3. **Dead code: `total_tokens()` method on `UsageSnapshot`** -- The method `total_tokens()` is defined in `signals.rs` but only used in tests. Either add `#[allow(dead_code)]` or mark it `pub(crate)` / use it somewhere, or gate it with `#[cfg(test)]`.

4. **Format check failure** -- There is a formatting issue in `dispatch.rs` around the `evaluate_learned_rules` function. Run `cargo fmt` to fix.

---

### Design Observations

#### 1. Scope Creep: Usage Tracking (Medium Concern)

This PR bundles **two orthogonal features** into one:
- **L3 post-mortem analysis** (the issue's scope)
- **Session usage tracking** (model name, token counts, cost estimation, tool call breakdown, edit/search ratios)

The usage tracking (~400 lines across `signals.rs`, `digest.rs`, `cmd_log.rs`) is unrelated to post-mortem analysis. It would be cleaner as a separate PR. The `estimate_cost()` function with hardcoded Anthropic pricing, `EDDA_MODEL_PRICING` env var parsing, `shorten_model_name()`, `format_token_count()`, and the enriched `edda log --type session` formatting are all standalone features.

**Recommendation:** Consider splitting usage tracking into a separate PR. If keeping together, document why in the PR description.

#### 2. `propose_rule()` Borrow Pattern (Low Concern, Code Smell)

In `rules.rs`, `propose_rule()` iterates `&self.rules` (immutable borrow) then tries to do `self.rules.iter_mut().find(|r| r.id == rule_id)` (mutable borrow) inside the same loop. This works because the `return` exits before the second borrow overlaps, but it is fragile and confusing. The pattern of cloning `rule_id` to break the borrow is a code smell.

**Suggestion:** Restructure as a two-pass approach: first pass collects the matching ID (or contradiction list), then second pass mutates. This is what the contradiction path already does correctly.

#### 3. `matches_trigger()` for `command_failure:` (Bug?)

In `hooks.rs`, the `command_failure:` trigger matching logic seems wrong:

```rust
if let Some(cmd) = trigger.strip_prefix("command_failure:") {
    return ctx.tool_name == "Bash" && ctx.cwd.contains(cmd);
}
```

This checks if the **current working directory** contains the command name, which does not make semantic sense. If the trigger is `command_failure:npm`, it should match when a Bash tool call invokes `npm`, not when the cwd contains "npm".

**Recommendation:** This should probably match against the tool input/command string, not `ctx.cwd`. Since `HookContext` does not carry the command string for Bash calls, either:
- Add a `command` field to `HookContext`, or
- Change the trigger to use the file_path mechanism, or
- Document this as a known limitation

#### 4. `run_postmortem()` Re-extracts Stats (Efficiency)

In `dispatch.rs`, `run_postmortem()` calls `digest::extract_stats()` again on the session ledger. But `dispatch_session_end()` already runs `run_auto_digest()` which calls `extract_stats()` internally. The stats are computed twice.

**Suggestion:** Pass the already-computed `SessionStats` to `run_postmortem()` from the digest flow, or cache the result. The implementation plan comment #3 on the issue even identified this and chose "option (c) -- pass SessionStats directly" but the actual implementation re-reads from disk.

#### 5. CLAUDE.md Sync Race Condition (Low Concern)

`sync_to_claude_md()` uses `write_atomic()` but reads the file non-atomically. If another process (e.g., the user or another agent) modifies CLAUDE.md between the read and write, changes will be lost. This is acceptable for a best-effort feature but worth noting.

#### 6. `decisions_superseded` and `had_conflict` are Always 0/false

In `dispatch.rs`, `run_postmortem()`:
```rust
decisions_superseded: 0, // TODO: count from workspace ledger
had_conflict: false,     // TODO: detect from coordination state
```

These are two of the five trigger reasons from the spec. Without implementing them, `DecisionSuperseded` and `MultiAgentConflict` triggers will never fire. The implementation plan (issue comment #3) described reading from `coordination.jsonl`, but this was not done.

**Recommendation:** Either implement the coordination log reading or add a tracking issue. As-is, 40% of the trigger system is dead code.

#### 7. `SessionStats` Struct Growing Large

`SessionStats` now has 20+ fields. Several new fields (`model`, `input_tokens`, `output_tokens`, `cache_read_tokens`, `cache_creation_tokens`, `estimated_cost_usd`, `tool_call_breakdown`, `file_edit_counts`) were added. Consider grouping related fields into sub-structs (e.g., `UsageStats`, `EditStats`) to keep the struct manageable.

---

### Testing

The test coverage is solid:
- **38 tests** in `edda-postmortem` covering all modules
- Usage/cost tests in `signals.rs`
- Session digest formatting tests in `cmd_log.rs`
- Round-trip serialization tests for rules and lessons stores

Missing test coverage:
- No integration test for `run_postmortem()` end-to-end flow
- No test for `evaluate_learned_rules()` in the bridge dispatch path
- No test for the `edda rules` CLI subcommands (though these are leaf commands)

---

### Positive Highlights

- **Clean crate boundary**: `edda-postmortem` has minimal dependencies and clean public API
- **Best-effort design**: All post-mortem operations swallow errors gracefully -- session-end never fails due to L3 logic
- **Rate-limited decay**: Only runs once per calendar day, preventing unnecessary computation
- **Contradiction detection**: Same trigger + different action automatically supersedes the old rule -- elegant design
- **CLAUDE.md section management**: Uses HTML comment markers for clean section replacement, preserves surrounding content
- **Active window cap**: Limits to 15 active rules, preventing cognitive overload
- **All existing tests converted to `..Default::default()`** pattern -- good hygiene for the new fields

---

### Summary

| Aspect | Status |
|--------|--------|
| CI | Failing (3 minor issues) |
| Architecture | Clean, follows spec |
| Code quality | Good, with minor concerns |
| Test coverage | Strong for new crate, gaps in integration |
| Scope | Slightly broader than issue (usage tracking) |

**Verdict:** Fix the CI failures, address the `command_failure:` trigger matching bug, and consider splitting usage tracking into a separate PR. After those, this is ready to merge.

fagemx (commented):
## PR Review: feat(L3): post-mortem analysis with learned rules and TTL decay

### Overall Assessment

A well-designed feature implementing the immune-system lifecycle for learned rules. The `edda-postmortem` crate is cleanly bounded with minimal dependencies. The bridge integration is non-invasive and the best-effort error handling ensures session-end never fails due to L3 logic. Good test coverage (38 tests across the new crate).

The previous review covered the CI failures and the `matches_trigger` bug well, so I will focus on additional findings.

---

### Issues Found

#### 1. `propose_rule()` Iterates Immutable + Mutable Borrows Unsafely (Medium)

In `crates/edda-postmortem/src/rules.rs`, `propose_rule()` iterates `&self.rules` (lines ~305-320 in diff) then immediately does `self.rules.iter_mut().find(...)` while still inside the outer loop's scope. The `return` makes this compile, but the pattern is fragile:

```rust
for rule in &self.rules {
    if rule.trigger == trigger && rule.is_alive() {
        if rule.action == action {
            let rule_id = rule.id.clone();
            if let Some(existing) = self.rules.iter_mut().find(|r| r.id == rule_id) {
                existing.record_hit();
            }
            return rule_id;
```

This works because `return` exits before the outer borrow is reused, but any future refactor that removes the `return` will cause a borrow checker error or logic bug. **Suggestion**: Restructure as a two-phase approach -- collect the matching index/ID first, then mutate in a second pass.

#### 2. `evaluate_learned_rules()` Reads Rules Store From Disk on Every Tool Call (Performance)

In `crates/edda-bridge-claude/src/dispatch.rs`, `evaluate_learned_rules()` calls `RulesStore::load_project()` on every `PreToolUse` hook invocation. This means every single tool call (Read, Edit, Bash, Grep, etc.) triggers a full JSON file read and parse of `rules.json`. For sessions with hundreds of tool calls, this adds up.

**Suggestion**: Cache the rules store in memory per session (perhaps in the process state), or at minimum add a timestamp-based cache that reloads at most once per minute.

#### 3. `record_matched_hits()` is Never Called (Dead Code Path)

In `crates/edda-postmortem/src/hooks.rs`, the `record_matched_hits()` function (line ~1988 in diff) is defined as a public API to update hit counts when rules match during evaluation. However, `evaluate_learned_rules()` in `dispatch.rs` calls `evaluate_rules()` and `format_warnings()` but **never calls `record_matched_hits()`**. This means:
- Active rules that match during `PreToolUse` never get their hit counter incremented
- TTL is never reset by hook-time hits, only by proposal-time confirmations
- The decay mechanism will eventually kill rules that are actively triggering because they appear "unused"

This is a significant oversight in the rule lifecycle -- rules that are actively enforced will still decay to dead.

#### 4. Missing `failed_commands` Field Mapping in `run_postmortem()` (Data Loss)

In `dispatch.rs` `run_postmortem()`, the `SessionSummary` is built from `SessionStats`, but `failed_commands` is populated from `stats.failed_commands.clone()`. However, `SessionStats.failed_commands` is `Vec<String>` while `SessionSignals.failed_commands` is `Vec<FailedBashCmd>` (which has `cmd`, `stderr_snippet`, `count` fields). The `failed_commands` in `SessionStats` are the digest-extracted list -- they may not match the transcript-extracted `FailedBashCmd` entries used in the trigger evaluation. Worth verifying this field actually contains data when the trigger check runs.

#### 5. Cost Estimation Does Not Account for Cache Tokens

In `crates/edda-bridge-claude/src/signals.rs`, `estimate_cost()` calculates:
```rust
let input_cost = (usage.input_tokens as f64 / 1_000_000.0) * pricing.input_per_m;
let output_cost = (usage.output_tokens as f64 / 1_000_000.0) * pricing.output_per_m;
```

But `input_tokens` from the API includes the full prompt size. Anthropic charges differently for cache-read vs. cache-write vs. uncached tokens. The `cache_read_tokens` and `cache_creation_tokens` are tracked but not factored into cost. For Sonnet with heavy caching, this could overestimate cost by 2-3x since cache reads are charged at 10% of the input rate.

**Suggestion**: Either document that the cost is an upper-bound estimate, or factor in cache pricing:
- Cache read: 10% of input price
- Cache creation: 25% of input price
- Uncached input: full input price

#### 6. `format_token_count()` Has Redundant Branches

In `crates/edda-cli/src/cmd_log.rs`:
```rust
if k >= 100.0 {
    format!("{:.0}k", k)
} else if k >= 10.0 {
    format!("{:.0}k", k)
} else {
    format!("{:.1}k", k)
}
```

The first two branches (`k >= 100.0` and `k >= 10.0`) produce identical format strings. They can be collapsed into a single `if k >= 10.0 { ... }`.

#### 7. CLAUDE.md Sync Uses `!` as Markdown List Marker for High Severity

In `crates/edda-postmortem/src/lessons.rs`:
```rust
let severity_marker = match lesson.severity.as_str() {
    "high" => "!",
    "medium" => "*",
    _ => "-",
};
```

`!` is not a valid markdown list marker. It will render as plain text rather than a bullet point. `*` is a valid unordered list marker in markdown, but `!` is not -- it's the image syntax prefix. This means high-severity lessons will break the markdown formatting in CLAUDE.md.

**Suggestion**: Use `-` for all and add a severity prefix like `[HIGH]`, `[MEDIUM]` etc., or use standard list markers with bold/emoji for emphasis.

---

### Minor Observations

- The `is_session_digest()` helper function in `cmd_log.rs` (line ~975 in diff) is defined but could be inlined or turned into a method on `Event` for reusability
- The `Enforcement` enum in `hooks.rs` (with `Block` and `Warn` variants) is defined but never constructed -- all enforcement is warn-only via `format_warnings()`. The `Block` variant is dead code.
- The `format_token_count()` and `shorten_model_name()` tests are thorough and well-chosen

---

### Summary

| Aspect | Assessment |
|--------|-----------|
| Architecture | Clean crate boundary, follows spec well |
| CI | 3 failures (derivable impl, collapsible if, dead code) -- trivial fixes |
| Bug: `record_matched_hits` not called | Rules will decay despite active enforcement |
| Bug: `!` markdown marker | High-severity lessons will break CLAUDE.md formatting |
| Performance: rules loaded per tool call | Should be cached for hot path |
| Cost estimation accuracy | Overestimates due to ignoring cache pricing |
| Test coverage | Strong for new crate; gaps in integration paths |

**Verdict**: Fix the CI failures, the `record_matched_hits` omission (rules decaying despite being active is a semantic correctness issue), and the `!` markdown marker. The performance concern (loading rules per tool call) can be addressed in a follow-up.

---
*Reviewed by Claude Opus 4.6*

PR #269, opened 2026-03-12
feat(bridge): real-time coordination diff injection (#146)

### LINKED ISSUE #146: feat(bridge): real-time coordination diff injection via UserPromptSubmit
## Context

When Agent A claims a file or makes a decision, Agent B only sees it when its own PreToolUse fires — which may be too late (conflict already happened).

Currently UserPromptSubmit injects workspace ledger state but does NOT diff coordination.jsonl for new entries since last injection.

## Value

- Near-real-time cross-session awareness: "Agent A just claimed src/auth.rs — avoid editing this file"
- Reduce file edit conflicts in multi-agent scenarios
- Surface new decisions/bindings from peers immediately, not on next PreToolUse
- Minimal overhead: only inject when coordination.jsonl has new entries

## Implementation

1. Track last-seen coordination.jsonl offset per session (in phase file or heartbeat)
2. In \`dispatch_user_prompt_submit()\`, read new entries since last offset
3. Format relevant new entries (claims, decisions, requests) as compact additionalContext
4. Apply cooldown/dedup to avoid repetitive injection
5. Reset offset after injection

Example injection:
```
[coordination update]
- peer:cli claimed src/auth.rs (2 min ago)
- peer:cli decided "auth.strategy=JWT" (5 min ago)
```

## Complexity

Low (~40 lines production code). Leverages existing coordination.jsonl reader.

## Research Source

Inspired by Claude Code's internal command queue (cY) auto-wake pattern — events push to queue → immediately visible on next turn. This brings similar real-time awareness to edda's hook-based architecture.

fagemx:
## Summary
- Add per-session byte offset tracking for `coordination.jsonl` so `UserPromptSubmit` injects only **new** coordination events since last check
- `render_coord_diff()` reads tail bytes from offset, filters own/low-priority events, renders compact diff (`[coordination update]` section)
- Offset seeded at `SessionStart`, cleaned up at `SessionEnd`, with compaction guard and lazy seeding for robustness

## Changes
| File | What |
|------|------|
| `state.rs` | `read_coord_offset` / `write_coord_offset` + test |
| `peers/render_coord.rs` | `render_coord_diff()` with offset tracking, filtering, budget |
| `peers/mod.rs` | Re-export + `coordination_path` visibility bump |
| `dispatch/session.rs` | Integration in dispatch + SessionStart seed + SessionEnd cleanup |
| `peers/tests.rs` | 4 test cases (render, filter, compaction, lazy seed) |

## Design Decisions
- **Byte offset** (not timestamp) for O(1) tail read without full file re-parse
- **Compaction guard**: if file size < stored offset, reset to 0
- **Lazy seeding**: if no offset file exists (SessionStart not called), seed and skip to avoid replaying history
- **200-char budget**, max 5 events per diff injection
- **Self-event filter**: skip events from own session_id
- **Priority filter**: only Claim, Binding, Request (skip Unclaim, TaskCompleted, SubagentCompleted, RequestAck)

## Test plan
- [x] `coord_offset_round_trip` — state file read/write
- [x] `coord_diff_renders_new_events` — renders peer events, returns None on second call
- [x] `coord_diff_filters_own_events` — own session events excluded
- [x] `coord_diff_compaction_guard` — offset reset when file shrinks
- [x] `coord_diff_skips_when_no_offset_file` — lazy seeding behavior
- [x] `user_prompt_submit_dedup_skips_identical_state` — existing dedup test still passes
- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo fmt` passes

Closes #146

### DIFF
--- a/crates/edda-bridge-claude/src/dispatch/session.rs
+++ b/crates/edda-bridge-claude/src/dispatch/session.rs
@@ -163,6 +163,14 @@ pub(super) fn dispatch_with_workspace_only(
         }
     }
 
+    // Coordination diff: inject new events since last check (#146)
+    if let Some(diff) = crate::peers::render_coord_diff(project_id, session_id) {
+        ws = Some(match ws {
+            Some(w) => format!("{w}\n{diff}"),
+            None => diff,
+        });
+    }
+
     if let Some(ws) = ws {
         let wrapped = wrap_context_boundary(&ws);
         // Dedup: skip if identical to last injection
@@ -483,6 +491,8 @@ pub(super) fn cleanup_session_state(project_id: &str, session_id: &str, peers_ac
     let _ = fs::remove_file(state_dir.join(format!("signal_count.{session_id}")));
     // Late peer detection counter (#11)
     let _ = fs::remove_file(state_dir.join(format!("peer_count.{session_id}")));
+    // Coordination diff offset (#146)
+    let _ = fs::remove_file(state_dir.join(format!("coord_offset.{session_id}")));
     // Auto-claim state file (#24)
     crate::peers::remove_autoclaim_state(project_id, session_id);
     // Agent phase state file (#55)
@@ -618,6 +628,12 @@ pub(super) fn dispatch_session_start(
     // re-inject the full protocol on the first prompt after SessionStart (#11).
     let peers = crate::peers::discover_active_peers(project_id, session_id);
     write_peer_count(project_id, session_id, peers.len());
+
+    // Seed coordination offset so first UserPromptSubmit only diffs new events (#146).
+    let coord_path = crate::peers::coordination_path(project_id);
+    if let Ok(meta) = std::fs::metadata(&coord_path) {
+        crate::state::write_coord_offset(project_id, session_id, meta.len());
+    }
 
     // Remaining body sections (truncatable — nice-to-have context).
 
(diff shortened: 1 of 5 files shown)


fagemx (commented):
## Review: feat(bridge): real-time coordination diff injection (#146)

**Verdict: APPROVE**

### Scope
- 🟢 OK — 5 files, all within `edda-bridge-claude`, tightly scoped to #146. No unrelated changes.

### Reality
- 🟢 OK — Byte-offset approach is sound for O(1) tail reads. Compaction guard handles file truncation. Lazy seeding avoids replaying history when `SessionStart` was missed.
- 🟢 OK — No `.unwrap()` in production code. Error handling uses `.ok()?`, `.unwrap_or()`, `.unwrap_or_default()` throughout — consistent with SI-1 and SI-2.
- 🟢 OK — Self-event filtering and priority filtering logic is correct; the `match` arms are exhaustive against the filtered types, and the `_ => continue` catch-all in the render match is safe because the earlier filter already excluded the other variants.
- 🟢 OK — `coordination_path` visibility bump from `fn` to `pub(crate) fn` is minimal and appropriate for cross-module access within the crate.
- 🟢 OK — Offset update (`write_coord_offset`) happens after reading tail bytes, so no data loss on partial read failure (function returns `None` via `?` before reaching the write).

### Testing
- 🟢 OK — 5 tests covering: round-trip offset, new event rendering + idempotency, self-event filtering, compaction guard, and lazy seeding. Both happy and edge paths covered.

### YAGNI
- 🟢 OK — Constants (`COORD_DIFF_BUDGET=200`, `COORD_DIFF_MAX_EVENTS=5`) are reasonable and not over-configurable. No unnecessary abstractions.

### Minor observations (no action needed)
- The `let _ = fs::write(...)` pattern in `write_coord_offset` silently drops write errors, but this is consistent with the existing `write_peer_count` pattern in the same file — acceptable for non-critical state tracking.

Clean, focused PR. Good work.

PR #282, opened 2026-03-13
feat(api): add DecideSnapshot storage endpoint (#272)

fagemx:
## Summary

- Adds `decide_snapshot` event type with governance/milestone taxonomy classification
- Schema v8 migration creates `decide_snapshots` materialized view with indexes on `context_hash`, `village_id`, `engine_version`
- 3 HTTP endpoints: `POST /api/snapshot`, `GET /api/snapshots`, `GET /api/snapshots/:context_hash`
- Blob offload for payloads >8KB using `DecisionEvidence` classification
- Hash chain integrity maintained via existing `finalize()` + `append_event()` flow

### L1 (edda-core)
- `classify_event_type("decide_snapshot")` returns `(governance, milestone)`
- `new_snapshot_event()` builder function

### L2 (edda-ledger)
- `blob_put_if_large()` — conditional blob offload helper
- `SCHEMA_V8_SQL` — `decide_snapshots` table with FK to events
- `insert_snapshot()`, `query_snapshots()`, `snapshots_by_context_hash()` CRUD
- Schema migration v7->v8 with backfill from existing events
- Ledger facade pass-through methods

### L4 (edda-serve)
- `POST /api/snapshot` — accepts DecideSnapshot, offloads large payloads to blob store
- `GET /api/snapshots?village_id=&engine_version=&limit=` — filtered listing
- `GET /api/snapshots/:context_hash` — find all versions for same context (A/B comparison)
- `reconstruct_snapshot()` resolves inline/blob data for GET responses

## Test plan

- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo fmt` clean
- [x] Existing tests pass (3 pre-existing failures unrelated to this PR)
- [x] Schema migration test updated for v8
- [ ] Manual: POST snapshot with small payload (inline)
- [ ] Manual: POST snapshot with large payload (blob offload)
- [ ] Manual: GET snapshots with village_id filter
- [ ] Manual: GET snapshots/:context_hash returns multiple versions

### DIFF
--- a/crates/edda-core/src/event.rs
+++ b/crates/edda-core/src/event.rs
@@ -667,6 +667,38 @@ pub fn new_execution_event(
     Ok(event)
 }
 
+/// Create a new `decide_snapshot` event.
+///
+/// The payload contains compact metadata; large context/result data should
+/// be offloaded to blob store by the caller and referenced via `blob_refs`.
+pub fn new_snapshot_event(
+    branch: &str,
+    parent_hash: Option<&str>,
+    payload: serde_json::Value,
+    blob_refs: Vec<String>,
+) -> anyhow::Result<Event> {
+    let mut event = Event {
+        event_id: new_event_id(),
+        ts: now_rfc3339(),
+        event_type: "decide_snapshot".to_string(),
+        branch: branch.to_string(),
+        parent_hash: parent_hash.map(|s| s.to_string()),
+        hash: String::new(),
+        payload,
+        refs: Refs {
+            blobs: blob_refs,
+            ..Default::default()
+        },
+        schema_version: SCHEMA_VERSION,
+        digests: Vec::new(),
+        event_family: None,
+        event_level: None,
+    };
+
+    finalize(&mut event)?;
+    Ok(event)
+}
+
 /// Parameters for creating an `approval_policy_match` event.
 #[derive(Debug, Clone)]
 pub struct ApprovalPolicyMatchParams {
--- a/crates/edda-core/src/types.rs
+++ b/crates/edda-core/src/types.rs
@@ -72,6 +72,7 @@ pub fn classify_event_type(event_type: &str) -> (Option<&'static str>, Option<&'
             Some(event_level::GOVERNANCE),
         ),
         "device_pair" | "device_revoke" => (Some(event_family::ADMIN), Some(event_level::INFO)),
+        "decide_snapshot" => (Some(event_family::GOVERNANCE), Some(event_level::MILESTONE)),
         _ => (None, None),
     }
 }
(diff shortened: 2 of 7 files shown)


fagemx (commented):
## Review: feat(api): add DecideSnapshot storage endpoint (#272)

### Scope — OK
Changes are well-scoped to L1 (event type + classification), L2 (schema + CRUD), and L4 (HTTP endpoints). Layer boundaries respected. 527 additions for a new materialized view + 3 endpoints is reasonable.

### Reality — 2 Issues

**[MUST FIX] Silent blob error swallowing (SI-2 violation)**
`crates/edda-serve/src/lib.rs` — `post_snapshot` handler:
```rust
let context_blob = edda_ledger::blob_put_if_large(...)
    .ok()
    .flatten();
```
`.ok()` silently discards blob write errors. If the blob store is full or the write fails, the snapshot will appear to succeed but lose context/result data — a silent data loss. This must propagate the error with `?` or `.context("blob offload for context failed")?`. Same applies to `result_blob`.

**[SHOULD FIX] Dynamic SQL construction in `query_snapshots`**
`crates/edda-ledger/src/sqlite_store.rs` — The `query_snapshots` method builds SQL via `format!()`:
```rust
let sql = if conditions.is_empty() {
    format!("{base} ORDER BY created_at DESC LIMIT ?")
} else {
    format!("{base} WHERE {} ORDER BY created_at DESC LIMIT ?", conditions.join(" AND "))
};
```
While the `WHERE` clause values use `?` placeholders (SI-5 satisfied), the column names in conditions are hardcoded string literals (`"village_id = ?"`, `"engine_version = ?"`), not user input, so this is not an injection risk. However, the pattern deviates from the project convention of static SQL strings. Consider splitting into 4 pre-written queries (no filter / village only / engine only / both) for clarity and consistency with the rest of `sqlite_store.rs`.

### Testing — Insufficient

- Schema migration test updated (v7->v8 assertion) — good.
- Classification test added for `decide_snapshot` — good.
- **No integration tests for the 3 new HTTP endpoints.** Every other endpoint in `edda-serve` has at least one `#[tokio::test]` exercising the happy path. This PR adds `POST /api/snapshot`, `GET /api/snapshots`, and `GET /api/snapshots/:context_hash` with zero test coverage. At minimum, add:
  1. POST a snapshot, assert 201 + event_id returned
  2. GET /api/snapshots returns the posted snapshot
  3. GET /api/snapshots/:context_hash returns correct results
  4. GET /api/snapshots/:context_hash with unknown hash returns 404

### YAGNI — Minor

- `redaction_level` and `cycle_id` fields appear unused by any consumer in this PR or elsewhere in the codebase. Not a blocker, but worth noting — if these are speculative, consider deferring until needed.

### Summary

The architecture and layering are solid. The two blocking items are:
1. **Fix blob error swallowing** — change `.ok().flatten()` to propagate errors
2. **Add HTTP endpoint tests** — maintain the project's test coverage standard

Grade: **C+** — good structure, needs the above fixes before merge.

PR #410, opened 2026-07-15
fix(search): CJK bigram tokenizer — CJK queries >2 chars now match (GH-402)

### LINKED ISSUE #402: search: CJK queries longer than 2 chars silently return zero — no CJK tokenization (P0)
## Problem

Full-text search silently returns zero results for any CJK query longer than ~2 characters, even when the exact string sits verbatim in the ledger. For a corpus that is majority Chinese, `edda search` currently *lies*: "No results found" reads as "this was never discussed", so agents doing recall will confidently duplicate prior work or contradict prior rulings.

Empirical evidence (2026-07-15, fleet workspace, 2147 events + 1232 turns freshly indexed):

| query | expected | actual |
|---|---|---|
| `watermark` (EN) | hits | 5 hits ✓ |
| `task rail` (EN, two terms) | hits | 20 hits ✓ |
| `收據` (2 chars) | hits | 20 hits ✓ |
| `驗收` (2 chars) | hits | 20 hits ✓ |
| `權威事實` (4 chars) | hits — string exists verbatim in a decision reason | **0** |
| `機器推論` (4 chars) | hits — same document | **0** |
| `洗成權威事實` (6 chars) | hits — same document | **0** |

## Root cause

The text fields use tantivy's default/simple tokenization with no CJK handling: a contiguous CJK run between punctuation marks is emitted as **one giant token**. Two-character hits are luck — those words happen to appear punctuation-delimited (「收據」, (收據)) somewhere in the corpus. Any word that only occurs inside a longer run (e.g. 權威事實 inside 把機器推論洗成權威事實) is unreachable. (`crates/edda-search-fts/src/schema.rs` — also note the `raw` tokenizer on at least one field.)

## Fix

- Register a CJK-capable tokenizer (bigram is the cheap robust choice; lindera if we want dictionary segmentation) and apply it to every indexed text field **at index time and query time symmetrically** — the asymmetry is the killer, not the choice of segmenter.
- Mixed-script text (CJK + ASCII identifiers like `ENV_LOCK`, `task_nudge.rs`) must keep matching on the ASCII terms.
- Bump an index schema version so `edda search index` knows a full rebuild is required, and say so in the CLI output instead of silently mixing old/new tokenization.

## Acceptance

- All seven queries in the table above return the expected hits on the same corpus
- A phrase that only exists inside a longer CJK run is findable
- English/mixed-script queries keep working
- House TDD: failing tokenizer/query tests first

fagemx:
Closes #402. A P0: `edda search` silently returned **zero** for CJK queries longer than ~2 characters, because tantivy's default tokenizer emits a contiguous CJK run as a single token — so any phrase that only appears *inside* a longer run was unreachable. On a majority-Chinese corpus, "No results" read as "never discussed," and recall failed.

## The fix: a CJK bigram tokenizer

- **`crates/edda-search-fts/src/tokenizer.rs`** (new): `CjkBigramTokenizer` — overlapping 2-grams for CJK runs (`把機器` → `把機`, `機器`), lowercased ASCII/Latin words, punctuation as separators. Registered on every index (`register_tokenizers`), so index-time and query-time (`QueryParser::for_index`) tokenize **symmetrically** — the asymmetry was the killer, not the segmenter choice. `權威事實` → `[權威, 威事, 事實]`, all present in a doc containing `…洗成權威事實`.
- **Precision**: a pure-CJK query defaults to **AND** over its bigrams, so a phrase requires all of them and isn't outranked by title-boosted partial matches. English keeps OR.
- **Fuzzy**: skipped for pure-CJK (Levenshtein-1 on 2-char bigrams is noise; bigrams already give exact recall); kept when the query has ASCII to correct.
- **Schema version → v2**: `edda search index` detects an outdated index and rebuilds tantivy + the turns watermark from scratch (announced); `edda search query` / `edda ask` refuse a stale index and hint the rebuild — old and new tokenizations never mix. Rebuild is crash-safe (marker written only after commit; a fresh/empty index always clears the watermark) and holds an index-keyed lock so concurrent indexers can't race.

## Acceptance — live-verified on the fleet corpus (2211 events + 1232 turns)

| query | before | after |
|---|---|---|
| `權威事實` (4-char, inside a run) | **0** | 1 ✓ |
| `機器推論` (4-char, inside a run) | **0** | 1 ✓ |
| `洗成權威事實` (6-char, inside a run) | **0** | 1 ✓ |
| `收據` (2-char) | 20 | 20 ✓ |
| `watermark` / `task rail` (English) | 5 / 20 | 5 / 20 ✓ |

Each previously-broken query now finds the exact decision that contains it verbatim (the match is even snippet-highlighted). English and 2-char CJK are unchanged.

## Review

House TDD (failing tokenizer/search tests first), then **four rounds of fresh-context adversarial review** (Codex), which hardened the rebuild path against crash-mid-wipe, concurrent-indexer races, corrupt-index-wipe-on-read, and CJK ranking precision. Final verdict: merge-ready, zero findings.

## Documented trade-offs of bigram indexing (not silent)

- Multi-char CJK **regex** (`/外部.*事實/`) can't match a 2-gram term dictionary; English regex is unaffected. CLI help qualified.
- **Single-char** CJK queries can't match multi-char runs (inherent to bigrams; not a regression — the default tokenizer failed this too).
- CJK **alternatives** (軟體 vs 軟件) need an explicit `OR` since pure-CJK defaults to AND.
- Mixed `postgre 中文` also fuzzes the CJK bigrams (tantivy fuzzy is per-field, not per-term).

## Existing installs

Run `edda search index` once after upgrading — it auto-detects the v1 index and rebuilds.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1046,6 +1046,7 @@ dependencies = [
  "edda-core",
  "edda-index",
  "edda-store",
+ "fs2",
  "rusqlite",
  "serde",
  "serde_json",
--- a/crates/edda-cli/src/cmd_ask.rs
+++ b/crates/edda-cli/src/cmd_ask.rs
@@ -60,6 +60,12 @@ fn build_transcript_callback(repo_root: &Path) -> Option<Box<TranscriptSearchFn>
     if !index_dir.exists() {
         return None;
     }
+    // GH-402: don't search a stale-schema index — its CJK results would be
+    // silently wrong. Skip transcript search and hint the rebuild instead.
+    if edda_search_fts::schema::index_is_outdated(&index_dir) {
+        eprintln!("(search index is out of date; run `edda search index` to include transcripts)");
+        return None;
+    }
 
     let index = edda_search_fts::schema::ensure_index(&index_dir).ok()?;
     let pid = project_id.clone();
--- a/crates/edda-cli/src/cmd_ask.rs
+++ b/crates/edda-cli/src/cmd_ask.rs
@@ -67,7 +67,7 @@ fn build_transcript_callback(repo_root: &Path) -> Option<Box<TranscriptSearchFn>
         return None;
     }
 
-    let index = edda_search_fts::schema::ensure_index(&index_dir).ok()?;
+    let index = edda_search_fts::schema::open_index(&index_dir)?;
     let pid = project_id.clone();
 
     Some(Box::new(move |query: &str, limit: usize| {
--- a/crates/edda-cli/src/cmd_search.rs
+++ b/crates/edda-cli/src/cmd_search.rs
@@ -20,7 +20,8 @@ pub enum SearchCmd {
     },
     /// Search for events and transcript turns
     Query {
-        /// Search query (supports fuzzy, "exact", /regex/)
+        /// Search query (fuzzy for ASCII; "exact"; /regex/ over indexed terms —
+        /// note: regex matches tokenized terms, so CJK regex only spans 2 chars)
         query: String,
         /// Project ID (defaults to current repo)
         #[arg(long)]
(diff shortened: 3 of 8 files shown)


fagemx:
## Code Review: PR #410 (Round 1)

### Summary

The CJK tokenizer, symmetric registration, version/rebuild orchestration, and query-side AND/fuzzy logic are correct and well-tested — four prior fresh-context rounds hardened correctness. This pass is code quality + testing conventions. One cross-crate coupling issue (P1); no correctness or test problems.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

1. **Watermark-clear SQL duplicates schema knowledge across crates** (`crates/edda-cli/src/cmd_search.rs:210`). The rebuild path hardcodes `DELETE FROM turns_meta; DELETE FROM index_watermark;`, but those table names are defined in a *different* crate (`edda-search-fts::schema::ensure_meta_db`). If the meta-DB schema is renamed or restructured, cmd_search breaks at **runtime**, not compile time — the two crates silently disagree. Extract a `schema::clear_index_watermark(&conn)` helper next to `ensure_meta_db` so the table names live in one place.

### Testing Review

#### Coverage
- Tokenizer (bigrams, mixed script, ASCII identifiers, offsets), version/fresh-flag transitions (incl. crash-mid-wipe), pure-CJK AND precision, and the acceptance table are all covered with real (RAM) indexes.

#### Convention Compliance
No violations — `#[test]`, real Tantivy/SQLite, no internal mocking, `cargo test` green.

Fixing the P1 and will re-review.

---

fagemx:
## Code Review: PR #410 (Round 2) — LGTM

The round-1 P1 is resolved. `clear_index_watermark` now lives in `edda-search-fts::schema` (beside `ensure_meta_db`), and `cmd_search` calls it — verified the inline `DELETE FROM turns_meta; DELETE FROM index_watermark;` no longer appears anywhere outside that one helper. The meta-DB table names are in a single crate again.

### Final state
- **Correctness** (from 4 prior fresh-context rounds + codex): CJK bigram tokenizer, symmetric index/query registration, v1->v2 forced rebuild, crash-safe rebuild (`created_fresh` + `IndexLock` + read-only `open_index`), pure-CJK AND / fuzzy guard — all sound.
- **Maintainability**: no cross-crate schema coupling, no duplication, doc comments consistent.
- **Testing**: 36 tests green in `edda-search-fts` (tokenizer, version/fresh-flag transitions, CJK AND precision, acceptance table); CLI builds clean.

No P0/P1 remaining. Approving.

---
*Automated review-fix loop converged in 2 rounds.*

PR #411, opened 2026-07-15
ci: cut Windows test time (debug-info trim + platform-sensitive crate scope)

fagemx:
Windows Test was 7-8 min vs ~1.5 min on Linux/macOS with the same `rust-cache`. The cached deps aren't the cost — it's recompiling and **linking 23 crates' test binaries with full debug info** on the slow MSVC linker.

Two changes:

**1. Trim debug info in CI** — `CARGO_PROFILE_{DEV,TEST}_DEBUG=line-tables-only`. Keeps readable backtraces (function name + line) but skips the heavy PDB generation/linking that dominates Windows build time (~30-50% link win). CI-only env; local dev unaffected.

**2. Narrow the Windows matrix to platform-divergent crates.** Windows is already `continue-on-error` (non-blocking), so instead of all 23 crates it now tests only the ones with genuinely OS-specific behaviour:
- `edda-store` — `%APPDATA%` vs `~/.edda` path resolution
- `edda-ledger` — file locks, paths, bundled SQLite
- `edda-search-fts` — tantivy mmap + fs2 index lock
- `edda-bridge-claude` — process spawn, hooks, heartbeat files, paths
- `edda` (CLI) — integration, spawn, paths

Linux and macOS keep full `--workspace` coverage, so pure-logic crates (edda-core, etc.) are still exercised every run — just not recompiled on the slow Windows runner.

Combined, this should bring the Windows job well under the current 7-8 min. If Windows coverage ever needs to be broader, add crates to the matrix list.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,10 @@ on:
 env:
   CARGO_TERM_COLOR: always
   RUSTFLAGS: -Dwarnings
+  # Trim debug info in CI: keeps readable backtraces (function name + line) but
+  # skips the heavy PDB generation and linking that dominate Windows build time.
+  CARGO_PROFILE_DEV_DEBUG: line-tables-only
+  CARGO_PROFILE_TEST_DEBUG: line-tables-only
 
 jobs:
   fmt:
@@ -39,9 +43,24 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
+          - os: ubuntu-latest
+            args: --workspace
+          - os: macos-latest
+            args: --workspace
+          # Windows is non-blocking (continue-on-error) and its build/link is
+          # ~5x slower, so it runs only the platform-divergent crates
+          # (filesystem paths, file locks, process spawn, mmap) instead of all
+          # 23. Linux/macOS keep full-workspace coverage.
+          - os: windows-latest
+            args: >-
+              -p edda-store
+              -p edda-ledger
+              -p edda-search-fts
+              -p edda-bridge-claude
+              -p edda
     steps:
       - uses: actions/checkout@v4
       - run: rustup toolchain install stable --profile minimal
       - uses: Swatinem/rust-cache@v2
-      - run: cargo test --workspace
+      - run: cargo test ${{ matrix.args }}


fagemx:
## Code Review: PR #411 (retroactive — this PR was merged without one)

Reviewed after the fact. #411 was merged on 2026-07-15 alongside #410 and is the **only** merged PR in this run that never went through the review loop — 17 others did. Filing what the loop would have caught.

It cannot be "changes requested" now; findings go to an issue instead.

### Summary

The debug-info trim (`CARGO_PROFILE_DEV_DEBUG: line-tables-only`) is well-judged: backtraces keep function name and line, and the heavy PDB link that dominates Windows time goes away. No objection.

The matrix change is where the problem is.

### Key Findings

#### High Priority (P1)

**P1-1 — the Windows crate list contradicts the criterion written directly above it.**

```yaml
# Windows is non-blocking (continue-on-error) and its build/link is
# ~5x slower, so it runs only the platform-divergent crates
# (filesystem paths, file locks, process spawn, mmap) instead of all 23.
- os: windows-latest
  args: -p edda-store -p edda-ledger -p edda-search-fts -p edda-bridge-claude -p edda
```

The stated criterion includes **process spawn**. `edda-conductor` is the most spawn-heavy crate in the workspace — `Command::new("edda")` for `ensure_init`, the Claude launcher, tmux control, shell checks — and it is **not in the list**. The rule names the thing; the list omits the crate that does it most.

Measured on #432's run, rather than inferred from the args:

| job | `test result:` lines |
|---|---|
| Test (ubuntu-latest) | **45** |
| Test (windows-latest) | **9** |

Windows runs ~20% of the suite. `edda-conductor`, `edda-ask`, `edda-pack`, `edda-core`, `edda-index` are among the 18 crates it never touches.

**This already cost something concrete.** GH-417 was a bug that manifested on the operator's own Windows machine — `cargo test` filed 13 dead temp dirs into the real registry, every run. The fix landed in #432; its regression guard, `ensure_init_sends_its_child_to_a_throwaway_store`, lives in `edda-conductor` and has **never executed on Windows CI**. A Windows-specific regression in that guard would ship green.

#### Medium (P2)

**P2-1 — Windows cannot block a merge, on the operator's own platform.**

`continue-on-error: ${{ matrix.os == 'windows-latest' }}` (lines 28 and 42) pre-dates this PR, so it is not #411's doing — but #411 compounds it. Windows now covers 5 of 23 crates *and* is advisory. edda's operator runs Windows; that is the platform where a regression is felt first and blocks least.

### Unverified, and worth saying so

I merged 15 PRs today on the strength of `gh pr checks` showing zero non-passing lines. Whether a **failing** `continue-on-error` job would still report as `pass` there, I could not establish — no Windows job has failed in the last 25 runs, so there is no case to inspect. If it does mask, my verification method has a hole in it and every "all seven green" this session was weaker than I stated. Flagging it as unknown rather than assuming either way.

### Testing Review

No tests, correctly — this is CI configuration. It has been exercised empirically: several dozen green runs today, all of which built on it.

#### Testing Verdict: N/A

### Verdict: Findings filed as an issue (PR already merged)

The trim is good. The matrix needs `edda-conductor` at minimum, since the comment already argues for it.

---
*Retroactive review — #411 was merged without the review loop. That was my omission, not a process exception.*

PR #412, opened 2026-07-15
search: auto-index — turns read early-out + incremental events + SessionEnd hook (GH-403)

### LINKED ISSUE #403: search: index is never built or updated automatically — flagship retrieval ships cold (P0)
## Problem

The Tantivy index is only ever built by a manual `edda search index`. Nothing in the hook lifecycle builds or updates it. Consequence observed 2026-07-15: the fleet's daily-driver workspace — months of transcripts, 2147 events — had **never been indexed at all**. `edda search` answered "No search index found" since forever, and nobody noticed because the SessionStart pack covered daily recall. The flagship retrieval feature shipped cold in its own home.

Even after a manual build the index starts rotting immediately: every new event/turn is invisible until someone remembers to re-run the command. A search surface that silently omits recent history is worse than none — recency is what recall queries usually want.

## Fix

- **Incremental reindex on SessionEnd** (hook already dispatches there; auto-digest shows the pattern) — index only events/turns past a stored watermark (`events_after_rowid` exists for exactly this cursor shape).
- **Build-if-missing**: `edda search query` against a missing index should trigger (or offer) a first build rather than a dead-end error.
- **Staleness honesty**: query output states the index watermark ("indexed through 2026-07-15T09:40, N newer events not yet indexed") so silence is never mistaken for absence.
- Budget guard per the house rule: incremental step must be cheap (the full build was ~20s for 3.4k docs; incremental should be sub-second) and must never block a session's exit — background/best-effort like auto-digest.

## Acceptance

- Fresh workspace: first `search query` works without a manual index step
- After a session writes events, a new session's `search query` finds them without manual reindex
- Query output shows the indexed-through watermark
- Kill the process mid-reindex: next run recovers (idempotent by watermark), no corrupt index

Depends on / relates to the CJK tokenizer issue: whichever lands second must handle the schema-version rebuild path.

fagemx:
Closes #403.

## The headline

A no-op reindex on the real fleet corpus (2261 events / 1329 turns / 100 sessions):

| | before | after |
|---|---|---|
| `edda search index`, nothing new | **24.7s** | **0.139s** |

~178x. That is what makes a SessionEnd hook shippable at all — the issue's own rule is that it must never block a session's exit, and the 45s background-join budget could not absorb 25s.

## The issue's premise was wrong, and so was mine

#403 assumed the cost was event re-indexing and that "turns are already incremental". I built on that assumption and it was **falsified by measurement**:

| run | result | time |
|---|---|---|
| old code, no-op | 2227 events + 0 turns | 24.7s |
| **events made fully incremental**, no-op | **0** events + 0 turns | **24.8s** |
| 1 session vs 100 sessions | 0 events + 0 turns | **0.111s vs 24.848s** |

Making events incremental bought **nothing**. The cost was `index_session`, linear in session count: it read every index record, built a uuid map, walked each assistant's parent chain, and fetched + JSON-parsed that turn's transcript line off disk — all *before* the per-turn `turns_meta` check that decides the turn is already indexed. 100 sessions / 24MB read, parsed, and discarded on every run.

Turns were incremental in what they **write**, never in what they **read**. `index_watermark.last_offset` — the byte cursor built for exactly this — was never consulted in `indexer.rs` at all. Now it is: a session whose index file has not grown is skipped without being opened.

The spec ([SEARCH_AUTO_INDEX_V1.md](docs/plan/search-auto-index/SEARCH_AUTO_INDEX_V1.md)) marks the wrong diagnosis in place rather than quietly rewriting it, so the error stays legible.

## What landed

- **`edda-search-fts::sync`** — one home for lock/open/cursor/commit, shared by all three callers (manual `search index`, query-time cold build, SessionEnd), so the dance is not re-implemented per call site. Events arrive via an injected closure, keeping the crate unaware of `edda-ledger`.
- **Turns read early-out** — the actual fix. Skip sessions whose index file has not grown.
- **Events incremental** — `events_watermark` rowid cursor over the existing `Ledger::events_after_rowid`; `index_events_since` replaces per event (`delete_term` on doc_id) instead of delete-all-then-re-add. Kept because it is correct and crash-safe on the rebuild path, but it is not what buys the budget.
- **Build-if-missing** — a missing or schema-outdated index announces itself and builds, instead of the dead-end error the issue is about. (This supersedes #410's refuse-on-outdated: both states are an unusable index one command away, and fixing beats complaining.)
- **Staleness honesty** — every query reports the timestamp it is indexed through and the count of newer events it has not seen.
- **SessionEnd `bg_index`** — gated spawn + mpsc join, following the existing `bg_*` pattern. No cooldown (a no-op sync is a cursor read), no cold build (a session's exit must never pay ~25s).
- **No new Tantivy field** — `INDEX_VERSION` stays at 2, so nobody eats a second forced rebuild days after #410 forced one.

## Crash safety

`writer.commit()` happens **before** `write_events_cursor`. A crash between them re-runs the batch, which the per-event `delete_term` makes idempotent — never duplicates, never misses.

## Adversarial review (codex) found two real bugs, both fixed with failing tests first

1. **A rebuild left orphans searchable.** The old bulk `index_events` deleted every event doc before re-adding — that was not only dedup, it was *also* purging events the ledger no longer had. Going incremental dropped that safety net: truncate or replace a ledger and the old docs answered queries forever. `delete_all_event_docs` now runs on the rebuild path only.
2. **The cursor was validated on rowid alone, which is not identity.** A ledger restored from a different backup can occupy the same rowid range with different events; the probe would trust the cursor and skip every event in the replacement forever. It now matches the boundary event's timestamp too, using the `last_ts` already stored (no schema change).

## Live verification (installed binary, real corpus)

- Budget: **24.7s → 0.139s**, stable across runs
- Cold build-if-missing: announces, builds 2261 events + 1329 turns, answers — 25.3s one-time
- Staleness: `(indexed through …; 1 newer event(s) not yet indexed)` after writing a note
- SessionEnd: dispatched a real `SessionEnd` hook → watermark advanced, 0 newer, note findable with **no manual index**
- #410 CJK regression intact: `權威事實`/`機器推論`/`洗成權威事實` → 2/2/2, `provenance` → 20

## Known limitations, filed separately, not folded in here

- `index_session` commits `turns_meta` (and now the session offset) *before* the caller's Tantivy commit. A crash there marks turns indexed that Tantivy never received. **Pre-existing** — the per-turn `turns_meta` dedup already had this hole — and the events cursor deliberately does not share it.
- Rewritten turns are not reconciled (`turns_meta` dedups by `turn_id` and has no update path). Also pre-existing.
- The early-out uses file length as the change signal, so a same-length in-place rewrite of an append-only index file would be missed.

Tests: 51 in `edda-search-fts`, workspace 1931 passed / 0 failed, fmt + clippy clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -773,6 +773,7 @@ dependencies = [
  "edda-notify",
  "edda-pack",
  "edda-postmortem",
+ "edda-search-fts",
  "edda-store",
  "edda-transcript",
  "globset",
--- a/crates/edda-bridge-claude/Cargo.toml
+++ b/crates/edda-bridge-claude/Cargo.toml
@@ -17,6 +17,7 @@ edda-transcript = { path = "../edda-transcript", version = "0.2.0" }
 edda-index = { path = "../edda-index", version = "0.2.0" }
 edda-pack = { path = "../edda-pack", version = "0.2.0" }
 edda-ledger = { path = "../edda-ledger", version = "0.2.0" }
+edda-search-fts = { path = "../edda-search-fts", version = "0.2.0" }
 edda-notify = { path = "../edda-notify", version = "0.2.0" }
 edda-postmortem = { path = "../edda-postmortem", version = "0.2.0" }
 edda-derive = { path = "../edda-derive", version = "0.2.0" }
--- a/crates/edda-bridge-claude/src/bg_index.rs
+++ b/crates/edda-bridge-claude/src/bg_index.rs
@@ -1,0 +1,69 @@
+//! Incremental search reindex at SessionEnd (GH-403).
+//!
+//! Design: non-blocking, idempotent, cheap.
+//!
+//! Two deliberate departures from the sibling background tasks:
+//!
+//! - **No cooldown or interval gate.** `bg_scan`/`bg_detect` are gated because
+//!   they spend LLM calls; an incremental sync with nothing new is a cursor read
+//!   and a no-op commit. Gating it would only reintroduce the staleness this
+//!   exists to remove.
+//! - **No cold build.** A missing index is left alone and picked up by
+//!   `edda search query`'s build-if-missing. A first build costs ~25s on a real
+//!   corpus, and a session's exit must never pay that.
+
+use anyhow::Result;
+use edda_store::project_dir;
+use std::path::Path;
+
+/// Run only when there is already an index to top up.
+pub fn should_run(project_id: &str) -> bool {
+    if std::env::var("EDDA_BG_ENABLED").unwrap_or_else(|_| "1".into()) == "0" {
+        return false;
+    }
+    index_exists(&project_dir(project_id))
+}
+
+/// Whether a project has an index worth topping up.
+///
+/// Split out from `should_run` so it can be tested against a temp dir: resolving
+/// a project dir means reading the process-wide store root, and a test that
+/// redirected it would corrupt every other test sharing this process.
+fn index_exists(proj_dir: &Path) -> bool {
+    proj_dir.join("search").join("tantivy").exists()
+}
+
+/// Bring the index up to date with events written during this session.
+pub fn run_index(project_id: &str, cwd: &str) -> Result<()> {
+    let ledger = edda_ledger::Ledger::open(std::path::Path::new(cwd))?;
+    let proj = project_dir(project_id);
+    let stats = edda_search_fts::sync::sync(&proj, project_id, None, |after| {
+        ledger.events_after_rowid(after)
+    })?;
+    tracing::debug!(
+        events = stats.events,
+        turns = stats.turns,
+        "search index synced"
+    );
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // Deliberately tests `index_exists` rather than `should_run`: the latter
+    // resolves the store root from the environment, and setting that here would
+    // redirect it for every other test in this process (cargo runs them in
+    // parallel threads), breaking bg_detect/bg_scan/bg_digest.
+    #[test]
+    fn index_exists_only_when_the_tantivy_dir_is_present() {
+        let tmp = tempfile::tempdir().unwrap();
+
+        // Cold builds belong to `edda search query`, never to a session's exit.
+        assert!(!index_exists(tmp.path()));
+
+        std::fs::create_dir_all(tmp.path().join("search").join("tantivy")).unwrap();
+        assert!(index_exists(tmp.path()));
+    }
+}
--- a/crates/edda-bridge-claude/src/dispatch/session.rs
+++ b/crates/edda-bridge-claude/src/dispatch/session.rs
@@ -357,6 +357,21 @@ pub(super) fn dispatch_session_end(
         bg_count += 1;
     }
 
+    // 2j. Background incremental search reindex (GH-403). Ungated by cooldown:
+    // with nothing new this is a cursor read and a no-op commit.
+    if crate::bg_index::should_run(project_id) {
+        let tx = bg_tx.clone();
+        let pid = project_id.to_string();
+        let cwd_owned = cwd.to_string();
+        std::thread::spawn(move || {
+            if let Err(e) = crate::bg_index::run_index(&pid, &cwd_owned) {
+                tracing::warn!(error = %e, "search reindex failed");
+            }
+            let _ = tx.send("bg_index");
+        });
+        bg_count += 1;
+    }
+
     // Drop the original sender so the channel closes when all threads finish.
     drop(bg_tx);
 
--- a/crates/edda-bridge-claude/src/lib.rs
+++ b/crates/edda-bridge-claude/src/lib.rs
@@ -2,6 +2,7 @@ pub mod agent_phase;
 pub mod bg_detect;
 pub mod bg_digest;
 pub mod bg_extract;
+pub mod bg_index;
 pub mod bg_scan;
 pub mod controls_suggest;
 pub mod digest;
--- a/crates/edda-cli/src/cmd_search.rs
+++ b/crates/edda-cli/src/cmd_search.rs
@@ -195,6 +195,14 @@ pub fn query(
 /// Report how current the index is (GH-403), so silence is never mistaken for
 /// absence. Best-effort: a broken watermark must not fail a query that already
 /// produced results.
+///
+/// `indexed through` is read from the project's own meta.sqlite and is therefore
+/// correct for any project. The "newer events" count is not: it can only be
+/// computed against a ledger, and the only ledger we have is `repo_root`'s. For
+/// a `--project` that is not this repo, that count would compare an unrelated
+/// project's cursor against these events — a fabricated number. Since the point
+/// of this line is honesty about staleness, a confidently wrong count is worse
+/// than none, so it is omitted rather than guessed.
 fn print_watermark(repo_root: &Path, proj_dir: &Path, project_id: &str) {
     let meta_path = proj_dir.join("search").join("meta.sqlite");
     let Ok(conn) = schema::ensure_meta_db(&meta_path) else {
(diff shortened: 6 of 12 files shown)


fagemx:
## Code Review: PR #412 (Round 1)

### Summary

Correctness was already pressed hard by codex (two real bugs found and fixed with failing tests first). This round is the maintainability/DRY/convention pass that a correctness reviewer skips. The implementation is clean — no stringly-typed state, no unwrap in lib code, no unsafe, error handling is `anyhow` + `?` throughout, and the crash-safety ordering is well commented. Two P1s.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — `index_watermark` SQL is split across two modules, reintroducing the exact coupling #410 fixed.**

`crates/edda-search-fts/src/indexer.rs:350,366` (`read_session_offset` / `write_session_offset`) hand-write `index_watermark` column names, while that table's DDL (`META_DDL`), its `clear_index_watermark`, and the sibling `events_watermark` accessors all live in `schema.rs`. Rename `last_offset` and `indexer.rs` breaks at **runtime, not compile time**.

This is the same finding that was raised against #410, whose fix commit says: *"so the table names live in one crate"*. The events cursor followed that rule (`schema::read_events_cursor`); the session offset didn't. The two watermarks in the same database should not live in different modules.

**P1-2 — `query`'s build-if-missing can index the wrong project's events.**

`crates/edda-cli/src/cmd_search.rs:121-124`:

```rust
let ledger = Ledger::open(repo_root)?;               // the repo we are standing in
let stats = sync::sync(&proj_dir, project_id, ...)   // the project that was asked for
```

`run_cmd` resolves `pid` from `--project` independently of `repo_root`, so these can diverge. `edda search query "x" --project OTHER` against a missing index silently builds **OTHER's index out of the current repo's ledger** — cross-project contamination, written without warning.

`index()` has the same shape and predates this PR, but `query` previously never opened a ledger at all, so this PR is what makes the flaw reachable from the read path — the command users actually run.

Suggested minimal fix: only auto-build when the requested project is the one this repo resolves to; otherwise keep the explicit "run `edda search index --project X`" message, which is honest for a project that isn't here.

### Testing Review

#### Coverage

| Change | Tests |
|---|---|
| `events_watermark` cursor + `META_DDL` extraction | 3 — roundtrip/default, clear-also-clears, both builders agree |
| `index_events_since` | 2 — idempotent re-run, appends without touching neighbours |
| `sync` | 8 — incl. the RED test, crash-recovery, cursor-ahead, wiped-index-surviving-meta |
| Turns read early-out | 3 — skip-without-reading, offset recorded, grown file reprocessed |
| Rebuild purge + cursor identity (codex findings) | 2 — both written failing first |
| `bg_index` | 1 |

Adequate. The RED test is the executable form of the 24.7s measurement, which is the right thing to pin.

#### Convention Compliance

No violations. Real SQLite throughout (`ensure_meta_db_memory`), no internal mocking, plain `#[test]`, `cargo test` clean. The `FakeLedger` test double injects through the crate's existing closure seam rather than mocking internals — correct.

Two P2s, not worth a round:
- `mk_event` (sync.rs:162) and `mk_test_event` (indexer.rs:803) are near-identical `Event` constructors in one crate. Normal Rust per-module test-helper duplication at 2 uses.
- `bg_index::run_index` is untested while siblings carry 11–32 tests. It is 12 lines of delegation to `sync` (8 tests), and testing it needs the process-wide store root that `index_exists` was deliberately split out to avoid touching. `cmd_search.rs` has 0 tests as its pre-existing baseline (as does `cmd_ask.rs`), so the untested query branch matches that file's convention.

Fixing both P1s and will re-review.

---

fagemx:
## Code Review: PR #412 (Round 2)

### Summary

Both round-1 P1s verified fixed. Round 1's P1-2 fix exposed a sibling in the same class that round 1 missed — the read path had two ledger opens, and only one of them got guarded.

### Round 1 verification

| Finding | Status |
|---|---|
| P1-1 `index_watermark` SQL split across modules | **Fixed.** Production code in `indexer.rs` no longer contains any `index_watermark` SQL; `read/write_session_offset` now live in `schema.rs` beside the DDL and the event cursor. |
| P1-2 build-if-missing indexes the wrong project | **Fixed.** `cmd_search.rs:120` guards on `project_id == resolve_project_id(repo_root)` and returns before `Ledger::open`. |

The `turns_meta` SQL still in `indexer.rs` is pre-existing (`main` has 10 such mentions) and out of scope here.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-3 — `print_watermark` reports a fabricated "newer events" count for a non-local `--project`.**

`crates/edda-cli/src/cmd_search.rs:198-217`:

```rust
let Ok(cursor) = schema::read_events_cursor(&conn, project_id) else { ... };  // THAT project's cursor
let newer = Ledger::open(repo_root)                                           // THIS repo's ledger
    .and_then(|l| l.events_after_rowid(cursor.rowid))
```

The cursor comes from `project_id`'s meta.sqlite; the count comes from `repo_root`'s ledger. For `edda search query "x" --project OTHER`, `newer` is the number of local events after an unrelated project's rowid — a meaningless number, printed as fact:

```
(indexed through 2026-07-15T...; 47 newer event(s) not yet indexed)
```

Round 1's P1-2 guarded the *write* path (build-if-missing) against exactly this repo_root/project_id divergence, but `print_watermark` runs on **every** query, guarded by nothing. It does not corrupt data, but this PR's stated purpose is staleness honesty — "silence is never mistaken for absence" — and a confidently wrong count is a worse failure of that goal than printing nothing.

`indexed through <ts>` is correct for any project (it is read from that project's own meta.sqlite). Only the `newer` count depends on the ledger, so only that part needs suppressing when the project is not local.

### Testing Review

#### Coverage

No new source behaviour added this round beyond the two fixes; existing coverage stands (51 in `edda-search-fts`, 189 in `edda`). `cmd_search.rs` carries 0 tests as its pre-existing baseline, so the guard is verified by inspection and the live run rather than a unit test — consistent with that file's convention, and noted as a P2 rather than a gate.

#### Convention Compliance

No violations. Clippy clean at `-D warnings` (the move made `OptionalExtension` unused in `indexer.rs`; import removed).

Fixing P1-3 and will re-review.

---

fagemx:
## Code Review: PR #412 (Round 3) — LGTM :tada:

### Summary

Three P1s found and fixed across two rounds, all in the same family: **things that diverge when `--project` names a project that isn't the repo you're standing in**, plus one module-boundary regression.

| Round | Finding | Fix |
|---|---|---|
| 1 | `index_watermark` SQL hand-written in `indexer.rs` while its DDL and sibling cursor lived in `schema.rs` — the coupling #410 fixed, reintroduced | `read/write_session_offset` moved beside the DDL |
| 1 | `query`'s build-if-missing could build project OTHER's index from **this** repo's ledger | auto-build limited to the locally-resolved project |
| 2 | `print_watermark` printed a **fabricated** "newer events" count for a non-local `--project` | count omitted when non-local; `indexed through` kept (it is read from that project's own meta.sqlite, so it stays correct) |

Round 1's fix is what exposed round 2's finding: guarding one of the two ledger opens on the read path made the unguarded one obvious. That is the loop earning its keep — a single pass would have shipped the fabricated count.

### Verification

Both branches of the final guard exercised live, not just inspected (`cmd_search.rs` carries no tests, by its pre-existing convention):

```
local project:      (indexed through 2026-07-15T15:14:14Z; 7 newer event(s) not yet indexed)
non-local project:  (indexed through 2026-07-15T15:14:14Z)
```

Before the fix, the second line carried a count computed from the wrong repo's ledger.

Tests: 51 `edda-search-fts` + 189 `edda`, workspace 1931 passed / 0 failed. Clippy clean at `-D warnings`, fmt clean.

### Deliberately not fixed here (filed instead)

- **`index()` has the same cross-project flaw** on the write path (`edda search index --project OTHER` indexes this repo's events into OTHER). It predates this PR — `query` never opened a ledger before, which is why the read path was in scope and this isn't. Worth its own change, and the registry already stores a path per project, so the flag could be made to genuinely work rather than merely stop lying.
- **`turns_meta` commits before the Tantivy commit** (crash there marks turns indexed that Tantivy never got). Pre-existing; the event cursor deliberately does not share it.
- **`turns_meta` SQL still in `indexer.rs`** — same class as P1-1 but pre-existing (`main` has 10 such mentions), so out of scope.

P2s not worth a round: duplicate `Event` test constructors (`mk_event`/`mk_test_event`), `bg_index::run_index` untested (12 lines of delegation to a `sync` that has 8 tests; testing it needs the process-wide store root that `index_exists` was split out to avoid).

No critical or high-priority issues remaining. Ready for merge — operator's gate.

---
*Completed after 3 round(s) of automated review-fix loop*

PR #416, opened 2026-07-15
search: defer turns_meta writes until after the Tantivy commit (GH-413)

### LINKED ISSUE #413: search: turns marked indexed before Tantivy commit — crash makes them permanently invisible (P1)
## Problem

`indexer::index_session` marks turns as indexed in SQLite **before** the Tantivy writer commits. A crash in that window makes those turns permanently invisible to search, with no self-healing path.

The order today (`crates/edda-search-fts/src/indexer.rs`):

```
turns_meta rows inserted → tx.commit()     ← SQLite now says "these turns are indexed"
                                           ← ★ crash here
writer.commit()                            ← Tantivy only now receives the documents
```

`writer.commit()` is the caller's job (`sync::sync`, `crates/edda-search-fts/src/sync.rs`). Because the dedup check is `SELECT 1 FROM turns_meta WHERE turn_id = ?1`, every later run **skips** those turns. They never enter the index and nothing ever notices.

This predates GH-403 — the per-turn `turns_meta` dedup has always had it — but GH-403 made it load-bearing by putting an incremental sync on the SessionEnd hot path, and its own acceptance criterion says:

> Kill the process mid-reindex: next run recovers (idempotent by watermark), no corrupt index

The turns path does not satisfy that. The events path deliberately does: `sync` commits Tantivy **first**, then advances `events_watermark`, so a crash re-runs the batch and `index_events_since` makes the re-add idempotent via `delete_term` on `doc_id`. That ordering is the reference pattern.

GH-403 also added the session offset (`index_watermark.last_offset`) into the same pre-commit transaction, so a crash now skips the whole file rather than just the turn — same failure mode, wider blast radius.

## Fix

Write the SQLite side only after the Tantivy commit succeeds. Likely shape: `index_session`/`index_project` buffer their `turns_meta` rows and session offsets and return them, so `sync` writes them after `writer.commit()`. Watch the borrow — `index_session` takes `&Connection` and uses `unchecked_transaction()`.

## Acceptance

- A crash between the SQLite commit and the Tantivy commit leaves the turn re-indexable: the next sync indexes it and `edda search` finds it.
- Test modelled on `sync::tests::crash_between_commit_and_cursor_write_recovers_without_duplicates` (`crates/edda-search-fts/src/sync.rs`), but for turns.
- No duplicate documents on the re-run.

Found during #412 review; deliberately not folded into that PR.

fagemx:
Closes #413.

## Problem

`index_session` recorded turns in SQLite and committed that transaction **before** its caller committed the Tantivy writer:

```
turns_meta rows → tx.commit()     ← SQLite: "these turns are indexed"
                                  ← ★ crash here
writer.commit()                   ← Tantivy only now receives the documents
```

Because the dedup is `SELECT 1 FROM turns_meta WHERE turn_id = ?1`, every later run **skipped** those turns. Permanently invisible to search, no self-healing path — and #403's own acceptance criterion says a kill mid-reindex must recover.

This predates #403, but #403 made it load-bearing (SessionEnd now syncs on every session exit) and widened the blast radius by putting the session offset in the same pre-commit transaction: a crash then skipped the whole file, not just the turn.

## Fix

`index_session`/`index_project` return a `PendingMeta` holding the `turns_meta` rows and session offsets. They touch SQLite only for **reads** (the early-out offset, the per-turn dedup). `sync` flushes it **after** `writer.commit()` succeeds.

This is the ordering the events cursor already used — commit, then advance `events_watermark`. The turns path now shares it.

## Adversarial review found two bugs in the fix itself — both fixed, tests first

The deferral moved the crash window rather than removing it, and the first version handled the new window badly:

1. **Duplicate turn documents.** Events survive that window because `index_events_since` deletes the `doc_id` before re-adding. Turns were still doing a plain `add_document`, so the recovery run added a *second copy of every turn*, permanently, with no cleanup path in normal syncing. Turns now `delete_term` first — the property that makes the ordering safe in the first place. (`crash_between_commit_and_flush_does_not_duplicate_turns`, failed with `left: 2, right: 1`.)

2. **A duplicate `turn_id` within one run aborted the entire flush.** The old dedup read from its own open transaction, so same-run duplicates collapsed for free. Deferring means the check only sees *flushed* rows, so both records reach the flush — where a plain `INSERT` raised `PRIMARY KEY constraint failed` and rolled back the whole transaction, recording nothing and failing identically on every retry. **Worse than the window being closed.** `INSERT OR IGNORE` restores the old semantics, and `flush` now reports rows actually inserted rather than attempted. (`duplicate_turn_id_within_one_run_does_not_poison_the_flush`, failed with `Error code 1555`.)

## Tests

`crash_before_tantivy_commit_leaves_turns_reindexable` is the executable form of the issue: it drops the writer without committing — the only way the crash actually happens — and asserted **0** re-indexed turns before the fix.

54 tests in `edda-search-fts` (was 52), workspace 1934 passed / 0 failed, clippy clean at `-D warnings`, fmt clean.

## Known limitation (not fixed here)

`PendingMeta` buffers a full rebuild's `turns_meta` rows in memory before flushing — O(total project turns). At the real corpus's ~1300 turns this is trivial, but it is unbounded in principle. Chunking would mean multiple commits, which is a different design; flagged rather than folded in.

## Not yet verified live

The deferral touches the hot path, so the #403 budget (no-op sync 0.157s, down from 24.7s) must still hold. Release build + live check on the real corpus pending — numbers will be posted here before this is ready to merge.

### DIFF
--- a/crates/edda-search-fts/src/indexer.rs
+++ b/crates/edda-search-fts/src/indexer.rs
@@ -946,6 +946,47 @@ mod tests {
         assert_eq!(reader.searcher().num_docs(), 1, "and to one document");
     }
 
+    /// `index_project` is the production path — `bg_index` passes no session id,
+    /// so every SessionEnd goes through here rather than `index_session`. It is
+    /// also the only caller of `PendingMeta::merge`, so without this the merge is
+    /// never executed by any test.
+    #[test]
+    fn index_project_merges_turns_and_offsets_across_sessions() {
+        let tmp = tempfile::tempdir().unwrap();
+        let project_dir = tmp.path();
+        write_session_fixture(project_dir, "s1", "1");
+        write_session_fixture(project_dir, "s2", "2");
+
+        let index = ensure_index_ram().unwrap();
+        let schema = index.schema();
+        let mut writer = index_writer(&index).unwrap();
+        let meta_conn = ensure_meta_db_memory().unwrap();
+
+        let pending = index_project(&writer, &schema, &meta_conn, project_dir, "p1").unwrap();
+        assert_eq!(
+            pending.turns(),
+            2,
+            "both sessions' turns must survive merge"
+        );
+
+        writer.commit().unwrap();
+        assert_eq!(pending.flush(&meta_conn).unwrap(), 2);
+
+        // Every session's offset must survive too. Losing them here would not
+        // fail anything loudly — it would just re-read every session on every
+        // sync forever, which is exactly the GH-403 cost this is built to avoid.
+        for sid in ["s1", "s2"] {
+            assert!(
+                crate::schema::read_session_offset(&meta_conn, "p1", sid).unwrap() > 0,
+                "offset for {sid} must survive merge"
+            );
+        }
+
+        // With the offsets recorded, a second pass does nothing.
+        let again = index_project(&writer, &schema, &meta_conn, project_dir, "p1").unwrap();
+        assert_eq!(again.turns(), 0, "unchanged sessions must all skip");
+    }
+
     #[test]
     fn indexing_a_session_records_its_consumed_offset() {
         let tmp = tempfile::tempdir().unwrap();
--- a/crates/edda-search-fts/src/indexer.rs
+++ b/crates/edda-search-fts/src/indexer.rs
@@ -125,13 +125,94 @@ fn extract_event_tags(event: &edda_core::Event) -> String {
         .unwrap_or_default()
 }
 
+/// SQLite writes held back until the Tantivy commit succeeds (GH-413).
+///
+/// `turns_meta` is the dedup source: once a turn is in it, every later run skips
+/// that turn. Writing it before the documents are committed means a crash in
+/// between marks turns as indexed that Tantivy never received — invisible to
+/// search forever, with no self-healing path. So nothing in here touches SQLite
+/// until the caller has committed the writer.
+///
+/// This is the ordering the events cursor already uses: `sync::sync` commits the
+/// writer, *then* advances `events_watermark`.
+#[derive(Debug, Default)]
+pub struct PendingMeta {
+    turns: Vec<TurnMetaRow>,
+    /// `(project_id, session_id, consumed byte length)`
+    offsets: Vec<(String, String, i64)>,
+}
+
+#[derive(Debug)]
+struct TurnMetaRow {
+    turn_id: String,
+    project_id: String,
+    session_id: String,
+    ts: String,
+    user_uuid: String,
+    assistant_uuid: String,
+    user_store_offset: i64,
+    user_store_len: i64,
+    assistant_store_offset: i64,
+    assistant_store_len: i64,
+}
+
+impl PendingMeta {
+    /// How many turns will be recorded once flushed.
+    pub fn turns(&self) -> usize {
+        self.turns.len()
+    }
+
+    /// Absorb another batch (used by `index_project` across sessions).
+    pub fn merge(&mut self, other: PendingMeta) {
+        self.turns.extend(other.turns);
+        self.offsets.extend(other.offsets);
+    }
+
+    /// Write everything in one transaction, returning the number of turns
+    /// recorded.
+    ///
+    /// Call this ONLY after the Tantivy writer has committed. Flushing early
+    /// reintroduces GH-413.
+    pub fn flush(self, conn: &Connection) -> anyhow::Result<usize> {
+        let tx = conn.unchecked_transaction()?;
+        for r in &self.turns {
+            tx.execute(
+                "INSERT INTO turns_meta (turn_id, project_id, session_id, ts, \
+                 user_uuid, assistant_uuid, \
+                 user_store_offset, user_store_len, \
+                 assistant_store_offset, assistant_store_len) \
+                 VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9, ?10)",
+                params![
+                    r.turn_id,
+                    r.project_id,
+                    r.session_id,
+                    r.ts,
+                    r.user_uuid,
+                    r.assistant_uuid,
+                    r.user_store_offset,
+                    r.user_store_len,
+                    r.assistant_store_offset,
+                    r.assistant_store_len,
+                ],
+            )
+            .context("insert turns_meta")?;
+        }
+        for (project_id, session_id, offset) in &self.offsets {
+            crate::schema::write_session_offset(conn, project_id, session_id, *offset)?;
+        }
+        tx.commit()?;
+        Ok(self.turns.len())
+    }
+}
+
 /// Index a single session's transcript records into Tantivy + turns_meta.
 ///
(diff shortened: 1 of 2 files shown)


fagemx:
## Code Review: PR #416 (Round 1)

### Summary

Correctness was already pressed by codex, which found two real bugs in the fix itself (duplicate docs through the new crash window; a same-run `turn_id` collision aborting the whole flush) — both fixed with tests that failed first. This round is the maintainability/coverage pass. The ordering rationale is well commented, `sync` is the only caller of the changed API, and error handling is `anyhow` + `?` throughout. One P1.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — `index_project` has no tests, and it owns the new `merge`.**

`crates/edda-search-fts/src/indexer.rs:459-478`. Nothing in the suite calls `index_project` directly. The `sync` tests do reach it (they pass `session_id: None`), but only against an empty index dir, so it returns `PendingMeta::default()` at the `!index_dir.exists()` guard — **`PendingMeta::merge` is never executed by any test**.

That matters more than a normal coverage gap for two reasons:

1. `merge` is new code in this PR, and it is the only thing joining per-session results. If it dropped `offsets` (extending only `turns`), every session would be re-read in full on every sync forever — silently reintroducing the exact GH-403 cost this all started from — and the suite would stay green.
2. `index_project` is the production path. `bg_index::run_index` passes `None`, so **every SessionEnd** goes through `index_project`, not `index_session`. The tested function is the one that runs rarely; the untested one is the one that runs every session exit.

Needs a test that indexes two sessions in one call and asserts both the merged turns *and* both session offsets survive, plus that a second run is a no-op. It should be verified to actually fail when `merge` is broken, rather than trusted to be meaningful.

### Testing Review

#### Coverage

| Change | Tests |
|---|---|
| Deferral (`PendingMeta`, flush-after-commit) | `crash_before_tantivy_commit_leaves_turns_reindexable` — the issue's executable form; asserted 0 re-indexed before the fix |
| `delete_term` for turns (codex #1) | `crash_between_commit_and_flush_does_not_duplicate_turns` — failed `left: 2, right: 1` |
| `INSERT OR IGNORE` (codex #2) | `duplicate_turn_id_within_one_run_does_not_poison_the_flush` — failed `Error code 1555` |
| Early-out / offsets / grown file | 3, adapted to the new API without weakening their assertions |
| **`index_project` + `merge`** | **none — see P1-1** |

#### Convention Compliance

No violations. Real SQLite (`ensure_meta_db_memory`), no internal mocking, plain `#[test]`, clippy clean at `-D warnings`.

Two P2s, not worth a round:

- `PendingMeta::turns()` is used only by tests; production reads `flush`'s return value. A pub accessor that exists for tests is a mild smell, but it is a legitimate query on a value object and the tests genuinely need pre-flush state.
- The "flush only after commit" invariant is enforced by a doc comment, not the type system. Tightening `index_session`/`index_project`/`flush` to `pub(crate)` would shrink the footgun to one module — `sync` is their only caller — but this crate is published, so that is a semver break and belongs in its own change.

#### Testing Verdict: Insufficient Coverage — the production path (`index_project`) is untested

---

fagemx:
## Code Review: PR #416 (Round 2) — LGTM :tada:

### Summary

| Round | Finding | Fix |
|---|---|---|
| 1 | `index_project` untested, and it owns the `merge` this PR adds | direct test asserting merged turns **and** both session offsets survive, plus a no-op second pass |

The gap was worth a round on its own merits: `bg_index` passes no session id, so **every SessionEnd runs `index_project`**, not `index_session` — the tested function was the rare one. And a `merge` that dropped offsets would have kept the suite green while re-reading every session on every sync forever, silently restoring the GH-403 cost this exists to avoid.

The new test was verified to have teeth rather than assumed: breaking `merge` to drop offsets makes it fail on `offset for s1 must survive merge`. A coverage test that has never been seen failing proves nothing.

### Final state

Correctness (codex, before this loop): two real bugs in the fix itself — duplicate documents through the newly-moved crash window, and a same-run `turn_id` collision aborting the whole flush. Both fixed, tests first, both failing first (`left: 2, right: 1`; `Error code 1555`).

Ordering, the point of the PR: `writer.commit()` → `write_index_version` → `pending.flush()` → `write_events_cursor`. Every SQLite write is post-commit. A failure anywhere after the commit leaves both paths re-runnable and idempotent — events via `delete_term` on `doc_id`, turns now likewise.

55 tests in `edda-search-fts` (from 51 on main), clippy clean at `-D warnings`, fmt clean.

### Known limits, deliberately not fixed here

- **The sync-level ordering is guarded by inspection, not a test.** Moving `pending.flush()` above `writer.commit()` would silently restore GH-413 and every unit test would still pass — the invariant only manifests under a real crash, which is not injectable at this seam without a test hook that would itself be a design smell. The unit tests pin the mechanics (`index_session` writes nothing; a re-run does not duplicate), and the live budget check catches a missing flush loudly, since unrecorded turns make the no-op sync fall back to ~25s.
- `PendingMeta` buffers a full rebuild's rows in memory — O(total project turns). Trivial at the real corpus's ~1300 turns; chunking means multiple commits, a different design.
- `PendingMeta::turns()` is used only by tests; production reads `flush`'s return. Mild smell, but a legitimate query, and the tests need pre-flush state.
- `index_session`/`index_project`/`flush` could be `pub(crate)` — `sync` is their only caller — which would shrink the "flush only after commit" footgun to one module. The crate is published, so that is a semver break for its own change.

Ready for merge once the live budget check is posted — the deferral touches the hot path and #403's 0.157s no-op must still hold. Merge is the operator's gate.

---

fagemx:
## Live verification (installed binary, real corpus)

The PR body promised these before merge, since the deferral touches the hot path.

### Budget — the #403 number must not be paid back for crash safety

| | no-op `edda search index` |
|---|---|
| before #403 | 24.7s |
| after #403 | 0.157s |
| **after this PR** | **0.171s** |

Same ballpark as #403; the deferral costs nothing measurable. The SessionEnd hook stays well inside the 45s background-join budget.

Incremental still works: 5 new events picked up in 0.276s.

### The turns path end-to-end, through the new deferred flush

```
edda note "gh413 verify zqxj413probe"     → evt_01kxkcfmgs3r7sjqdjsryzpnkv
SessionEnd hook dispatched
edda search query "zqxj413probe"          → Found 2 result(s)
  1. [note] evt_01kxk5fg9jw7b7ck3xnas57pcd    ← older, similar name (fuzzy)
  2. [note] evt_01kxkcfmgs3r7sjqdjsryzpnkv    ← the new one
  (indexed through 2026-07-15T17:16:39.3210519Z)
```

The watermark advanced to the new note's own timestamp with **0 newer**, so SessionEnd indexed it with no manual command — through `PendingMeta`, flushed after the commit.

(`Found 2` is fuzzy ASCII matching against the earlier `zqxj403probe` probe, one character apart — not a miss. Checked rather than assumed: the new event id is present in the results.)

A missing flush would have shown up loudly here, since unrecorded turns would drag the no-op sync back toward ~25s.

Issue #418, opened 2026-07-15
search: a missing ledger.db silently purges the whole search index and reports success


fagemx:
## Problem

If a project's `.edda/ledger.db` goes missing while `.edda/` survives, `edda search index` **silently deletes that project's entire event index** and reports success.

The chain:

1. `Ledger::open` gates on `EddaPaths::is_initialized()`, which is only `edda_dir.is_dir()` (`crates/edda-ledger/src/paths.rs:62`).
2. It then calls `SqliteStore::open_or_create(&paths.ledger_db)` — **creates** an empty `ledger.db` rather than failing.
3. `sync` reads the stored `events_watermark` cursor, probes the (now empty) ledger, finds the boundary event gone, and correctly concludes the cursor has run ahead → `rebuilt = true`.
4. The rebuild path calls `delete_all_event_docs` and re-adds from the ledger — which has nothing.

Result: every event document purged, cursor reset to 0, and the command prints:

```
Rebuilt index from scratch.
Indexed 0 event(s) + 0 turn(s) for project <id>
```

The output reads like a success. Nothing warns that a populated index just became empty.

## Not a regression

The pre-#412 bulk path did the same thing more directly: `index_events` unconditionally issued `delete_term(doc_type="event")` and then re-added whatever the ledger yielded — nothing, for an empty ledger. So an empty ledger has always purged the index. #412 narrowed it (the purge now only runs on the rebuild path), but preserved the outcome.

Arguably "index mirrors the ledger; empty ledger → empty index" is correct semantics. The problem is that an *accidentally* empty ledger is indistinguishable from a deliberately empty one, and the destructive interpretation is chosen silently.

## Fix (sketch)

Options, roughly increasing in strength:

- Require `.edda/ledger.db` to exist before opening, rather than letting `open_or_create` invent one. `Ledger::open` arguably should not be a creating call at all — `open_or_init` already exists for callers that want creation.
- Refuse the purge when the ledger is empty but the index is not, unless something explicit is passed. A rebuild that deletes N documents and adds 0 is worth a confirmation or at least a loud warning.
- Report it honestly: "rebuilt from scratch: 2261 documents removed, 0 added" makes the destruction visible instead of framing it as routine success.

## Acceptance

- Deleting `.edda/ledger.db` (leaving `.edda/`) and running `edda search index` does not silently empty the index.
- Whatever it does instead is visible in the output.

[...]

Issue #417, opened 2026-07-15
registry: 1628 of 1644 project entries are dead — tests register temp dirs and never clean up


fagemx:
## Problem

The project registry on a real dev machine currently holds **1644 entries, of which 1628 point at paths that no longer exist**:

```
GONE  1628
OK      16
```

And several of the 16 survivors are themselves temp/smoke dirs. The genuinely real projects number about half a dozen. 436 entries are literally named `test*`; most of the rest look like `.tmpK3o9Lx`.

Tests call `edda_store::register_project` (directly, or via `edda init` / smoke paths) against `tempfile` dirs. The temp dir is cleaned up; the registry entry it created is not. Every test run leaves more. CI never notices because containers start clean — this only accumulates on developer machines.

This is the same class as #415 (tests writing to the developer's real store rather than a temp root), but with a four-digit blast radius rather than two directories.

## Why it matters

- `registry.json` is read and rewritten on `register_project` / `touch_project`. It is now ~99% dead weight on every write.
- `list_projects()` returns 1644 entries, essentially all garbage. Anything built on top of it — fleet grouping (`group set fleet`), the cross-repo read fan-out specced in #407, `edda peers` style views — inherits the noise. #407's fan-out over a registry in this state would try 1628 dead paths.
- It quietly misrepresents the fleet: the registry is supposed to be the list of projects that exist.

`registry.rs` already has the right notion — an entry is stale when its path no longer contains a `.edda/` directory. Nothing calls it as a sweep.

## Fix

- Stop tests polluting the real registry. `EDDA_STORE_ROOT` overrides the store root (`edda_store::store_root`), but it is process-wide and cargo runs tests in parallel threads, so setting it per-test corrupts siblings — see the comment on `bg_index::tests` for that trap. Prefer threading a root/registry path through the API the tests exercise.
- Add a prune: drop entries whose path is gone or no longer contains `.edda/`. Either an explicit verb (`edda project prune`) or opportunistically on registry write — the staleness rule already exists, it just needs a caller.
- Sweep the existing 1628 dead entries.

## Acceptance

- `cargo test --workspace` adds no entries to the real `registry.json`.
- A prune leaves only projects whose paths still exist and still contain `.edda/`.
- Whatever #407 fans out over is not 99% dead paths.

Found while assessing the registry for #414 (which needs `get_project(pid).path` to resolve a project's repo). #414 is unaffected — it looks up a single id and checks the path — but it is what surfaced this.

[...]

PR #419, opened 2026-07-15
search: index --project X from X's own repo, not the current one (GH-414)

### LINKED ISSUE #414: search: `index --project X` indexes the current repo's events into X (cross-project contamination)
## Problem

`edda search index --project OTHER` run from repo X silently fills **OTHER's** search index with **X's** events.

`run_cmd` resolves the project id from `--project` independently of `repo_root` (`crates/edda-cli/src/cmd_search.rs`):

```rust
SearchCmd::Index { project, session } => {
    let pid = project.as_deref().unwrap_or(&default_pid);
    index(repo_root, pid, session.as_deref())
}
```

and `index()` then pairs the two:

```rust
let ledger = Ledger::open(repo_root)?;                       // THIS repo's ledger
let stats = sync::sync(&proj_dir, project_id, session_id, |after| {
    ledger.events_after_rowid(after)                          // THAT project's index
})?;
```

The events land under OTHER's `project_id` and become searchable there. No warning, no error — the command reports success. Normally `project_id` is derived from `repo_root` so they agree; only an explicit `--project` makes them diverge, which is exactly when this fires.

This predates #412 (the old `index()` had the same `Ledger::open(repo_root)` + `project_dir(project_id)` shape). #412 fixed the two equivalents on the **read** path and left this one, since `query` never opened a ledger before and `index` did:

- `query`'s build-if-missing now returns early unless `project_id == resolve_project_id(repo_root)`
- `print_watermark` omits the "newer events" count for a non-local project (the `indexed through` timestamp stays, since it is read from that project's own meta.sqlite)

## Fix

Two options:

1. **Mirror the #412 guard** — refuse `index --project X` unless X is the project this repo resolves to, telling the user to run it from X's repo. Simple and consistent, but removes the flag's only real use.
2. **Make the flag actually work** — look up the project's repo path in the registry (`crates/edda-store/src/registry.rs` stores a `path` per entry) and open *that* ledger, so `--project X` genuinely indexes X from X's own repo. Registry already has a staleness notion (an entry is stale when its path no longer contains `.edda/`).

Option 2 is the better fit for the `fleet.cross-repo-reads1` ruling in the ledger (truth stays in each repo; cross-repo knowledge is a query-time fan-out over the registry). Run `edda ask "cross-repo"` for that reasoning before choosing.

## Acceptance

- `edda search index --project X` from an unrelated repo either indexes X from X's own ledger, or refuses — never writes this repo's events into X.
- Covered by a test. `cmd_search.rs` has no tests today, so the logic likely needs splitting into a testable unit — see how `bg_index::index_exists` was split out of `should_run` to avoid touching the process-wide store root.

Found during #412 review.

fagemx:
Closes #414.

## Problem

`edda search index --project X` run from repo Y silently filled **X's** search index with **Y's** events, and reported success.

`run_cmd` resolves `--project` independently of `repo_root`, and `index()` paired `Ledger::open(repo_root)` with `project_dir(project_id)`. Normally those agree (the project id is derived from the repo you are in); only an explicit `--project` makes them diverge — which is exactly when it fires. #412 fixed the two equivalents on the read path and left this one, since `query` had never opened a ledger before and `index` always had.

## Fix — resolve, don't refuse

The registry already knows where each project lives, so `--project X` now genuinely indexes X **from X's own repo** rather than being rejected.

That keeps `--project` meaning "operate on X" the way it does for `query` and `show`, instead of making `index` the one subcommand that refuses. It also matches the shape of the `fleet.cross-repo-reads1` ruling: truth stays home, so X's events come from X's repo.

`query`'s build-if-missing goes through the same resolution. #412 had guarded it by telling the user to run `index --project X` "from that project's repo" — with index now working from anywhere, that message would have become a lie.

## Guarding the registry rather than trusting it

Adversarial review (codex) turned up that resolving through the registry needs more care than the registry's own staleness rule provides:

- **`.edda/` presence is too weak.** `Ledger::open` only checks the *directory* (`is_initialized()` is `edda_dir.is_dir()`) and then calls `SqliteStore::open_or_create` — it **creates** an empty `ledger.db`. An orphan `.edda/` would hand back an empty ledger, and an empty ledger makes `sync` see the cursor as ahead and **purge every event document the project has**, reported as a successful rebuild. Resolution now requires `.edda/ledger.db` itself.
- **The registry's claim is verified, not trusted.** The repo it names must actually hash to the project asked for. Acting on a wrong path is precisely this issue's bug; sourcing it from bad data rather than bad logic is no improvement. And the registry is demonstrably unreliable — 1628 of 1644 entries on a real machine point at paths that no longer exist (#417).

## Tests

`cmd_search.rs` had **no tests**. It now has six. The resolution is split into `ledger_root_for` with an injected lookup — resolving for real reads the process-wide store root, which a test cannot redirect without corrupting parallel siblings (the same shape as `bg_index::index_exists`).

| Case | Behaviour |
|---|---|
| local project | uses this repo; registry never consulted (test panics the lookup to prove it) |
| foreign, registered, live | resolves to **its** repo |
| foreign, unregistered | refused |
| foreign, path gone | refused |
| foreign, orphan `.edda/` without a ledger | refused |
| registry names a repo that isn't that project | refused |

195 tests in `edda`, clippy clean at `-D warnings`, fmt clean.

## Filed rather than folded in

- **#418** — a missing `ledger.db` silently purges the index and reports success. This PR closes the *foreign* path into it, but the local path (your own repo, `ledger.db` gone) still purges. Not a regression: the pre-#412 bulk path did the same thing more directly.
- **#417** — the registry is 99% dead entries because tests register temp dirs and never clean up.

Two further codex findings judged out of scope, both pre-existing and neither worsened here: `project_id()` deliberately maps every git worktree of a repo to the same id, so "same id ⇒ same ledger" is not strictly sound (but the local fast path only preserves what `index` already did); and `query` returns exit 0 when it could not search at all, which predates this work and applies to every "no index" path.

### DIFF
--- a/crates/edda-cli/src/cmd_search.rs
+++ b/crates/edda-cli/src/cmd_search.rs
@@ -3,7 +3,7 @@ use edda_index::fetch_store_line;
 use edda_ledger::Ledger;
 use edda_search_fts::{schema, search, sync};
 use edda_store::project_dir;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 // ── CLI Schema ──
 
@@ -112,26 +112,26 @@ pub fn query(
     // command is the complaint this issue exists to answer.
     let missing = !index_dir.exists();
     let outdated = schema::index_is_outdated(&index_dir);
-    // Only auto-build the project this repo actually resolves to. `--project`
-    // names a project independently of where we are standing, and repo_root's
-    // ledger is not that project's ledger — building from it would write these
-    // events into someone else's index. For a project that isn't here, the
-    // explicit message is the honest answer.
-    let is_local_project = project_id == resolve_project_id(repo_root);
-    if (missing || outdated) && !is_local_project {
-        eprintln!(
-            "No usable search index for project {project_id}. Run \
-             `edda search index --project {project_id}` from that project's repo."
-        );
-        return Ok(());
-    }
     if missing || outdated {
         if missing {
             println!("No search index — building now (one-time)…");
         } else {
             println!("Search index schema is outdated — rebuilding now (one-time)…");
         }
-        let ledger = Ledger::open(repo_root)?;
+        // Build from the ledger that actually backs this project, which may not
+        // be the repo we are standing in (GH-414). Degrade rather than hard-fail:
+        // a query that cannot work out where to build from should say so, not
+        // abort with an error.
+        let ledger_root = match ledger_root_for(repo_root, project_id, |pid| {
+            edda_store::registry::get_project(pid).map(|e| e.path)
+        }) {
+            Ok(root) => root,
+            Err(e) => {
+                eprintln!("{e}");
+                return Ok(());
+            }
+        };
+        let ledger = Ledger::open(&ledger_root)?;
         let stats = sync::sync(&proj_dir, project_id, None, |after| {
             ledger.events_after_rowid(after)
         })?;
@@ -229,6 +229,46 @@ fn print_watermark(repo_root: &Path, proj_dir: &Path, project_id: &str) {
     }
 }
 
+/// Decide whose ledger backs `project_id` (GH-414).
+///
+/// `--project` names a project independently of where the user is standing, so
+/// the two can diverge. Pairing this repo's ledger with another project's index
+/// writes these events into that project — silently, and reported as success.
+///
+/// The registry knows where each project lives, so a foreign project is resolved
+/// to its own repo rather than refused. `lookup` is injected (production passes
+/// `edda_store::get_project`) because resolving it for real reads the
+/// process-wide store root, which a test cannot redirect without corrupting its
+/// siblings.
+fn ledger_root_for(
+    repo_root: &Path,
+    project_id: &str,
+    lookup: impl FnOnce(&str) -> Option<String>,
+) -> anyhow::Result<PathBuf> {
+    if project_id == resolve_project_id(repo_root) {
+        return Ok(repo_root.to_path_buf());
+    }
+
+    let Some(path) = lookup(project_id) else {
+        anyhow::bail!(
+            "Project {project_id} is not in the registry, so there is no way to tell \
+             whose events it holds. Run `edda search index` from that project's repo."
+        );
+    };
+
+    // Same staleness rule the registry itself uses: a live repo has a `.edda/`.
+    // Falling back to repo_root here would be the bug this exists to fix.
+    let root = PathBuf::from(path);
+    if !root.join(".edda").is_dir() {
+        anyhow::bail!(
+            "Project {project_id} is registered at {} but that path no longer holds \
+             an edda ledger. Run `edda search index` from that project's repo.",
+            root.display()
+        );
+    }
+    Ok(root)
+}
+
 /// Execute `edda search index` — build/update the Tantivy index for a project.
 pub fn index(repo_root: &Path, project_id: &str, session_id: Option<&str>) -> anyhow::Result<()> {
     let proj_dir = project_dir(project_id);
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: PR #419 (Round 1)

### Summary

Correctness was pressed by codex before this loop, which found the `.edda/`-is-too-weak hole (an orphan directory resolves to an empty ledger, and an empty ledger purges the index). That is fixed. This round is the maintainability pass, and it turned up one issue of the same family: the guard is right, but the message it prints sends the user around it.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — the orphan-`.edda` refusal advises an action that triggers the very bug it is refusing.**

`crates/edda-cli/src/cmd_search.rs:274-280`:

```
Project X is registered at P but that path holds no edda ledger.
Run `edda search index` from that project's repo.
```

Follow that literally:

1. `cd P` and run `edda search index`
2. `project_id == resolve_project_id(repo_root)` → the **local fast path**, which returns `repo_root` without any `ledger.db` check
3. `Ledger::open` sees `.edda/` is a dir, calls `SqliteStore::open_or_create` → **invents an empty ledger.db**
4. `sync` finds the cursor ahead of an empty ledger → `delete_all_event_docs` → **the project's index is emptied**, printed as `Rebuilt index from scratch. Indexed 0 event(s)`

So this PR refuses to do the destructive thing remotely, then tells the user to go do it locally, where the new guard does not reach. That asymmetry — foreign path guarded, local path not — is created by this PR, so the misdirection is too.

The guard itself is right and the local hole is correctly out of scope (#418, and not a regression: the pre-#412 bulk path purged the same way). But the message must not steer into it. State the fact and stop; do not prescribe a reindex that would empty the index.

### Testing Review

#### Coverage

Six tests where `cmd_search.rs` previously had zero: local wins without consulting the registry (the lookup panics to prove it), foreign resolves to its own repo, and unregistered / path-gone / orphan-`.edda` / registry-names-the-wrong-project each refuse. The injected `lookup` is the right seam — resolving for real reads the process-wide store root, which a parallel test cannot redirect safely.

The messages themselves are asserted on (`contains("no edda ledger")` etc.), so P1-1's fix is covered by an existing test once the text changes.

#### Convention Compliance

No violations. Clippy clean at `-D warnings`, real filesystem in tests via `tempfile`, no mocking of internals.

Two P2s, not worth a round:

- The production lookup closure `edda_store::registry::get_project(pid).map(|e| e.path)` appears twice (`:126` in query, `:303` in index). One line, same file, and divergence would be compile-visible — but both call sites of the injection point wire it identically, so a named `registry_lookup` fn would be the honest shape.
- `:262` says "that path no longer holds an edda ledger" while the condition it guards is `!root.exists()` — the path is gone entirely, which is a different (and more informative) fact than "no ledger".

---

fagemx:
## Code Review: PR #419 (Round 2) — LGTM :tada:

### Summary

| Round | Finding | Fix |
|---|---|---|
| 1 | The orphan-`.edda` refusal prescribed the destruction it had just refused | message states the fact and stops; test asserts the safety property, not a substring |

The guard was right; the message sent the user around it. `Run `edda search index` from that project's repo` takes the **local fast path**, which never reaches the `ledger.db` check — `Ledger::open` invents an empty ledger and the rebuild empties the index (#418), printed as success. This PR guards the foreign path and leaves the local one, so the asymmetry, and therefore the misdirection, was this PR's to own.

Round 2 swept the other four `Run `edda search index`` sites for the same trap. They are clean: "could not be opened → rebuild" is correct advice (corrupt index, ledger intact), and "not in the registry → go to that repo" is honest, since an unregistered project is exactly the case we know nothing about.

### Final state

Correctness (codex, before this loop): `.edda/` presence is too weak to resolve a ledger — `is_initialized()` is only `edda_dir.is_dir()` and `Ledger::open` then calls `open_or_create`, so an orphan directory yields an empty ledger and an empty ledger purges the index. Resolution now requires `.edda/ledger.db`, and the registry's claim is verified rather than trusted (the named repo must actually hash to the project asked for).

Six tests where `cmd_search.rs` had none. 195 in `edda`, clippy clean at `-D warnings`, fmt clean.

### P2s, deliberately not gated

- `:263` says "no longer holds an edda ledger" while guarding `!root.exists()` — the path is gone entirely, a different and more useful diagnosis, and the message then advises going to a repo that does not exist. Wrong about its own condition, but harmless: the advice cannot be followed, so it cannot do damage. Flagged in round 1 and left as flagged rather than promoted to justify a round.
- The production lookup closure `registry::get_project(pid).map(|e| e.path)` appears at `:126` and `:303`. One line, same file, divergence would be compile-visible.

No critical or high-priority issues remaining. Merge is the operator's gate.

---

PR #420, opened 2026-07-15
search: refuse to let an empty ledger empty a populated index (GH-418)

### LINKED ISSUE #418: search: a missing ledger.db silently purges the whole search index and reports success

fagemx:
Closes #418.

## Problem

If a project's `.edda/ledger.db` went missing while `.edda/` survived, `edda search index` **silently deleted that project's entire event index** and reported success:

```
Rebuilt index from scratch.
Indexed 0 event(s) + 0 turn(s) for project <id>
```

`Ledger::open` gates only on `EddaPaths::is_initialized()`, which is `edda_dir.is_dir()`, then calls `SqliteStore::open_or_create` — it **creates** an empty ledger rather than failing. `sync` then saw its cursor as ahead of that empty ledger, concluded the ledger had been replaced, purged every event document, and re-added nothing.

At SessionEnd this is completely silent: `bg_index` only `tracing::warn`s.

**Not a regression.** The pre-#412 bulk path did the same thing more directly — `index_events` unconditionally deleted every event doc then re-added whatever the ledger yielded, which for an empty ledger is nothing.

## Fix — refuse, don't warn

Events are append-only. A ledger that was populated and now reports nothing means **the ledger is gone**, not that its events were removed. Acting on the other reading destroys the only surviving copy.

A warning was considered and rejected: in the SessionEnd hook a warning goes to `tracing` and nobody ever sees it. Only a failure stops the destruction.

## Two destructive steps, not one

The first version of this fix guarded `delete_all_event_docs` — and adversarial review (codex) showed that was the wrong place for the path that matters most.

`sync` also wipes the whole index dir earlier, on a schema upgrade (`remove_dir_all`). For an outdated index whose ledger had vanished: the wipe destroyed the populated index, `open_or_create_index` made a fresh empty one, and the guard then counted **zero** event docs and waved the rebuild through. **The check passed precisely because the damage was already done.**

That is the worst case, not an exotic one: the upgrade path discards the index *on purpose*, to rebuild it from the ledger. If the ledger is gone too, the wipe is the loss. The original test could not catch it — it never marked the index outdated.

The ledger is now checked before the wipe as well, with the ledger half evaluated first (it is the definitive half, and that read only happens on the rare upgrade path). Both sites share one refusal.

## Reconciling a real conflict with #412

#412's review demanded that a rebuild purge events the ledger no longer has — and its test proved that **using an empty ledger**, which is exactly the case now refused. Its actual concern was a *replaced* ledger leaving orphans, so the test now uses a replacement and asserts the stronger property: the old events go **and** the new one arrives.

## Root cause, deliberately not fixed here

`Ledger::open` inventing a ledger is the root cause, but it has **163 call sites** across the workspace and "`.edda/` exists means initialized" is a standing design assumption. `sync` is the only consumer that acts *destructively* on an empty ledger — every other one just reads nothing — so the guard belongs here. Fixing `Ledger::open` is its own change.

## Tests

57 in `edda-search-fts` (from 55 on main). Both new tests were seen failing first, with the bug's own signature (`Ok(SyncStats { events: 0, rebuilt: true })` — success, having just purged).

| Test | Pins |
|---|---|
| `an_empty_ledger_must_not_empty_a_populated_index` | refusal, and the index is untouched afterwards |
| `an_outdated_index_is_not_wiped_when_the_ledger_has_nothing_to_rebuild_from` | the wipe path — refusing after the wipe is not refusing |

The guard costs nothing on the hot path: it only evaluates when `rebuilt && batch.is_empty()`, which never happens on a normal incremental sync, so #403's 0.157s no-op budget is untouched by construction.

Clippy clean at `-D warnings`, fmt clean. (`secret_guard::tests::hot_path_under_ten_ms_for_realistic_length` in `edda-core` is a known timing flake under parallel load — verified passing in isolation and on `main`, and untouched by this change.)

### DIFF
--- a/crates/edda-search-fts/src/sync.rs
+++ b/crates/edda-search-fts/src/sync.rs
@@ -9,6 +9,7 @@
 //! `edda-ledger` (the same inversion `index_events` already used).
 
 use crate::{indexer, schema};
+use anyhow::Context;
 use std::path::Path;
 
 /// What a sync did. `indexed_through` is the timestamp of the newest indexed
@@ -57,19 +58,29 @@ where
         // copy. The ledger is checked first because it is the definitive half and
         // this read only happens on the rare upgrade path.
         if events_after(0)?.is_empty() {
-            let existing = schema::open_index(&index_dir)
-                .and_then(|old| {
-                    let old_schema = old.schema();
-                    event_doc_count(&old, &old_schema).ok()
-                })
-                .unwrap_or(0);
-            if existing > 0 {
-                return Err(empty_ledger_refusal(
-                    project_id,
-                    existing,
-                    proj_dir,
-                    &search_dir,
-                ));
+            // An index that will not open has nothing to lose, so a missing one
+            // is fine to wipe. But one that opens and cannot be counted is the
+            // opposite: we would be deleting an unknown quantity. Assuming zero
+            // there fails open on the only guard against silent, unrecoverable
+            // loss — it would reproduce GH-418 through the guard instead of
+            // around it. So the count propagates rather than defaulting.
+            if let Some(old) = schema::open_index(&index_dir) {
+                let old_schema = old.schema();
+                let existing = event_doc_count(&old, &old_schema).with_context(|| {
+                    format!(
+                        "refusing to wipe project {project_id}'s outdated index: its \
+                         ledger reports no events and the index could not be counted, \
+                         so wiping it might destroy the only copy"
+                    )
+                })?;
+                if existing > 0 {
+                    return Err(empty_ledger_refusal(
+                        project_id,
+                        existing,
+                        proj_dir,
+                        &search_dir,
+                    ));
+                }
             }
         }
         std::fs::remove_dir_all(&index_dir)?;
--- a/crates/edda-search-fts/src/sync.rs
+++ b/crates/edda-search-fts/src/sync.rs
@@ -67,10 +67,36 @@ where
     let cursor = schema::read_events_cursor(&meta_conn, project_id)?;
     let (batch, rebuilt) = resolve_batch(&events_after, &cursor)?;
 
+    // GH-418: refuse to let an empty ledger empty a populated index.
+    //
+    // `Ledger::open` creates an empty `ledger.db` when one is missing, so a
+    // deleted ledger is indistinguishable from a legitimately empty one. Events
+    // are append-only: a ledger that was populated and now reports nothing means
+    // the ledger is gone, not that its events were removed. Purging on that
+    // reading destroys the only surviving copy of the index, and reports it as a
+    // successful rebuild — silently, since at SessionEnd nobody reads the log.
+    //
+    // Checked BEFORE the delete: a refusal that fires after the documents are
+    // gone is not a refusal.
+    if rebuilt && batch.is_empty() {
+        let existing = event_doc_count(&index, &tantivy_schema)?;
+        if existing > 0 {
+            anyhow::bail!(
+                "Refusing to rebuild project {project_id}: its ledger reports no \
+                 events, but the index holds {existing}. Events are append-only, so \
+                 this almost certainly means the ledger is missing rather than empty \
+                 — check {}/.edda/ledger.db. Rebuilding would empty the index. To \
+                 discard it deliberately, delete {}.",
+                proj_dir.display(),
+                search_dir.display()
+            );
+        }
+    }
+
     // A rebuild must purge events the ledger no longer has. Re-adding the
-    // current ones is not enough: a truncated or replaced ledger would leave
-    // the old documents searchable forever, answering queries with events that
-    // no longer exist. (No-op when the index was just created fresh.)
+    // current ones is not enough: a replaced ledger would leave the old
+    // documents searchable forever, answering queries with events that no longer
+    // exist. (No-op when the index was just created fresh.)
     if rebuilt {
         indexer::delete_all_event_docs(&writer, &tantivy_schema)?;
     }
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: PR #420 (Round 1)

### Summary

codex already found the big one before this loop — the guard sat after the schema-upgrade wipe, so it counted zero and waved through exactly the destruction it existed to stop. That is fixed, and both destructive sites now share one refusal. This round found one issue of the same family: the two guards disagree about which way to fail.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — the wipe guard fails OPEN; the purge guard fails CLOSED.**

`crates/edda-search-fts/src/sync.rs:59-75` vs `:104-114`:

```rust
// wipe guard — swallows the error, assumes zero, wipes
let existing = schema::open_index(&index_dir)
    .and_then(|old| { let s = old.schema(); event_doc_count(&old, &s).ok() })
    .unwrap_or(0);

// purge guard — propagates
let existing = event_doc_count(&index, &tantivy_schema)?;
```

`.ok().unwrap_or(0)` collapses two very different cases:

| case | current | correct |
|---|---|---|
| `open_index` → `None` (unopenable/corrupt) | wipe | **wipe** — nothing to lose |
| opens, but `event_doc_count` → `Err` | **wipe** | refuse — we do not know what we are destroying |

The second is a fail-open on a guard whose only job is to prevent silent, unrecoverable data loss. If we cannot count the index we are about to delete, "assume it is empty" is the one assumption we must not make — it reproduces GH-418 through the guard rather than around it.

Concretely: an older index whose stored schema does not resolve `doc_type` (the field `event_doc_count` queries) makes the count `Err`, so a populated index is wiped with the ledger gone. That is precisely the upgrade-path scenario codex flagged, surviving in the error branch.

The purge guard has it right. The two should agree, and they should agree on the safe direction: distinguish "cannot open" (proceed) from "opened but cannot count" (refuse).

### Testing Review

#### Coverage

57 tests (from 55 on main). Both new ones were seen failing first with the bug's own signature — `Ok(SyncStats { events: 0, rebuilt: true })`, i.e. success immediately after purging:

| Test | Pins |
|---|---|
| `an_empty_ledger_must_not_empty_a_populated_index` | refusal, and the index survives it |
| `an_outdated_index_is_not_wiped_when_the_ledger_has_nothing_to_rebuild_from` | the wipe path — refusing after the wipe is not refusing |

Neither covers the error branch in P1-1, which is why it survived.

#### Convention Compliance

No violations. Real tantivy + real SQLite in temp dirs, no internal mocking, clippy clean at `-D warnings`.

One P2, not worth a round: on the upgrade path the whole ledger is read twice — once by the new pre-wipe check (`events_after(0)`), then again by `resolve_batch` once the cursor is cleared. Only on a schema bump, so a fraction of a second on a 2.2k-event corpus, but it is avoidable by threading the first read through.

---

fagemx:
## Code Review: PR #420 (Round 2) — LGTM :tada:

### Summary

| Round | Finding | Fix |
|---|---|---|
| 1 | the wipe guard failed **open**, the purge guard failed **closed** | the count propagates with context; both now fail the same, safe way |

`.ok().unwrap_or(0)` had collapsed two cases that must not be collapsed: an index that will not open has nothing to lose, but one that opens and cannot be counted is an unknown quantity. Assuming zero there reproduced GH-418 *through* the guard rather than around it — in the very upgrade path codex had just flagged.

Round 2 swept the file: no `unwrap_or` / `.ok()` remains on any guard path.

### Final state

Both destructive steps are now guarded, and both refuse rather than warn — a warning at SessionEnd goes to `tracing` and is seen by nobody, so only a failure stops the loss.

57 tests (from 55 on main). Both new tests were seen failing first, with the bug's own signature: `Ok(SyncStats { events: 0, rebuilt: true })` — success, immediately after purging.

The guard costs nothing on the hot path (it evaluates only when `rebuilt && batch.is_empty()`), so #403's 0.157s no-op budget is untouched by construction.

### One unexplained observation, recorded rather than buried

During this loop a single `cargo test -p edda-search-fts` run reported `error: test failed` and the failing test name was not captured. It has not recurred in **20 subsequent local runs**, and CI is green 7/7 across Linux/macOS/Windows on clean containers. codex had specifically asked whether `event_doc_count` opening an `index.reader()` alongside a live `IndexWriter` could misbehave, which is the right shape for such a flake — but nothing reproduces it, so this is logged as unexplained, not as diagnosed and not as absent.

### P2s, deliberately not gated

- On the upgrade path the ledger is read twice: once by the pre-wipe check, then again by `resolve_batch` after the cursor is cleared. Only on a schema bump; avoidable by threading the first read through.
- `Ledger::open` inventing an empty ledger is the root cause. 163 call sites, and "`.edda/` exists means initialized" is a standing assumption — its own change.

Merge is the operator's gate.

---

Issue #415, opened 2026-07-15
test: bg_digest audit_log_appends writes to the real store and fails permanently once it fails once


fagemx:
## Problem

`bg_digest::tests::audit_log_appends` writes into the developer's **real** store and only cleans up on the happy path, so one failure makes it fail forever on that machine.

`crates/edda-bridge-claude/src/bg_digest.rs:450`:

```rust
let pid = "test_digest_audit";
let _ = edda_store::ensure_dirs(pid);        // real store, not a temp dir
...
append_audit_log(pid, &entry).unwrap();
...
assert_eq!(content2.lines().count(), 2);     // requires exactly 2 lines
...
// Cleanup
let _ = fs::remove_dir_all(edda_store::project_dir(pid));   // last line — skipped on panic
```

The cleanup is the final statement, so any earlier assertion failure panics past it. The audit log keeps its rows, the next run appends 2 more, `assert_eq!(..., 2)` sees 4, panics again, skips cleanup again. **The failure is self-perpetuating** — once red on a machine, red forever, and the error message ("expected 2, got 4") describes the previous run rather than the actual defect.

CI never sees this because containers start clean. It only bites developers.

Test residue in a real store is not hypothetical — this is what is sitting in `%APPDATA%\Roaming\edda\projects` on one dev machine right now:

```
test_state_rt
test_wmfail_23084
```

`test_wmfail_23084` even carries a pid suffix, i.e. it was *meant* to be unique per run and still was not cleaned up. So other tests leak too; `audit_log_appends` is just the one whose assertion turns a leak into a permanent failure.

## Fix

- Point the test at a temp dir. `EDDA_STORE_ROOT` overrides the store root (`edda_store::store_root`), but it is process-wide and cargo runs tests in parallel threads, so setting it here would corrupt sibling tests — see the comment on `bg_index::tests` for that trap. Prefer refactoring `append_audit_log`/`audit_log_path` to take a root (or a project dir) so the test can pass a `tempfile::tempdir()`.
- Failing that, at minimum make the assertion independent of prior state (assert the last N lines / that the entries are present) so a leak cannot make the failure permanent.
- Sweep the other tests that write real project dirs (`test_state_rt`, `test_wmfail_*`) for the same pattern.

## Acceptance

- The test passes twice in a row on a machine where it previously failed, with no manual cleanup.
- No `test_*` project dirs appear in the real store after `cargo test --workspace`.

Found while running the workspace suite during #403.

PR #421, opened 2026-07-15
test(cli): stop edda-cli tests writing to the developer's real store (GH-417)

fagemx:
Part of #417 (does not close it — see "What remains").

## Problem

edda-cli's tests wrote into the **developer's real per-user store** and left it that way. Measured: one `cargo test -p edda` added **11 registry entries**, and 1644 had accumulated — 1628 of them pointing at paths that no longer exist. CI never notices, because its containers start empty. Only the developer's machine accumulates.

Two sources, both reaching `registry::register_project` via `cmd_init::execute`:

- `cmd_init`'s own 7 tests
- `cmd_draft`'s `init_workspace` helper, used by 11 more

Neither redirected the store.

## Fix

`edda-store` already solved this for itself (`with_isolated_store` + `ENV_STORE_LOCK`), but that lock is `#[cfg(test)] pub(crate)` — edda-cli cannot reach it. This adds the equivalent in `test_support.rs`, with two improvements over the original:

- the redirect is undone by **`Drop`**, not a trailing statement, so a panicking test cannot leave `EDDA_STORE_ROOT` pointing at a `TempDir` that is about to be deleted
- a **poisoned lock is taken, not unwrapped**, so one failing test does not cascade into every later one

### The env approach is all-or-nothing

The store root is process-wide, so isolating *some* tests breaks the siblings that still use the real store. Isolating init and draft duly broke two `cmd_bridge` tests — the exact trap already documented in `bg_index`'s own comment.

`cmd_bridge` (9 tests) and `cmd_scan` (1) also write to the real store; they are what leaves `test_state_rt` / `test_wmfail_*` in `projects/` — the #415 class. Isolating them fixes the breakage and their pollution in one move.

## Verified by measurement, not inspection

| | before | after |
|---|---|---|
| `cargo test -p edda` registry delta | **+11** | **+0** |
| tests | 193 | **195 green** |

The registry itself was also swept, using a command that already existed — see the correction below.

## Two corrections to the issue, both mine

**"The staleness rule just needs a caller" was false.** `edda gc --global` phase 4d already calls `validate_projects()` and unregisters stale entries. It found all 1788; with a dry-run first confirming it would touch nothing else (0 blobs, 0 transcripts), the sweep took **1644 → 16 entries, all live**. No prune needed building.

**The culprits were wrong.** `registry.rs`'s own tests are already isolated; the real sources were `cmd_init` and `cmd_draft`.

## What remains — `edda-conductor` adds 13/run

Bisected across all 85 workspace members; it is the only remaining source, and it resists the obvious explanations: it depends only on `edda-store`, uses just `project_dir`/`project_id`/`write_atomic`, never calls `register_project`, and reproducing its `Command::new("edda")` subprocess by hand registers nothing. Left open on #417 with the evidence.

## A measurement trap, recorded

Two false "+0 pollution" readings occurred while working this, both because **`cargo test --workspace` aborts the remaining crates once one fails** — so a broken run reports less pollution *and* a smaller passed count. The first "+0" was a run with 2 failures, not a fix. A sudden drop in the total passed count is a red flag, not noise.

### DIFF
--- a/crates/edda-cli/src/cmd_bridge.rs
+++ b/crates/edda-cli/src/cmd_bridge.rs
@@ -1497,6 +1497,7 @@ mod tests {
 
     #[test]
     fn decide_writes_binding_to_coordination_log() {
+        let _store = crate::test_support::isolated_store();
         let _env = env_guard();
         let (tmp, _ledger) = setup_workspace();
         let pid = edda_store::project_id(&tmp);
@@ -1541,6 +1542,7 @@ mod tests {
 
     #[test]
     fn decide_writes_structured_ledger_event() {
+        let _store = crate::test_support::isolated_store();
         let _env = env_guard();
         let (tmp, ledger) = setup_workspace();
         let pid = edda_store::project_id(&tmp);
@@ -1588,6 +1590,7 @@ mod tests {
 
     #[test]
     fn ratify_records_separate_event_and_makes_decision_binding() {
+        let _store = crate::test_support::isolated_store();
         let _env = env_guard();
         let (tmp, ledger) = setup_workspace();
         let pid = edda_store::project_id(&tmp);
@@ -1639,6 +1642,7 @@ mod tests {
 
     #[test]
     fn ratify_unknown_key_errors() {
+        let _store = crate::test_support::isolated_store();
         let _env = env_guard();
         let (tmp, _ledger) = setup_workspace();
         let pid = edda_store::project_id(&tmp);
@@ -1656,6 +1660,7 @@ mod tests {
 
     #[test]
     fn decide_supersedes_prior_decision_same_key() {
+        let _store = crate::test_support::isolated_store();
         let _env = env_guard();
         let (tmp, ledger) = setup_workspace();
         let pid = edda_store::project_id(&tmp);
@@ -1702,6 +1707,7 @@ mod tests {
 
     #[test]
     fn resolve_session_id_tiers() {
+        let _store = crate::test_support::isolated_store();
         let _env = env_guard();
         let pid = "test_resolve_sid_tiers";
         let _ = edda_store::ensure_dirs(pid);
@@ -1822,6 +1828,7 @@ mod tests {
 
     #[test]
     fn render_coordination_solo_no_bindings() {
+        let _store = crate::test_support::isolated_store();
         let pid = "test_render_coord_solo";
         let _ = edda_store::ensure_dirs(pid);
         let result = edda_bridge_claude::render::coordination(pid, "solo-session");
@@ -1835,6 +1842,7 @@ mod tests {
 
     #[test]
     fn render_pack_no_pack_file() {
+        let _store = crate::test_support::isolated_store();
         let pid = "test_render_pack_empty";
         let _ = edda_store::ensure_dirs(pid);
         let result = edda_bridge_claude::render::pack(pid);
@@ -1844,6 +1852,7 @@ mod tests {
 
     #[test]
     fn heartbeat_write_touch_remove_lifecycle() {
+        let _store = crate::test_support::isolated_store();
         let pid = "test_hb_lifecycle";
         let sid = "sess-lifecycle-1";
         let _ = edda_store::ensure_dirs(pid);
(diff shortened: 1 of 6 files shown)


fagemx:
## Code Review: PR #421 (Round 1)

### Summary

The fix is verified by measurement rather than inspection (+11 → +0 entries per run, 195 green), and the all-or-nothing nature of env redirection was found the hard way and handled — isolating init/draft alone broke two `cmd_bridge` tests, so `cmd_bridge` and `cmd_scan` are isolated too. All 28 call sites correctly bind `let _store` (a `let _ =` would drop instantly and silently disable the isolation). One P1.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — `Drop` clobbers a pre-existing `EDDA_STORE_ROOT` instead of restoring it.**

`crates/edda-cli/src/test_support.rs:20-24`:

```rust
impl Drop for IsolatedStore {
    fn drop(&mut self) {
        std::env::remove_var("EDDA_STORE_ROOT");
    }
}
```

It removes unconditionally, never restoring what was there before.

Consider the person most likely to hit it: a developer who, knowing about exactly this pollution, runs `EDDA_STORE_ROOT=/tmp/scratch cargo test` to keep the suite off their real store. The first isolated test finishes, `Drop` removes the variable, and every subsequent test in that process silently falls back to the **real** store — writing precisely the entries they redirected to avoid.

So the mechanism defeats the protection the user explicitly asked for, and does it quietly. That is worse than not having the guard, because they believe they are covered.

The fix is to save the previous value on acquire and put it back on drop:

```rust
prev: Option<OsString>,   // captured before set_var
// drop: match prev { Some(v) => set_var(.., v), None => remove_var(..) }
```

`edda-store`'s original has the same flaw, but this is new code and should not inherit it — the whole point of writing a fresh copy was to improve on it (this version already fixes its panic-safety and lock-poisoning gaps).

### Testing Review

#### Coverage

The verification here is a measurement, not a unit test, and that is the right call: the property is "the suite does not touch the real store", which cannot be asserted from inside the suite without reaching for the real store to check. Recorded numbers: +11 → +0 for `cargo test -p edda`, 195 green, registry swept 1644 → 16.

No test covers P1-1's restore behaviour. Worth adding with the fix, but note it cannot assert on the process-wide variable without racing its own siblings — asserting the struct's captured `prev` is the honest seam.

#### Convention Compliance

No violations. Clippy clean at `-D warnings`. The isolation mirrors the pattern the codebase already adopted in `edda-store`.

Two P2s, not worth a round:

- There are now **two** implementations of this idea in the workspace — `edda_store::registry::tests::with_isolated_store` and this one — and the new one is strictly better (Drop-safe, poison-tolerant, and after P1-1, restore-correct). The next person copies whichever they find first. Sharing it would need a dev-only support crate, which is its own change.
- `edda-conductor` still adds 13 entries per run, so `cargo test --workspace` is not yet clean. Bisected and documented on #417 with the evidence; deliberately not chased here.

---

fagemx:
## Code Review: PR #421 (Round 2) — LGTM :tada:

| Round | Finding | Fix |
|---|---|---|
| 1 | `Drop` clobbered a pre-existing `EDDA_STORE_ROOT` instead of restoring it | previous value captured on acquire (inside the lock) and put back on drop |

That one mattered because of *who* it hit: the developer already defending against this bug by running `EDDA_STORE_ROOT=/somewhere cargo test` would have had their redirect removed by the first isolated test, silently dropping every later test back onto the real store — writing precisely the entries they redirected to avoid.

### Final state

- edda-cli: **+11 registry entries per run → +0**, measured, not inspected
- 195 tests green across three consecutive runs; clippy clean at `-D warnings`
- registry swept 1644 → 16, all live, using `edda gc --global` which already existed
- all 28 call sites bind `let _store` (a `let _ =` would drop instantly and silently disable the isolation)

### A test that was deleted rather than shipped

A test for the restore path was written, failed, and removed. Every seam available to it is racy: the assertion must compare against the process-wide variable, which is only stable while the lock is held — but acquiring it means calling `isolated_store()`, whose lock is not reentrant, so a nested read deadlocks and an unnested one races whichever sibling is mid-redirect. The first version failed exactly that way.

A flaky test here would be worse than none: it trains the next person to re-run until green, which is the habit that lets real failures through. The reasoning is recorded in the file so nobody re-attempts it blind.

### Not covered here

`edda-conductor` still adds 13 entries per run, so `cargo test --workspace` is not yet clean and **#417 stays open**. It is bisected and documented there with the evidence, including why the obvious explanations do not hold.

Merge is the operator's gate.

---

PR #422, opened 2026-07-15
test(bridge): stop two tests leaving residue in the real store (GH-415)

### LINKED ISSUE #415: test: bg_digest audit_log_appends writes to the real store and fails permanently once it fails once

fagemx:
Closes #415.

## Problem

Two tests in `edda-bridge-claude` wrote into the **developer's real store** and left it that way. One of them failed permanently once it failed once.

### `bg_digest::tests::audit_log_appends`

It asserts an exact row count while appending to a log in the real store, and cleans up only as its **last statement**:

```rust
let pid = "test_digest_audit";
let _ = edda_store::ensure_dirs(pid);          // real store
...
assert_eq!(content2.lines().count(), 2);       // exact count
...
let _ = fs::remove_dir_all(edda_store::project_dir(pid));   // last line — skipped on panic
```

A run that panics before reaching that line leaves rows behind. The next run counts those too, panics again, and skips the cleanup again. **Self-perpetuating**: once red on a machine, red forever — and the message (`expected 2, got 3`) describes the leftovers rather than the defect, so the next person debugs the wrong thing.

Reproduced before fixing, rather than taken on faith:

```
one leftover row  →  assertion failed: left: 3, right: 2
second run        →  fails identically
```

During that exercise the leftover log grew to **5 rows** — the accumulation made visible.

### `digest::tests::digest_state_round_trip`

No cleanup at all. That is why `projects/test_state_rt` sits in the store of every machine that has ever run the suite.

CI never saw either: its containers start empty. Only developers accumulate this.

## Fix

**Clear first, not only last.** The trailing cleanup still handles the happy path; clearing at the start recovers the unhappy one, so a poisoned state self-heals instead of persisting.

For `digest_state_round_trip`, the cleanup is placed **before** the assertions rather than after, so a failing assertion cannot skip it.

## Verified

| Criterion | Result |
|---|---|
| passes twice in a row on a machine where it previously failed, no manual cleanup | ✅ verified from a deliberately dirtied store, then re-dirtied and run twice |
| no `test_*` project dirs in the real store after a full run | ✅ clean |
| tests | **542 green**, clippy clean at `-D warnings` |

## Scope note

These tests still *write* to the real store — they just no longer poison themselves or leave residue. Redirecting them properly is the `EDDA_STORE_ROOT` isolation work, and that turned out to be all-or-nothing: the store root is process-wide, so isolating some tests breaks siblings that still use the real store (proven in #421, where isolating two edda-cli files duly broke two `cmd_bridge` tests). This crate has **91+ real-store touches across 10 files**, so doing it here would be a change of a different size — and it belongs with #417's remaining work, not with a two-test hygiene fix.

### DIFF
--- a/crates/edda-bridge-claude/src/bg_digest.rs
+++ b/crates/edda-bridge-claude/src/bg_digest.rs
@@ -449,6 +449,19 @@ mod tests {
     #[test]
     fn audit_log_appends() {
         let pid = "test_digest_audit";
+
+        // Start from a known state (GH-415). This test asserts an exact row
+        // count, and it appends to a log in the real store — so a previous run
+        // that panicked before reaching its cleanup leaves rows behind and this
+        // one counts them too. That failure is self-perpetuating: it panics
+        // again, skips cleanup again, and the log grows by two every attempt.
+        // Once red on a machine, red forever, with a message ("expected 2, got
+        // 3") describing the leftovers rather than the defect.
+        //
+        // Cleaning up first rather than only last is what makes it self-healing:
+        // the trailing cleanup still runs on the happy path, and this one
+        // recovers the unhappy one.
+        let _ = std::fs::remove_dir_all(edda_store::project_dir(pid));
         let _ = edda_store::ensure_dirs(pid);
 
         let entry = AuditEntry {
--- a/crates/edda-bridge-claude/src/bg_scan.rs
+++ b/crates/edda-bridge-claude/src/bg_scan.rs
@@ -1179,6 +1179,12 @@ End of analysis."#;
     fn audit_log_appends() {
         let pid = "test_scan_audit";
         let _ = edda_store::ensure_dirs(pid);
+        // Start from a known state (GH-415), as bg_detect's copy of this test
+        // already does. Appending to a log in the real store while asserting an
+        // exact count means a run that panicked before its cleanup leaves rows
+        // for this one to count — which panics, skips cleanup, and leaves more.
+        // Once red on a machine, red forever.
+        let _ = fs::remove_file(audit_log_path(pid));
 
         let entry = AuditEntry {
             ts: "2026-03-12T10:00:00Z".to_string(),
--- a/crates/edda-bridge-claude/src/digest/tests.rs
+++ b/crates/edda-bridge-claude/src/digest/tests.rs
@@ -653,6 +653,11 @@ fn digest_permanent_failure_after_3_retries() {
 #[test]
 fn digest_state_round_trip() {
     let project_id = "test_state_rt";
+    // Writes to the real per-user store and, unlike its neighbours, never
+    // cleaned up — so it left `projects/test_state_rt` on every developer
+    // machine that ever ran the suite (GH-415). Clearing first also makes it
+    // independent of whatever a previous run left behind.
+    let _ = std::fs::remove_dir_all(edda_store::project_dir(project_id));
     let _ = edda_store::ensure_dirs(project_id);
 
     let state = DigestState {
@@ -666,6 +671,9 @@ fn digest_state_round_trip() {
     save_digest_state(project_id, &state).unwrap();
 
     let loaded = load_digest_state(project_id);
+
+    let _ = std::fs::remove_dir_all(edda_store::project_dir(project_id));
+
     assert_eq!(loaded.session_id, "sess-123");
     assert_eq!(loaded.event_id, "evt_abc");
     assert_eq!(loaded.retry_count, 0);


fagemx:
## Code Review: PR #422 (Round 1)

### Summary

The bug was reproduced before fixing, the fix is verified from a deliberately dirtied store, and the cleanup ordering (before the assertions, not after) is right. But the PR fixed one of **two** live instances of the same defect.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — `bg_scan::tests::audit_log_appends` has the identical self-perpetuating bug and is untouched.**

There are three tests named `audit_log_appends` in this crate. They are near-copies:

| test | clears first? | exact-count asserts | status |
|---|---|---|---|
| `bg_digest` (`:450`) | now yes | `count() == 2` | **fixed by this PR** |
| `bg_detect` (`:964`) | **yes** — `fs::remove_file(audit_log_path(pid))` at `:967` | `count() == 1`, `== 2` | already immune |
| `bg_scan` (`:1179`) | **no** | `count() == 1`, `== 2` | **same bug, still live** |

Reproduced, not inferred — one leftover row in `test_scan_audit`'s log:

```
test bg_scan::tests::audit_log_appends ... FAILED
  left: 2
 right: 1
```

Same shape as the one this PR fixes: appends to a log in the real store, asserts an exact count, cleans up only on the happy path, and so poisons itself permanently on any failure.

`bg_detect` makes the omission worse rather than better: the fix idiom **already exists in this crate**, one file over. Shipping the fix for `bg_digest` while its identical neighbour keeps the bug leaves the codebase in a state where three copies of one test have three different behaviours — and the next person to hit the `bg_scan` failure will find a fixed `bg_digest` and a fixed `bg_detect` and reasonably assume theirs is a real defect in the code under test.

The issue itself asked for this: *"Sweep the other tests that write real project dirs … for the same pattern."* The sweep found `test_state_rt`; it should have found this too.

### Testing Review

#### Coverage

The verification is empirical and appropriate — the property is "does not poison itself across runs", which is a claim about state between runs, not something an in-suite assertion can hold. Recorded: reproduced at `left: 3, right: 2`, leftover log observed growing to 5 rows, then passes from a dirtied store and twice more after re-dirtying. 542 green, no `test_*` residue.

#### Convention Compliance

No violations. Clippy clean at `-D warnings`. `digest_state_round_trip`'s cleanup is correctly placed before its assertions.

One P2, not worth a round: 215 `remove_dir_all(project_dir(..))` cleanup sites exist in this crate and only 27 clear at the start. Most are harmless — permanence needs the append + exact-count combination — but the "clean only at the end" habit is crate-wide, and the next test to add an exact-count assertion inherits the trap. Worth a convention note rather than 188 edits.

---

fagemx:
## Code Review: PR #422 (Round 2) — LGTM :tada:

| Round | Finding | Fix |
|---|---|---|
| 1 | `bg_scan::tests::audit_log_appends` had the identical self-perpetuating bug, unfixed | cleared first, using the idiom `bg_detect` already had one file over |

Three tests in this crate share the name `audit_log_appends` and are near-copies. The PR fixed one; `bg_detect` was already immune; `bg_scan` still poisoned itself. Reproduced rather than inferred (`left: 2, right: 1`), and fixed.

That `bg_detect` already had the fix is what made the omission worth a round: leaving it would have given three copies of one test three different behaviours, with the fix idiom sitting visible in a neighbouring file.

### Final state — all three verified from a deliberately dirtied store

```
bg_digest    dirty → PASS
bg_detect    dirty → PASS
bg_scan      dirty → PASS
```

- 542 tests green; clippy clean at `-D warnings`
- no `test_*` directories left in the real store after a full run
- `digest_state_round_trip`'s cleanup sits before its assertions, so a failing assertion cannot skip it

Round 2 swept the crate for a third instance of the combination that causes permanence (append + exact-count + cleanup-only-at-end). There isn't one — the only other `lines().count()` is production code in `dispatch/helpers.rs`.

### P2, not gated

215 `remove_dir_all(project_dir(..))` cleanup sites exist here and only 27 clear at the start. Most are harmless, since permanence needs the append + exact-count combination — but "clean only at the end" is the crate's habit, and the next exact-count assertion added to a real-store test inherits the trap. That is a convention note, not 188 edits.

These tests still *write* to the real store; they just no longer poison themselves or leave residue. Redirecting them is all-or-nothing `EDDA_STORE_ROOT` work (91+ touches across 10 files here) and belongs with #417's remainder.

Merge is the operator's gate.

---

Issue #407, opened 2026-07-14
fleet reads: --fleet fan-out on ask/search/log/task-list + empty-result honesty (P1)


fagemx:
## Problem

Every read verb — `ask`, `search query`, `log`, `task list` — sees exactly one workspace: the cwd's ledger. A fleet is cross-repo by definition, and the knowledge is already split: task-rail P1's decisions and receipts live in the fleet workspace's ledger, while rail P2+ work items live in the edda repo's own ledger. Standing in either workspace, the other half does not exist. Standing in a third repo, none of it exists.

Observed 2026-07-15: `edda ask "rail"` answers fully in the fleet workspace and would answer nothing in any sibling repo. The split-brain is invisible because a local miss renders exactly like true absence — the same silent-empty failure mode as the CJK tokenizer bug (#402).

The federation primitives already exist and are unused: `edda_store::registry` has `list_projects()` / `list_group_members()` / `validate_projects()`, every ledger is a local SQLite file, and per-project Tantivy indexes are addressable. Cross-repo read is a fan-out loop, not a distributed system.

## Fix

**Principle: truth stays home, reads federate, sync is only for enforcement.** No storage centralization; queries fan out at read time.

1. `--fleet` flag on `ask`, `search query`, `log`, and `task list`:
   - fan out over registered projects (group members when a group exists, else all registered projects), open each ledger read-only, run the same query, merge
   - every hit is tagged with its home project name: `[foundry]`, `[edda]`, …
   - per-project failures (missing repo on this machine, no search index) are reported per project, never silently dropped — `[dazun] index not built` is a result line, not an omission
2. **Empty-result honesty (default, no flag)**: when a local query misses, probe the fleet cheaply and say so: `no hits in this workspace; 3 hits in [foundry] — rerun with --fleet`. A bare "No results" may no longer mean "checked everywhere".
3. `edda task list --fleet --status ready` is the operator's morning board — one command, every repo's ready queue. (Mini-preview of the P4 watch board; P4 should consume the same fan-out.)

## Acceptance

- From a sibling repo, `ask --fleet "rail"` returns the fleet workspace's rail decisions, tagged
- `task list --fleet --status ready` lists ready tasks from ≥2 workspaces in one output
- A project whose search index is missing shows an explicit per-project notice in `search query --fleet`
- Local miss without `--fleet` mentions fleet hits when they exist
- House TDD; multi-workspace fixtures via the existing temp-workspace pattern

## Not in scope

[...]

PR #423, opened 2026-07-15
fix(ask): check a decision's glob against the directory it names (GH-405)

### LINKED ISSUE #405: ask: stale-code hint false-positives on absolute cross-repo paths (P2)
## Problem

`edda ask` renders a stale-code hint per decision by checking the decision's `--paths` globs for existence — but it resolves them against the current repo root. Decisions recorded with **absolute paths into another repo** (normal in a fleet: the Foundry ledger governs work in `C:/ai_agent/edda/*`) are all reported as `⚠ stale-code hint: ... (Missing)` even though the paths exist. Every cross-repo decision currently carries a false warning, which trains readers to ignore the hint entirely — killing the signal for genuinely stale paths.

Observed 2026-07-15: `ask "rail"` flags `C:/ai_agent/edda/crates/edda-bridge-claude/*` and `C:/ai_agent/edda/crates/edda-core/*` as Missing; both exist.

## Fix

- Detect absolute path patterns (drive letter / leading `/`) and check them as absolute, not repo-relative
- If the absolute target doesn't exist on this machine, say `(unreachable from this workspace)` instead of `(Missing)` — a fleet ledger is legitimately read on machines that don't have every repo checked out
- Relative patterns keep the current repo-root behavior

## Acceptance

- A decision with existing absolute cross-repo paths shows no warning
- A decision whose relative path was deleted still warns `Missing`
- Test covering both, plus the drive-letter form on Windows

fagemx:
Closes #405.

## Problem

Every decision recorded with `edda claim --paths` carried a false `⚠ stale-code hint: … (Missing)`.

Reproduced live before touching anything:

```
$ ls -d C:/ai_agent/edda/crates/edda-bridge-claude C:/ai_agent/edda/crates/edda-core
→ both exist

$ edda ask "cross-repo reads federate"
  ⚠ stale-code hint: C:/ai_agent/edda/crates/* (Missing)
```

A hint that cries wolf on **every** decision trains readers to ignore it — which destroys it for the paths that really are stale. That is the actual damage, not the noise.

## The issue's diagnosis was wrong

#405 blamed absolute cross-repo paths being resolved against the current repo root, and proposed detecting absolute patterns. But `check_paths_staleness` **already** does that:

```rust
if p.is_absolute() { p.to_path_buf() } else { root.join(p) }
```

So the proposed fix would have changed nothing. The proof the model was wrong is that the **relative** form false-positives identically — from inside the repo that contains it:

```
⚠ stale-code hint: crates/edda-search-fts/* (Missing)
```

#405 asserts relative patterns "keep the current repo-root behavior", implying they work. They don't.

**The axis is glob vs literal, not absolute vs relative.** `edda claim --paths "crates/foo/*"` is the documented way to record a scope, so the stored patterns are globs — and the probe was `fs::metadata` on the pattern itself. Nothing is named `*`, so the check could only ever fail, for every decision, on both path forms.

## Fix

Probe the **deepest wildcard-free ancestor** — which is what the hint actually asserts: *the area this decision governs still exists*.

- `crates/foo/*` → probes `crates/foo`
- `a/*/c.rs` → probes `a`
- no wildcard → returned unchanged, so literal paths keep their exact previous behaviour, including per-file mtime

No new dependency; the existing absolute/relative handling is untouched and now actually reachable.

## Tests

Seen failing first. The deletion signal is pinned so the fix cannot quietly swallow the thing the hint exists for:

| Test | Pins |
|---|---|
| `a_glob_is_checked_against_the_directory_it_names_not_literally` | the relative form (which #405 believed was fine) |
| `an_absolute_glob_into_another_repo_that_exists_is_not_missing` | the observed drive-letter form |
| `a_glob_whose_directory_is_gone_still_reports_missing` | **a real deletion still warns** |

43 green in `edda-ask`, clippy clean at `-D warnings`.

## Acceptance

- ✅ a decision with existing absolute cross-repo paths shows no warning
- ✅ a decision whose relative path was deleted still warns `Missing`
- ✅ tests covering both, including the Windows drive-letter form

## Not done: `(unreachable from this workspace)`

#405's Fix section also wanted absolute paths that are absent to say *unreachable* rather than *Missing* — a fleet ledger legitimately gets read on machines without every repo checked out.

Deliberately skipped: it adds a variant to a `Serialize`d public enum (an API change) for a distinction none of the three acceptance criteria require, and the idea was premised on the diagnosis this PR just disproved. Worth its own issue if the fleet work (#407) needs it, where the "absent repo vs deleted code" distinction actually has a consumer.

### DIFF
--- a/crates/edda-ask/src/staleness.rs
+++ b/crates/edda-ask/src/staleness.rs
@@ -9,10 +9,15 @@
 //! query-time derivation, not a mutation. Best-effort: unreadable repo /
 //! missing file returns "unknown" rather than false-positive stale.
 //!
+//! Paths are recorded as globs (`edda claim --paths "crates/foo/*"`), so what
+//! gets probed is the deepest wildcard-free ancestor — the area the decision
+//! governs — not the pattern itself, which resolves to nothing. See
+//! [`probe_target`]. Literal paths are probed exactly as written.
+//!
 //! Vocabulary alignment:
-//! - `fresh`: the path exists and its mtime is at or before the decision ts.
-//! - `stale_modified`: file exists but its mtime is strictly after decision ts.
-//! - `missing`: path does not resolve on disk (repo-relative or absolute).
+//! - `fresh`: the probed path exists and its mtime is at or before the decision ts.
+//! - `stale_modified`: it exists but its mtime is strictly after decision ts.
+//! - `missing`: it does not resolve on disk (repo-relative or absolute).
 //! - `unknown`: repo root not supplied, or path attributes unreadable.
 
 use crate::DecisionHit;
--- a/crates/edda-ask/src/staleness.rs
+++ b/crates/edda-ask/src/staleness.rs
@@ -46,6 +46,36 @@ pub struct DecisionStaleness {
     pub paths: Vec<PathStaleness>,
 }
 
+/// What to actually probe on disk for a recorded path pattern.
+///
+/// Scopes are recorded as globs — `edda claim --paths "crates/foo/*"` is the
+/// documented form — and a glob cannot be probed literally, because nothing is
+/// named `*`. Doing so reported every such decision as `Missing` regardless of
+/// the truth (GH-405), which trains readers to ignore the hint and so destroys
+/// it for genuinely stale paths.
+///
+/// The deepest wildcard-free ancestor is what the hint is really asserting:
+/// "the area this decision governs still exists". `crates/foo/*` probes
+/// `crates/foo`; `a/*/c.rs` probes `a`. A pattern with no wildcard is returned
+/// unchanged, so literal paths keep their exact previous behaviour — including
+/// per-file mtime.
+fn probe_target(pattern: &Path) -> PathBuf {
+    let mut out = PathBuf::new();
+    for comp in pattern.components() {
+        let text = comp.as_os_str().to_string_lossy();
+        if text.contains('*') || text.contains('?') || text.contains('[') {
+            break;
+        }
+        out.push(comp);
+    }
+    // An all-wildcard pattern (`*`) leaves nothing to probe; keep it literal so
+    // it resolves against the repo root rather than silently becoming "".
+    if out.as_os_str().is_empty() {
+        return pattern.to_path_buf();
+    }
+    out
+}
+
 /// Filesystem oracle abstracts real fs so tests can be deterministic.
 pub trait FsOracle {
     /// Return (exists, mtime_rfc3339 when available).
@@ -84,8 +114,11 @@ pub fn check_paths_staleness<F: FsOracle>(
     let mut out = Vec::with_capacity(affected_paths.len());
     let mut any_stale = false;
     for rel in affected_paths {
+        // Reduce a glob to the directory it names before resolving; the reported
+        // path stays the pattern the decision actually recorded (GH-405).
+        let pattern = probe_target(Path::new(rel));
         let resolved: PathBuf = {
-            let p = Path::new(rel);
+            let p = pattern.as_path();
             if p.is_absolute() {
                 p.to_path_buf()
             } else {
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: PR #423 (Round 1)

### Summary

The diagnosis correction is well evidenced (the relative form false-positives identically, which disproves the absolute-vs-relative model), tests were seen failing first, and the deletion signal is explicitly pinned. One P1: the fix changes what the hint can detect, and neither the module contract nor the PR says so.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — probing a directory silently narrows the contract, and the module still claims the old one.**

The module header says:

> check whether those **files** have been modified since the decision was recorded

For a glob, this PR now probes the directory instead. A directory's mtime moves when entries are added or removed — **not when a file inside is edited**. Verified rather than assumed:

```
edit foo/bar.rs   → foo/ mtime unchanged   (edit not detected)
touch foo/new.rs  → foo/ mtime changed     (add/remove detected)
```

Editing a file in scope is the most common way a decision goes stale, and that is now invisible for every glob-scoped decision — which is nearly all of them, since `edda claim --paths` takes globs.

This is still a large improvement (100% false positives → none), and it is **aligned with the module's stated philosophy**, which explicitly prefers false-negatives:

> Best-effort: unreadable repo / missing file returns "unknown" rather than false-positive stale

So the trade is defensible — a hint that rarely fires beats one that always lies. But it must be stated. As written, the header promises file-level modification detection that no longer happens for globs, and the PR body claims the fix with no mention of the narrowing. A future reader debugging "why didn't the hint fire when I edited this file" has nothing to find.

Needs: the limitation recorded in the module doc (where the contract lives) and in the PR, plus a follow-up for accurate glob staleness — which means a directory walk taking the max mtime of matching entries, and therefore an `FsOracle` that can enumerate, not just `probe`.

### Testing Review

#### Coverage

Three tests, all seen failing/passing appropriately:

| Test | Pins |
|---|---|
| `a_glob_is_checked_against_the_directory_it_names_not_literally` | the relative form #405 believed was fine |
| `an_absolute_glob_into_another_repo_that_exists_is_not_missing` | the observed drive-letter form |
| `a_glob_whose_directory_is_gone_still_reports_missing` | the deletion signal survives the fix |

That last one is the important one — a fix that removed false positives by never warning at all would pass the first two.

No test covers the narrowing in P1-1, which is consistent with it being an accepted limitation rather than a bug, but it should be documented in place.

`annotate_hits` has exactly one consumer (`cmd_ask`), so the fix reaches every caller — checked.

#### Convention Compliance

No violations. `MockFs` keeps the tests deterministic; no new dependency; the existing absolute/relative branch is untouched and now actually reachable.

One P2: `probe_target` treats `*`, `?` and `[` as wildcard markers. That matches the globs edda records in practice, but it is a hand-rolled predicate, not a glob parser — a literal path legitimately containing `[` would be truncated. Vanishingly unlikely in a repo path, and the failure mode is a coarser probe rather than a wrong answer.

---

fagemx:
## Code Review: PR #423 (Round 2) — LGTM :tada:

| Round | Finding | Fix |
|---|---|---|
| 1 | probing a directory narrows what the hint detects, while the module still advertised file-level detection | limitation recorded in the header where the contract lives; #424 filed for accurate detection |

### Live verification on the real ledger

Not just "the false positive is gone" — the hint started telling the truth in both directions:

| | before | after |
|---|---|---|
| `C:/ai_agent/edda/crates/*` (absolute, cross-repo, exists) | `(Missing)` ×2 — **false** | **no warning** |
| `crates/edda-search-fts/*` (relative, exists, edited since) | `(Missing)` — **false** | **`(StaleModified)`** — **true** |

That second row is the point. The `search.tokenizer` decision was recorded at 05:24 and that crate has been edited since (directory mtime 14:56), so it *is* stale — and for the first time the hint says so instead of crying wolf. A fix that merely silenced the warning would have shown nothing there.

It also re-confirms the diagnosis correction: the relative form was false-positiving exactly like the absolute one, which is what disproved #405's absolute-vs-relative model.

### Final state

- 43 tests green in `edda-ask`; clippy clean at `-D warnings`
- three tests, all seen failing/passing appropriately, including the one that matters most — `a_glob_whose_directory_is_gone_still_reports_missing`, which a fix that silenced warnings would fail
- `annotate_hits` has one consumer (`cmd_ask`), so every caller is covered
- no new dependency; the existing absolute/relative branch untouched and now actually reachable

### Known limits, recorded not buried

- **Edits inside a glob are invisible** — a directory's mtime does not move when a file is rewritten in place (verified). Deliberate: it matches this module's stated preference for false-negatives, and a hint that rarely fires beats one that always lies. Documented in the header; **#424** filed for accurate detection, including the open question of whether walking the tree on every interactive `ask` is worth it.
- **`(unreachable from this workspace)`** from #405's Fix section is not implemented: it adds a variant to a `Serialize`d public enum for a distinction none of the three acceptance criteria need, and it was premised on the diagnosis this PR disproved. Belongs with #407 if the fleet work gives it a consumer.

Merge is the operator's gate.

---

PR #425, opened 2026-07-15
feat(fleet): read fan-out + ask --fleet (GH-407)

fagemx:
Part of #407 — delivers the fan-out foundation and the first verb. Does **not** close it (see "What remains").

## Problem

Every read verb sees exactly one workspace: the cwd's ledger. A fleet is cross-repo by definition, and the knowledge is already split. The failure is silent — a local miss renders identically to true absence.

Reproduced from the edda repo, asking for a ruling that lives in the Foundry ledger:

```
$ edda ask "cross-repo reads federate"
No results found.
```

That is a bare "No results" for a decision that exists one repo over — the exact thing the `fleet.cross-repo-reads1` ruling forbids, and the same silent-empty shape as the CJK bug (#402).

## Fix — the same query, asked of every project in scope

```
$ edda ask "cross-repo reads federate" --fleet
── Decisions ──────────────────────────
  fleet.cross-repo-reads1 = truth-stays-home-reads-federate-sync-only-for-enforcement — …
```

**Live on the real registry:** 4 of 16 projects answered, 0 unreadable, whole fan-out in **0.331s**.

### `registry::fleet_scope`

The group when the cwd project has one, else the whole registry. Deliberately unlike `list_group_members` in two ways:

- **Home is included** — a fleet read tags its own workspace's hits like any other; excluding it would silently omit the repo the user is standing in.
- **Absent repos are kept, not filtered** — the caller must be able to say "did not look here". Dropping them makes the fan-out quietly incomplete, which is the failure this issue exists to remove.

### `fleet::fan_out`

Owns the loop, the tagging and the failure accounting; takes the per-project query as a closure, because `ask` reads decisions, `search` reads a Tantivy index, and `log`/`task` read the ledger. Misses are **results, not omissions**: an absent repo or a failing project is reported with its name and reason, and one project failing does not take the others down.

## A design call worth flagging

Output is **per-project sections, not one merged ranked list.**

Relevance is TF-IDF *against a corpus*. A score from the foundry ledger and one from the edda ledger are computed over different document sets and are **not comparable** — interleaving them by rank would invent an ordering that means nothing. Since the entire point of this issue is to stop read verbs implying things they have not established, a fabricated ranking would be self-defeating. Sections also keep every hit unambiguously attributed.

## #417 was a prerequisite in practice

Before that sweep the registry held 1788 entries whose repos are gone. This fan-out would have reported 1788 `repo not on this machine` misses and been slow doing it — technically correct, useless in practice. Worth knowing if this is ever backported.

## Tests

Pin the promises, not the plumbing. 199 in `edda`, 28 in `edda-store`, clippy clean at `-D warnings`.

| Test | Pins |
|---|---|
| `fleet_scope_is_every_project_when_no_group_and_includes_home` | home is not omitted |
| `fleet_scope_narrows_to_the_group_when_one_is_set` | group scoping |
| `fleet_scope_keeps_projects_whose_repo_is_gone` | absent repos reach the caller |
| `hits_carry_the_project_they_came_from` | tagging |
| `an_absent_repo_is_a_reported_miss_not_a_silence` | **the core promise** |
| `a_failing_project_is_a_miss_and_the_rest_still_answer` | isolation |

## What remains (#407 stays open)

- `search query --fleet` + the per-project "index not built" notice (acceptance 3)
- `task list --fleet --status ready` — the operator's morning board (acceptance 2)
- `log --fleet`
- Empty-result honesty without the flag: a local miss should mention fleet hits (acceptance 4)

Shipping the foundation first so each verb lands reviewable on its own rather than as one unreadable change.

### DIFF
--- a/crates/edda-cli/src/cmd_ask.rs
+++ b/crates/edda-cli/src/cmd_ask.rs
@@ -6,6 +6,7 @@ use edda_ledger::Ledger;
 use std::path::Path;
 
 /// `edda ask [query]` — query project decisions, history, and conversations.
+#[allow(clippy::too_many_arguments)]
 pub fn execute(
     repo_root: &Path,
     query: Option<&str>,
@@ -14,8 +15,8 @@ pub fn execute(
     all: bool,
     branch: Option<&str>,
     impact: bool,
+    fleet: bool,
 ) -> anyhow::Result<()> {
-    let ledger = Ledger::open(repo_root)?;
     let q = query.unwrap_or("");
 
     let opts = AskOptions {
@@ -25,6 +26,12 @@ pub fn execute(
         impact,
         ..Default::default()
     };
+
+    if fleet {
+        return execute_fleet(repo_root, q, &opts, json);
+    }
+
+    let ledger = Ledger::open(repo_root)?;
 
     // Build transcript search callback
     let transcript_cb = build_transcript_callback(repo_root);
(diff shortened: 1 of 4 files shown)


fagemx:
## Code Review: PR #425 (Round 1)

### Summary

The fan-out is verified live on the real registry (4 of 16 projects answered, 0 unreadable, 0.331s), the tests pin promises rather than plumbing, and the non-fleet path is untouched — `execute` returns early into `execute_fleet` and the original body is unchanged. The per-project-sections call is well argued. One P1, and it is this PR's own principle turned against it.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — the stale-index warning is anonymous and repeats once per project.**

`crates/edda-cli/src/cmd_ask.rs:138`, reached from `execute_fleet` via `build_transcript_callback`:

```rust
eprintln!("(search index is out of date; run `edda search index` to include transcripts)");
```

That message was written for a single workspace, where "the index" is unambiguous. Fanned out, it is wrong in three ways at once:

1. **Unattributed.** With 16 projects in scope it says nothing about *which* one is stale. This issue's rule is that per-project failures are reported *per project* — an anonymous line is not that.
2. **Repeats.** Three stale indexes print the same sentence three times, and the reader cannot tell whether that is three projects or one bug.
3. **The remedy is now ambiguous.** "Run `edda search index`" meant the cwd's project when it was written. After #414 the honest instruction is `--project <that one>`, which requires knowing which one — see (1).

It also goes to stderr while the fan-out's own output goes to stdout, so it interleaves unpredictably with the sections it is talking about.

Latent today only by luck: none of the 16 projects currently has an outdated index, and a *missing* index returns `None` silently without this message. It fires the moment a fleet project's index falls behind a schema bump — i.e. exactly the situation #402/#410 created and will create again.

The fix is to attribute it: give `build_transcript_callback` the project label the fan-out already has, and render `[name] search index is out of date` — which is also the shape acceptance 3 needs for `search query --fleet`'s "index not built" notice, so it is work the next verb wants anyway.

### Testing Review

#### Coverage

| Test | Pins |
|---|---|
| `fleet_scope_*` ×3 | home included, group narrowing, absent repos preserved |
| `hits_carry_the_project_they_came_from` | tagging |
| `an_absent_repo_is_a_reported_miss_not_a_silence` | the core promise |
| `a_failing_project_is_a_miss_and_the_rest_still_answer` | one failure does not sink the fan-out |

Adequate for what ships here. No test covers P1-1 — consistent with it being latent, and a reason it survived.

`execute_fleet` itself is untested (it is CLI glue that prints, in a file with no test module — the crate's pre-existing convention), and is covered instead by the live run against the real 16-project registry, which is recorded in the PR.

#### Convention Compliance

No violations. Clippy clean at `-D warnings` — worth noting the foundation commit alone would *not* have been: `fan_out` was dead code until a verb used it, so the two had to land together.

---

fagemx:
## Code Review: PR #425 (Round 2) — LGTM :tada:

| Round | Finding | Fix |
|---|---|---|
| 1 | the stale-index notice was anonymous and repeated once per project | attributed and made actionable; verified live by marking a real index stale |

```
[edda] search index is out of date; run `edda search index --project <id>` to include its transcripts
```

Worth recording why that instruction is safe to print at all: only **#414**, earlier in this same loop, made `index --project X` resolve X's own repo through the registry. Before that, this message would have been telling the user to fill X's index with the wrong repo's events. The honest remedy exists because of the previous fix.

### Live verification on the real registry

| | |
|---|---|
| the split-brain, from the edda repo | `No results found.` → **`[AI Delivery Foundry]` + the actual ruling** |
| coverage | 4 of 16 projects answered, **0 unreadable** |
| cost | **0.331s** for the whole fan-out |
| attributed notice | fired correctly with a deliberately staled index, then restored (v2, 125 events) |

The first row is the point: that bare `No results` is exactly what `fleet.cross-repo-reads1` forbids, and it was the real behaviour of this repo an hour ago.

### #417 was a prerequisite in practice

Before that sweep the registry held 1788 entries whose repos are gone. This fan-out would have printed 1788 `repo not on this machine` lines — correct by the rules, useless in practice. Worth knowing if this is ever backported.

### Final state

199 tests in `edda`, 28 in `edda-store`, clippy clean at `-D warnings`. Note the foundation could not have shipped alone: `fan_out` was dead code until a verb consumed it, so `-D warnings` forced them to land together.

### What remains — #407 stays open

- `search query --fleet` + per-project "index not built" notice (acceptance 3). **Landmine already located**: `cmd_search.rs:146` prints `Search index could not be opened…` anonymously — the same defect this round just fixed in `ask`, waiting for the fan-out to reach it.
- `task list --fleet --status ready` — the operator's morning board (acceptance 2)
- `log --fleet`
- Empty-result honesty without the flag (acceptance 4)

Merge is the operator's gate.

---

PR #426, opened 2026-07-16
feat(fleet): search query --fleet with per-project index notices (GH-407)

fagemx:
Part of #407 — delivers acceptance 3. Does not close it (see "What remains").

Follows #425, which landed `fleet_scope` + `fan_out` + `ask --fleet`.

## What this adds

`search query --fleet` — the same search, run against every project's **own** index, tagged with its home.

Live from the edda repo, searching a CJK term that only exists elsewhere:

```
$ edda search query "權威事實" --fleet
  [note] evt_01kxh47tz… ts=2026-07-14T20:14:06Z
     供掃。起因:外部評論指出投影把機器推論洗成«權威事實»。…
  [note] evt_01kxkaxn7… ts=2026-07-15T16:49:21Z

  [dazun]  index not built — run `edda search index --project 10bfe3f9…`
  [yushan] index not built — run `edda search index --project 97f7c3be…`
```

**0.180s** across 16 projects.

That output is four of tonight's fixes compounding: #410's CJK tokenizer makes the term findable at all, #403's auto-index keeps those indexes current, #414 makes the prescribed remedy actually work from another repo, and #425's fan-out ties them together.

## Acceptance 3, demonstrated

> A project whose search index is missing shows an explicit per-project notice in `search query --fleet`

Two notices above, each naming its project and carrying a command that runs from anywhere.

## The notice needed no new machinery

`fan_out` already turns a per-project error into an attributed line, so `bail!` *is* the notice. That also disposes of the landmine found while reviewing #425:

`cmd_search.rs`'s `"Search index could not be opened. Run \`edda search index\` to rebuild."` is the same anonymous shape just fixed in `ask` — it would have repeated unattributed, once per project, the moment this fanned out. The fleet path carries the project id in every message. The single-workspace path keeps its original wording, where naming the project would be noise.

## `--fleet` never builds

A cold build costs ~25s (measured in #403). Doing that for each of N projects because someone asked a question would be a trap that turns one question into a multi-minute stall. So an unusable index reports why and the others still answer — which is also what makes the notice above possible instead of a hang.

## Grouped, not merged

Same reasoning recorded in `cmd_ask`: relevance is TF-IDF against a corpus, so scores from different indexes are not comparable, and interleaving them by rank would invent an ordering. The whole point of this issue is to stop read verbs implying things they have not established.

## Tests

199 in `edda`, clippy clean at `-D warnings`. The fan-out's promises are pinned by #425's unit tests (`an_absent_repo_is_a_reported_miss_not_a_silence`, `a_failing_project_is_a_miss_and_the_rest_still_answer`); this verb is CLI glue that prints, in a file whose test module covers the pure logic (`ledger_root_for`), so it is verified by the live run above — recorded here rather than asserted in-process.

## What remains (#407 stays open)

- `task list --fleet --status ready` — the operator's morning board (acceptance 2)
- `log --fleet`
- Empty-result honesty without the flag (acceptance 4)

### DIFF
--- a/crates/edda-cli/src/cmd_search.rs
+++ b/crates/edda-cli/src/cmd_search.rs
@@ -41,6 +41,9 @@ pub enum SearchCmd {
         /// Maximum results (default: 20)
         #[arg(long, default_value_t = 20)]
         limit: usize,
+        /// Search every project in the fleet, not just this workspace
+        #[arg(long)]
+        fleet: bool,
     },
     /// Show full content of a specific turn
     Show {
@@ -70,6 +73,7 @@ pub fn run_cmd(cmd: SearchCmd, repo_root: &Path) -> anyhow::Result<()> {
             event_type,
             exact,
             limit,
+            fleet,
         } => {
             let pid = project.as_deref().unwrap_or(&default_pid);
             query(
@@ -81,6 +85,7 @@ pub fn run_cmd(cmd: SearchCmd, repo_root: &Path) -> anyhow::Result<()> {
                 event_type.as_deref(),
                 exact,
                 limit,
+                fleet,
             )
         }
         SearchCmd::Show { turn, project } => {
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: PR #426 (Round 1)

### Summary

Acceptance 3 is demonstrated on real data, the notices are attributed and actionable, and the #425 landmine is disposed of rather than inherited. One P1 — and it is this issue's own principle, broken by the new flag.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — `--project X --fleet` silently ignores `--project`.**

`crates/edda-cli/src/cmd_search.rs:217`:

```rust
if fleet {
    return query_fleet(repo_root, query_str, session_id, doc_type, event_type, exact, limit);
    //                 ^ project_id is never passed — it is dropped on the floor
}
```

Confirmed, not inferred:

```
$ edda search query "權威事實" --project 10bfe3f9…(dazun) --fleet
  [note] evt_01kxh47tz… …
```

The user asked for **dazun**. They got **the whole fleet**, including a project they did not name, and nothing said the flag had been discarded.

The two flags are contradictory by construction — "only this project" and "every project" — so one of them must lose. Losing silently is the problem: this is a read verb doing something other than what it was asked, quietly, which is precisely the failure mode #407 exists to remove. It is the same shape as the bare "No results", one level up: the output looks like a legitimate answer to a question nobody asked.

`ask --fleet` is unaffected — that command has no `--project` flag, so the contradiction cannot be expressed there.

The fix belongs at the CLI boundary rather than in the function: clap's `conflicts_with` turns it into a parse error, so the user is told, before anything runs, that the two cannot be combined. Any runtime precedence rule would still be a guess about which flag they meant.

### Testing Review

#### Coverage

The fan-out's promises are pinned by #425's unit tests, which this reuses unchanged. This verb is CLI glue that prints, in a file whose test module covers the pure logic (`ledger_root_for`), so it is verified by the live run recorded in the PR — including the per-project notices and the 0.180s fan-out cost across 16 projects.

No test covers P1-1, which is why it survived to review. A clap-level fix is verifiable by `--help`/parse behaviour rather than a unit test.

#### Convention Compliance

No violations. Clippy clean at `-D warnings`.

One P2, not worth a round: `group_by_project` is O(n·groups) — a linear scan of the accumulator per hit. At ≤20 results × 16 projects against ~16 groups this is nothing, and the scan is what preserves visit order, which a `HashMap` would lose. Worth a comment if the fleet ever gets large; not worth a data structure today.

---

fagemx:
## Code Review: PR #426 (Round 2)

### Summary

P1-1 is fixed at the CLI boundary. `--project X --fleet` is now refused before any project is touched:

```
$ edda search query "權威事實" --project 10bfe3f9… --fleet
error: the argument '--project <PROJECT>' cannot be used with '--fleet'
```

Both flags still work alone — `--fleet` fans out across 16 projects in **0.186s**, `--project` on a non-local project builds its index on demand (26 events, via GH-414's ledger resolution) and answers.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

None. P1-1 resolved.

### Testing Review

#### Coverage

The new test was **falsified before it was trusted**: with `conflicts_with` removed it fails at the `.expect`, because the parse *succeeds* — the bug, reproduced. Restored, it passes. That matters more than usual here, since `conflicts_with` takes an arg id as a string, so a typo or a later rename of the `project` field would stop enforcing silently and bring the bug back exactly as quietly as it arrived. A test that cannot fail would not have caught that.

`cmd_search::query` has no caller other than the CLI, so the clap guard covers every reachable path. The `project_id` parameter it still accepts and ignores under `--fleet` is unreachable dead weight rather than a defect — worth folding into the signature only if a second caller ever appears.

Full suite: **200 passed, 0 failed**. fmt clean. clippy **0** warnings at `--all-targets` (counted, not inferred from empty output).

#### Convention Compliance

No violations.

### Verdict: LGTM

Both flags do what they say, and saying "no" is one of the things they can say.

---

PR #427, opened 2026-07-16
feat(fleet): task list --fleet — the operator's morning board (GH-407)

fagemx:
## What

`edda task list --fleet` — one board across every project in scope, each row tagged with the repo it came from. GH-407 acceptance 2, the operator's morning board.

```
edda task list --fleet --status ready
```

## Why the project tag is load-bearing

Task ids restart at 1 in every repo. An untagged merged board shows two unrelated `#1` as if they were one row of one list — so the tag is what makes the board readable at all, not decoration on top of it.

The same reasoning drives the JSON shape. `--fleet --json` groups by project and carries an `unavailable` array rather than flattening:

```json
{
  "projects": [ { "project": "edda", "tasks": [ … ] } ],
  "unavailable": [ { "project": "dazun", "reason": "repo not on this machine (…)" } ]
}
```

A flat array would silently merge unrelated tasks under one id, and a consumer could not tell *"nothing ready in dazun"* from *"dazun could not be read"* — the second is the silent-empty failure GH-407 exists to remove. It is deliberately a different shape from the local `[TaskView]`, because it answers a different question; conflating them is the same trap.

## How

- `do_list_fleet` takes the **scope as a parameter** rather than looking it up. `fleet_scope` reads the global registry — process-wide state — so an injected scope is what lets the tests run against temp workspaces without fighting their neighbours. Same inversion as `ledger_root_for`'s lookup and `sync`'s `events_after`.
- Reuses `fleet::fan_out` from #425, so absent repos and per-project failures are already accounted for as misses.
- `task_row` is extracted and shared by both boards so they cannot drift into two dialects of the same list.
- No `--project` flag exists on `task list`, so the `--project`/`--fleet` conflict fixed in #426 has no analogue here.

## Tests

Three new, each written RED first:

- `fleet_list_tags_each_task_with_the_project_it_came_from` — asserts both tasks are `#1` in their own repo, so the tag is the only thing telling them apart.
- `fleet_list_applies_the_status_filter_in_each_project` — a filter must reach every project, not just the one being stood in.
- `fleet_list_reports_an_unreadable_project_instead_of_dropping_it` — "did not look" must not render as "nothing ready there".

`203 passed; 0 failed`. clippy 0 warnings at `--all-targets` (counted, not inferred from empty output). fmt clean.

Live verification on the real fleet to follow as a comment.

Closes part of #407.

### DIFF
--- a/crates/edda-cli/src/cmd_ask.rs
+++ b/crates/edda-cli/src/cmd_ask.rs
@@ -108,8 +108,11 @@ fn execute_fleet(repo_root: &Path, q: &str, opts: &AskOptions, json: bool) -> an
     // look", which is the failure this whole verb exists to remove.
     crate::fleet::print_misses(&misses);
 
-    if answered == 0 && misses.is_empty() {
-        println!("No results across {} project(s) for: {q}", scope.len());
+    if answered == 0 {
+        println!(
+            "{}",
+            crate::fleet::empty_summary("results", &format!(" for: {q}"), scope.len(), &misses)
+        );
     }
 
     Ok(())
--- a/crates/edda-cli/src/cmd_ask.rs
+++ b/crates/edda-cli/src/cmd_ask.rs
@@ -83,13 +83,11 @@ fn execute_fleet(repo_root: &Path, q: &str, opts: &AskOptions, json: bool) -> an
     });
 
     if json {
-        let payload = serde_json::json!({
-            "projects": hits.iter().map(|h| serde_json::json!({
-                "project": h.project,
-                "result": h.item,
-            })).collect::<Vec<_>>(),
-            "unavailable": crate::fleet::misses_json(&misses),
-        });
+        let projects = hits
+            .iter()
+            .map(|h| serde_json::json!({ "project": h.project, "result": h.item }))
+            .collect();
+        let payload = crate::fleet::json_envelope(projects, &misses);
         println!("{}", serde_json::to_string_pretty(&payload)?);
         return Ok(());
     }
--- a/crates/edda-cli/src/cmd_ask.rs
+++ b/crates/edda-cli/src/cmd_ask.rs
@@ -84,14 +84,11 @@ fn execute_fleet(repo_root: &Path, q: &str, opts: &AskOptions, json: bool) -> an
 
     if json {
         let payload = serde_json::json!({
-            "fleet": hits.iter().map(|h| serde_json::json!({
+            "projects": hits.iter().map(|h| serde_json::json!({
                 "project": h.project,
                 "result": h.item,
             })).collect::<Vec<_>>(),
-            "unreadable": misses.iter().map(|m| serde_json::json!({
-                "project": m.project,
-                "reason": m.reason,
-            })).collect::<Vec<_>>(),
+            "unavailable": crate::fleet::misses_json(&misses),
         });
         println!("{}", serde_json::to_string_pretty(&payload)?);
         return Ok(());
@@ -111,9 +108,7 @@ fn execute_fleet(repo_root: &Path, q: &str, opts: &AskOptions, json: bool) -> an
     // Misses are printed as results, not swallowed: a fleet read that quietly
     // skipped a repo would answer "nothing there" when the truth is "did not
     // look", which is the failure this whole verb exists to remove.
-    for miss in &misses {
-        println!("  [{}] unreadable: {}", miss.project, miss.reason);
-    }
+    crate::fleet::print_misses(&misses);
 
     if answered == 0 && misses.is_empty() {
         println!("No results across {} project(s) for: {q}", scope.len());
--- a/crates/edda-cli/src/cmd_search.rs
+++ b/crates/edda-cli/src/cmd_search.rs
@@ -156,7 +156,7 @@ fn query_fleet(
     });
 
     let mut total = 0;
-    for (project, results) in group_by_project(&hits) {
+    for (project, results) in crate::fleet::group_by_project(&hits) {
         if results.is_empty() {
             continue;
         }
@@ -178,9 +178,7 @@ fn query_fleet(
 
     // Per-project failures are results, never omissions: "did not look" must not
     // render as "nothing there".
-    for miss in &misses {
-        println!("  [{}] {}", miss.project, miss.reason);
-    }
+    crate::fleet::print_misses(&misses);
 
     if total == 0 && misses.is_empty() {
         println!(
@@ -190,21 +188,6 @@ fn query_fleet(
     }
 
     Ok(())
-}
-
-/// Collect a fan-out's hits back into per-project groups, preserving the order
-/// projects were visited in.
-fn group_by_project(
-    hits: &[crate::fleet::FleetHit<search::SearchResult>],
-) -> Vec<(String, Vec<&search::SearchResult>)> {
-    let mut out: Vec<(String, Vec<&search::SearchResult>)> = Vec::new();
-    for hit in hits {
-        match out.iter_mut().find(|(p, _)| *p == hit.project) {
-            Some((_, items)) => items.push(&hit.item),
-            None => out.push((hit.project.clone(), vec![&hit.item])),
-        }
-    }
-    out
 }
 
 /// Execute `edda search <query>` — full-text search over the Tantivy index.
--- a/crates/edda-cli/src/cmd_search.rs
+++ b/crates/edda-cli/src/cmd_search.rs
@@ -180,10 +180,15 @@ fn query_fleet(
     // render as "nothing there".
     crate::fleet::print_misses(&misses);
 
-    if total == 0 && misses.is_empty() {
+    if total == 0 {
         println!(
-            "No results across {} project(s) for: {query_str}",
-            scope.len()
+            "{}",
+            crate::fleet::empty_summary(
+                "results",
+                &format!(" for: {query_str}"),
+                scope.len(),
+                &misses
+            )
         );
     }
 
--- a/crates/edda-cli/src/cmd_task.rs
+++ b/crates/edda-cli/src/cmd_task.rs
@@ -76,6 +76,9 @@ pub enum TaskCmd {
         /// Output as JSON
         #[arg(long)]
         json: bool,
+        /// List tasks from every project in the fleet, not just this workspace
+        #[arg(long)]
+        fleet: bool,
     },
     /// Show one task in full (user verb)
     Show {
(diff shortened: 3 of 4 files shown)


fagemx:
## Code Review: PR #427 (Round 1)

### Summary

The verb itself is right, and the project tag is correctly identified as load-bearing rather than decorative. Both findings are the same defect seen twice: `fleet.rs` says it "owns the loop, the tagging, and the failure accounting", but each verb has been quietly keeping its own copy — and the copies have started to disagree.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — `group_by_project` is duplicated byte-for-byte, and `log --fleet` would make it three.**

`cmd_search.rs:197` and `cmd_task.rs:385` are identical but for the type:

```rust
let mut out: Vec<(String, Vec<&T>)> = Vec::new();
for hit in hits {
    match out.iter_mut().find(|(p, _)| *p == hit.project) {
        Some((_, items)) => items.push(&hit.item),
        None => out.push((hit.project.clone(), vec![&hit.item])),
    }
}
```

Nothing in it is verb-specific — it is `fan_out`'s output reshaped, which makes it `fleet.rs`'s job by that module's own description. One generic `group_by_project<T>` retires both copies and pre-empts the third.

**P1-2 — the three fleet verbs have drifted into three dialects of "this project did not answer".**

| | `ask --fleet` | `search --fleet` | `task list --fleet` |
|---|---|---|---|
| text | `[p] unreadable: {reason}` | `[p] {reason}` | `[p] {reason}` |
| JSON hits key | `fleet` | — | `projects` |
| JSON miss key | `unreadable` | — | `unavailable` |

Two problems, not one aesthetic one:

1. **`ask`'s `unreadable:` prefix is factually wrong for the most common miss.** `fan_out`'s dominant reason is `repo not on this machine (…)` — that repo is absent, not unreadable. The prefix contradicts the sentence it is prefixing.
2. **A script crossing verbs must learn each spelling.** `unreadable` vs `unavailable` for the same array is exactly the kind of thing that is free to fix now and expensive once three verbs ship it.

`fan_out` already returns complete reason phrases, so `[project] {reason}` needs no prefix. `unavailable` is the word that covers both misses `fan_out` actually produces (absent **and** errored); `unreadable` only covers one.

### Fix

Move all three into `fleet.rs`, which already claims the responsibility:

- `group_by_project<T>` — one generic, retires both copies
- `print_misses` — one text form, no wrong prefix
- `misses_json` — one key

This touches `ask` (merged in #425) and `search` (merged in #426). That is deliberate and in scope: the divergence was introduced by me across those two PRs and this one, all three are still fresh, and `log --fleet` is the next increment. Unifying at two copies is cheaper than at four, and no consumer exists outside this repo.

### Testing Review

#### Coverage

Three new tests, each written RED first — `do_list_fleet` genuinely absent, not a typo. The tagging test earns its place by asserting both tasks are `#1` in their own repo, which is the property that makes the tag necessary rather than ornamental.

Injecting the scope instead of calling `fleet_scope` is the right seam: the registry is process-wide state, and `EDDA_STORE_ROOT` isolation is all-or-nothing under cargo's thread-parallel tests.

---

fagemx:
## Code Review: PR #427 (Round 2)

### Summary

Round 1's two findings are fixed: one `group_by_project<T>`, one miss vocabulary, and `ask`'s wrong `unreadable:` prefix is gone. Verified by grep — exactly one definition of each survives.

But the fix itself has a defect, and it is the same species as the one it was fixing.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-3 — `misses_json` documents a guarantee it does not provide.**

`fleet.rs:102` says the misses are rendered *"under the one key every fleet verb uses"*. They are not. The helper returns a bare `Vec<Value>`; the key is typed by hand at each call site:

```rust
// cmd_ask.rs:91
"unavailable": crate::fleet::misses_json(&misses),
// cmd_task.rs:342
"unavailable": crate::fleet::misses_json(&misses),
```

Nothing stops `log --fleet` from writing `"unreachable":` next week — which is precisely the divergence P1-2 existed to end. The helper deduplicated the *mapping* and left the part that actually diverged in the callers' hands, then claimed credit for it in a doc comment. A comment asserting an invariant the code does not hold is worse than no comment: the next reader trusts it.

Both call sites already build the same envelope — `{projects: [...], unavailable: [...]}`, differing only in the per-project inner shape. So the envelope is the honest seam:

```rust
pub fn json_envelope(projects: Vec<serde_json::Value>, misses: &[FleetMiss]) -> serde_json::Value
```

That fixes both keys in one place, is shorter at both call sites than what it replaces, and makes the doc true.

### Testing Review

#### Coverage

`grouping_keeps_each_project_together_in_the_order_visited` uses non-adjacent hits (`edda`, `dazun`, `edda`), so it would catch a rewrite that only coalesced neighbours — the bug a naive fold would introduce.

`a_miss_renders_as_project_and_reason_with_nothing_invented` pins the absent-repo reason verbatim, which is the case `ask`'s old prefix contradicted.

Gap, once the envelope lands: nothing asserts the key name itself. The envelope test should read `envelope["unavailable"]`, so a rename cannot pass silently.

#### Testing Verdict: Adequate, pending the envelope test

---

fagemx:
## Code Review: PR #427 (Round 3)

### Summary

P1-3 is fixed — `json_envelope` now owns both keys and the test reads them, so a rename cannot pass quietly. Live-verified below.

Round 3 found a P1 by running the verb instead of reading it, and it is not confined to this PR: **`search --fleet`, merged an hour ago in #426, has it too.**

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-4 — the empty summary is suppressed exactly when it matters most.**

All three fleet verbs guard the summary with `&& misses.is_empty()`:

```rust
if total == 0 && misses.is_empty() {
    println!("No results across {} project(s) for: {query_str}", scope.len());
}
```

So when some project *did* fail, the summary vanishes. Measured on the real fleet, `search query "zzz-no-such-term-zzz-9999" --fleet` prints **one line, total**:

```
  [yushan] index not built — run `edda search index --project 97f7c3be…`
```

That is the entire output. Three other projects were searched and found nothing, and the operator is told none of it. They cannot distinguish "the other three had no results" from "only yushan was looked at" — which is the silent-empty failure this issue exists to remove, reintroduced by the guard meant to serve honesty.

The guard has the logic backwards. `misses.is_empty()` is the case where a bare summary is *safe*; a non-empty `misses` is when the reader most needs to be told what **was** covered:

```rust
if total == 0 {
    let answered = scope.len() - misses.len();
    // …"no results in the N that answered", with the misses above accounting
    // for the rest.
}
```

Applies to `search` (#426, merged), `task list` (this PR), and `ask` (#425, merged) — same shape in all three, so the fix lands in all three.

### Live Verification

`task list --fleet --status ready`, real fleet, from the edda repo:

```
  #2 [ready] search CJK 分詞:index/query 兩端對稱,>2 字中文查詢可中(P0)
  #3 [ready] search 自動索引:SessionEnd 增量 reindex + build-if-missing + 水位誠實(P0-)
  #4 [ready] 檢索覆蓋與 UX:receipt 進 FTS/ask + snippet + ask 排序(P1)
  #5 [ready] ask stale-code hint 跨 repo 絕對路徑誤報(P2)
  #7 [ready] fleet 讀取聯邦:--fleet 扇出四讀動詞 + 空結果誠實(P1)
```

0.468s warm. `--json` emits `"projects"` and `"unavailable"` — both keys present, envelope holding.

Scope note: this reads **4** projects, not 16, because `edda` belongs to the `fleet` group and `fleet_scope` prefers the group over the whole registry. That is the documented intent, verified against the registry directly (16 entries, 4 grouped `fleet`, 12 ungrouped) — worth stating because "across 4 project(s)" looks like under-coverage until you know why.

### Testing Review

#### Coverage

The three `do_list_fleet` tests hold. The envelope test now pins the key names and asserts the retired spellings (`unreadable`, `fleet`) cannot return.

**P1-4 is a coverage gap as much as a logic one**: every existing test drives the path where `misses` is empty or where hits exist. Nothing exercised "nothing found *and* something failed", which is why three verbs shipped with it. The fix needs a test at that intersection.

#### Testing Verdict: Inadequate — the failing combination is untested in all three verbs

---
*Round 3 of automated review-fix loop*

fagemx:
## Code Review: PR #427 (Round 4)

### Summary

P1-4 is fixed in all three verbs, and the fix is visible where the bug was. Same query, same fleet, before and after:

```
# before — this was the entire output
  [yushan] index not built — run `edda search index --project 97f7c3be…`

# after
  [yushan] index not built — run `edda search index --project 97f7c3be…`
No results in the 3 project(s) that answered for: zzz-no-such-term-zzz-9999; 1 could not be read (above)
```

The reader is now told what was covered *and* what was not. `scope.len()` no longer appears anywhere as a coverage claim — it is only passed into the helper that subtracts the misses, so the arithmetic that caused this cannot be re-done by hand in the next verb.

The no-miss path still reads plainly: `No tasks on the rail across 4 project(s) (assignee: nobody-9999).`

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

None. P1-1 through P1-4 resolved.

### What this PR settled

`fleet.rs` now holds the five things every fleet verb needs — `fan_out`, `group_by_project`, `print_misses`, `empty_summary`, `json_envelope`. That is not tidying for its own sake: three of the four findings in this review were verbs quietly keeping private copies of exactly these, and the copies disagreeing. `log --fleet` inherits all five and cannot re-introduce any of them.

Two of the four defects were in already-merged code (#425, #426) and were only found by running the verb against the real fleet rather than reading the diff. Worth recording: the diff looked fine every time.

### Testing Review

#### Coverage

`206 passed; 0 failed`. clippy **0** warnings at `--all-targets` (counted, not inferred from empty output). fmt clean.

Each new test was written RED first and failed for the right reason — missing function, not a typo. The two that matter most:

- `an_empty_fleet_read_never_counts_projects_it_could_not_reach` sits on the intersection every prior test missed — nothing found **and** something failed — which is why three verbs shipped the same backwards guard. It asserts the negative (`!contains("across 4")`), so a regression that re-claims full coverage fails loudly.
- `the_json_envelope_names_what_answered_and_what_did_not` pins the key names and asserts the retired spellings cannot return.

Remaining gap, P2, not blocking: `print_misses` has no test — its format is pinned only by the `json_envelope` test's verbatim reason. It is one `println!`, and stdout capture would cost more than it protects.

### Verdict: LGTM

The board says what it looked at, and what it couldn't.

---
*Round 4 of automated review-fix loop — approved*

PR #428, opened 2026-07-16
feat(fleet): log --fleet — every project's ledger, tagged (GH-407)

fagemx:
## What

`edda log --fleet` — the fourth and last read verb to federate. GH-407.

```
edda log --fleet --type note --limit 5
```

## Why it is small

It inherits all five shared pieces from `fleet.rs` (`fan_out`, `group_by_project`, `print_misses`, `empty_summary`, `json_envelope`), so it cannot re-introduce any of the four defects the earlier verbs shipped and the review loop caught: untagged rows, dropped per-project failures, a third spelling of "unavailable", or an empty summary that counts projects it never reached.

That is what the #427 consolidation was for. This verb is the test of it.

## Two decisions worth stating

**`--limit` caps each project's answer, not the merged total.** A log is a timeline, so merging by timestamp and applying a global limit was tempting — and I started there. But `search query --fleet` already applies `--limit` per project, and it has a good reason to: Tantivy scores are per-index and not comparable across projects, so "top N from each" is the only ranking a fan-out can honestly offer. A flag that means one thing in one verb and something else in the next is worse than either meaning alone.

Measured on the real fleet, `search query "edda" --fleet --limit 3` returns **7** hits while the flag said "Maximum results". That was true before this PR and is now said out loud: the help text reads *"per project when --fleet"*.

**`--fleet --json` emits the envelope, not `log`'s bare JSONL.** The projects that could not be read have to land somewhere a script will look, and a trailing object appended to a stream of lines is exactly what a consumer skips. Same reasoning as `task list --fleet --json` in #427, same shape.

## How

`execute` grew two seams it never had:

- `collect_matching(params) -> Result<Vec<Event>>` — one workspace's answer, without the printing
- `collect_fleet(scope, params)` — the same question asked of every project, scope injected so the tests never touch the global registry

`LogParams` derives `Clone, Copy` (all fields already were), so the fan-out closure rebuilds it with a different `repo_root` and nothing else.

## Tests

Three, each written RED first — `collect_fleet` genuinely absent, not a typo:

- `fleet_log_tags_each_event_with_the_project_it_came_from`
- `fleet_log_applies_the_limit_to_each_project_separately` — pins the semantics argued above, so a later "helpful" global limit fails loudly
- `fleet_log_reports_an_unreadable_project_instead_of_dropping_it`

`209 passed; 0 failed`. clippy **0** at `--all-targets` (counted, not inferred from empty output). fmt clean.

Live verification to follow as a comment.

## Not in this PR

[...]

PR #429, opened 2026-07-16
feat(fleet): a local miss asks the fleet before reporting absence (GH-407)

### LINKED ISSUE #407: fleet reads: --fleet fan-out on ask/search/log/task-list + empty-result honesty (P1)

fagemx:
## What

Acceptance 4, and the reason #407 exists: **a local miss asks the fleet before reporting absence.**

Live, from the edda repo, for a term that lives in sibling repos:

```
$ edda log --keyword "desk-operator"
No events match the filter.
  9 event(s) in [AI Delivery Foundry], 1 event(s) in [dazun] — rerun with --fleet

$ edda search query "desk-operator"
No results found for: desk-operator
  8 result(s) in [AI Delivery Foundry], 1 result(s) in [dazun]; [yushan] could not be checked — rerun with --fleet
```

Every read verb sees one workspace, so `No results` has always meant *"not in this repo"* while reading as *"nowhere"*. The issue's own framing: knowledge really is split across repos, so that reading was routinely wrong, and *"the split-brain is invisible because a local miss renders exactly like true absence."*

A bare miss may now only stay bare once the fleet has been asked and had nothing to add. `elsewhere_hint` returns `None` in exactly that case — verified: `log --keyword "zzz-nowhere-in-any-repo-9999"` prints the bare message with no hint, because the hint corrects lies rather than decorating facts.

## The `ask` story, corrected

> **An earlier version of this description claimed `ask` "printed nothing at all" on an empty result. That was wrong, and I had not checked it.** It prints `No results found.`, from `edda-ask/src/lib.rs:830`. The correction matters because everything I first built for `ask` was gated on that false belief.

`format_human` renders its own message for an empty result, so `body.trim().is_empty()` is a question that is **never answered yes**. One wrong assumption, three dead branches, across three PRs — each read correctly, none ever ran:

| branch | shipped in | symptom |
|---|---|---|
| local fleet hint | this PR | `ask` never mentioned the fleet, while `[AI Delivery Foundry]` held 8 hits for the same term |
| `--fleet` empty-project skip | #425 (merged) | every silent project got a header and a line saying it had nothing |
| `--fleet` empty summary | #427 (merged) | `answered` counted the silent projects, so `answered == 0` was unreachable |

All three now use `hit_count`, which the probe needed anyway. `ask` has **one** definition of empty, shared by both ends of the probe.

## Two silences are deliberate, and tested

- **The home project is skipped.** It is the one that just missed; re-probing it would report its own silence back as news.
- **A repo absent from this machine is left out.** It cannot be probed, `rerun with --fleet` would not make it appear, and `--fleet` reports it anyway. A project that *exists* but fails to answer is named — that one might have hits once its index is built. `a_repo_absent_from_this_machine_is_left_out_of_the_hint` uses `panic!` as the probe body, so it asserts the absent repo is never *reached*, not merely absent from the output.

## Cost

`search`'s probe **never builds an index**. A ~25s cold build (GH-403) triggered by someone else's miss would be a trap; it reports `could not be checked` instead — true and free. All probes are read-only and only run on a miss.

The counts are capped by `--limit`, which is correct here rather than an undercount: `--limit` is *per project* under `--fleet` (#428), so the hint reports exactly what the command it recommends will show.

## How

One helper in `fleet.rs`, scope injected so tests never touch the global registry:

```rust
pub fn elsewhere_hint<F>(scope: &[ProjectEntry], home_project_id: &str, what: &str, count_in: F) -> Option<String>
```

Four call sites: `cmd_search.rs`, `cmd_log.rs`, `cmd_task.rs`, `cmd_ask.rs`. No hint can reach a `--json` stream — `ask` and `task list` return on `json` before the empty branch, `log` guards with `!params.json`, and `search` has no `--json`.

## Tests

`216 passed; 0 failed`. clippy **0** at `--all-targets` (counted, not inferred from empty output). fmt clean.

Seven new, each RED first. The two that exist because of what went wrong:

- `a_hit_of_any_kind_counts_not_just_decisions_and_timeline` — `AskResult` has six hit collections; the probe counted two, so a commit-only or transcript-only hit probed as absence.
- `an_empty_result_still_renders_text_so_emptiness_must_be_counted` — pins the contract all three dead branches depended on and none checked: an empty result counts **zero**, and its render is **not blank**.

Every finding in this review came from running the verb. The diff read fine three times, and 215 tests passed throughout, because every test injected its own notion of a hit and not one of them ever called `format_human`.

Closes #407.

### DIFF
--- a/crates/edda-cli/src/cmd_ask.rs
+++ b/crates/edda-cli/src/cmd_ask.rs
@@ -50,11 +50,40 @@ pub fn execute(
 
     if json {
         println!("{}", serde_json::to_string_pretty(&result)?);
-    } else {
-        print!("{}", format_human(&result));
+        return Ok(());
     }
 
+    let body = format_human(&result);
+    if body.trim().is_empty() {
+        // Printing nothing at all was the worst form of the silent miss: it did
+        // not even claim absence, it just ended. A workspace with no decisions
+        // yet — a sibling repo, a fresh clone — is exactly where the fleet's
+        // answer lives, and exactly where the old behaviour said least.
+        println!("No results for: {q}");
+        if let Some(hint) = fleet_hint_for_ask(repo_root, q, &opts) {
+            println!("{hint}");
+        }
+        return Ok(());
+    }
+    print!("{body}");
+
     Ok(())
+}
+
+/// Ask the rest of the fleet whether a local miss is really absence (GH-407,
+/// acceptance 4).
+fn fleet_hint_for_ask(repo_root: &Path, q: &str, opts: &AskOptions) -> Option<String> {
+    if q.trim().is_empty() {
+        return None; // no question was asked; there is nothing to look for elsewhere
+    }
+    let scope = edda_store::registry::fleet_scope(repo_root);
+    let home = edda_store::project_id(repo_root);
+    crate::fleet::elsewhere_hint(&scope, &home, "result", |entry| {
+        let root = Path::new(&entry.path);
+        let ledger = Ledger::open(root)?;
+        let result = ask(&ledger, q, opts, None)?;
+        Ok(result.decisions.len() + result.timeline.len())
+    })
 }
 
 /// `edda ask --fleet` — the same question, asked of every project in scope.
--- a/crates/edda-cli/src/cmd_ask.rs
+++ b/crates/edda-cli/src/cmd_ask.rs
@@ -53,19 +53,15 @@ pub fn execute(
         return Ok(());
     }
 
-    let body = format_human(&result);
-    if body.trim().is_empty() {
-        // Printing nothing at all was the worst form of the silent miss: it did
-        // not even claim absence, it just ended. A workspace with no decisions
-        // yet — a sibling repo, a fresh clone — is exactly where the fleet's
-        // answer lives, and exactly where the old behaviour said least.
-        println!("No results for: {q}");
+    // Emptiness is counted, not detected from the rendering: `format_human`
+    // prints its own "No results found." for an empty result, so asking whether
+    // the render is blank is a question that is never answered yes.
+    print!("{}", format_human(&result));
+    if hit_count(&result) == 0 {
         if let Some(hint) = fleet_hint_for_ask(repo_root, q, &opts) {
             println!("{hint}");
         }
-        return Ok(());
     }
-    print!("{body}");
 
     Ok(())
 }
@@ -141,15 +137,18 @@ fn execute_fleet(repo_root: &Path, q: &str, opts: &AskOptions, json: bool) -> an
         return Ok(());
     }
 
+    // Counted, not detected from the rendering: `format_human` prints its own
+    // "No results found." for an empty result, so a blank-body test never fires
+    // and every project — including the silent ones — got a section header and a
+    // line saying it had nothing.
     let mut answered = 0;
     for hit in &hits {
-        let body = format_human(&hit.item);
-        if body.trim().is_empty() {
+        if hit_count(&hit.item) == 0 {
             continue;
         }
         answered += 1;
         println!("── [{}] ──────────────────────────", hit.project);
-        print!("{body}");
+        print!("{}", format_human(&hit.item));
     }
 
     // Misses are printed as results, not swallowed: a fleet read that quietly
@@ -259,6 +258,36 @@ mod tests {
             hit_count(&r),
             1,
             "a note-only hit must not probe as absence"
+        );
+    }
+
+    /// The contract three branches were built on, and none of them checked.
+    ///
+    /// `format_human` prints its own "No results found." for an empty result, so
+    /// `body.trim().is_empty()` is a question that is never answered yes. That
+    /// one wrong assumption shipped three dead branches across three PRs — the
+    /// local hint, `--fleet`'s empty-project skip, and `--fleet`'s empty summary
+    /// — each of which read correctly and never ran. Emptiness is counted here,
+    /// never inferred from the rendering.
+    #[test]
+    fn an_empty_result_still_renders_text_so_emptiness_must_be_counted() {
+        let empty = edda_ask::AskResult {
+            query: "q".to_string(),
+            input_type: "keyword".to_string(),
+            decisions: Vec::new(),
+            timeline: Vec::new(),
+            related_commits: Vec::new(),
+            related_notes: Vec::new(),
+            conversations: Vec::new(),
+            dependents: Vec::new(),
+            override_risk: None,
+        };
+
+        assert_eq!(hit_count(&empty), 0, "nothing was found");
+        assert!(
+            !format_human(&empty).trim().is_empty(),
+            "the render is NOT blank — it says 'No results found.' — so any \
+             branch gated on a blank body is dead code"
         );
     }
 }
--- a/crates/edda-cli/src/cmd_ask.rs
+++ b/crates/edda-cli/src/cmd_ask.rs
@@ -70,8 +70,27 @@ pub fn execute(
     Ok(())
 }
 
+/// Everything `format_human` would render as a hit.
+///
+/// The count has to span every collection, not the obvious two: a project whose
+/// only hit is a commit or a transcript turn must not probe as empty, or the
+/// hint reports absence where `--fleet` finds answers — the very failure this
+/// hint exists to remove, one level down.
+fn hit_count(r: &edda_ask::AskResult) -> usize {
+    r.decisions.len()
+        + r.timeline.len()
+        + r.related_commits.len()
+        + r.related_notes.len()
+        + r.conversations.len()
+        + r.dependents.len()
+}
+
 /// Ask the rest of the fleet whether a local miss is really absence (GH-407,
 /// acceptance 4).
+///
+/// Probes exactly as `execute_fleet` reads, transcript callback included. A line
+/// that says "rerun with --fleet" is a promise about what that command will
+/// show, so anything the probe declines to look at is a promise it cannot keep.
 fn fleet_hint_for_ask(repo_root: &Path, q: &str, opts: &AskOptions) -> Option<String> {
     if q.trim().is_empty() {
         return None; // no question was asked; there is nothing to look for elsewhere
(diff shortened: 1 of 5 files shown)


fagemx:
## Code Review: PR #429 (Round 1)

### Summary

The helper and its silences are right, and `ask`'s bare-silence fix is the most valuable line in the PR. One finding — and it is the same bug this PR exists to fix, reintroduced inside the fix.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — `ask`'s probe counts two of six hit collections, so it can report absence where the fleet has hits.**

`AskResult` carries six:

```rust
pub decisions: Vec<DecisionHit>,
pub timeline: Vec<DecisionHit>,
pub related_commits: Vec<CommitHit>,
pub related_notes: Vec<NoteHit>,
pub conversations: Vec<ConversationHit>,
pub dependents: Vec<DependentHit>,
```

`fleet_hint_for_ask` counts `decisions.len() + timeline.len()`. A project whose hits are commits, notes, conversations, or dependents therefore counts as **0**, `elsewhere_hint` returns `None`, the bare `No results for: q` stands unqualified — and `ask --fleet` would have shown those hits all along.

That is precisely the failure this PR is written to remove: a miss rendering as absence. It is now one level down, inside the probe.

Two things make it worse than an off-by-one:

1. **The two ends disagree about what "empty" means.** Local emptiness is decided by `format_human(&result).trim().is_empty()` — the full render, all six collections. The probe uses two. So the verb can call the local result non-empty on evidence the probe would not even look at, and vice versa.
2. **The probe passes `None` for the transcript callback** while `execute_fleet` passes a real one. So `conversations` is structurally always 0 in the probe, and `--fleet` finds transcript hits the hint promises do not exist. Symmetry with `--fleet` is the whole contract of a line that says *"rerun with --fleet"*.

Fix: count every collection `format_human` would render, and build the transcript callback as `execute_fleet` does. The callback's staleness notice goes to `eprintln!`, so it does not pollute stdout, and *"[dazun] search index is out of date"* on a local miss is honest rather than noisy — it is the same class of statement as the rest of this PR.

**The counts are capped by `--limit`, and that is correct here.** `search`'s probe passes `limit`, so a sibling with 37 hits and `--limit 20` reports "20 result(s)". That is not an undercount in context: `--limit` is *per project* under `--fleet` (#428), so the hint reports exactly what `rerun with --fleet` will show. The hint predicts the command it recommends. Worth stating because it looks like a bug until you line the two up.

**`search`'s probe never builds.** A ~25s cold build (GH-403) triggered by someone else's miss would be a trap; `could not be checked` is true and free.

### Testing Review

#### Coverage

Five tests, each RED first — `elsewhere_hint` genuinely absent, not a typo.

`a_repo_absent_from_this_machine_is_left_out_of_the_hint` uses `panic!` as the probe body, so it asserts the absent repo is never *reached* rather than merely absent from the output — a stronger claim, and the one that matters.

**P1-1 is a coverage gap first.** Every `elsewhere_hint` test injects its own counting closure, so the helper is well covered while all four real probes — the code that decides what "a hit" means — are untested. That is exactly where the defect is. A test that a project with only a `related_notes` hit produces a hint would have caught it.

`214 passed; 0 failed`. clippy **0** at `--all-targets`. fmt clean.

#### Testing Verdict: Inadequate — the helper is tested, the probes are not, and the bug is in a probe

---

fagemx:
## Code Review: PR #429 (Round 2)

### Summary

P1-1 fixed — `hit_count` spans all six collections and the probe builds the transcript callback. Three of the four verbs are live-verified and correct:

```
$ edda log --keyword "desk-operator"
No events match the filter.
  9 event(s) in [AI Delivery Foundry], 1 event(s) in [dazun] — rerun with --fleet

$ edda search query "desk-operator"
No results found for: desk-operator
  8 result(s) in [AI Delivery Foundry], 1 result(s) in [dazun]; [yushan] could not be checked — rerun with --fleet
```

The `search` line is the design working exactly as intended: hits named, and the one project that *could not* be checked named separately rather than counted as empty.

The negative case holds too — `log --keyword "zzz-nowhere-in-any-repo-9999"` prints the bare message with **no** hint, because the fleet genuinely has nothing and the hint corrects lies rather than decorating facts.

Then there is `ask`.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-2 — `ask`'s hint is dead code. The feature does not exist for the verb that needed it most.**

```
$ edda ask "desk-operator"
No results found.
```

No hint — while `[AI Delivery Foundry]` holds 8 hits for that exact term.

The cause is one wrong assumption. `format_human` does **not** return an empty string when there is nothing:

```rust
// edda-ask/src/lib.rs:830
out.push_str("No results found.\n");
```

So `body.trim().is_empty()` is never true, the branch never runs, and every claim this PR made about `ask` — the bare-silence fix, the hint, the "worst of the four" framing — was written against behaviour that does not occur. The PR body's assertion that `ask` "printed nothing at all" was **wrong**: it printed `No results found.`, from inside `edda-ask`, which is why nothing I added downstream ever fired.

**P1-3 — the same assumption is already merged in #425, and it is live.**

`execute_fleet` skips empty projects with the identical predicate (`cmd_ask.rs:147`). It never fires either:

```
$ edda ask "desk-operator" --fleet
── [dazun] ──────────────────────────
No results found.
── [yushan] ──────────────────────────
No results found.
```

Empty projects render as a header plus `No results found.` — exactly the noise the skip exists to suppress. And because `answered` only increments for non-empty bodies, `answered == 0` is unreachable, which means **the `empty_summary` fix from #427 is dead for `ask`** as well. One bad predicate, three dead branches, across three PRs.

### Fix

`hit_count(&result) == 0` is the honest predicate and this PR already introduces it. Use it at all three sites, so `ask` has exactly one definition of "empty" — the same one on both ends of the probe.

### What this says about the method

Every finding in rounds 1 and 2 came from running the verb, not reading the diff. The diff looked right three times: in #425 when the skip was written, in #427 when the summary was "fixed" for a path that could not reach it, and in this PR's own body, which confidently described behaviour I had not checked. The tests passed throughout — 215 of them — because every one of them injected its own notion of a hit and none ever called `format_human`.

### Testing Review

#### Coverage

`a_hit_of_any_kind_counts_not_just_decisions_and_timeline` is good and catches P1-1's class.

The gap that produced P1-2 and P1-3 is unchanged: **nothing asserts what `ask` prints when it finds nothing.** That is a claim about `format_human`'s contract, and it is the claim all three dead branches depend on. It needs a test that pins `hit_count(&empty) == 0` *and* that the render is non-empty — the two facts whose confusion caused this.

`215 passed; 0 failed`. clippy **0**. fmt clean.

#### Testing Verdict: Inadequate — the predicate three branches depend on is untested

---

fagemx:
## Code Review: PR #429 (Round 3)

### Summary

P1-2 and P1-3 fixed. All three previously-dead branches now run, verified live rather than asserted.

### Live Verification

**The local hint fires** — the branch that never once executed:

```
$ edda ask "desk-operator"
No results found.
  23 result(s) in [AI Delivery Foundry], 3 result(s) in [dazun] — rerun with --fleet
```

**The `--fleet` skip fires** — empty projects are gone, which is what #425 intended and never did:

```
$ edda ask "desk-operator" --fleet
── [dazun] ──────────────────────────
```

Before: `[edda]` and `[yushan]` each got a header and a line saying they had nothing.

### A near-miss worth recording

The first run of this verification showed **no change at all** — identical output to the broken version. The binary's mtime (11:33:05) was newer than the fix's commit (11:27:21), which looked like proof it contained the fix. It did not.

`cargo` reads source at build **start** and writes the exe at the end. Two builds were queued; the one that finished at 11:33:05 had compiled pre-fix source at ~11:18. A newer mtime than the commit proves nothing.

Had I trusted that reading, the conclusion would have been "the fix doesn't work" — and the next move would have been rewriting code that was already correct. The only reliable proof is waiting for every `cargo`/`rustc` to exit, then re-running `cargo build --release` and seeing `Finished in 0.60s` — nothing to do — before touching the binary. That is what turned this round around.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

None. P1-1 through P1-3 resolved.

### What #407 closes with

Four read verbs federate; the fifth thing they share is honesty about what they did not see. Every acceptance criterion is met and live-verified:

1. tagged cross-project reads — `ask` / `search` / `task list` / `log`
2. `task list --fleet --status ready`, the morning board
3. an explicit per-project notice for a missing index
4. a local miss that names the projects holding the answer

The count moving from 8 to 23 between rounds is P1-1's fix visible: the probe now counts all six hit collections and searches transcripts, as `--fleet` always did.

### Testing Review

#### Coverage

`216 passed; 0 failed`. clippy **0** at `--all-targets` (counted, not inferred from empty output). fmt clean.

`an_empty_result_still_renders_text_so_emptiness_must_be_counted` is the test this PR exists to leave behind. It pins both halves of the confusion that produced three dead branches across three PRs — an empty result counts **zero**, and its render is **not blank** — so the next person to reach for `body.is_empty()` gets a red test instead of code that reads fine and never runs.

The honest summary of this review: **the diff read correctly every time it was wrong.** 215 tests passed while three branches were dead, because every test injected its own notion of a hit and none ever called `format_human`. Everything found here came from running the verb.

### Verdict: LGTM

---
*Round 3 of automated review-fix loop — approved*

Issue #408, opened 2026-07-14
pack: budget-capped Fleet section — sibling shared decisions + ready-task counts (P2)


fagemx:
## Problem

The SessionStart pack renders only the cwd project's ledger and coordination state. A session opening in one repo learns nothing about sibling repos' recent rulings or waiting work — cross-repo awareness currently depends on the agent thinking to run a fleet query, which the retrieval review showed never happens unprompted (pull demand ≈ zero; the push channel is what actually works).

## Fix

Add a compact, budget-capped **Fleet** section to the pack (and to the UserPromptSubmit workspace refresh where budget allows):

- recent binding decisions from sibling projects with scope `shared`/`global` (query-time view over sibling ledgers via the registry — fresher than sync-imported copies, and no duplication)
- one line per sibling with waiting work: `[edda] 5 ready task(s)` — derived from each sibling's task projection
- hard budget: the section competes inside the existing context budget and truncates first; solo projects (no registry siblings / no group) render nothing at all

Depends on the fan-out core from the `--fleet` read-verbs issue — this is its push-side consumer. Same per-project failure honesty: an unreachable sibling is one short line, not an omission.

## Acceptance

- Session in repo A shows repo B's shared-scope decision recorded today, tagged `[B]`
- Session in repo A shows `[B] N ready task(s)` when B's queue is non-empty
- Pack size stays within the existing budget; Fleet section truncates before core sections
- No Fleet section when there are no siblings

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #430, opened 2026-07-16
refactor(fleet): move the fan-out core to edda-store so the pack can use it

fagemx:
## What

Move the fan-out core out of the CLI binary so the SessionStart pack can use it. Pure move, no behaviour change. Groundwork for #408.

## Why it is necessary, not tidying

#408 adds a Fleet section to the pack and states it *"depends on the fan-out core from the `--fleet` read-verbs issue — this is its push-side consumer."*

It cannot depend on it. The core lives in `crates/edda-cli/src/fleet.rs`, and `edda-cli` is:

```toml
[[bin]]
# ...and no [lib]
```

Nothing else in the workspace can link a binary crate. The pack is rendered in `edda-bridge-claude`, a lib. So as it stands, #408's only options are to duplicate the fan-out or to not use it — and a second copy of this loop is exactly what four review rounds on #427 were spent removing.

## The split: mechanism vs sentence

**To `edda_store::fleet`** — beside the `registry` it already reads:

- `FleetHit<T>`, `FleetMiss`
- `fan_out` — the loop, the tagging, the absent-repo accounting
- `group_by_project`

No new dependencies: `edda-store` already had `anyhow`, `serde_json`, and `tempfile`.

**Stays in `edda-cli`** — how the CLI talks, not how a fan-out works:

- `print_misses`, `empty_summary`, `elsewhere_hint`, `json_envelope`

`elsewhere_hint` stays despite being fan-out-shaped: it composes an English sentence ending in `— rerun with --fleet`, which is advice about a CLI flag. The pack has no flag to recommend.

## Verification

**Test accounting is exact** — the property that matters for a move, since a silently dropped test is the failure mode:

| | before | after |
|---|---|---|
| `edda` bin | 216 | **211** |
| `edda-store` | — | **+5** |

The five that moved are `hits_carry_the_project_they_came_from`, `an_absent_repo_is_a_reported_miss_not_a_silence`, `a_failing_project_is_a_miss_and_the_rest_still_answer`, `an_empty_scope_yields_nothing_rather_than_erroring`, `grouping_keeps_each_project_together_in_the_order_visited`.

Workspace: **1972 passed, 0 failed**. clippy **0** at `--workspace --all-targets` (counted, not inferred from empty output). fmt clean.

> A `link.exe` 1104 appeared on one intermediate run and was **not** a real failure — a file lock from a concurrent release build. `cargo test -p edda-ask` alone: 43 passed. Noted rather than omitted, because a compile failure makes cargo abort the remaining crates and can silently deflate the total.

Part of #408.

### DIFF
--- a/crates/edda-cli/src/fleet.rs
+++ b/crates/edda-cli/src/fleet.rs
@@ -1,91 +1,17 @@
-//! Fan-out for fleet reads (GH-407).
+//! How the CLI renders a fleet read (GH-407).
 //!
-//! Truth stays home: nothing is centralised. A `--fleet` read visits each
-//! project in scope, runs the same query against that project's own ledger or
-//! index, and merges the answers — tagged with where each came from.
+//! The fan-out itself lives in `edda_store::fleet`: the SessionStart pack needs
+//! it too (GH-408), and the CLI is a binary crate, so nothing else could ever
+//! link it from here.
 //!
-//! The shape is shared, the query is not: `ask` reads decisions, `search` reads
-//! a Tantivy index, `log` and `task list` read the ledger. So this owns the loop,
-//! the tagging, and the failure accounting, and takes the per-project work as a
-//! closure — the same inversion `edda-search-fts::sync` uses for its events.
+//! What stays is the part that is genuinely the CLI's — the sentences. Every
+//! `--fleet` verb reports misses the same way, spells "unavailable" the same
+//! way, and counts an empty read the same way, because each of those diverged
+//! once already and the copies disagreed.
 
+pub use edda_store::fleet::{fan_out, group_by_project, FleetHit, FleetMiss};
 use edda_store::registry::ProjectEntry;
 use std::path::Path;
-
-/// One project's contribution to a fleet read.
-#[derive(Debug, Clone, PartialEq)]
-pub struct FleetHit<T> {
-    /// The project's registered name, rendered as `[name]` beside the hit.
-    pub project: String,
-    pub item: T,
-}
-
-/// A project that could not be read.
-///
-/// These are results, not omissions. A fleet read that quietly skipped a repo
-/// would answer "nothing there" when the truth is "did not look" — the exact
-/// silent-empty failure GH-407 exists to remove, so every miss carries the
-/// project it belongs to and the reason it failed.
-#[derive(Debug, Clone, PartialEq)]
-pub struct FleetMiss {
-    pub project: String,
-    pub reason: String,
-}
-
-/// Run `query` against every project in scope, tagging hits and collecting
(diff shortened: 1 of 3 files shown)


fagemx:
## Code Review: PR #430 (Round 1)

### Summary

A move PR's only real risk is that it is not actually a move. This one is, and it was checked rather than assumed.

### Verification of purity

Each relocated item diffed against `main`'s copy:

```
FleetHit:          逐字相同 (byte-identical)
fan_out:           逐字相同
group_by_project:  逐字相同
```

Each item that stayed, diffed the same way — `print_misses`, `empty_summary`, `elsewhere_hint`, `json_envelope`: **未動**, untouched.

Each moved test present exactly once, nowhere twice:

| test | edda-store | edda-cli |
|---|---|---|
| `hits_carry_the_project_they_came_from` | 1 | 0 |
| `an_absent_repo_is_a_reported_miss_not_a_silence` | 1 | 0 |
| `a_failing_project_is_a_miss_and_the_rest_still_answer` | 1 | 0 |
| `an_empty_scope_yields_nothing_rather_than_erroring` | 1 | 0 |
| `grouping_keeps_each_project_together_in_the_order_visited` | 1 | 0 |

That table is the point. A move's characteristic failure is a test that quietly stops running — present in neither place, or present in the old place where nothing calls it any more. Counts of 1/0 across the board, plus the bin dropping 216 → 211 and the store gaining exactly five, closes that gap arithmetically rather than by inspection.

Workspace: **1972 passed, 0 failed**. clippy **0** at `--workspace --all-targets`.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

None.

**Is `edda-store` the right home?** Its description is *"Per-user store with atomic writes"*, and a fan-out loop is not storage. But `registry` lives there, `fleet_scope` lives in `registry`, and `fan_out`'s whole input is `&[ProjectEntry]` — the thing this crate already owns. A dedicated `edda-fleet` crate for ~90 lines and one dependency would be ceremony around a cohesion that already exists.

**The `pub use` re-export** keeps `crate::fleet::fan_out` working, so no call site churned and the diff stays a move. It also means `edda-cli`'s `fleet` module remains the single import point for everything fleet-shaped in the CLI, which is the coherent reading rather than an accident.

**Why `elsewhere_hint` stayed** while being fan-out-shaped: it composes a sentence ending in `— rerun with --fleet`. That is advice about a CLI flag. The pack has no flag to recommend, so the line would be a lie there.

### Testing Review

#### Coverage

No new tests, correctly — a pure move should add none. The five that moved carry their assertions unchanged, verified by diff.

### Verdict: LGTM

The pack can now link the core it was specified to depend on.

---

PR #432, opened 2026-07-16
fix(conductor): tests must not file temp dirs in the operator's registry (GH-417)

### LINKED ISSUE #417: registry: 1628 of 1644 project entries are dead — tests register temp dirs and never clean up

fagemx:
## What

`cargo test -p edda-conductor` filed **13 dead entries in the operator's real project registry**, every run, on the machine of whoever ran the tests. GH-417.

```
cargo test -p edda-conductor   +13  ->  +0
cargo test --workspace         +13  ->  +0     (1980 passed, 0 failed)
```

## Why the trail went cold

The crate never calls `register_project`. It shells out:

```rust
// runner/edda.rs — ensure_init, called by runner/sequential.rs:57
let _ = Command::new("edda")     // the *installed* binary, from PATH
    .arg("init")
    .current_dir(cwd)
    .status();
```

`edda init` registers its cwd (`cmd_init.rs:112`). Under test that cwd is a fresh `tempfile::tempdir()`, no test sets `EDDA_STORE_ROOT`, the child inherits that absence — and the temp path lands in the real registry.

My earlier comment on #417 concluded the conductor could not be the source, because it "never calls `register_project`" and hand-reproducing the subprocess "registered nothing". Both observations were correct. The conclusion was not. (`ensure_init` returns early when `.edda` exists, so re-running it by hand in an initialised directory exercises the guard, not the path.)

## Evidence

The registry grew **16 → 55** during one working session, purely from running tests. Every added entry a dead tempdir:

```
total: 55   alive: 16   dead: 39
 39  C:\Users\synvoke\AppData\Local\Temp
     e.g. .tmplTU4Ei, .tmp1mwIWH, .tmpJGfIYH
```

Bisected by crate → **conductor +13**, everything else +0. By module → **runner +13**, everything else +0. Then the controlled pair, which is the actual proof:

| test | runs the runner? | delta |
|---|---|---|
| `cancellation_stops_runner` | yes | **+1** |
| `truncate_str_ascii` | no | **+0** |

## The fix

Isolation at the **seam**, not in the test bodies. There is one call site and thirteen callers today; guarding callers would be thirteen chances for the fourteenth to forget. `EDDA_STORE_ROOT` points the child at a throwaway store, so the whole path still runs — it just cannot reach the operator's registry.

Whole-process isolation is safe *here*, which it usually is not: the variable is process-wide and cargo runs tests in parallel threads, so this is normally all-or-nothing. It can be all — no test in this crate wants the real store.

The store is created once and leaked deliberately: it must outlive every test in the process, and with tests on parallel threads there is no later moment at which dropping it would be safe. It is an OS temp dir; the OS reclaims it.

## The test

`ensure_init_cannot_reach_the_operators_real_store` asserts the **guard**, not the absence. Reading the real registry to prove we did not touch it would depend on the developer's machine having one, and would pass for the wrong reason on a machine that has none.

It also cannot ask `store_root()` — that function *returns* `EDDA_STORE_ROOT` when set, so it would compare the temp store with itself and pass for the wrong reason again. It asserts the store lives under the OS temp dir instead.

## Operator's registry

Swept with the pre-existing `edda gc --global`: **95 → 16, all alive, zero dead**. The sweep in the issue title needed no code; the leak did.

Closes #417.

### DIFF
--- a/crates/edda-conductor/src/runner/edda.rs
+++ b/crates/edda-conductor/src/runner/edda.rs
@@ -9,10 +9,27 @@ use std::process::Command;
 
 /// Ensure `.edda/` ledger exists in the working directory.
 /// No-op if already initialized or if `edda` is not available.
+///
+/// This shells out to the *installed* `edda`, and `edda init` registers its cwd
+/// in the operator's global project registry. That is right in production and
+/// wrong under test: the runner's tests each hand this a fresh
+/// `tempfile::tempdir()`, so every test run filed another dead temp path in the
+/// real registry — 13 per `cargo test -p edda-conductor`, which is GH-417.
+///
+/// The tests are isolated at this seam rather than in each test body: there is
+/// one call site and thirteen callers today, so guarding the callers would be
+/// thirteen chances for the fourteenth to forget. `EDDA_STORE_ROOT` points the
+/// child at a throwaway store — the whole path still runs, it just cannot reach
+/// the operator's registry. Whole-process isolation is safe here, which it
+/// usually is not: the variable is process-wide, but no test in this crate wants
+/// the real store.
 pub fn ensure_init(cwd: &Path) {
     if cwd.join(".edda").exists() {
         return;
     }
+    #[cfg(test)]
+    let _isolation = tests::isolate_store_for_this_process();
+
     let _ = Command::new("edda")
         .arg("init")
         .current_dir(cwd)
@@ -97,6 +114,63 @@ pub fn record_phase_failed(cwd: &Path, phase_id: &str, error: &str) {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    /// Point every `edda` child this test binary spawns at a throwaway store.
+    ///
+    /// Returns a guard the caller holds for the duration of the spawn. The store
+    /// itself is created once and leaked: it must outlive every test in the
+    /// process, and tests run in parallel threads, so there is no later moment
+    /// at which dropping it would be safe. It is an OS temp dir — the OS
+    /// reclaims it.
+    ///
+    /// The lock is what makes this correct rather than merely usual:
+    /// `set_var` is process-wide, so a concurrent test that reads
+    /// `EDDA_STORE_ROOT` mid-write would see a torn value. Everything here wants
+    /// the same store, so serialising the spawns costs nothing worth having.
+    pub(super) fn isolate_store_for_this_process() -> std::sync::MutexGuard<'static, ()> {
+        use std::sync::{Mutex, Once, OnceLock};
+        static LOCK: OnceLock<Mutex<()>> = OnceLock::new();
+        static SET: Once = Once::new();
+
+        SET.call_once(|| {
+            let store = tempfile::tempdir().expect("temp store for edda-conductor tests");
+            std::env::set_var("EDDA_STORE_ROOT", store.path());
+            std::mem::forget(store);
+        });
+        LOCK.get_or_init(|| Mutex::new(()))
+            .lock()
+            .unwrap_or_else(|e| e.into_inner())
+    }
+
+    /// GH-417: the runner shells out to the installed `edda`, and `edda init`
+    /// registers its cwd. Under test that cwd is a `tempfile::tempdir()`, so
+    /// every run filed another dead path in the operator's real registry — 13
+    /// per `cargo test -p edda-conductor`, forever, on the machine of whoever
+    /// ran the tests.
+    ///
+    /// Asserts the guard rather than the absence: reading the real registry to
+    /// prove we did not touch it would depend on the developer's own machine
+    /// having one, and would pass for the wrong reason on a machine that has
+    /// none.
+    #[test]
+    fn ensure_init_cannot_reach_the_operators_real_store() {
+        let _guard = isolate_store_for_this_process();
+        let root = std::env::var("EDDA_STORE_ROOT").expect("isolation must be in force");
+        assert!(
+            !root.is_empty(),
+            "an empty EDDA_STORE_ROOT resolves back to the real store"
+        );
+
+        // Asserted against the OS temp dir rather than against `store_root()`:
+        // that function *returns* EDDA_STORE_ROOT when set, so asking it here
+        // would compare the temp store with itself and pass for the wrong
+        // reason. A store under the OS temp dir is a store the operator's
+        // registry is not in.
+        assert!(
+            Path::new(&root).starts_with(std::env::temp_dir()),
+            "the child must write to a throwaway store, got: {root}"
+        );
+    }
 
     #[test]
     fn get_context_returns_empty_on_missing_edda_dir() {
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: PR #432 (Round 1)

### Summary

The root cause is right and the fix works — `+13 → +0`, measured, on both `-p edda-conductor` and `--workspace` (1980 passed). The registry is swept, 95 → 16, zero dead.

One finding: **the test cannot fail.**

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — `ensure_init_cannot_reach_the_operators_real_store` never calls `ensure_init`.**

It calls `isolate_store_for_this_process()` directly and asserts the helper set a temp path. Delete the wiring line from `ensure_init` — the entire fix — and the test still passes, because it never goes near the function it is named after.

So the only thing standing between this bug and its return is a `+13 → +0` measurement I took by hand once. That is proof the fix works **today**; it is not a guard. This is the same shape as the three dead branches in #429: a test that reads correctly, asserts something true, and is about the wrong thing.

**Fix:** call `ensure_init` and assert the *child landed in the throwaway store*. Verified what that looks like:

```
$ EDDA_STORE_ROOT=$S edda init      # in a temp repo
$S/registry.json     ← 1 entry
$S/registry.lock
```

So `EDDA_STORE_ROOT` set + `registry.json` present under it means the child registered *there*. Remove the wiring and the variable is unset → the test dies on the `expect`. Falsifiable, which the current one is not.

It needs a skip when `edda` is not on PATH — `ensure_init` is best-effort, so with no binary nothing spawns and there is no child to have misdirected. CI has no `edda` installed; the developer's machine does, and the developer's machine is exactly where the damage lands.

**`get_context` is not a second leak — checked, not assumed.** It also shells out (`edda context`), and my fix does not guard it:

| command, in an empty temp dir | registry delta |
|---|---|
| `edda context` | **+0** |
| `edda init` | **+1** |

`edda context` does not register. `ensure_init` is the only path, which is a third independent confirmation of the mechanism.

**`#[cfg(test)]` is per-crate.** `edda-cli` depends on `edda-conductor`, so an `edda-cli` test that drove the runner would get the unguarded `ensure_init`. None does today (bisect: `edda-cli +0`), and a runtime guard would be worse than this. Worth knowing rather than worth fixing.

**Leaking the temp store is deliberate**, and the doc says so: it must outlive every test in the process, and with tests on parallel threads there is no later moment at which dropping it is safe.

### Testing Review

#### Coverage

The empirical bisect is the strongest evidence in this PR — by crate, then by module, then a controlled pair (`cancellation_stops_runner` +1 / `truncate_str_ascii` +0). None of that lives in the test suite, which is precisely P1-1.

`edda-conductor` 148 passed; workspace **1980 passed, 0 failed**. clippy **0**.

#### Testing Verdict: Inadequate — the fix is unguarded against regression

---

fagemx:
## Code Review: PR #432 (Round 2)

### Summary

P1-1 fixed. The test now drives `ensure_init` and checks where its child landed, and it was **falsified before being trusted**. With the fix removed:

```
ensure_init_sends_its_child_to_a_throwaway_store ... FAILED
  ensure_init must isolate the store before it spawns anything: NotPresent
registry: 17 -> 18  (+1)
```

Both ends agree: the test goes red *and* the registry gets polluted again. Restored: 148 passed, registry **+0**, clippy 0.

That pairing is what makes it a guard rather than a claim. The old test passed with the fix deleted; this one fails, and fails for the reason the bug exists.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

None. P1-1 resolved.

### Final state

| | before | after |
|---|---|---|
| `cargo test -p edda-conductor` | +13 entries | **+0** |
| `cargo test --workspace` | +13 entries | **+0** (1980 passed) |
| operator's registry | 95 entries, 79 dead | **16, zero dead** |

Swept with the pre-existing `edda gc --global` — including the two entries this review's own falsification created, which is a small demonstration that the leak was real and reproducible on demand.

### What this issue turned on

The trail went cold for one reason: the crate never calls `register_project`, so grepping for it finds nothing and the obvious conclusion is "not here". It shells out to the installed binary instead, one process removed from anything the source tree mentions. My earlier note on the issue drew exactly that wrong conclusion from two correct observations.

What broke it open was measuring instead of reading — bisect by crate, then by module, then a controlled pair. `cancellation_stops_runner` +1 versus `truncate_str_ascii` +0 is the whole diagnosis in two lines, and no amount of reading the diff would have produced it.

### Testing Review

#### Coverage

`edda-conductor` 148 passed. Workspace **1980 passed, 0 failed**. clippy **0** at `--all-targets` (counted, not inferred from empty output). fmt clean.

The test skips when `edda` is not on PATH — deliberate and documented. `ensure_init` is best-effort, so with no binary nothing spawns and there is no child to misdirect. CI has no `edda`; the developer's machine does, and that is the machine whose registry was being filled.

### Verdict: LGTM

Running the tests no longer costs you your registry.

---

Issue #433, opened 2026-07-16
ci: Windows runs 9 of 45 test binaries and cannot block — including none of edda-conductor, where GH-417's guard lives (P1)


fagemx:
## Problem

Windows CI runs **9 of 45** test binaries and **cannot block a merge**. edda's operator runs Windows.

`.github/workflows/ci.yml` since #411:

```yaml
# Windows is non-blocking (continue-on-error) and its build/link is
# ~5x slower, so it runs only the platform-divergent crates
# (filesystem paths, file locks, process spawn, mmap) instead of all 23.
- os: windows-latest
  args: -p edda-store -p edda-ledger -p edda-search-fts -p edda-bridge-claude -p edda
```

**The list contradicts the criterion written above it.** The rule names *process spawn*; `edda-conductor` is the most spawn-heavy crate in the workspace — `Command::new("edda")` in `ensure_init`, the Claude launcher, tmux control, shell checks — and it is not in the list.

Measured on run 29472530485 rather than inferred from the args:

| job | `test result:` lines |
|---|---|
| Test (ubuntu-latest) | **45** |
| Test (windows-latest) | **9** |

Uncovered on Windows: `edda-conductor`, `edda-ask`, `edda-pack`, `edda-core`, `edda-index`, and 13 others.

## Why this is not hypothetical

GH-417 was a Windows-manifested bug: `cargo test` filed 13 dead temp dirs into the operator's real registry, every run, because `edda-conductor` shells out to `edda init` in a tempdir. Fixed in #432.

Its regression guard — `ensure_init_sends_its_child_to_a_throwaway_store` — lives in `edda-conductor` and **has never executed on Windows CI**. The bug was found on Windows, felt on Windows, and its guard runs everywhere except Windows. The test is also the most platform-sensitive kind there is: it spawns a subprocess and asserts where the child wrote.

`continue-on-error: ${{ matrix.os == 'windows-latest' }}` (lines 28, 42) compounds it — even the 9 that do run are advisory. That predates #411; the matrix change did not create it, but it turned "Windows is advisory" into "Windows is advisory *and* mostly absent".

## Fix

1. Add `-p edda-conductor` to the Windows matrix at minimum. The existing comment already argues for it: it is the *only* crate matching "process spawn", which the comment names as a selection criterion.
[...]

PR #434, opened 2026-07-16
ci: derive the Windows crate list from its criterion, add edda-conductor + edda-transcript (GH-433)

fagemx:
## What

Fixes the coverage half of #433: the Windows CI matrix now runs the crates its own comment said it should. Derived from the stated criteria, not hand-curated.

## The bug in the old matrix

The comment named four selection criteria — process spawn, file locks, mmap, filesystem paths — then listed 5 crates. Deriving from the criteria instead of trusting the list turns up two matches that were absent:

| criterion | crates that match | was in list? |
|---|---|---|
| process spawn (`Command::new`) | edda, edda-bridge-claude, **edda-conductor** | conductor **no** |
| file locks (`fs2`) | edda-store, edda-ledger, edda-search-fts, **edda-transcript** | transcript **no** |
| mmap (`tantivy`) | edda-search-fts | yes |

`edda-conductor` has **7** files calling `Command::new` — the most spawn-heavy crate in the workspace — and the comment named "process spawn" as a criterion. It was the crate most obviously in scope and the one omitted.

## Why it mattered

`edda-conductor` is where GH-417 lived: `cargo test` filed 13 dead temp dirs into the operator's real registry, every run, on Windows. Fixed in #432. Its regression guard, `ensure_init_sends_its_child_to_a_throwaway_store`, **had never run on Windows CI** — the platform the bug was found on. It is also the most platform-sensitive kind of test there is: it spawns a subprocess and asserts where the child wrote.

## The fix

Add `edda-conductor` and `edda-transcript`, and rewrite the comment to list each crate against the criterion it satisfies, so the list is read off the rule rather than maintained beside it. That is how it drifted: a curated list and a stated rule with nothing tying them together.

## Verification

- Both added crates green on a **real Windows machine** before commit: conductor **148**, transcript **21**.
- The exact folded-scalar command CI will run, reconstructed and compiled with `--no-run`: `Finished`, all 7 crate names resolve.
- Windows test count **1059 → 1228** (+16%). conductor is logic-heavy and fast (6.6s), not the slow-linking kind that motivated the trim.

## Deliberately not in this PR

- **`continue-on-error` on Windows** (lines 28, 42): a policy call — should Windows block a merge on the operator's own platform? Left for the operator to decide, per #433.
- **The `line-tables-only` debug trim from #411**: sound, stays.
- **The open question** in #433 — whether `gh pr checks` reports a *failing* `continue-on-error` job as `pass` — is unresolved. It needs a deliberately-failing Windows job to observe, which does not belong on this branch. Still worth settling: if it masks, every "CI green" merge gate in this repo is weaker than it reads.

Addresses the coverage part of #433.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,16 +48,25 @@ jobs:
             args: --workspace
           - os: macos-latest
             args: --workspace
-          # Windows is non-blocking (continue-on-error) and its build/link is
-          # ~5x slower, so it runs only the platform-divergent crates
-          # (filesystem paths, file locks, process spawn, mmap) instead of all
-          # 23. Linux/macOS keep full-workspace coverage.
+          # Windows build/link is ~5x slower, so it runs only the crates whose
+          # behaviour actually diverges by platform instead of all 23.
+          # Linux/macOS keep full-workspace coverage.
+          #
+          # Each crate below earns its place against a concrete criterion; the
+          # list is derived from these, not curated, so it cannot silently drift
+          # from the rule (it did — GH-433 caught edda-conductor and
+          # edda-transcript missing while the comment named their criteria):
+          #   process spawn (Command::new):  edda, edda-bridge-claude, edda-conductor
+          #   file locks (fs2):              edda-store, edda-ledger, edda-search-fts, edda-transcript
+          #   mmap (tantivy):                edda-search-fts
           - os: windows-latest
             args: >-
               -p edda-store
               -p edda-ledger
               -p edda-search-fts
+              -p edda-transcript
               -p edda-bridge-claude
+              -p edda-conductor
               -p edda
     steps:
       - uses: actions/checkout@v4


fagemx:
## Code Review: PR #434 (Round 1)

### Summary

A PR that fixes "the list drifted from its rule" has one way to fail: **under-derive, and reproduce the exact bug.** So the review is the derivation, checked with a tool that does not corrupt output — my first probe (`grep` inside `execSync`) mangled CJK error text and returned empty, which made every crate look like it matched nothing. Re-run through the Grep tool:

**Every crate that spawns a process:**

```
edda-conductor      (6 files)   ← in list
edda-bridge-claude  (4 files)   ← in list
edda-cli / edda     (6 files)   ← in list
```

Checked both `Command::new` and the broader `process::Command | tokio::process` surface — same three crates, no fourth. The derivation is complete: no crate matches a criterion and sits outside the list.

**Internal consistency:** every crate now in the list matches at least one criterion (`edda`/bridge = spawn; store/ledger/search-fts/transcript = fs2; search-fts = mmap). Nothing is carried for a reason that no longer holds.

23 crates in the workspace, confirmed — so the old comment's "all 23" was accurate about the denominator; it was the numerator that was wrong.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

None.

### What makes this the right fix rather than a patch

Adding `-p edda-conductor` alone would have closed the specific gap and left the mechanism that produced it. The mechanism was a curated list beside a stated rule with nothing binding them. Listing each crate against its criterion means the next reader derives instead of trusts — the same move as pinning a key name in a test so a rename cannot pass silently.

### Verification (from the PR, spot-checked)

- Both added crates green on a **real Windows machine**: conductor 148, transcript 21. Re-confirmed here.
- The folded-scalar `args` reconstructs to the exact 7-crate command and compiles `--no-run` clean — no crate-name typo that would make CI fail to select a package. That failure mode matters specifically for a YAML list.
- +16% Windows test count; conductor is fast (6.6s), not slow-linking.

### Scope discipline

Three things correctly left out and named: `continue-on-error` (policy, operator's call), the #411 debug trim (sound), and #433's open question about whether a failing advisory job reads as `pass` (needs a deliberately-red job to observe). The last is the one that, if true, weakens every merge gate here — right to leave it stated and unresolved rather than guess.

### Testing Review

No tests — CI configuration. Exercised by construction: the added crates were run on the target platform before commit, which is the strongest check available for "will this job go green".

#### Testing Verdict: N/A

### Verdict: LGTM

The list is now read off the rule, and the rule's most obvious member is finally in it.

---

Issue #404, opened 2026-07-14
retrieval coverage & UX: task receipts unreachable, no snippets, ask unranked (P1)


fagemx:
## Problem

Retrieval coverage and result UX have three gaps that all answer the same operator question badly: "what did we actually do about X?"

1. **Task receipts are invisible to every retrieval surface.** `task.done` receipts are the fleet's answer to "what shipped, verified how" — and they are unreachable: FTS indexes note text and turns but not task payloads (verified: `ENV_LOCK` sits verbatim in rail task #1's receipt, `edda search query "ENV_LOCK"` → 0 hits), and `edda ask` has no Tasks section (its sections are Decisions / Timeline / Related Commits / Related Notes / Conversations — `ask "drill"` shows the drill only via an incidental session note).
2. **Search results carry no snippets.** Hits render as bare event ids + timestamps. Neither a human nor an agent can judge relevance without running `search show` per hit — which in practice means nobody does, and the search surface loses most of its value.
3. **`ask` ordering under `--limit` is not relevance-ranked** — appears chronological/arbitrary, so the best match can fall below the cut.

## Fix

- Index `task.*` payload fields (title, receipt, failure reason, brief_ref) in FTS; tag hit type so results show `[task.done #3]`
- Add a Tasks section to `edda ask` (title, status, receipt excerpt, evidence paths), reusing the existing `task_views()` projection
- Snippets: render a ±~10-word window around the first match per hit (tantivy has snippet support built in)
- `ask` ranking: exact-key > key-substring > value/reason-substring, recency as tiebreaker — cheap heuristic, no scoring engine needed

## Acceptance

- `search query "ENV_LOCK"` returns the rail task.done hit with a readable snippet
- `ask "drill"` shows the drill task with its receipt
- Every FTS hit line includes a snippet
- An exact decision-key query always ranks that decision first

🤖 Generated with [Claude Code](https://claude.com/claude-code)

fagemx:
## Done — all four acceptance criteria met, and two of the three "gaps" were narrower than filed

Retrieval coverage and UX are fixed. As is now usual, measuring the code first showed the three problems were not three equal-sized holes:

### Acceptance, each verified

1. **`search query "ENV_LOCK"` returns the task.done hit with a readable snippet** ✅ — live in the fleet workspace, the exact string the issue names:
   ```
   Found 1 result(s) for: ENV_LOCK
     [task.done] evt_01kxgfnc6scy…
[...]

Issue #401, opened 2026-07-14
decision provenance: recorded ≠ ratified — stop the projection from laundering agent inference into operator authority


fagemx:
## Problem: the projection launders authority

Three kinds of "decision" enter the ledger through the same verb and come out wearing the same uniform. An operator ruling made after deliberate convergence, an agent's transcription of that ruling, and an agent's own mid-implementation working call all render in the SessionStart pack under one undifferentiated **Binding Decisions** list.

The laundering pipeline, step by step:

1. t0 — agent writes `edda decide` (machine inference, zero human review; the PostToolUse hook actively prompts this after every commit)
2. t1 — next session start: it renders as a Binding Decision, same standing as operator rulings
3. t2 — downstream agents treat it as a constraint and build on it, citing it as justification
4. t3 — by the time a human sees it, work has hardened around it; reversal costs a layer, not a line

The hash chain proves the record was never altered — it cannot prove the record deserved its authority. Every hop in the pipeline adds authority; no hop adds verification.

## Fix: recorded ≠ ratified (same shape the task rail already ships for tasks: done ≠ accepted)

The schema already carries the field — `DecisionPayload.authority` exists and defaults to `None`. This is a defaults + rendering + promotion-path change, not a storage change.

1. **Honest default**: `edda decide` marks the decision `authority=agent` (recorded, unratified). No self-declared upgrades — an agent cannot mint operator authority on write.
2. **Ratification is a separate event** (candidate-style, mirroring the existing draft/approval flow): `edda decide ratify <key>` writes an approval event referencing the decision event. Supersede semantics unchanged. Ratifying is an operator act; the promotion is itself hash-chained and replayable.
3. **Two-tier pack rendering**: SessionStart / UserPromptSubmit context splits into **Operator rulings (binding)** and **Agent working decisions (unratified — challenge with reasons)**. The downstream contract for tier 2 is "current best guess", not "constitution".
4. **Operator sweep surface**: `edda ask --unratified` (or equivalent) lists pending agent decisions; ratify or reject-with-reason. A rejection should read like the gate's rework loop: the reason travels back.

## Acceptance

- An agent-written decision starts unratified and renders in tier 2
- A ratify event promotes it to tier 1; the promotion chain replays from the ledger
- Existing decisions are not retro-tagged (archive rule: history is never retrofitted); they render under a third "legacy (pre-provenance)" label or the unratified tier — decide during implementation
- House TDD: failing tests first for default tagging, ratify event, rendering split

[...]

PR #436, opened 2026-07-16
feat(search): index task receipts so they are searchable (GH-404)

fagemx:
## What

Make task receipts (and reasons, and titles) searchable. First of the #404 fixes.

Live, before → after in the edda repo:

```
# before
$ edda search query "GH-401"
No results found for: GH-401          # ← but that string is verbatim in rail task #1's receipt

# after (one-time rebuild of 143 events)
$ edda search query "GH-401"
Found 1 result(s) for: GH-401
  [task.done] evt_01kxhb50wh35…
  decision-provenance «GH»«401» shipped: recorded≠ratified 全鏈…
```

## Root cause (measured, not the issue's framing)

The issue lists three gaps. Reading the code, two of them share one cause and the third is narrower than stated:

- **Task receipts unreachable (real).** `task.*` events already reach the indexer, but `extract_event_title_body` had no task branch — they fell to the generic `body = payload.text`, and no task event has a `text` field. They indexed **empty**.
- **"Search results carry no snippets" (mostly already solved).** `SnippetGenerator` already runs and both render paths already print `r.snippet`. The "bare event ids" the issue saw is an **empty body → empty snippet**, i.e. the *same* root cause. Fixing task bodies gives task hits their snippets, as the live output shows.
- **"ask unranked" (narrower than stated).** `rank_decisions_by_similarity` already does TF-IDF. That's a separate follow-up (exact-key-first), not part of this PR.

So this one change closes the receipt gap *and* the snippet gap for task hits.

## How

A `task.*` branch in the extractor:

| event | title | body |
|---|---|---|
| `task.done` | `task #N done` | receipt + evidence paths |
| `task.failed` | `task #N failed` | reason |
| `task.created` | the task title | brief_ref |

`INDEX_VERSION` 2→3 so existing indexes rebuild once and pick up the task text. Without it the fix only helps *new* events and every already-recorded receipt stays invisible — the exact thing the issue is about. The rebuild is the standard `index_is_outdated` path (GH-402/412); one-time, and the GH-403 turns-early-out keeps it from being the old ~25s.

## Tests

`task_events_expose_their_text_to_the_index` — written RED first (failed with `title="" body=""`), asserts the receipt, reason, title, and brief each reach the indexed text.

`edda-search-fts` 58 passed, `edda` bin 211 passed. clippy 0 at `--all-targets` (counted). fmt clean. Live-verified above.

## Not in this PR (the rest of #404)

- Tasks section in `edda ask` (genuinely missing — `AskResult` has no tasks field).
- Exact-key-first ranking in `ask`.

Part of #404.

### DIFF
--- a/crates/edda-search-fts/src/indexer.rs
+++ b/crates/edda-search-fts/src/indexer.rs
@@ -105,6 +105,29 @@ fn extract_event_title_body(event: &edda_core::Event) -> (String, String) {
         return (title.to_string(), body.to_string());
     }
 
+    // task_intake: a task ingested from an external source — a GitHub issue,
+    // via `edda intake`. Named with an underscore, so it is *not* part of the
+    // `task.` family below and needs its own arm; miss it and the issue a task
+    // came from — its title and intent, the richest text any task event carries
+    // — is unsearchable (GH-404).
+    if event.event_type == "task_intake" {
+        let s = |k: &str| payload.get(k).and_then(|v| v.as_str()).unwrap_or("");
+        let arr = |k: &str| {
+            payload
+                .get(k)
+                .and_then(|v| v.as_array())
+                .map(|a| {
+                    a.iter()
+                        .filter_map(|x| x.as_str())
+                        .collect::<Vec<_>>()
+                        .join(" ")
+                })
+                .unwrap_or_default()
+        };
+        let body = format!("{} {} {}", s("intent"), arr("labels"), arr("constraints"));
+        return (s("title").to_string(), body.trim().to_string());
+    }
+
     // Task events: the searchable payload is the receipt / reason / title, and
     // none of it lives in a `text` field — so without this branch every task
     // event indexed empty and the receipt (the fleet's record of what shipped)
@@ -1151,6 +1174,33 @@ mod tests {
         assert!(
             joined.contains("docs/plan/x.md"),
             "task.created brief_ref must be indexed: {joined:?}"
+        );
+
+        // task_intake is named with an underscore, not a dot, so a `task.`
+        // prefix check silently skips it — yet it carries the richest text of
+        // any task event (a GitHub issue's title and intent). It must index too.
+        let intake = edda_core::event::new_task_intake_event(&edda_core::event::TaskIntakeParams {
+            branch: "main".to_string(),
+            parent_hash: None,
+            source: "github".to_string(),
+            source_id: "404".to_string(),
+            source_url: "https://example/issues/404".to_string(),
+            title: "retrieval coverage and UX".to_string(),
+            intent: "make task receipts searchable".to_string(),
+            labels: vec!["P1".to_string()],
+            priority: "high".to_string(),
+            constraints: vec!["no scoring engine".to_string()],
+        })
+        .unwrap();
+        let (t, b) = extract_event_title_body(&intake);
+        let joined = format!("{t} {b}");
+        assert!(
+            joined.contains("retrieval coverage"),
+            "task_intake title must be indexed: {joined:?}"
+        );
+        assert!(
+            joined.contains("make task receipts searchable"),
+            "task_intake intent must be indexed: {joined:?}"
         );
     }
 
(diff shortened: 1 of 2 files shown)


fagemx:
## Code Review: PR #436 (Round 1)

### Summary

The receipt fix is correct and live-proven (`GH-401`: 0 → 1, tagged + snippet). But enumerating the task event types turns up one this PR misses — and it is the richest-text one.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — `task_intake` is a task event with the most searchable text of all, and this fix skips it.**

The branch keys on `event_type.strip_prefix("task.")`. There are six dotted types (created/done/failed/requeued/session/started), but a seventh task event is named with an **underscore**:

```rust
// event.rs — new_task_intake_event
event_type: "task_intake".to_string()   // ← not "task.intake"
```

`strip_prefix("task.")` returns `None` for `"task_intake"`, so it falls to the generic branch and indexes **empty** — exactly the bug this PR fixes for the others.

It is not vestigial. `cmd_intake.rs` ingests a **GitHub issue** into the ledger as a `task_intake`, and `cmd_pipeline.rs` reads it back. Its payload is the richest of any task event:

```
title, intent, labels, priority, constraints, source_url
```

So after this PR, a task's receipt is searchable but the **issue it came from** — its title and intent — is not. For a verb whose whole point is "what did we do about X", missing the intake text is the gap that matters most.

The naming split (`task_intake` vs `task.done`) is the trap: the dotted convention makes the underscore one look like a different family, and `strip_prefix` silently agrees.

### Fix

Handle `task_intake` alongside the dotted types: title = the intake title, body = intent + labels + constraints (the "what and why", searchable). One extra arm, and the RED test extends to cover it so the naming split cannot re-hide it.

- `task.requeued` (task_id + attempt) and `task.started`/`task.session` genuinely have no prose — the empty `_ =>` arm is correct for them. `task_intake` is the only miss.
- Snippets come from the `body` field only (pre-existing). For `created`/`intake` the title lives in the `title` field, so a title-only match snippets from the body. Acceptable and unchanged by this PR; worth knowing.

### Testing Review

#### Coverage

`task_events_expose_their_text_to_the_index` was RED-first and is falsifiable. But it tests only the three types the code happens to handle — it would not have caught P1-1, because it never constructs a `task_intake`. The fix's test must add that case; that is the exact gap that let the miss through.

`edda-search-fts` 58, `edda` bin 211, clippy 0.

#### Testing Verdict: Inadequate — the untested event type is the one with the bug

---

fagemx:
## Code Review: PR #436 (Round 2)

### Summary

P1-1 fixed. `task_intake` now indexes its title + intent + labels + constraints, and the test constructs one — it was RED first (`title must be indexed: " "`).

### Coverage is now complete for the scope

All seven task event types accounted for:

| type | indexed text |
|---|---|
| `task_intake` | title + intent + labels + constraints |
| `task.done` | receipt + evidence paths |
| `task.failed` | reason |
| `task.created` | title + brief_ref |
| `task.started` / `.session` / `.requeued` | none — correct: these carry only ids/counters, no prose |

The `_ =>` empty arm is right for the last three; verified their payloads (`task_id` + `attempt`/lease, no text).

### Key Findings

None.

### Deliberately out of scope

Other non-task events also hit the generic branch and index only their `text` field: `decision_ratify` (a ratify reason), `review_bundle`, `cmd` (argv). These are real, but they are not #404 — which is task retrieval — and folding them in would be an unreviewed scope expansion. Named here so the boundary is a decision, not an oversight.

### Verification

`task_intake` is **unit-tested, not live-verified**: the edda ledger has no `task_intake` event to search (none has been ingested via `edda intake`), and manufacturing one mutates the real ledger. It shares the exact mechanism the receipt case proved end-to-end (`GH-401`: 0 → 1 with snippet), so the risk is low — but I am not calling it live-proven when it is not.

`edda-search-fts` 58 passed, clippy **0** at `--all-targets` (counted). fmt clean.

### Testing Review

The test now covers the case whose absence let P1-1 through. That is the specific lesson: the first test exercised only the three types the code happened to name, so it agreed with the bug. It now constructs every text-bearing task type.

### Verdict: LGTM

Every task event with words in it now has those words in the index.

---

PR #437, opened 2026-07-16
feat(ask): add a Tasks section to ask (GH-404)

fagemx:
## What

Add a **Tasks section** to `edda ask`. Second of the #404 fixes.

Before, `ask` had six hit sections — Decisions, Timeline, Related Commits, Related Notes, Conversations, Dependents — but none for tasks. So a finished task's **receipt**, the record of what shipped and how it was verified, never surfaced however well it matched.

Live (edda repo):

```
$ edda ask "provenance"
…
── Tasks ──────────────────────────────
  #1 [done] decision provenance: recorded ≠ ratified…
     receipt: decision-provenance GH-401 shipped: recorded≠ratified 全鏈…
     evidence: https://github.com/fagemx/edda/pull/409
```

That is the issue's acceptance — "ask shows the task with its receipt" — met.

## How

- A task matches when the query appears in its **title, receipt, or failure reason** (case-insensitive). The receipt is the point: that is where verification lives.
- Newest first (`task_id` desc), capped at `--limit`.
- Reuses the existing `task_views()` projection — no new query path, no new event reads.
- `TaskHit` carries `task_id`, `title`, `status`, receipt excerpt, evidence paths — exactly the fields the issue names.

`hit_count` (the fleet-hint emptiness check from GH-407) now counts tasks too, so a **task-only** match is not mistaken for "nothing here" and does not trigger a spurious `--fleet` hint.

## Design notes

- **Empty query (Overview mode) shows no tasks.** `ask` with no query lists active decisions; dumping every task there would be noise. A task surfaces only when something is asked. Stated so it reads as a choice.
- **JSON**: `tasks` is `skip_serializing_if = "Vec::is_empty"`, so existing `--json` consumers see the field only when there are hits — no shape change for the empty case.

## Tests

`ask_surfaces_a_task_whose_receipt_matches_the_query` — RED first (empty section, assertion failed). Builds a `created`+`done` task, queries a receipt token, asserts the hit carries id, title, `status == "done"`, the receipt, and evidence.

`edda-ask` 44 passed, `edda` bin 211 passed. clippy **0** at `--all-targets` (counted). fmt clean. Live-verified above.

## Not in this PR (the rest of #404)

Exact-key-first ranking in `ask` — the last #404 piece.

Part of #404.

### DIFF
--- a/crates/edda-ask/src/lib.rs
+++ b/crates/edda-ask/src/lib.rs
@@ -375,6 +375,10 @@ pub fn ask(
     // shipped about X, verified how", so a task matches when the query appears
     // in its title, receipt, or failure reason — the receipt being the point.
     // Newest first, since a rail answer is usually a recent one (GH-404).
+    //
+    // Deliberately not branch-filtered, unlike decisions above: the task rail is
+    // a workspace-global queue, not a per-branch record, so `--branch` narrowing
+    // the decisions but not the tasks is the intended asymmetry.
     let tasks: Vec<TaskHit> = if q.is_empty() {
         vec![]
     } else {
@@ -1017,17 +1021,41 @@ mod tests {
         .unwrap();
         ledger.append_event(&done).unwrap();
 
+        // A second task that does NOT mention "drill". Without it, the test
+        // could not tell a working filter from one that returns every task —
+        // with a single task, "the match is present" and "everything is
+        // present" are the same assertion.
+        let other =
+            edda_core::event::new_task_created_event(&edda_core::event::TaskCreatedParams {
+                branch: "main",
+                parent_hash: None,
+                task_id: 8,
+                title: "write the changelog",
+                assignee: None,
+                agent_kind: None,
+                after: &[],
+                plan_id: None,
+                work_unit_ref: None,
+                brief_ref: None,
+                idempotency_key: None,
+            })
+            .unwrap();
+        ledger.append_event(&other).unwrap();
+
         let opts = AskOptions {
             limit: 10,
             ..Default::default()
         };
         let result = ask(&ledger, "drill", &opts, None).unwrap();
 
-        let hit = result
-            .tasks
-            .iter()
-            .find(|t| t.task_id == 7)
-            .expect("a task whose title and receipt say 'drill' must appear");
+        assert_eq!(
+            result.tasks.len(),
+            1,
+            "only the drill task matches — the changelog task must be filtered out: {:?}",
+            result.tasks
+        );
+        let hit = &result.tasks[0];
+        assert_eq!(hit.task_id, 7);
         assert!(hit.title.contains("drill"), "title: {}", hit.title);
         assert_eq!(hit.status, "done");
         assert!(
--- a/crates/edda-ask/src/lib.rs
+++ b/crates/edda-ask/src/lib.rs
@@ -384,7 +384,7 @@ pub fn ask(
     } else {
         let needle = q.to_lowercase();
         let mut views = ledger.task_views()?;
-        views.sort_by(|a, b| b.task_id.cmp(&a.task_id));
+        views.sort_by_key(|t| std::cmp::Reverse(t.task_id)); // newest first
         views
             .into_iter()
             .filter(|t| {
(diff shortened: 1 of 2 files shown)


fagemx:
## Code Review: PR #437 (Round 1)

### Summary

The Tasks section works and is live-proven (`ask "provenance"` → task #1 with receipt + evidence). The code is correct. The **test is not strong enough to know that**.

### Key Findings

#### Critical Issues (P0)

None.

#### High Priority (P1)

**P1-1 — the test cannot distinguish "matches the query" from "returns every task".**

`ask_surfaces_a_task_whose_receipt_matches_the_query` builds **one** task (#7) and asserts it appears. But a broken filter that returned *all* tasks would pass this identically — with one task in the ledger, "the match is present" and "everything is present" are the same assertion.

So the test proves the section renders a task; it does not prove the **filter** works. That is the exact shape this session keeps hitting — a green test that is about the wrong property.

The precision is real: live, `ask "provenance"` returned only task #1, though the edda ledger has a second `done` task that does not mention provenance. So the code excludes non-matches. But that evidence lives in a manual run, not the suite, and a future refactor that broke the filter would stay green.

**Fix:** add a second task that does **not** match the query and assert it is absent. One task present + one task excluded is what makes the filter the thing under test.

- **`--fleet` inherits this for free.** `execute_fleet` calls `ask()` (now populating tasks) and renders via `format_human` (now with the section), and `hit_count` counts tasks — so a task-only match on a sibling both renders and is not skipped as empty. Verified by reading; consistent.
- **Branch filter asymmetry.** `opts.branch` filters decisions but not tasks — `ask --branch X "q"` would show tasks from all branches. Tasks are rail-global in practice, so this is defensible, but it is an inconsistency worth a one-line acknowledgement in the code rather than silence.
- **Cost.** `task_views()` projects *all* task events on every `ask`, then filters in memory. Fine at today's task counts; if a rail ever holds thousands, a SQL-side filter would be the move. `ask` is interactive, so worth noting as the place that would bite first.

### Testing Review

#### Coverage

The match case is covered but weakly (P1-1). Missing: a non-matching task excluded (precision), and empty-query → no tasks (the Overview choice, which the code makes deliberately). The precision case is the one that matters.

`edda-ask` 44, `edda` bin 211, clippy 0.

#### Testing Verdict: Inadequate — one task cannot test a filter

---

fagemx:
## Code Review: PR #437 (Round 2)

### Summary

P1-1 fixed. The test now builds two tasks — one matching, one not — and asserts exactly one hit. It was **falsified before being trusted**: with the filter forced to `true`, it fails —

```
only the drill task matches — the changelog task must be filtered out:
  [TaskHit { task_id: 8, title: "write the changelog", … },
   TaskHit { task_id: 7, title: "run the onboarding drill", … }]
```

— restored, it passes. So the filter is now the thing under test, not just the rendering.

### Key Findings

None.

### What this closes

Two of #404's three parts are now done and live-verified:

- **#436** — task receipts searchable (`GH-401`: 0 → 1 with snippet), plus the snippet gap for task hits.
- **#437** (this) — Tasks section in `ask` (`ask "provenance"` surfaces task #1 with its receipt + evidence).

The branch-filter asymmetry is now documented in the code (task rail is workspace-global, so `--branch` narrows decisions but not tasks) rather than left implicit.

### Design notes (unchanged from round 1, not blocking)

- **Cost**: `task_views()` projects all task events per `ask` and filters in memory. Fine at current scale; a SQL-side filter is the move if a rail ever holds thousands. Named as the place that would bite first, since `ask` is interactive.
- **Empty query** shows no tasks (Overview mode) — a deliberate choice, guarded by `q.is_empty()`, reasoned rather than tested.

### Testing Review

`edda-ask` 44, `edda` bin 211, clippy **0** at `--all-targets` (counted). fmt clean. Live-verified.

The strengthening is the point: this session has repeatedly hit tests that pass for the wrong reason, and a one-task test of a filter is exactly that shape. Two tasks is the minimum that makes "filter" falsifiable.

### Verdict: LGTM

`ask` now answers "what shipped about X" with the task that shipped it.

---

PR #438, opened 2026-07-16
feat(ask): glob-scoped decisions detect in-place edits (GH-424)

### LINKED ISSUE #424: ask: glob-scoped decisions cannot detect edits, only add/remove
## Problem

A glob-scoped decision cannot detect the most common way it goes stale: a file inside the scope being **edited**.

#405 fixed the false positives — `crates/foo/*` was probed literally, matched nothing, and reported `Missing` for every decision. It now probes the deepest wildcard-free ancestor (`crates/foo`). But a directory's mtime moves only when entries are added or removed, not when their contents are rewritten:

```
edit foo/bar.rs   → foo/ mtime unchanged   (not detected)
touch foo/new.rs  → foo/ mtime changed     (detected)
```

Since `edda claim --paths` takes globs, nearly every decision is glob-scoped, so in practice `stale_modified` almost never fires. The hint is honest but weak.

This is a deliberate trade, recorded in `staleness.rs`'s header: it matches the module's stated preference for false-negatives ("returns unknown rather than false-positive stale"), and a hint that rarely fires beats one that always lies. But weak is not the goal.

## Fix

Enumerate the glob's matches and take the newest mtime among them.

The blocker is the seam: `FsOracle` can only `probe(&Path) -> (bool, Option<String>)` — one path, no enumeration. Detecting edits needs something like `newest_mtime_under(&Path) -> Option<String>`, or a `list(&Path)`, so the mock stays deterministic.

Worth considering while there:

- **Cost.** This runs per decision per `edda ask`. A scope like `crates/*` over a large tree would walk a lot; the current probe is one `stat`. A depth cap or a "first N entries" bound may be needed, and `ask` is on the interactive path.
- **Semantics.** `crates/foo/*` means the direct children; `crates/foo/**` (if ever recorded) means the subtree. The current reduction treats both as `crates/foo`.
- **Whether it is worth it.** The hint exists so an agent does not cite a decision about code that has since changed. If the answer is "walk the tree on every ask", the honest alternative might be to drop `stale_modified` for globs entirely and keep only `missing` — a narrower promise that is fully kept.

## Acceptance

- Editing a file inside a glob scope, after the decision ts, reports `stale_modified`
- The cost is bounded well enough for `edda ask` to stay interactive
- `MockFs` can still drive every case deterministically

Found during #405's review (PR #423).

fagemx:
## What

Glob-scoped decisions now detect the most common way they go stale: a file **edited** inside the scope. Implements #424 (the "implement edit detection" fork you chose).

## The problem

#405 fixed the false positives — `crates/foo/*` was probed literally, matched nothing, and reported `Missing` for everything. It now probes the deepest wildcard-free ancestor (`crates/foo`). But a directory's mtime moves only on add/remove, not on an in-place rewrite:

```
edit foo/bar.rs   → foo/ mtime unchanged   (was: not detected)
touch foo/new.rs  → foo/ mtime changed     (detected)
```

Since `edda claim --paths` takes globs, nearly every decision is glob-scoped, so `stale_modified` almost never fired.

## The fix

The blocker was the seam: `FsOracle::probe(&Path)` takes one path, no enumeration. Adds:

```rust
fn newest_mtime_under(&self, dir: &Path) -> Option<String>;
```

For a glob, staleness is now the **newer of** the directory's own mtime (catches add/remove) and its newest child (catches edit). A literal path is unchanged — it names one file and uses its own mtime.

## The three things #424 asked to weigh

**Cost** — this runs per glob-scoped decision per `edda ask`, on the interactive path. Bounded: one level of `read_dir`, capped at `MAX_GLOB_ENTRIES` (512). A normal source directory is scanned whole; only a pathological one is truncated, where a bounded, possibly-incomplete answer beats a slow `ask`.

**Semantics** — `crates/foo/*` (direct children) is fully handled. `crates/foo/**` (subtree) still misses edits nested deeper than one level. That is a narrower promise than `**` implies — documented in the code — but a fully-kept one for the common `*` case, and it never lies, which keeps the module's stated false-negative bias.

**Worth it** — yes: the common glob is `*` (one `read_dir`), so the feared "walk the tree" cost is really one directory listing, and edit-detection is the whole reason the hint exists.

## Tests (RED first)

- `a_glob_detects_a_file_edited_inside_it` — the headline: directory mtime *before* the decision, a child edited *after*, must be `StaleModified`. Was RED (reported `Fresh` with the dir's old mtime).
- `a_literal_path_uses_its_own_mtime_not_its_siblings` — the glob path must not leak into literal scopes.
- `stdfs_newest_mtime_reads_a_real_directory` — MockFs proves the *logic*; this proves `StdFs` actually reads the filesystem (a `read_dir` returning nothing would leave the feature inert while every MockFs test stayed green). Asserts presence, not an exact time, so there is no clock race.

`MockFs` gained `set_newest_under`, so every case stays deterministic — the third #424 acceptance.

`edda-ask` 47 passed. Fresh `clippy --workspace --all-targets` (`-Dwarnings`) **0** — run fresh, not cached, after the last session's lesson that a cached clippy misses lints CI catches.

Closes #424.

### DIFF
--- a/crates/edda-ask/src/staleness.rs
+++ b/crates/edda-ask/src/staleness.rs
@@ -95,10 +95,56 @@ fn probe_target(pattern: &Path) -> PathBuf {
     out
 }
 
+/// Whether a recorded path is a glob (has a wildcard) rather than a literal
+/// file. Same characters `probe_target` splits on, so the two agree on what
+/// "glob" means.
+fn is_glob(pattern: &str) -> bool {
+    pattern.contains('*') || pattern.contains('?') || pattern.contains('[')
+}
+
+/// The later of two optional rfc3339 timestamps. Unparseable values are ignored
+/// rather than treated as newest, keeping the module's false-negative bias: a
+/// timestamp we cannot read must not manufacture staleness.
+fn newer_of(a: Option<String>, b: Option<String>) -> Option<String> {
+    let parse = |s: &str| OffsetDateTime::parse(s, &Rfc3339).ok();
+    match (a, b) {
+        (Some(x), Some(y)) => match (parse(&x), parse(&y)) {
+            (Some(dx), Some(dy)) => Some(if dy > dx { y } else { x }),
+            (Some(_), None) => Some(x),
+            (None, Some(_)) => Some(y),
+            (None, None) => Some(x),
+        },
+        (Some(x), None) => Some(x),
+        (None, Some(y)) => Some(y),
+        (None, None) => None,
+    }
+}
+
+/// Most entries scanned when finding the newest mtime under a glob's directory.
+///
+/// This runs per glob-scoped decision per `edda ask`, on the interactive path,
+/// so the walk is bounded: a directory with more entries than this reports the
+/// newest among the first `MAX_GLOB_ENTRIES` it yields. Normal source
+/// directories are far smaller, so the cap only ever bites a pathological tree —
+/// where a bounded, possibly-incomplete answer beats a slow `ask`.
+const MAX_GLOB_ENTRIES: usize = 512;
+
 /// Filesystem oracle abstracts real fs so tests can be deterministic.
 pub trait FsOracle {
     /// Return (exists, mtime_rfc3339 when available).
     fn probe(&self, path: &Path) -> (bool, Option<String>);
+
+    /// Newest mtime (rfc3339) among the direct entries of `dir`, or `None` if it
+    /// is not a readable directory.
+    ///
+    /// This is what lets a glob detect an *edit*: a directory's own mtime moves
+    /// when entries are added or removed, but not when a file inside is rewritten
+    /// in place — so the freshest child, not the directory, is the signal
+    /// (GH-424). Direct entries only (one level); see [`check_paths_staleness`]
+    /// for how `**` subtree scopes are handled.
+    fn newest_mtime_under(&self, _dir: &Path) -> Option<String> {
+        None
+    }
 }
 
 pub struct StdFs;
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: PR #438 (Round 1)

### Summary

The seam is right and the logic holds for every case, not just the headline. I traced all four ways a glob scope changes:

| change | dir mtime | newest child | `newer_of` | verdict |
|---|---|---|---|---|
| edit in place | old | **new** | new | StaleModified ✓ |
| add file | new | new | new | StaleModified ✓ |
| remove file | new | old | new | StaleModified ✓ |
| nothing | old | old | old | Fresh ✓ |

Add and remove still work because they move the directory's own mtime; edit works because the newest child now contributes. `newer_of` taking the later of the two is what makes all four fall out correctly.

### Key Findings

None.

### Completeness checks that passed

- **Only two `FsOracle` impls** — `StdFs` (production) and `MockFs` (test), both override `newest_mtime_under`. The `None` default is an unreached safety net, and `annotate_hits` — the one production entry point — uses `StdFs`. So no path silently gets edit-blindness from the default.
- **`Missing` still short-circuits** before the glob mtime logic, so a deleted directory is still `Missing`, not a confusing `Fresh`.
- **`is_glob` keys on the recorded pattern** (`rel`), not the resolved path, so a literal path never enumerates — guarded by `a_literal_path_uses_its_own_mtime_not_its_siblings`.

### The three things #424 asked to weigh, each resolved

- **Cost**: one `read_dir` for the common `*` glob, capped at 512 entries. Bounded for the interactive path; a normal source dir is scanned whole.
- **Semantics**: `*` (direct children) fully handled; `**` (subtree) misses edits nested past one level — documented in the code, a narrower promise than `**` implies but one that never lies.
- **Determinism**: `MockFs` gained `set_newest_under`; every case is driven without touching a real clock.

### Testing Review

- `a_glob_detects_a_file_edited_inside_it` — RED first (reported `Fresh` with the dir's stale mtime), the exact bug #424 describes.
- `stdfs_newest_mtime_reads_a_real_directory` — the one that matters beyond the mock: it proves `StdFs::read_dir` actually returns something, so the feature is not inert behind a working-looking mock. Asserts presence, not a timestamp, so no clock race — deliberately, after this session's flaky-timing lessons.

`edda-ask` 47 passed. Fresh `clippy --workspace --all-targets` (`-Dwarnings`) **0** — run fresh rather than trusting a cached pass, after #437's cached-clippy miss.

### Verdict: LGTM

The hint that almost never fired now fires on the change it was built to catch — and still never lies.

---

Issue #447, opened 2026-08-08
test(bridge-claude): dispatch/digest tests flake under parallel execution — 23 env::set_var sites mutate process-global state


fagemx:
## Summary

On main (c581b85, clean tree), `cargo test -p edda-bridge-claude` with default parallelism fails 9 tests; the same suite is fully green with `-- --test-threads=1` (549 passed).

Failing under parallelism:
- `digest::tests::*` (7 tests)
- `dispatch::tests::subagent_stop_writes_summary_and_removes_heartbeat`
- `dispatch::tests::task_completed_skips_when_task_id_empty`

## Cause (corrected after deeper audit)

Not "missing serialization" — **asymmetric locking**. An `ENV_LOCK` already exists (`lib.rs`) and 10 of the `set_var` sites in `dispatch/tests.rs` correctly take it. But the 9 failing tests take no lock at all — they never write env themselves (`digest/tests.rs` has zero `ENV_LOCK` and zero `set_var`); they are *readers* of env-dependent code paths, poisoned by concurrent writers.

Concrete chain: `session_end_unclaim_only_with_active_peers` (`dispatch/tests.rs:1368-1370`) holds `ENV_LOCK` while setting `EDDA_BRIDGE_AUTO_DIGEST=0` and `EDDA_PLANS_DIR=/nonexistent`; `digest::tests::*` run in parallel and read `EDDA_BRIDGE_AUTO_DIGEST`-gated paths. A mutex only excludes other lock-holders — unlocked readers race straight past it.

## Impact

- Anyone running the standard pre-commit command (`cargo test --workspace`) sees spurious failures depending on scheduling.
- CI results depend on test-thread count rather than code correctness.

## Suggested direction

- Correct fix shape: **env-reading tests must take the same `ENV_LOCK`** (read-side locking), or better, inject the env dependency as a parameter/config so digest paths don't consult process env in tests.
- Wrong fix shape (what this issue's first draft implied): wrapping more `set_var` writers in the lock — the writers largely already lock; the readers don't.
- Ban new bare `set_var` in tests either way.

Found by the verifier session during the baseline audit for #442-#445; team convention meanwhile is `-- --test-threads=1` for this crate.

fagemx:
[...]

Issue #128, opened 2026-02-27
Agents don't auto-claim scope — need SessionStart hook or built-in auto-claim


fagemx:
## Problem

When `edda watch` is running, peers show up as `unknown` because agents never execute `edda claim`. The CLAUDE.md coordination section instructs agents to claim at session start:

```
**Claim your scope** at session start: `edda claim "label" --paths "src/scope/*"`
```

But this is just a text instruction — agents often skip it when they jump straight into a task (e.g. `/issue-plan #7`).

## Current Workaround

Manually create `.claude/hooks.json` in each project:

```json
{
  "hooks": {
    "SessionStart": [
      {
        "command": "edda claim \"agent-$RANDOM\" --paths \"**/*\"",
        "timeout": 5000
      }
    ]
  }
}
```

## Proposed Solution

Consider one or more of:

1. **`edda init` generates hooks.json** — When running `edda init` in a project, automatically create/update `.claude/hooks.json` with a SessionStart hook for `edda claim`
2. **Auto-claim in bridge** — The `edda-bridge-claude` SessionStart hook injection could include an automatic claim if the agent hasn't claimed yet
3. **Document the hook requirement** — At minimum, the README/setup docs should mention that `.claude/hooks.json` is needed for coordination to work properly

## Impact

Without auto-claim:
- `edda watch` shows all peers as "unknown"
- Coordination rules (off-limits, scope boundaries) can't be enforced
- The whole multi-agent coordination flow is silently broken

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #448, opened 2026-08-08
fix(bridge-claude): repo-wide auto-claims and orphan claims (GH-444, GH-445)

### LINKED ISSUE #444: bug(bridge-claude): branch-fallback auto-claim claims "**/*" — under enforce_offlimits a fresh session blocks every peer write
## Summary

When a session has no file edits yet, `maybe_auto_claim` falls back to claiming the entire repo under the branch name (`crates/edda-bridge-claude/src/peers/autoclaim.rs:148-151`, introduced for #128 so the peer shows up in `edda watch` immediately):

```rust
match detect_git_branch_in(cwd) {
    Some(branch) => (branch, vec!["**/*".to_string()]),
    None => return,
}
```

With advisory-only claims this is mostly cosmetic noise ("everything is off-limits" in peers' protocol render). But with enforcement enabled (`EDDA_ENFORCE_OFFLIMITS=1` or `bridge.enforce_offlimits`), the PreToolUse hook blocks any Edit/Write matching an active peer's claim globs (`crates/edda-bridge-claude/src/dispatch/tools.rs:60-93`). A `**/*` claim from a fresh session therefore hard-blocks **every** Edit/Write from **every** peer for as long as that session's heartbeat is alive.

## Reproduction

1. Enable `bridge.enforce_offlimits` in a repo with two Claude Code sessions.
2. Session A: start it and let it idle (submit a prompt, make no edits) — auto-claim writes `(branch, ["**/*"])`.
3. Session B: any Edit or Write → blocked with "Off-limits: file ... is claimed by agent '<branch>' (paths: **/*)".

Two fresh sessions on the same branch make it mutual: both claim `**/*`, both are fully blocked.

## Impact

Turning on the (documented, planned-to-become-default) hard enforcement makes multi-agent startup deadlock-prone. The failure reads as "everything is claimed by <branch-name>", which is confusing because nobody ran `edda claim`.

## Suggested direction

- The `**/*` fallback exists only for `edda watch` visibility — record it as a *presence* signal (heartbeat label), not a *scope claim*. E.g. write the heartbeat label without a claim event, or claim an empty path set.
- Alternatively, have `check_offlimits` ignore auto-claims whose paths are `["**/*"]` (or ignore all branch-fallback claims) when enforcing.

Found during a coordination-layer audit.

### LINKED ISSUE #445: bug(bridge-claude): sessions that end without active peers never write Unclaim — orphan claims accumulate in coordination.jsonl
## Summary

`cleanup_session_state` removes the heartbeat unconditionally but only writes the `Unclaim` event when peers are active at session end (`crates/edda-bridge-claude/src/dispatch/session.rs:653-655`):

```rust
crate::peers::remove_heartbeat(project_id, session_id);
if peers_active {
    crate::peers::write_unclaim(project_id, session_id);
}
```

A session that ends solo — the common case for single-agent work — leaves its claim in `coordination.jsonl` with no terminating `Unclaim`. Compaction preserves all claims (`compute_board_state_for_compaction` keeps every claim/binding/request), so these orphans survive `edda gc` indefinitely.

## Impact

- `compute_board_state` folds orphan claims into `board.claims` on every hook invocation forever; consumers stay correct only because each one independently re-filters against live heartbeats.
- Any consumer that forgets that join sees phantom scope claims (e.g. label resolution in `render_coordination_protocol` reads `board.claims` by session_id first).
- The log accumulates one permanent claim per solo session per scope change, growing compaction state without bound.

## Suggested direction

Write the `Unclaim` unconditionally — same rationale as the adjacent comment for `remove_heartbeat` ("keep ... unconditional as idempotent cleanup"). An `Unclaim` with no matching claim is already a no-op in the fold. Alternatively, make compaction drop claims whose session has no heartbeat and no matching Unclaim older than the state-file GC horizon (7 days).

Found during a coordination-layer audit.

fagemx:
Both bugs are claim-lifecycle bugs in the Claude bridge, so they land together. Route for GH-444 follows the controller/verifier ruling (presence signal on the heartbeat); see "How this landed" at the bottom for what changed from the first push.

## GH-444 — branch-fallback auto-claim claimed `**/*`

`maybe_auto_claim` fell back to `(branch, ["**/*"])` when a session had no edits yet, so the peer would show up in `edda watch` (#128). Under `enforce_offlimits` that glob matches every file, so one idle session hard-blocked every peer's `Edit`/`Write` until its heartbeat went stale.

**Fix — presence moves to the heartbeat.** The branch name reaches `edda watch` and my-label resolution today *only* through that claim: `auto_label` (`peers/helpers.rs:76-79`) returns `""` for a session with no edits, and `write_claim` never touches the heartbeat. So `write_heartbeat` now falls back to the git branch when `auto_label` yields nothing, and `maybe_auto_claim` emits no claim at all for a session that has edited nothing. Nothing is left for enforcement to match, and identity lives where every consumer already looks:

- TUI peer list reads `hb.label` (`discovery.rs:79`) → non-empty, so `active_peers()` (`edda-cli/src/tui/app.rs:110`) no longer hides the session after 120s. The old `**/*` claim never fixed this — the claim only populated the Claims panel.
- My-label resolution and request delivery already fall back to the heartbeat (`render_coord.rs:33`, `helpers.rs:18`), so `edda request "<branch>"` keeps landing.

**Also kept: `check_offlimits` skips any claim containing `**/*`.** Not redundant with the fallback removal — `coordination.jsonl` is append-only, so sessions that claimed `**/*` before the upgrade keep folding into the board and keep blocking peers until they end. Without the guard the deadlock survives the fix.

**Trade-off, as flagged in the issue:** a deliberate `edda claim "x" --paths "**/*"` no longer hard-blocks peers. It still renders in the coordination protocol and on the board, so it stays advisory. A claim on everything can't express a coordination *boundary* anyway, and its enforcement failure mode is indistinguishable from enforcement being broken. Honouring explicit repo-wide claims would need provenance on `ClaimEntry` (auto vs manual) plus a `board.rs` fold change — out of scope here.

## GH-445 — solo sessions never wrote `Unclaim`

`cleanup_session_state` wrote `Unclaim` only when `peers_active`. A session ending solo left its claim in `coordination.jsonl` with no terminator, and since compaction preserves every claim, `edda gc` never reclaimed it.

**Fix:** write it unconditionally, for the same reason `remove_heartbeat` is unconditional. That leaves `peers_active` unused, so it is removed along the caller chain — `cleanup_session_state`, `dispatch_session_end`, and the `SessionEnd` arm — which also retires dispatch's private mtime-based `has_active_peers` (its only non-test caller was that gate). Its three tests now assert the same liveness behaviour against `discover_active_peers`, which owns liveness for the peers layer.

`remove_heartbeat` → `write_unclaim` ordering is unchanged.

## Same bug elsewhere (not touched)

`edda-bridge-codex`, `-cursor`, `-hermes`, and `-openclaw` each carry their own `cleanup_session_state` with the identical `if peers_active` gate, so GH-445 reproduces in all four. Outside this PR's claimed scope — worth a follow-up issue.

## Verification

Against the stated acceptance criteria:

| Criterion | Test |
|---|---|
| Fresh session claims nothing | `auto_claim_writes_no_claim_for_a_fresh_session` |
| Claiming resumes on first edit | `auto_claim_starts_claiming_once_files_are_edited` |
| Fresh session stays visible with a non-empty label | `heartbeat_label_falls_back_to_git_branch_for_a_fresh_session` |
| `edda request "<branch>"` still delivers | `fresh_session_receives_requests_addressed_to_its_branch` (plus the pre-existing `request_delivered_via_heartbeat_label_no_claim`) |
| Two fresh sessions on one branch don't block each other | `two_fresh_sessions_on_one_branch_do_not_block_each_other` |
| A real claim still blocks under enforcement | `offlimits_blocks_peer_claimed_file` (unchanged, still green) |
| Legacy `**/*` claims don't block | `offlimits_ignores_repo_wide_peer_claim` |
| Solo end releases the claim, board no longer lists it | `session_end_writes_unclaim_for_every_claimed_session` (the reversed-and-renamed `session_end_unclaim_only_with_active_peers`, whose `assert_eq!(unclaim_count, 0)` encoded the bug) |
| Unclaim with no claim is a fold no-op | `unclaim_without_a_claim_is_a_fold_noop` |
| GC compaction self-heals — verified, not assumed | `compaction_drops_released_claims_and_never_emits_unclaim` |

Gates: `cargo fmt --check` clean, `RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets` clean, `cargo test -p edda-bridge-claude --lib -- --test-threads=1` 557/557.

Per the fleet test convention, everything ran with `--test-threads=1`; no new test mutates the environment. `solo_to_multi_session_transition` now ages a heartbeat timestamp directly instead of sleeping 1.1s and overriding `EDDA_PEER_STALE_SECS`, removing one env mutation from the suite (GH-447).

Unrelated pre-existing failures on this machine, reproduced on unmodified `main`: `edda-cli cmd_search::tests::foreign_project_uses_its_own_registered_repo` fails even in isolation; `edda-store` registry/skill_registry/user_config tests fail only under parallel execution and pass with `--test-threads=1`. Some `edda-bridge-claude` digest tests also flake under machine load (three consecutive serial runs of this branch gave 7, 1, then 0 failures; a serial run of clean `main` gave 0) — they pass in isolation and this branch touches nothing they read.

## How this landed

The first push recorded the fallback as a zero-path claim, keeping the branch label on the claim event. That preserved label resolution and request delivery but still used a claim to mean "present", and left the TUI's 120s hide in place. The verifier's route — put the label on the heartbeat, emit no claim — is strictly better, and this branch now does that.

Fixes #444
Fixes #445

### DIFF
--- a/crates/edda-bridge-claude/src/dispatch/mod.rs
+++ b/crates/edda-bridge-claude/src/dispatch/mod.rs
@@ -95,47 +95,6 @@ pub(crate) fn render_write_back_protocol(_cwd: &str) -> Option<String> {
     Some(render::writeback())
 }
 
-// ── L2 Solo Gate ──
-
-/// Check if any non-stale peer session heartbeats exist (excluding current session).
-/// Used to skip all L2 I/O when running solo.
-fn has_active_peers(project_id: &str, session_id: &str) -> bool {
-    let state_dir = edda_store::project_dir(project_id).join("state");
-    let stale_threshold: u64 = std::env::var("EDDA_PEER_STALE_SECS")
-        .ok()
-        .and_then(|v| v.parse().ok())
-        .unwrap_or(120);
-    let entries = match fs::read_dir(&state_dir) {
-        Ok(e) => e,
-        Err(_) => return false,
-    };
-    let now = std::time::SystemTime::now();
-    for entry in entries.flatten() {
-        let name = entry.file_name().to_string_lossy().to_string();
-        if !name.starts_with("session.") || !name.ends_with(".json") {
-            continue;
-        }
-        let sid = name
-            .strip_prefix("session.")
-            .and_then(|s| s.strip_suffix(".json"))
-            .unwrap_or("");
-        if sid.is_empty() || sid == session_id {
-            continue;
-        }
-        // Check file modification time as a lightweight staleness check
-        // (avoids parsing JSON — just stat the file)
-        if let Ok(meta) = entry.metadata() {
-            if let Ok(modified) = meta.modified() {
-                let age = now.duration_since(modified).unwrap_or_default();
-                if age.as_secs() <= stale_threshold {
-                    return true;
-                }
-            }
-        }
-    }
-    false
-}
-
 // ── Hook dispatch ──
 
 /// Main hook entrypoint: parse stdin, dispatch by hook_event_name.
@@ -235,16 +194,10 @@ pub fn hook_entrypoint_from_stdin(stdin: &str) -> anyhow::Result<HookResult> {
             Ok(HookResult::empty())
         }
         "SessionEnd" => {
-            // Solo gate: only used to skip coordination log writes (write_unclaim).
-            // Computed here (not at top) so non-SessionEnd hooks avoid the dir scan (#83).
-            let peers_active = !session_id.is_empty() && has_active_peers(&project_id, &session_id);
-            dispatch_session_end(
-                &project_id,
-                &session_id,
-                &transcript_path,
-                &cwd,
-                peers_active,
-            )
+            // The solo gate used to live here, to skip the write_unclaim on
+            // coordination.jsonl. It orphaned claims instead (#445) — cleanup is
+            // now unconditional, so the peer scan is gone with it.
+            dispatch_session_end(&project_id, &session_id, &transcript_path, &cwd)
         }
         "SubagentStart" => {
             // Inject peer context BEFORE writing heartbeat so the sub-agent
--- a/crates/edda-bridge-claude/src/dispatch/session.rs
+++ b/crates/edda-bridge-claude/src/dispatch/session.rs
@@ -622,6 +622,14 @@ pub(super) fn run_postmortem(project_id: &str, session_id: &str, cwd: &str) {
     );
 }
 
+/// True if the coordination board still folds a claim for this session.
+fn session_holds_claim(project_id: &str, session_id: &str) -> bool {
+    crate::peers::compute_board_state(project_id)
+        .claims
+        .iter()
+        .any(|c| c.session_id == session_id)
+}
+
 /// Remove session-scoped state files that are no longer needed.
 pub(super) fn cleanup_session_state(project_id: &str, session_id: &str, peers_active: bool) {
     let state_dir = edda_store::project_dir(project_id).join("state");
@@ -650,7 +658,13 @@ pub(super) fn cleanup_session_state(project_id: &str, session_id: &str, peers_ac
     crate::peers::cleanup_subagent_heartbeats(project_id, session_id);
     // Peer heartbeat + unclaim (L2 — keep remove_heartbeat unconditional as idempotent cleanup)
     crate::peers::remove_heartbeat(project_id, session_id);
-    if peers_active {
+    // Release the claim whenever this session still holds one. Gating on
+    // `peers_active` alone orphaned every solo session's claim in
+    // coordination.jsonl: compaction preserves claims, so `edda gc` never
+    // reclaimed them and the board fold carried them forever (#445).
+    // `peers_active` stays as the cheap path — with peers around, the claim
+    // write is the norm, so skip the board read.
+    if peers_active || session_holds_claim(project_id, session_id) {
         crate::peers::write_unclaim(project_id, session_id);
     }
     // NOTE: The following state files are intentionally NOT deleted here — they
--- a/crates/edda-bridge-claude/src/dispatch/session.rs
+++ b/crates/edda-bridge-claude/src/dispatch/session.rs
@@ -262,7 +262,6 @@ pub(super) fn dispatch_session_end(
     session_id: &str,
     transcript_path: &str,
     cwd: &str,
-    peers_active: bool,
 ) -> anyhow::Result<HookResult> {
     // 1. Final ingest so signals are up-to-date
     ingest_and_build_pack(project_id, session_id, transcript_path, cwd);
@@ -414,7 +413,7 @@ pub(super) fn dispatch_session_end(
     notify_session_end(project_id, cwd, session_id);
 
     // 3. Clean up session-scoped state files
-    cleanup_session_state(project_id, session_id, peers_active);
+    cleanup_session_state(project_id, session_id);
 
     // 4. Collect warnings (pending tasks)
     if let Some(warning) = collect_session_end_warnings(project_id) {
@@ -622,16 +621,8 @@ pub(super) fn run_postmortem(project_id: &str, session_id: &str, cwd: &str) {
     );
 }
 
-/// True if the coordination board still folds a claim for this session.
-fn session_holds_claim(project_id: &str, session_id: &str) -> bool {
-    crate::peers::compute_board_state(project_id)
-        .claims
-        .iter()
-        .any(|c| c.session_id == session_id)
-}
-
 /// Remove session-scoped state files that are no longer needed.
-pub(super) fn cleanup_session_state(project_id: &str, session_id: &str, peers_active: bool) {
+pub(super) fn cleanup_session_state(project_id: &str, session_id: &str) {
     let state_dir = edda_store::project_dir(project_id).join("state");
     // Dedup hash (Step 2)
     let _ = fs::remove_file(state_dir.join(format!("inject_hash.{session_id}")));
@@ -658,15 +649,12 @@ pub(super) fn cleanup_session_state(project_id: &str, session_id: &str, peers_ac
     crate::peers::cleanup_subagent_heartbeats(project_id, session_id);
     // Peer heartbeat + unclaim (L2 — keep remove_heartbeat unconditional as idempotent cleanup)
     crate::peers::remove_heartbeat(project_id, session_id);
-    // Release the claim whenever this session still holds one. Gating on
-    // `peers_active` alone orphaned every solo session's claim in
-    // coordination.jsonl: compaction preserves claims, so `edda gc` never
+    // Unconditional, for the same reason remove_heartbeat is: an Unclaim with no
+    // matching claim is a no-op in the fold, and compaction drops both halves.
+    // Gating this on active peers orphaned every solo session's claim in
+    // coordination.jsonl — compaction preserves claims, so `edda gc` never
     // reclaimed them and the board fold carried them forever (#445).
-    // `peers_active` stays as the cheap path — with peers around, the claim
-    // write is the norm, so skip the board read.
-    if peers_active || session_holds_claim(project_id, session_id) {
-        crate::peers::write_unclaim(project_id, session_id);
-    }
+    crate::peers::write_unclaim(project_id, session_id);
     // NOTE: The following state files are intentionally NOT deleted here — they
     // persist across sessions to provide continuity context at the next SessionStart:
     //   - active_tasks.json  → L1 narrative shows previous session's final task state
--- a/crates/edda-bridge-claude/src/dispatch/tests.rs
+++ b/crates/edda-bridge-claude/src/dispatch/tests.rs
@@ -2,11 +2,10 @@ use std::fs;
 
 // Imports from dispatch/mod.rs
 use super::{
-    apply_context_budget, context_budget, has_active_peers, hook_entrypoint_from_stdin,
-    increment_counter, is_same_as_last_inject, mark_nudge_sent, read_counter,
-    render_workspace_section, render_write_back_protocol, set_compact_pending,
-    take_compact_pending, wrap_context_boundary, write_inject_hash, write_peer_count, HookResult,
-    EDDA_BOUNDARY_END, EDDA_BOUNDARY_START,
+    apply_context_budget, context_budget, hook_entrypoint_from_stdin, increment_counter,
+    is_same_as_last_inject, mark_nudge_sent, read_counter, render_workspace_section,
+    render_write_back_protocol, set_compact_pending, take_compact_pending, wrap_context_boundary,
+    write_inject_hash, write_peer_count, HookResult, EDDA_BOUNDARY_END, EDDA_BOUNDARY_START,
 };
 // Imports from sub-modules
 use super::events::{
@@ -646,7 +645,7 @@ fn session_end_cleans_state() {
     fs::write(state_dir.join(format!("inject_hash.{sid}")), "abcd").unwrap();
     fs::write(state_dir.join("compact_pending"), "1").unwrap();
 
-    cleanup_session_state(pid, sid, false);
+    cleanup_session_state(pid, sid);
 
     assert!(
         !state_dir.join(format!("inject_hash.{sid}")).exists(),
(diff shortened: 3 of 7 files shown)


fagemx:
Controller review (first pass — adversarial verifier pass still pending):

**Boundary discipline: clean.** Touched files are exactly the claimed set; render_coord.rs / heartbeat.rs / board.rs untouched. The empty-paths-claim route is a better call than the heartbeat-label route I originally suggested — label still flows through the board, so watch/my-label/request addressing keep working with zero cross-boundary edits.

**Criteria walk (from the #444/#445 review checklists):**
- old gating test inverted AND renamed, with a board-fold assertion added — ✓
- no-claim solo end stays silent (noise test) — ✓
- legacy `**/*` claims neutralized under enforcement (append-only log argument in the code comment is correct) — ✓
- fallback→real-claim upgrade path covered — ✓
- new tests use real temp git repos instead of env mutation — ✓ (respects #447 discipline)

**Accepted tradeoffs (documented in body):** deliberate manual `**/*` claims become advisory-only — acceptable; the precise fix needs claim provenance, which belongs to board.rs (other worker's territory) and is out of scope here.

**Deferred items:** empty-paths protocol rendering (`Your scope: **main** ()`) reassigned to the request/ack PR where that file is already open; cross-bridge duplication filed as #449; machine-dependent test failures filed as #450.

Holding merge until the verifier's independent pass lands.

fagemx (commented):
## Independent verification (verifier lane)

Reviewed on a clean worktree at `1604abb`, without reading the controller's review first.

### Quality gates (PR branch)

| Gate | Result |
|---|---|
| `cargo fmt --check` | PASS |
| `RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets` | PASS |
| `cargo test -p edda-bridge-claude -- --test-threads=1` | PASS — 553 passed, 0 failed (main baseline: 549) |

### GH-445 — PASS on every criterion

- Solo session **with** a claim now writes `Unclaim`; the claim is gone from the board fold. Verified by the rewritten `session_end_writes_unclaim_for_every_claimed_session`, which also asserts the fold, not just the log line.
- Solo session **without** a claim writes nothing (`session_end_skips_unclaim_when_nothing_was_claimed`). This is better than the "unconditional write" the issue suggested — it fixes the orphan without adding a log line per solo session.
- The pre-existing test that asserted the *old* behaviour was inverted and renamed rather than deleted. That was the specific thing I was watching for.
- Ordering (`remove_heartbeat` → board read → `write_unclaim`) is safe: `session_holds_claim` reads `coordination.jsonl`, not heartbeats, so the earlier heartbeat removal can't affect it.

Minor, non-blocking: `session_holds_claim` adds a full `coordination.jsonl` parse at solo SessionEnd. Once per session end, on the path that previously did no work. Fine as-is; noting it so it isn't a surprise later.

### GH-444 — PASS on the enforcement and delivery criteria

The presence-signal route (claim kept, `paths` emptied) is the right call, and it is safe for the two downstream consumers I flagged as at-risk during prep — **because the claim event is retained**:

- **Request delivery**: `render_coord.rs:28-34` and `helpers.rs:13-19` both resolve `my_label` from `board.claims` first. The claim still carries `label = <branch>`, so a fresh session still resolves a non-empty label and `edda request "<branch>" "msg"` still lands. Had the claim been dropped entirely, `auto_label` returns `""` for a session with no edits (`helpers.rs:76-79`) and the session would have become unaddressable.
- **`edda watch` visibility**: the Claims panel renders `c.label` from the board, so the branch label still shows. The Peers panel reads `hb.label` and is unaffected either way (unchanged by this PR).
- **Off-limits render**: `render_coord.rs:219-222` filters on `!p.claimed_paths.is_empty()`, so a fresh peer correctly drops out of the "Off-limits" section. Improvement.
- **Enforcement still bites for real claims**: the `**/*` skip is narrow; `crates/foo/*` still blocks, covered by the existing `offlimits_blocks_peer_claimed_file`. The new `offlimits_ignores_repo_wide_peer_claim` covers the skip. The skip is also not an escape hatch — it only makes *your own* claim unenforceable against peers, it grants no write access into anyone else's scope.
- The upgrade path (fallback → real scope once files are edited) is covered by `auto_claim_replaces_branch_fallback_once_files_are_edited`, and the dedup in `maybe_auto_claim` handles the empty-paths state correctly.

### WARN — empty-paths claims make `/api/scope/check` deny everything

This is the one thing I'd fix before merge. `edda-serve` is outside the PR's diff but is a consumer of the shape it changed.

`crates/edda-serve/src/api/policy.rs:70-94` builds a `GlobSet` from `claim.paths` and answers `allowed: glob_set.is_match(f)`. With `paths == []` the builder produces an **empty** set, and an empty `GlobSet` matches nothing — globset 0.4.18 `lib.rs:351-352` returns `false` immediately when `is_empty()`, and its own `default_set_is_empty_works` test asserts exactly that.

So for a fresh session, this endpoint flips:

| | `no_claim` | `all_allowed` |
|---|---|---|
| before (`paths = ["**/*"]`) | false | **true** |
| after (`paths = []`) | false | **false** — every file denied |

The `None` arm directly above (`policy.rs:51-68`) is deliberately permissive with the comment "no claim means all files allowed". An empty-paths claim expresses the same thing — presence, no scope — but takes the `Some` arm and gets the opposite answer.

That is the GH-444 failure shape again (a fresh session that has touched nothing denies work), just relocated from the PreToolUse hook to the HTTP surface. Suggested fix, in `post_scope_check`:

```rust
Some(claim) if claim.paths.is_empty() => /* same permissive response as None, but no_claim: false */
```

No existing test covers it: `scope_check_no_claim_permissive`, `scope_check_wildcard_claim` and friends cover `None` and `["**/*"]`, so this stays green either way. Worth an `scope_check_presence_claim_is_permissive` case alongside the fix.

(For the record, `**/*` itself does *not* diverge between the two surfaces — the bridge ignoring a peer's repo-wide claim and serve allowing everything under your own repo-wide claim are both permissive. Only the empty case is inconsistent.)

### WARN (minor) — a deliberate repo-wide claim is now silently non-enforcing

`edda claim "x" --paths "**/*"` is now ignored by `check_offlimits` with no signal to the operator. The code comment justifies the guard well (pre-#444 logs keep emitting `**/*` from still-heartbeating sessions, so it has to be load-bearing beyond the fallback), and I agree with keeping it. But nothing surfaces the downgrade — the claim still renders as if enforced. A one-line note in `edda coord`, or a mention in the enforcement docs, would keep it from reading as a bug later.

### Note — cosmetic, and pre-existing

`render_coord.rs:174-178` now renders `Your scope: **main** ()` for a fresh session. Confirmed cosmetic-only: no behaviour depends on it. Worth stating that the adjacent "Claim your scope" nudge stays suppressed for fresh sessions (a claim exists, so the `else` branch never fires) — but that was equally true before this PR with `(**/*)`, so it is not a regression here. Flagging it as its own follow-up rather than scope creep, since `render_coord.rs` is another lane's file.

### Verdict

GH-444 and GH-445 are both correctly fixed within `edda-bridge-claude`, with tests that assert observable behaviour rather than shape. The one blocking item is the `edda-serve` consumer above.

fagemx:
Route ruling — FINAL: b126894 (heartbeat presence) ships. And a correction to my own earlier assessment on this thread:

I called the interim empty-paths-claim route (1604abb) "a better call" — that was wrong. Worker's counter-evidence, verified: `PeerSummary.label` comes from `hb.label` (`discovery.rs:79`), `App::active_peers` (`tui/app.rs:102-112`) hides peers with an empty label after 120s, and the Claims panel has no staleness filter at all. So under 1604abb a fresh session vanished from the Peers list after 120s leaving an unexpiring ghost row in Claims — failing the review criterion "fresh session stays visible in `edda watch` past 120s with a non-empty label". b126894 satisfies it (`heartbeat_label_falls_back_to_git_branch_for_a_fresh_session`), and the redo accidentally shipped the only correct design.

Remaining verifier findings routing: serve scope-check inversion + silent `**/*` downgrade → #453 (real bugs, reachable today via manual pathless `edda claim`, not introduced by this PR). render `()` cosmetics — moot on HEAD.

Process: push freeze on this branch until the verifier's delta verdict on b126894.

fagemx (commented):
## Delta re-review — `1604abb` → `b126894`

My earlier review was against `1604abb`, which has since been superseded. Re-reviewed the new route independently. HEAD confirmed unchanged at `b126894` at both open and close of this pass.

### Quality gates (`b126894`)

| Gate | Result |
|---|---|
| `cargo fmt --check` | PASS |
| `RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets` | PASS |
| `cargo test -p edda-bridge-claude -- --test-threads=1` | PASS — 557 passed, 0 failed |

### Correction: my `edda watch` PASS on `1604abb` was wrong

The counter-argument is right and I want to record that plainly. My own prep criterion was "a fresh session still has a non-empty label in `edda watch`, and is still listed after 120s". Under `1604abb` that was **not** satisfied: `PeerSummary.label` comes from `hb.label` (`discovery.rs:79`), the branch name lived only on the claim, and `tui/app.rs:102-112` hides a peer whose label is empty once it is older than 120s. `render_claims` has no staleness filter at all, so the fresh session got the worst of both — gone from Peers, lingering in Claims.

I checked the right code and then graded the wrong question: I concluded "unchanged from main, so not a regression" and marked the criterion met. Those are two different claims, and only the first one was true.

`b126894` satisfies it properly, and the route is the one my prep note said was safe *provided `hb.label` is non-empty* — which is now exactly what it guarantees.

### The new route holds up

**The branch fallback lands before anything resolves a label.** This was the load-bearing question, since a brand-new session has no claim and no transcript. `dispatch/mod.rs:201` calls `ensure_heartbeat_exists` *before* `dispatch_session_start` (line 202), and `ensure_heartbeat_exists` (`heartbeat.rs:117-128`) delegates to `write_heartbeat` with `label = None` and `SessionSignals::default()`. Empty signals make `auto_label` return `""`, so the new branch fallback fires and the heartbeat carries the branch name. Only then does `render_coordination_protocol` resolve `my_label`. The ordering is correct, and the comment at `mod.rs:197-200` explains exactly why `ensure_heartbeat_exists` has to exist at all.

(`write_heartbeat_minimal` does not carry the fallback, but it takes an explicit label and is the CLI path, so that is fine.)

**Test coverage for the new path is real, and it is the new tests carrying it.** Worth being precise about which test does what:

- `request_delivered_via_heartbeat_label_no_claim` passes **explicit** labels, so it protects the general "label from heartbeat, no claim → delivery" mechanism the new route depends on — but it does not itself exercise the branch fallback.
- `heartbeat_label_falls_back_to_git_branch_for_a_fresh_session` pins the label on a real git repo.
- `fresh_session_receives_requests_addressed_to_its_branch` is the end-to-end one and the direct answer to my criterion: real repo on a branch, heartbeat with `label = None`, `maybe_auto_claim` writing nothing, peer sends to the branch label, request lands.
- `two_fresh_sessions_on_one_branch_do_not_block_each_other` covers the mutual-deadlock case from the issue and asserts `claims.is_empty()`.

**Claims panel no longer listing a fresh session is an improvement, not a UX regression.** The panel is titled "Claims" and lists scope claims; a session that has touched nothing has no scope, and the old `main []` row asserted a claim that did not exist. The identity moved to the Peers panel, where it now shows with a non-empty label and survives the 120s stale-unlabeled filter. Layout is unaffected: `ui.rs:24` falls back to `has_peers`, which is true because the heartbeat exists.

**The three rewritten dispatch tests are equivalent for the behaviour that still exists.** `has_active_peers` was deleted along with the SessionEnd solo gate, so testing it directly was no longer possible; `any_active_peer` asserts the same observable through `discover_active_peers`. The only difference is mtime-based vs `last_heartbeat`-based staleness, and the peers layer is now the single owner of that question.

**`solo_to_multi_session_transition` still tests the original behaviour, better.** Ageing the heartbeat timestamp instead of overriding `EDDA_PEER_STALE_SECS` still drives the peer stale and still asserts the fall back to solo — and it exercises the real timestamp path rather than a degenerate zero threshold. It also removes one env mutation, which is the right direction given #447.

**`compaction_drops_released_claims_and_never_emits_unclaim` closes the item I had asked to verify rather than assume.** The unconditional `write_unclaim` now appends a line at every session end including sessions that never claimed; this test proves `edda gc` reclaims both halves, so the cost does not accumulate. That tradeoff — a line per session end, self-healing at GC, versus the board read the previous revision used — is a reasonable call and is now documented in the code.

### Verdict

PASS. No blocking findings on the delta. The remaining `edda-serve` item is no longer reachable from auto-claim under this route and is correctly tracked separately (see below).

Issue #450, opened 2026-08-08
test(workspace): tests read the real ~/.edda store — machine-state-dependent failures on clean main


fagemx:
Two symptom groups on a developer machine, both reproducible on clean main and unrelated to any PR:

1. **Deterministic**: `edda-cli cmd_search::tests::foreign_project_uses_its_own_registered_repo` fails even when run alone — it consults the real user store, so its result depends on what projects happen to be registered on the machine.
2. **Parallel-only**: several `edda-store` tests (registry / skill_registry / user_config) and some `edda-bridge-claude` digest tests fail under default parallelism but pass with `-- --test-threads=1` — they share the real `~/.edda` (or `%APPDATA%\edda`) as mutable global state and race on it.

This is a distinct root cause from #447 (env-var races): here the shared global is the **store directory**, not process env. The clean fix is the same shape though — tests must not touch the real user store. `EDDA_STORE_ROOT` already exists as an override; test setup should point it at a `tempfile::tempdir()` per test (and that env write must respect the #447 locking discipline, or better, `store_root()` could accept an injected override for tests).

Until fixed, workspace-wide green cannot be used as a review gate on machines with a lived-in edda store; per-crate serial runs are the honest baseline.

Reported by the fixer-claims worker during #448; needs independent confirmation on a second machine (verifier session is on it).

fagemx:
Lead from PR #451 review: edda-cli already has a `crate::test_support::isolated_store()` helper — the new GH-443 CLI test uses it and is machine-state-independent. The deterministic failure in `cmd_search::tests::foreign_project_uses_its_own_registered_repo` likely just predates that helper; first fix attempt should be wiring it (and the edda-store registry/skill_registry/user_config tests) through `isolated_store()` / `EDDA_STORE_ROOT` instead of inventing new isolation machinery.

fagemx:
Live evidence caught during the fleet run: the real store's project state dir for this repo contains `session.s1.json` — session_id `s1` (a test fixture id), `started_at: 2026-07-07` (a month old), `last_heartbeat: 2026-08-08T14:14:20Z` (refreshed a minute before capture, while a worker session was running `cargo test` in its worktree), `branch: fix/gh-444-445-claim-lifecycle` (that worker's branch).

[...]

Issue #446, opened 2026-08-08
feat(coordination): doorbell mode — machine-readable peer→session mapping and host-push hints so host messaging can wake peers


fagemx:
## Context

Edda's coordination layer is deliberately pull-only: requests, claims, and task assignments land in the shared store and a peer sees them at its next hook event (SessionStart / UserPromptSubmit / every-3rd PreToolUse nudge). Nothing wakes an idle peer — if nobody types into that session, new records sit unread. TASK_RAIL_V1 already names the missing piece as an optional L2 fast path ("triggers are pushed and may be dropped; truth is written in the ledger").

Meanwhile every major host now ships cross-session push messaging (Claude Code `send_message` between sessions, Codex Desktop `send_message_to_thread`). Field test today (3-session formation: controller + 2 fixers + verifier, driving #442-#445): a host-pushed message arrives as a user turn in the target session, which **triggers UserPromptSubmit — so edda's coord diff, peer state, and task nudge inject automatically alongside the message**. The doorbell already works end-to-end with zero edda changes; what's missing is that edda neither enables nor teaches it.

Proposal: edda stays transport-free, but makes borrowing the host's push channel easy, discoverable, and semi-automatic.

### 1. `edda peers --json` with host session ids (highest value / lowest cost)

Heartbeats already store `session_id` from the hook payload; it just isn't exposed. A controller agent today knows a peer's *label* but the host messaging tool needs a *session id* — closing that gap currently requires reading heartbeat files by hand.

```bash
edda peers --json
# [{"label":"fixer-claims","session_id":"d2365545-...","branch":"fix/gh-444-445","last_heartbeat":"...","stale":false}]
```

This makes label → session_id → host `send_message` fully scriptable.

### 2. Doorbell hint after `edda request`

`edda request` currently reports only "request recorded", hiding the delivery latency. Print the honest next step:

```
Request recorded for 'billing'.
The peer will see it at their next prompt. To wake them now, use your host's
cross-session messaging (target session: d2365545-...).
```

Same hint belongs in the injected coordination protocol block next to the existing "Message a peer:" line.

### 3. notify events: `request_pending` and `task_assigned`

edda-notify currently dispatches approval_pending / phase_change / session_end / anomaly. Add:
- `request_pending` on `write_request` — so at minimum a human gets pinged (Telegram/ntfy) and can poke the target session;
- `task_assigned` on `edda task new --assignee X` — today the assignee only learns about new work from its own Stop-hook nudge.

[...]

PR #451, opened 2026-08-08
fix(bridge-claude): ack requests per message, validate and expire targets

### LINKED ISSUE #442: bug(bridge-claude): request acks are per-label, not per-message — auto-ack on render suppresses later requests from the same peer
## Summary

Two compounding defects in the coordination request/ack flow make peer messaging unreliable:

**(a) Auto-ack means "rendered once", not "seen or acted on".**
`render_coordination_protocol` collects unacked requests addressed to me and unconditionally writes `RequestAck` events for all of them immediately after rendering (`crates/edda-bridge-claude/src/peers/render_coord.rs:56-60`). If the agent's context gets compacted, the session crashes mid-turn, or the model simply ignores the block, the request is still marked acked and never reappears. The sender has no way to distinguish "delivered and handled" from "rendered into a context that never acted on it".

**(b) Acks are keyed by `from_label` only — no timestamp, no message identity.**
The ack payload contains only `from_label` (`crates/edda-bridge-claude/src/peers/heartbeat.rs:298-306`), and the unacked filter matches any ack with the same `from_label` + acker session, with no timestamp comparison (`render_coord.rs:42-47`):

```rust
.filter(|r| {
    !board
        .request_acks
        .iter()
        .any(|a| a.from_label == r.from_label && a.acker_session == session_id)
})
```

Failure scenario: peer `billing` sends request A; I render + auto-ack it. Later `billing` sends request B. B matches the existing `(billing, me)` ack and is treated as already acked — it never appears in the coordination protocol render or the pending-request nudge. The coord diff tail (`render_coord_diff`) may surface it once as a transient `[coordination update]` line, but that channel is capped at 5 events / ~200 chars and advances the offset regardless.

## Impact

`edda request` — the only peer-to-peer messaging primitive — silently drops every request after the first one from a given peer within a session. Multi-agent workflows built on the documented pattern (`docs/guides/multi-agent.md` "Peer messaging") lose messages with no error on either side.

## Suggested direction

- Give requests an identity (ts or a request id) and ack that identity, not the label.
- Only suppress a request when an ack with `ack.ts >= request.ts` (or matching id) exists.
- Consider separating "rendered" from "acknowledged" (e.g. keep `edda request-ack` as the real ack and stop auto-acking on render, or track render-count separately for re-nudging).

Found during a coordination-layer audit comparing edda's pull model against push-messaging systems.

### LINKED ISSUE #443: bug(bridge-claude): coordination requests are routed by free-string label with no validation, TTL, or dead-letter
## Summary

`edda request <to> <message>` addresses the target by a free-string label. `write_request` appends the event with zero validation (`crates/edda-bridge-claude/src/peers/heartbeat.rs:277-295`):

- **No existence check**: a typo'd label (`edda request "biling" ...`) appends a request that no session will ever match — no error, no warning on the sender side.
- **No uniqueness guarantee**: labels come from `edda claim <label>`, `EDDA_SESSION_LABEL`, or auto-derivation from branch/paths. Two sessions can hold the same label; the request then renders into both (or whichever acks first suppresses nothing for the other).
- **No TTL / dead-letter**: requests are never expired. An unacked request to a label that no longer exists sits in `coordination.jsonl` indefinitely, and `compact_coordination_log` explicitly preserves all requests and acks (`crates/edda-cli/src/cmd_gc.rs`, `board.rs:198-226`), so it survives GC forever while being invisible to every renderer.

## Impact

The sender gets identical feedback ("request recorded") whether the message will reach one peer, several peers, or nobody. Combined with #442 (per-label auto-ack), delivery of peer requests is effectively unverifiable end-to-end.

## Suggested direction

- On `edda request`, resolve the target label against active heartbeats + claims; warn (or require `--force`) when it matches zero or >1 active sessions.
- Add an expiry horizon for unacked requests (even a generous one, e.g. 7 days) so GC/compaction can drop dead letters, and surface expired-unacked requests in `edda coord` / `coord-review` as a WARN instead of hiding them.

Related: #442. Found during a coordination-layer audit.

fagemx:
Fixes #442
Fixes #443

## The bug

`edda request` is the only peer-to-peer messaging primitive, and it silently dropped every request after the first one from a given peer. Two defects in the same code chain:

**(a) Auto-ack on render.** `render_coordination_protocol` wrote `RequestAck` events for everything it had just rendered. A compacted context, a crashed turn, or a model that skipped the block all counted as "handled", and the sender had no way to tell delivery from handling.

**(b) Acks keyed by `from_label`.** The ack payload named only the peer, so the ack for message A matched message B from that same peer and suppressed it permanently.

On the send side, `write_request` took a free-string label with no validation, no uniqueness check, and no expiry — a typo'd label produced "request sent", and the dead letter survived every compaction forever.

## What changed

**Per-message identity (GH-442).** Requests carry a ULID. `edda request-ack <from>` resolves what is actually outstanding for this session and records those ids, so an ack retires exactly the messages it saw and nothing that arrives later.

**Rendering is delivery, not acknowledgement (GH-442).** The auto-ack is gone. A request keeps appearing until an explicit `edda request-ack` or until it expires, and the render now carries the ack instruction so the receiving agent knows how to clear it. An ack in the log now genuinely means an agent acted.

**One place decides delivery.** `board::partition_requests_for_session` returns `(live, expired)` and is shared by both renderers, the pending-request nudge, and the ack writer. Previously four call sites each open-coded the same filter — which is how the renderers and the nudge could disagree about what had been seen.

**Target validation (GH-443).** `edda request` resolves the label against non-stale heartbeats + claims first. Zero matches is an error listing the labels that do exist; more than one is a warning, because the message really will land in several inboxes.

**TTL and dead letters (GH-443).** Unacked requests expire after 7 days (`EDDA_REQUEST_TTL_SECS`). Expired ones are surfaced in the coordination render as `### Expired requests (unacked, aged out)` rather than hidden, and compaction drops them instead of preserving them forever.

## Trade-offs

- **Typo is an error, not a warning.** The issue offered "warn (or require `--force`)". Warning-and-send keeps the exact failure mode being fixed — identical feedback whether or not anyone is listening. But sending to a peer that has not booted yet is a legitimate fleet pattern, so `--force` keeps it available and the error message says so. If this proves annoying in practice, downgrading to a warning is a one-line change.
- **ULID over `ack.ts >= request.ts`.** The issue suggested timestamp comparison as an alternative. Rejected: `now_rfc3339` collides within a second, so a request sent in the same second as an ack would be wrongly suppressed — the original bug at finer granularity. Timestamp comparison is kept only as the fallback for legacy acks that carry no ids, where it is strictly better than the unbounded label match it replaces.
- **Requests now persist until acked.** They no longer vanish after one render. This is the point — the sender needs a real signal — but it does mean an agent that never acks will keep seeing the request until the TTL. The render caps at 3 requests and the section is inside the existing char budget.

## Backward compatibility

Existing `coordination.jsonl` files keep working. Requests without an `id` get a synthetic `legacy:{session}:{ts}` one on read; acks without `request_ids` fall back to `from_label` matching **bounded by timestamp**, so a legacy ack can only retire requests that already existed when it was written — it can no longer swallow future messages.

## Files outside the original claim

The three claimed files (`render_coord.rs`, `heartbeat.rs`, `board.rs`) could not carry the fix alone:

- `peers/mod.rs` — `RequestEntry`/`RequestAckEntry` gain fields; TTL config.
- `peers/helpers.rs` — `pending_requests_for_session` had defect (b) verbatim; the nudge is one of the two channels the issue names. Also gained the shared label resolver that three sites had duplicated.
- `edda-cli/cmd_bridge.rs`, `main.rs` — `--force` and target validation for GH-443.
- `edda-cli/tui/ui.rs` — `edda watch` marked requests `[ack]` by sender label, the same defect on a display surface.
- `docs/reference/cli.md`, `docs/guides/multi-agent.md`, `skills/coord-request.md` — the ack is now load-bearing, so the docs and the skill that teach `edda request` have to teach `edda request-ack` too.

`cmd_gc.rs` needed no change: compaction goes through `compute_board_state_for_compaction`, where the TTL now applies.

Noticed in passing, **not** changed: `main.rs` declares dead duplicates of `BridgeCmd`/`BridgeClaudeCmd` that shadow nothing — the live definitions are in `cmd_bridge.rs`. I kept the dead copy in sync rather than let it drift further; deleting it is separate work.

## Verification

- New failing-first tests, in `peers/tests.rs` unless noted:
  - `second_request_from_same_peer_survives_first_ack` — reproduces (b)
  - `render_does_not_auto_ack_requests` — reproduces (a)
  - `resolve_request_targets_matches_only_live_labels` / `..._reports_ambiguous_labels`
  - `expired_requests_are_not_pending_and_are_compacted_away`
  - `render_surfaces_expired_requests_as_warning`
  - `cmd_bridge::tests::request_to_unknown_label_is_rejected_unless_forced`
- `full_lifecycle_multi_session` updated: it asserted the auto-ack contract this PR removes.
- `cargo fmt --check` clean, `cargo clippy --workspace --all-targets` zero warnings, `cargo test -p edda-bridge-claude peers::` 113/113.
- `cargo test --workspace` on this branch fails the same pre-existing, environment-dependent tests that fail on `main` at the merge base (shared-store parallel isolation in `digest::`, `registry::`, `skill_registry::`, `cmd_search::foreign_project_uses_its_own_registered_repo`) — 8 failures here vs 28 on `main` in the same run. None involve peers/requests, and they pass single-threaded.

### DIFF
--- a/crates/edda-bridge-claude/src/peers/board.rs
+++ b/crates/edda-bridge-claude/src/peers/board.rs
@@ -1,9 +1,11 @@
 use std::fs;
 
+use super::helpers::parse_rfc3339_to_epoch;
 use super::{
-    coordination_path, BindingEntry, BoardState, ClaimEntry, CoordEvent, CoordEventType,
-    RequestAckEntry, RequestEntry, SubagentCompletedEntry,
+    coordination_path, request_ttl_secs, BindingEntry, BoardState, ClaimEntry, CoordEvent,
+    CoordEventType, RequestAckEntry, RequestEntry, SubagentCompletedEntry,
 };
+use crate::parse::now_rfc3339;
 
 // ── Board State Computation ──
 
@@ -73,7 +75,14 @@ pub fn compute_board_state(project_id: &str) -> BoardState {
                     .to_string();
                 let to_label = event.payload["to_label"].as_str().unwrap_or("").to_string();
                 let message = event.payload["message"].as_str().unwrap_or("").to_string();
+                // Pre-GH-442 requests have no id; synthesize a stable one from
+                // the fields that already identify the event on disk.
+                let id = event.payload["id"]
+                    .as_str()
+                    .map(|s| s.to_string())
+                    .unwrap_or_else(|| format!("legacy:{}:{}", event.session_id, event.ts));
                 requests.push(RequestEntry {
+                    id,
                     from_session: event.session_id,
                     from_label,
                     to_label,
@@ -86,9 +95,18 @@ pub fn compute_board_state(project_id: &str) -> BoardState {
                     .as_str()
                     .unwrap_or("")
                     .to_string();
+                let request_ids: Vec<String> = event.payload["request_ids"]
+                    .as_array()
+                    .map(|arr| {
+                        arr.iter()
+                            .filter_map(|v| v.as_str().map(|s| s.to_string()))
+                            .collect()
+                    })
+                    .unwrap_or_default();
                 request_acks.push(RequestAckEntry {
                     acker_session: event.session_id,
                     from_label,
+                    request_ids,
                     ts: event.ts,
                 });
             }
@@ -158,6 +176,74 @@ pub fn compute_board_state(project_id: &str) -> BoardState {
     }
 }
 
+// ── Request Delivery Rules ──
+
+/// Current wall clock as epoch seconds, on the same coarse scale the peers
+/// module uses for every other event timestamp.
+fn now_epoch() -> u64 {
+    parse_rfc3339_to_epoch(&now_rfc3339()).unwrap_or(0)
+}
+
+/// True when `session_id` has acknowledged this specific request.
+///
+/// Acks name the request ids they cover (GH-442), so acking one message from a
+/// peer no longer suppresses that peer's later messages. Legacy acks carry no
+/// ids; they fall back to label matching bounded by timestamp, so they can only
+/// suppress requests that already existed when the ack was written.
+fn is_acked(board: &BoardState, request: &RequestEntry, session_id: &str) -> bool {
+    board.request_acks.iter().any(|a| {
+        if a.acker_session != session_id {
+            return false;
+        }
+        if a.request_ids.is_empty() {
+            a.from_label == request.from_label
+                && parse_rfc3339_to_epoch(&a.ts).unwrap_or(0)
+                    >= parse_rfc3339_to_epoch(&request.ts).unwrap_or(0)
+        } else {
+            a.request_ids.iter().any(|id| id == &request.id)
+        }
+    })
+}
+
+/// True when an unacked request has aged past the dead-letter horizon.
+/// An unparseable timestamp never expires — losing a message is worse than
+/// keeping a stale one.
+fn is_expired(request: &RequestEntry, now: u64, ttl: u64) -> bool {
+    match parse_rfc3339_to_epoch(&request.ts) {
+        Some(ts) => now.saturating_sub(ts) > ttl,
+        None => false,
+    }
+}
+
+/// Split the unacked requests addressed to `my_label` into `(live, expired)`.
+///
+/// The single place delivery is decided: every renderer and the pending-request
+/// nudge go through here, so they cannot drift apart on what counts as acked.
+pub fn partition_requests_for_session<'a>(
+    board: &'a BoardState,
+    session_id: &str,
+    my_label: &str,
+) -> (Vec<&'a RequestEntry>, Vec<&'a RequestEntry>) {
+    if my_label.is_empty() {
+        return (Vec::new(), Vec::new());
+    }
+    let now = now_epoch();
+    let ttl = request_ttl_secs();
+    let mut live = Vec::new();
+    let mut expired = Vec::new();
+    for r in board.requests.iter().filter(|r| r.to_label == my_label) {
+        if is_acked(board, r, session_id) {
+            continue;
+        }
+        if is_expired(r, now, ttl) {
+            expired.push(r);
+        } else {
+            live.push(r);
+        }
+    }
+    (live, expired)
+}
+
 /// Compact coordination.jsonl: compute current state and return as JSONL lines.
 /// Used by GC to shrink the append-only log.
 pub fn compute_board_state_for_compaction(project_id: &str) -> Vec<String> {
@@ -195,12 +281,20 @@ pub fn compute_board_state_for_compaction(project_id: &str) -> Vec<String> {
         }
     }
 
-    for request in &board.requests {
+    // Requests and acks past the dead-letter horizon are dropped rather than
+    // carried forward forever (GH-443). An ack is never older than the request
+    // it answers, so TTL-ing both on the same horizon cannot resurrect a
+    // request by outliving its ack.
+    let now = now_epoch();
+    let ttl = request_ttl_secs();
+
+    for request in board.requests.iter().filter(|r| !is_expired(r, now, ttl)) {
         let event = CoordEvent {
             ts: request.ts.clone(),
             session_id: request.from_session.clone(),
             event_type: CoordEventType::Request,
             payload: serde_json::json!({
+                "id": request.id,
                 "from_label": request.from_label,
                 "to_label": request.to_label,
                 "message": request.message,
@@ -211,13 +305,18 @@ pub fn compute_board_state_for_compaction(project_id: &str) -> Vec<String> {
         }
     }
 
-    for ack in &board.request_acks {
+    for ack in board.request_acks.iter().filter(|a| {
+        parse_rfc3339_to_epoch(&a.ts)
+            .map(|ts| now.saturating_sub(ts) <= ttl)
+            .unwrap_or(true)
+    }) {
         let event = CoordEvent {
             ts: ack.ts.clone(),
             session_id: ack.acker_session.clone(),
             event_type: CoordEventType::RequestAck,
             payload: serde_json::json!({
                 "from_label": ack.from_label,
+                "request_ids": ack.request_ids,
             }),
         };
         if let Ok(line) = serde_json::to_string(&event) {
--- a/crates/edda-bridge-claude/src/peers/heartbeat.rs
+++ b/crates/edda-bridge-claude/src/peers/heartbeat.rs
@@ -4,11 +4,11 @@ use std::io::Write;
 use crate::parse::now_rfc3339;
 use crate::signals::SessionSignals;
 
-use super::board::compute_board_state;
-use super::helpers::auto_label;
+use super::board::{compute_board_state, partition_requests_for_session};
+use super::helpers::{auto_label, parse_rfc3339_to_epoch, session_label_from_board};
 use super::{
-    coordination_path, detect_git_branch_in, env_label, heartbeat_path, BindingConflict,
-    CoordEvent, CoordEventType, SessionHeartbeat,
+    coordination_path, detect_git_branch_in, env_label, heartbeat_path, stale_secs,
+    BindingConflict, CoordEvent, CoordEventType, SessionHeartbeat,
 };
 
 // ── Heartbeat Write/Read ──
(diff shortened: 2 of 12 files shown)


fagemx:
Controller review (first pass — adversarial verifier pass pending):

**Structural win.** Delivery decisions collapse into one function (`partition_requests_for_session`) and all five previously-divergent ack-comparison sites route through it (both protocol renders, peer updates, the PreToolUse nudge) or mirror its semantics (TUI, with the legacy ts-bounded fallback). The class of "fixed 4 of 5 sites" bugs is now structurally impossible. Auto-ack removed with the right semantics: rendering is delivery, not acknowledgement.

**Compat verified against the checklist:** legacy requests get a synthetic stable id on read; legacy id-less acks fall back to label matching bounded by timestamp (cannot swallow future messages); compaction round-trips `id`/`request_ids` and drops both sides of the TTL horizon with a sound no-resurrection argument. Target validation ships with the `--force` escape and did-you-mean labels; expiry surfaces as dead letters instead of silent loss. Docs and the coord-request skill updated to match — including the ack-when-handled teaching, which closes half of #446's step 4.

**Two open items before merge:**
1. The three single-pair tests flagged in the #442 review comment (`compaction_preserves_request_acks`, `render_coordination_filters_acked_requests`, `peer_updates_filters_acked_requests`) were not extended. The new tests cover the render/pending paths via other routes, and the ts-bounded legacy fallback structurally protects the compaction case — but the fixated criterion (ack A → same peer sends B → compact → recompute → B still pending) has no direct test. Either add that one scenario to `compaction_preserves_request_acks` or argue in the PR body why the ts-bound makes it redundant. Verifier will rule on severity.
2. Interplay with #448: empty-paths presence claims now render as `Your scope: **<label>** ()` — the rendering guard requested alongside this PR's scope (suggest a "no paths claimed yet" hint) isn't in this diff. Small, same file, should land here.

Scope note: cmd_bridge/main/docs/skill changes exceed the original claimed file set but are the unavoidable CLI surface of GH-443 — accepted, and they collided with nobody.

fagemx:
Review criteria from the verifier audit addressed in 5559900. Point by point:

**5 ack-comparison sites** — all covered:

| # | Site | Now |
|---|------|-----|
| 1 | `render_coord.rs:42-47` (drives auto-ack) | deleted — render no longer acks |
| 2 | `render_coord.rs:267-272` | `partition_requests_for_session` |
| 3 | `render_coord.rs:386-391` | `partition_requests_for_session` |
| 4 | `helpers.rs:29-34` | `partition_requests_for_session` |
| 5 | `tui/ui.rs:273-277` | per-id match; legacy branch also requires the acker to be the addressee |

Sites 2–4 are now literally the same function call, so they cannot drift apart — that was the "acked but never rendered" hazard in the audit.

**Site 5, phantom `[ack]`** — the audit flagged that it ignores `acker_session`. For id-bearing acks the id is specific, so acker identity adds nothing. For legacy ackless-id acks it does matter: an ack written by a third session for its own mail from that peer would match on `from_label` + ts alone. The legacy branch now additionally requires the acker's claim label to equal `to_label`. That can under-report `[ack]` for a legacy heartbeat-only acker with no claim — deliberate, since a missing `[ack]` on a board view is recoverable and a phantom one is not.

**Old-log compatibility** — id-less legacy acks fall back to ts-compare, exactly the "not the third option" the audit called out: not "no id → not an ack" and not "no id → acks everything". `RequestAckEntry.ts` still comes from `event.ts` (`board.rs:92`), unchanged. Requests with no id get a synthetic `legacy:{session}:{ts}`.

**Same field, same clock** — expiry and the legacy ack comparison both go through `parse_rfc3339_to_epoch` on the event `ts`. Primary ack matching is by id, which sidesteps the clock entirely; ULID rather than pure ts-compare because `now_rfc3339` collides within a second, and a request sent in the same second as an ack would be wrongly suppressed — the original bug at finer granularity.

**Compaction round-trip** — `board.rs` re-serialization carries `id` and `request_ids`. `compaction_preserves_request_acks` now builds a log with **two** messages from the same peer (ack A, then B arrives), compacts, recomputes the board, and asserts the unacked set is unchanged — B still pending, A still acked. That is the test the audit required; without the identity fields in re-serialization it fails.

**False-green tests** — all three extended with the second-message scenario:
- `compaction_preserves_request_acks` (compact → recompute → assert)
- `render_coordination_filters_acked_requests`
- `peer_updates_filters_acked_requests`

**Auto-ack semantics** — the preferred resolution, implemented: render does not ack, `edda request-ack` is the real ack, unacked requests keep re-nudging until acked or expired.

**GH-443 criteria** — typo label exits non-zero via `anyhow::bail!` and never prints "sent"; two live sessions on one label warns and proceeds; happy path output unchanged; expired requests are dropped by compaction and rendered as `### WARN: N expired request(s) — unacked past the horizon, from <senders>`.

**Added scope: pathless claim render.** PR #448 turns branch-fallback auto-claim into a claim with empty paths, which `render_coord.rs` would print as `Your scope: **main** ()`. Now renders `Your scope: **main** (no paths claimed yet — \`edda claim "main" --paths "src/..."\`)`. Covered by `render_pathless_claim_says_what_is_missing`.

**Note on `render_coord.rs:28-34`** — that `my_label` block is gone rather than modified. It existed only to feed the auto-ack, and no other caller read it. The claim → heartbeat fallback **order** is preserved everywhere it still matters: `render_coordination_protocol_with` keeps its own resolution untouched, and the shared `helpers::session_label_from_board` uses the same order. Nothing for #448 to collide with.

**Verification** (per the `--test-threads=1` convention, GH-447):
- `cargo test -p edda-bridge-claude -- --test-threads=1` → **556 passed, 0 failed**
- `cargo test -p edda -- --test-threads=1` → 211 passed, 1 failed: `cmd_search::tests::foreign_project_uses_its_own_registered_repo`, which fails identically on `main` when run alone (tempdir path mismatch, unrelated to this change).
- `cargo fmt --check` clean, `cargo clippy --workspace --all-targets` zero warnings.
- No `std::env::set_var` in any test added by this PR.

fagemx:
Correction to my review two comments up: items 1 (compaction round-trip) and 2 (pathless-claim render guard) were already present in 5559900 — my review was written against the first push and crossed the worker's update. My miss, not theirs.

Worker's technical correction also accepted: the extended `compaction_preserves_request_acks` exercises the **id path** (ack carries `request_ids`), not the ts-bounded fallback — those are distinct code paths needing separate pinning. a7aff20 adds exactly that: `legacy_ackless_id_log_survives_compaction_without_swallowing_later_requests` builds an id-less, timestamp-separated log and asserts compaction neither fabricates ids nor changes the unacked set. Pure test addition, no production change.

Review target is now a7aff20. Push freeze on this branch until the verifier verdict.

fagemx (commented):
## Independent verification (verifier lane)

Reviewed on a clean worktree at `5559900`, without reading the controller's review first.

### Quality gates

| Gate | Result |
|---|---|
| `cargo fmt --check` | PASS |
| `RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets` | PASS |
| `cargo test -p edda-bridge-claude -- --test-threads=1` | PASS — 556 passed, 0 failed (main baseline: 549) |
| `cargo test -p edda --bin edda -- --test-threads=1` | 211 passed, 1 failed — `cmd_search::tests::foreign_project_uses_its_own_registered_repo`, which I reproduced standalone on unmodified `main`. Pre-existing (GH-450), unrelated to this PR. The new `request_to_unknown_label_is_rejected_unless_forced` passes. |

### The design holds up

Collapsing delivery into `partition_requests_for_session` is the right shape. I had flagged five independent ack-matching sites during prep; four of them (both renderers, the pending-request nudge, and `write_request_ack` itself) now call that one function, and `render_coordination_protocol_with` has no residual second computation — the old inline filter is gone from both renderers. Dropping auto-ack on render is the part that actually fixes "rendered ≠ handled", and the doc comment says so plainly.

Two things I checked adversarially and found correct:

**Legacy fallback is bounded in the safe direction.** `is_acked` falls back to `from_label` + `ack.ts >= request.ts` only when `request_ids` is empty, so a legacy ack can suppress messages that existed when it was written and nothing after. That is the behaviour the issue needed, and it works on old logs with no migration because `RequestAckEntry.ts` was always populated from the top-level `event.ts`.

**The compaction TTL invariant is sound.** Requests are dropped at `now - ts > ttl` while acks are kept at `now - ts <= ttl`. Since an ack is never older than the request it answers, any request that survives compaction (`r.ts >= now - ttl`) implies `a.ts >= r.ts >= now - ttl`, so its ack survives too. A request cannot be resurrected by outliving its ack. The comment claims this; it checks out.

Also confirmed: `id` and `request_ids` are both carried through `compute_board_state_for_compaction`, so `edda gc` no longer strips the identity — that was the specific way this bug could have come back.

### Test coverage — the three round-trip cases are covered

Noting this explicitly because it was raised as an open question: all three tests I had flagged as false-greens **were** extended, and with the exact shape needed.

- `compaction_preserves_request_acks` now sends a second message from the same peer *before* compacting, then asserts `pending_after.len() == 1` and that the survivor is the later message. A single request/ack pair renders identically under per-label and per-message matching; this one does not.
- `render_coordination_filters_acked_requests` and `peer_updates_filters_acked_requests` both gained the same second-message case.
- `legacy_ackless_id_log_survives_compaction_without_swallowing_later_requests` covers the pre-fix on-disk shape end to end, including the compaction round-trip — and correctly pins the ack to the timestamp-bounded path rather than letting it degrade to "acks nothing" or "acks everything".

### WARN (minor) — the TUI's timestamp comparison is not equivalent to the epoch path

`crates/edda-cli/src/tui/ui.rs` compares `a.ts >= r.ts` as strings, while `board.rs::is_acked` compares through `parse_rfc3339_to_epoch`. These are not the same ordering:

`parse_rfc3339_to_epoch` (`peers/helpers.rs:161-175`) reads fixed byte offsets up to `[17..19]` and **discards fractional seconds**, so it is second-granular. The `time` crate's RFC3339 formatter emits fractional digits whenever the nanosecond component is non-zero. Within a single second the two paths disagree: the epoch path sees the ack and the request as equal and `>=` reports acked, while the string compare is sub-second precise and reports not-acked. (There is also a nanos-exactly-zero case where `'Z'` sorts above `'.'` and the string compare calls an earlier ack newer — about 1 in 10^7, mentioned only for completeness.)

Impact is small and I would not block on it: display-only (`[ack]` in `edda watch`), legacy branch only, sub-second window. Worth fixing by exporting a shared comparison rather than reimplementing it in the TUI, so the two cannot drift further.

The same second-granularity applies to `is_acked`'s own legacy branch: an ack and a strictly-later request landing in the same second compare equal, so `>=` suppresses the later one. Same ≤1s window, same low severity — but it is the one place the legacy path can still swallow a message it should not.

### WARN (minor) — a new ack with no resolvable ids is indistinguishable from a legacy ack

When `write_request_ack` runs on a session whose label does not resolve, `partition_requests_for_session` returns nothing, `request_ids` is empty, and the event is written as `"request_ids": []`. On read, `#[serde(default)]` makes that identical to a legacy ack that has no `request_ids` key at all, so it takes the label+timestamp path.

The important half is safe — the timestamp bound means it still cannot swallow anything sent afterwards. The residual edge: an ack written while the session is unidentified will retroactively suppress messages once that session later acquires a label, if those messages predate the ack. Hard to reach in practice (an agent only acks what it saw, and seeing a request requires a label), but the fix is cheap — `Option<Vec<String>>` distinguishes "key absent, genuinely legacy" from "new ack, nothing pending", and only the former needs the fallback.

### Notes

- The `#448` interaction guard is present, not pending: `render_coord.rs` renders a pathless claim as `no paths claimed yet` with the fix-up command, covered by `render_pathless_claim_says_what_is_missing`.
- The TUI's legacy branch verifies the acker by joining `board.claims`. After #448, a solo session that ends writes `Unclaim` and its claim leaves the fold, so a previously-`[ack]`-marked legacy request reverts to showing unacked. Display-only, legacy-only. The id branch returns before that join, which is sound — ids are only ever recorded for requests addressed to the acking session.
- Expired requests surface as a WARN line in the protocol render, but the TUI request list has no expired marker; they simply sit unacked until GC drops them. Small gap in the dead-letter story.
- `request()` discards the id now returned by `write_request`. Harmless, just noting it is unused at that call site.

### Verdict

No blocking findings. GH-442 and GH-443 are both properly closed, the legacy-compatibility and compaction round-trip cases that worried me most are covered by tests that would actually fail if the fix regressed, and the two WARNs above are narrow enough to take as follow-ups.

fagemx (commented):
## Addendum — review re-confirmed at `a7aff20`, plus stacked smoke

My review above was written against `5559900`. The delta to `a7aff20` is test-only (+77 lines, no production change), so every finding stands unchanged.

The addition — `legacy_ackless_id_log_survives_compaction_without_swallowing_later_requests` — closes a gap worth naming: the earlier round-trip test drove the **id** path, so compaction was only proven to preserve identity for acks that carry `request_ids`. The legacy branch reaches the same conclusion by a different mechanism (timestamp bound), and "structurally protected" is not the same as "covered". This test drives that second path through compaction and pins the ack to the timestamp-bounded behaviour rather than letting it degrade to acking nothing or acking everything from the sender. Right call to add it.

### Stacked verification (`b126894` + `a7aff20`)

Merged both branches in a scratch worktree — **no conflicts**, including in `peers/heartbeat.rs` and `peers/tests.rs`, which both PRs touch.

| Gate (stacked) | Result |
|---|---|
| `cargo fmt --check` | PASS |
| `RUSTFLAGS=-Dwarnings cargo clippy --workspace --all-targets` | PASS |
| `cargo test -p edda-bridge-claude -- --test-threads=1` | PASS — 565 passed, 0 failed |
| `cargo test -p edda --bin edda -- --test-threads=1` | 211 passed, 1 failed — `cmd_search::tests::foreign_project_uses_its_own_registered_repo`, reproduced standalone on unmodified `main`. Pre-existing (GH-450). |

The interaction that mattered is directly exercised by the stacked run: #448's `fresh_session_receives_requests_addressed_to_its_branch` resolves its label through #451's `partition_requests_for_session` → `session_label_from_board`, which finds no claim and falls back to the heartbeat's branch label. Delivery to a claimless fresh session works with both changes applied.

The pathless-claim render guard keeps its value under #448's new route: auto-claim no longer produces an empty-paths claim, but `edda claim <label>` takes `--paths` as an optional `Vec<String>`, so a manual claim with no paths is still reachable and still renders the "no paths claimed yet" hint.

Issue #28, opened 2026-02-21
feat(ledger): decisions table with auto-extraction and supersede tracking


fagemx:
## Context

Decisions are currently stored as generic note events with tag `["decision"]` and a structured `payload.decision` field. Supersede resolution is **copy-pasted in 3 places** (`cmd_query.rs:43`, `context.rs:404`, `snapshot.rs:159`) — each building the same `HashSet<superseded_ids>` and filtering.

Domain grouping (e.g., all `db.*` decisions) requires full scan + string prefix matching. No index, no pre-extraction.

## Proposal

Add a `decisions` table to the SQLite ledger (#27) that auto-extracts structured fields:

```sql
CREATE TABLE decisions (
    event_id TEXT PRIMARY KEY REFERENCES events(event_id),
    key TEXT NOT NULL,           -- "db.engine"
    value TEXT NOT NULL,         -- "postgres"
    reason TEXT DEFAULT '',
    domain TEXT NOT NULL,        -- "db" (auto-extracted: key up to first '.')
    supersedes_id TEXT,          -- event_id of prior decision
    is_active BOOLEAN DEFAULT TRUE
);
CREATE INDEX idx_decision_key ON decisions(key);
CREATE INDEX idx_decision_domain ON decisions(domain);
CREATE INDEX idx_decision_active ON decisions(is_active);
```

### Auto-extraction trigger

On INSERT into `events` where payload contains `"decision"` tag:
1. Extract `key`, `value`, `reason` from `payload.decision`
2. Compute `domain` = substring before first `.`
3. Find prior decision with same `key` on same `branch` → set `supersedes_id`
4. Set prior decision's `is_active = FALSE`

### What this eliminates

- 3x copy-pasted supersede resolution → one SQL query: `WHERE is_active = TRUE`
- `find_prior_decision()` reverse scan → `SELECT FROM decisions WHERE key = ? AND is_active`
- Domain grouping full scan → `WHERE domain = ?`

### Queries enabled

```sql
-- All active decisions
SELECT key, value, reason, domain FROM decisions WHERE is_active ORDER BY domain, key

-- Domain browse
SELECT * FROM decisions WHERE domain = 'db' AND is_active ORDER BY key

-- Supersede timeline
SELECT d.*, e.ts FROM decisions d JOIN events e ON d.event_id = e.event_id
WHERE d.domain = 'auth' ORDER BY e.ts

-- Find what superseded what
SELECT d1.key, d1.value as current, d2.value as previous
[...]

Issue #389, opened 2026-07-13
feat(cli): intent-mapped unknown-subcommand errors and edda show <key> exact lookup


fagemx:
## Problem

Edda's primary users are coding agents, and agents arrive with CLI muscle
memory trained on git/kubectl/gh: `git show <sha>`, `kubectl get`,
`gh pr view`, `docker inspect`.

Real incident from a multi-agent fleet that runs edda on every session: an
agent wanted to inspect one signed decision by its key and instinctively ran

```
edda show "fleet.skill-usage-law"
→ error: unrecognized subcommand 'show'
```

then burned a full round reading the command table to self-correct to
`edda ask`. This will repeat for every fresh agent, forever — when your
user base is LLMs, discoverability means landing inside the distribution of
what models will guess.

Two distinct gaps hide in this one incident:

1. **Error affordance.** For an agent, `unrecognized subcommand` is a dead
   end — stderr *is* the documentation agents actually read. clap's built-in
   suggestions only catch string-distance typos (`shwo`), not intent misses
   (`show` → `ask`).
2. **A missing verb.** `edda ask` is *search* (keyword → related decisions).
   "Display exactly this record, I already know its key" is *lookup* — a
   different intent. git has both (`git log --grep` vs `git show <sha>`);
   edda has ask/log/export but no first-class exact-key lookup. `show`
   exists only as `edda draft show`.

## Proposed solution

1. **Intent-mapped unknown-subcommand errors** (cheapest, highest value): a
   small table of common wrong verbs appended to the error —

   ```
   error: unrecognized subcommand 'show'
   hint: to query decisions use `edda ask "<key or keyword>"`;
         for draft contents use `edda draft show`
   ```

   Map at least: `show` / `get` / `view` / `describe` / `inspect` → `ask`;
   `history` → `log`. One line of stderr saves every new agent one
   round-trip.

2. **`edda show <key>`** as a top-level command: exact-key lookup of a single
   decision/event — value, reason, dates, status, supersession chain.
   Clean division of labor: `ask` = search, `show` = lookup.

3. *(Optional, telemetry-flavored)* Record unrecognized-subcommand attempts
   as ledger events. Wrong guesses from agents are free UX research — over
   time the distribution of guessed verbs tells you which commands to add
   next.

## Alternatives considered

- **Document "use ask, not show" in agent guides / CLAUDE.md** — helps the
  fleet that reads the guide, does nothing for a fresh agent's first guess.
  Priors beat docs.
[...]

PR #459, opened 2026-08-14
fix: reliability wave 1 redaction and test isolation

### LINKED ISSUE #357: [chore] Add secret redaction patterns for Slack, Stripe, npm
`crates/edda-bridge-claude/src/redact.rs` covers API keys, GitHub/GitLab tokens, AWS keys, Bearer tokens.

Missing patterns for:
- Slack tokens (`xoxb-`, `xoxp-`)
- Stripe keys (`sk_live_`, `pk_live_`)
- npm tokens (`npm_`)

### LINKED ISSUE #447: test(bridge-claude): dispatch/digest tests flake under parallel execution — 23 env::set_var sites mutate process-global state

### LINKED ISSUE #450: test(workspace): tests read the real ~/.edda store — machine-state-dependent failures on clean main

### LINKED ISSUE #452: bug(edda-store): registry fleet_scope assertion fails, then ENV_STORE_LOCK poisoning cascades 1 real failure into 18
Split from #450 after the verifier's independent reproduction corrected the original diagnosis.

## Two distinct defects

**(a) The one real failure:** `registry.rs:346` `assert_eq!(scope.len(), 2)` gets 1 — deterministic on this machine, serial or parallel, but passes when the test runs alone. Root cause not yet identified; state leakage between tests in the same process is the leading suspect (the isolation helper `with_isolated_store` does point `EDDA_STORE_ROOT` at a tempdir, so it is NOT raw real-store sharing — something else persists across tests in-process).

**(b) The cascade:** that assertion panics while the test holds `ENV_STORE_LOCK`; `registry.rs:232` acquires it with `.lock().unwrap()`, so every subsequent test in the binary dies on `PoisonError` — 17 false failures obscuring the 1 real one. Serial execution cannot help; poisoning is process-permanent.

## Fix shape

- (b) first, it's one line and pure win: `.lock().unwrap_or_else(|e| e.into_inner())` (the guarded state is a store-root path env var — safe to recover). That alone turns 18 red into 1 red so (a) becomes debuggable.
- (a) then: bisect what in-process state survives `with_isolated_store` teardown (env var restore order? registry cache? cwd?).

Found by the verifier session during PR #448/#451 gate reruns; details and repro counts in #450's correction comment.

fagemx:
Closes #357
Closes #447
Closes #450
Closes #452

Summary:
- redact Slack, Stripe, and npm token formats
- isolate Claude digest/search/store fixtures from the real user store
- serialize all store-touching dispatch test boundaries with the existing poison-recovering guard
- keep default-parallel bridge tests deterministic without weakening coverage

Verification at `a94adc36a6c4381a775cfc23cc51a31c4e299eee`:
- GitHub CI: 7/7 green on Ubuntu, macOS, and Windows
- independent verifier task #28: PASS; 64/64 store-touching dispatch test boundaries guarded
- final exact-SHA synthetic integration with PR #460 and PR #461: fmt PASS, clippy `-D warnings` PASS, default-parallel `cargo test --workspace` PASS

### DIFF
--- a/crates/edda-bridge-claude/src/digest/tests.rs
+++ b/crates/edda-bridge-claude/src/digest/tests.rs
@@ -11,10 +11,60 @@ use super::*;
 use std::io::Write;
 use std::sync::{MutexGuard, PoisonError};
 
-fn env_guard() -> MutexGuard<'static, ()> {
-    crate::ENV_LOCK
+struct EnvGuard {
+    _lock: MutexGuard<'static, ()>,
+    previous_store_root: Option<std::ffi::OsString>,
+    _store_root: tempfile::TempDir,
+}
+
+impl Drop for EnvGuard {
+    fn drop(&mut self) {
+        match self.previous_store_root.take() {
+            Some(root) => std::env::set_var("EDDA_STORE_ROOT", root),
+            None => std::env::remove_var("EDDA_STORE_ROOT"),
+        }
+    }
+}
+
+impl EnvGuard {
+    fn new(lock: MutexGuard<'static, ()>) -> Self {
+        let previous_store_root = std::env::var_os("EDDA_STORE_ROOT");
+        let store_root = tempfile::tempdir().unwrap();
+        std::env::set_var("EDDA_STORE_ROOT", store_root.path());
+        Self {
+            _lock: lock,
+            previous_store_root,
+            _store_root: store_root,
+        }
+    }
+}
+
+fn env_guard() -> EnvGuard {
+    let lock = crate::ENV_LOCK
         .lock()
-        .unwrap_or_else(PoisonError::into_inner)
+        .unwrap_or_else(PoisonError::into_inner);
+    EnvGuard::new(lock)
+}
+
+#[test]
+fn env_guard_isolates_store_root_and_restores_on_drop() {
+    let lock = crate::ENV_LOCK
+        .lock()
+        .unwrap_or_else(PoisonError::into_inner);
+    let before = std::env::var_os("EDDA_STORE_ROOT");
+    let guard = EnvGuard::new(lock);
+    let inside = std::env::var_os("EDDA_STORE_ROOT");
+    assert_ne!(inside, before, "fixture needs a private store root");
+    drop(guard);
+
+    let _lock = crate::ENV_LOCK
+        .lock()
+        .unwrap_or_else(PoisonError::into_inner);
+    assert_eq!(
+        std::env::var_os("EDDA_STORE_ROOT"),
+        before,
+        "fixture must restore the caller environment"
+    );
 }
 
 fn write_session_ledger(dir: &Path, lines: &[serde_json::Value]) -> std::path::PathBuf {
--- a/crates/edda-bridge-claude/src/digest/tests.rs
+++ b/crates/edda-bridge-claude/src/digest/tests.rs
@@ -9,6 +9,13 @@ use super::render::*;
 use super::*;
 
 use std::io::Write;
+use std::sync::{MutexGuard, PoisonError};
+
+fn env_guard() -> MutexGuard<'static, ()> {
+    crate::ENV_LOCK
+        .lock()
+        .unwrap_or_else(PoisonError::into_inner)
+}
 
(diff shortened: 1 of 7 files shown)


fagemx:
Controller review at frozen head \28589a87697898b5c06e3eb6fe6a6ef398ea18bd\: **Changes requested**.\n\n- Scope: GH-357/#447/#450/#452 are in range; GH-389 correctly is not.\n- Reality: APIs and production call paths exist.\n- Testing blocker: \crates/edda-bridge-claude/src/digest/tests.rs\ still writes via the real \dda_store::project_dir\. The added \ENV_LOCK\ only serializes those tests; it does not redirect \EDDA_STORE_ROOT\, so #450's machine-state pollution remains. The reported workspace pass used an external TEMP/TMP redirection and does not close this gap.\n- YAGNI: no production abstraction or dependency is needed; isolate the digest fixture store, restore the prior env value safely, then rerun the default-parallel bridge and workspace gates and refreeze at a new full SHA.\n\nIndependent verifier verdict matches: Changes Requested. Any push voids this frozen-SHA verdict.

PR #479, opened 2026-08-17
fix(skills): own fleet-orchestrate global distribution

fagemx:
## Summary
- add an explicit Windows PowerShell Status/Sync surface for Codex, Claude, or both
- classify complete-directory installs as missing, current, stale, or locally modified using exact provenance and Git-history bootstrap
- stage and verify replacements, preserve forced backups, and document explicit distribution outside `edda init`

## Root cause
`skills/fleet-orchestrate/**` was canonical but had no owner for user-level distribution, so merged policy changes could leave both global installs stale.

## Validation
- RAN: `pwsh -NoProfile -File scripts/skills/self-test-fleet-orchestrate-sync.ps1`
- RAN: PowerShell parser validation for both scripts
- RAN: staged diff/link/scope check for the three issue-owned files
- RAN (read-only diagnostic): live Status classified both existing global installs as stale

No Cargo/workspace gates were run because this changes only the explicit PowerShell/docs distribution surface. No real global install was synced.

Closes #478

PR #480, opened 2026-08-18
test(edda-conductor): fix ETXTBSY flake in codex_app_server fake server

fagemx:
## Problem

CI run [32002599039](https://github.com/fagemx/edda/actions/runs/32002599039) for `main@7f5d1555b4c00e4fe6ac9a452a3c81c62289972c` failed **only** in `Test (ubuntu-latest)` (165 passed, 1 failed, 1 ignored in `edda-conductor`). PR #479, the merged head, touched only `scripts/skills/*.ps1` and docs, so this is a pre-existing Linux-only flake, not a regression from that PR.

The test name made it look like a PID-visibility problem, but the raw log shows it never reached the PID check — it failed on the very first line:

```
---- agent::codex_app_server::tests::dropping_concrete_client_makes_child_pid_disappear stdout ----
Error: failed to spawn Codex App Server

Caused by:
    Text file busy (os error 26)
```

## Root cause — ETXTBSY (commit 2, the actual fix)

The unix branch of `fake_app_server` writes `fake-app-server`, chmods it `0755`, and then `execve`s it. Linux fails `execve` with `ETXTBSY` while any process holds the target file open for writing. In a threaded test binary, a concurrent `Command::spawn` from another test can be sitting between `fork` and `exec` holding an **inherited write fd** for this very file — `O_CLOEXEC` clears it at exec, not at fork.

Whichever test happens to land inside that window loses, which is why **the failure moves between tests**: run 32002599039 hit `dropping_concrete_client_makes_child_pid_disappear`, and the first push on this branch hit `cancelling_concrete_method_makes_child_pid_disappear` with the byte-identical error.

**Fix:** feed the script to `sh` instead of exec'ing it, mirroring what the Windows branch already does with `powershell.exe -File`. `sh` opens the script read-only, so the window cannot arise — the race is removed by construction rather than retried around. The chmod is no longer needed, and the `#!/bin/sh` line is simply a comment when the script is passed as an argument.

Verified on real Linux — with a writer holding the fd open:

| | result |
|---|---|
| old: exec the script | **FAILED errno=26 Text file busy** |
| new: `/bin/sh script` | ok |

and once the writer closes, both succeed — which is precisely why the old path failed only rarely.

## Second, latent flake — zombie PIDs (commit 1)

While investigating I found a separate real hazard in the same helper, matching the failure mode the test name suggests. `wait_for_pid_exit` probed liveness with `kill -0`, which **succeeds for zombies**: `drop(server)` sends SIGKILL and hands the `Child` to tokio's *global* orphan queue, and nothing ever calls `waitpid` on it, since the drop consumes the handle. The killed child stays in the table as a zombie until that queue drains — which needs a runtime's process driver to park *and* the lazily-registered global SIGCHLD watch to fire, an event that has to cross runtimes in a parallel test binary.

`dropping_concrete_client_...` is the only pid-polling test here with no reap path; every other one reaps first via `terminate()`, `try_wait()`, or `wait_for_child_exit`. On Linux tokio 1.49 takes the `PidfdReaper` branch and never registers SIGCHLD at spawn, leaving that lazy path as the only reaper; macOS wires SIGCHLD up front; Windows evicts terminated processes outright.

The probe now asks `ps -o state=` for the state: a zombie has terminated so it no longer counts as running, while a child that drop failed to kill still does — the behavior contract is unchanged, and a genuine "drop didn't kill it" regression still fails. It also errors instead of reporting "gone" when it cannot run, so a broken environment fails loudly rather than passing vacuously.

Validated on Linux with the exact shipped logic:

| case | `process_is_alive` | probe |
|---|---|---|
| running child | **true** | rc=0, state `S` |
| zombie (SIGKILLed, unreaped) | **false** | rc=0, state `Z` |
| reaped | false | rc=1 |
| never existed | false | rc=1 |

To be clear about provenance: this one is a latent hazard I have **not** observed firing in CI. It is separable — happy to drop commit 1 if you'd rather keep the PR to the ETXTBSY fix alone.

Test-only change; no production code touched.

## Verification

- `cargo test -p edda-conductor codex_app_server` — 18 passed, 0 failed, 1 ignored
- `cargo clippy -p edda-conductor --all-targets -- -D warnings` — clean
- `cargo fmt -p edda-conductor -- --check` — clean

Windows compiles only the `#[cfg(windows)]` arms, so every `#[cfg(unix)]` / `#[cfg(not(windows))]` arm in the tests module was type-checked separately by temporarily swapping the cfg conditions and re-running clippy `-D warnings` — clean, then reverted.

The first push on this branch (commit 1 alone) is itself useful evidence: `dropping_concrete_client_makes_child_pid_disappear` **passed** on ubuntu-latest, and the ETXTBSY failure surfaced on a different test in the same binary.

### DIFF
--- a/crates/edda-conductor/src/agent/codex_app_server.rs
+++ b/crates/edda-conductor/src/agent/codex_app_server.rs
@@ -519,12 +519,15 @@ mod tests {
 
     async fn wait_for_pid_exit(pid: u32) -> anyhow::Result<()> {
         tokio::time::timeout(std::time::Duration::from_secs(5), async move {
-            while process_exists(pid) {
+            loop {
+                if !process_is_alive(pid)? {
+                    return Ok(());
+                }
                 tokio::time::sleep(std::time::Duration::from_millis(25)).await;
             }
         })
         .await
-        .context("fake app-server PID did not disappear")
+        .context("fake app-server PID kept running")?
     }
 
     async fn wait_for_child_exit(child: &mut Child) -> anyhow::Result<()> {
@@ -540,24 +543,40 @@ mod tests {
         .context("fake app-server child did not exit")?
     }
 
-    fn process_exists(pid: u32) -> bool {
+    /// Reports whether `pid` is still a *running* process.
+    ///
+    /// A killed child stays in the process table as a zombie until its parent
+    /// reaps it, and dropping the client cannot reap: it hands the `Child` to
+    /// tokio's global orphan queue, drained later off SIGCHLD. So "the PID is
+    /// absent" is not something drop can guarantee -- "the process no longer
+    /// runs" is. A zombie has already terminated, so it does not count as
+    /// alive; a child that drop failed to kill still does.
+    fn process_is_alive(pid: u32) -> anyhow::Result<bool> {
         #[cfg(windows)]
         {
-            std::process::Command::new("tasklist.exe")
+            // Windows drops terminated processes from the table outright, so
+            // being listed at all already means still running.
+            let output = std::process::Command::new("tasklist.exe")
                 .args(["/FI", &format!("PID eq {pid}"), "/FO", "CSV", "/NH"])
                 .output()
-                .is_ok_and(|output| {
-                    String::from_utf8_lossy(&output.stdout).contains(&format!("\"{pid}\""))
-                })
+                .context("failed to run tasklist.exe")?;
+            Ok(String::from_utf8_lossy(&output.stdout).contains(&format!("\"{pid}\"")))
         }
         #[cfg(unix)]
         {
-            std::process::Command::new("kill")
-                .args(["-0", &pid.to_string()])
-                .stdout(Stdio::null())
-                .stderr(Stdio::null())
-                .status()
-                .is_ok_and(|status| status.success())
+            // `kill -0` succeeds for zombies, so ask for the state instead: `ps`
+            // exits non-zero once the PID is gone, and reports `Z` (Linux) or
+            // `Z+` (macOS) while it is a terminated-but-unreaped zombie.
+            let output = std::process::Command::new("ps")
+                .args(["-o", "state=", "-p", &pid.to_string()])
+                .output()
+                .context("failed to run ps")?;
+            if !output.status.success() {
+                return Ok(false);
+            }
+            let stdout = String::from_utf8_lossy(&output.stdout);
+            let state = stdout.trim();
+            Ok(!state.is_empty() && !state.starts_with('Z'))
         }
     }
 
--- a/crates/edda-conductor/src/agent/codex_app_server.rs
+++ b/crates/edda-conductor/src/agent/codex_app_server.rs
@@ -454,14 +454,18 @@ mod tests {
 
         #[cfg(unix)]
         {
-            use std::os::unix::fs::PermissionsExt;
-
-            let launcher = dir.path().join("fake-app-server");
-            std::fs::write(&launcher, shell_fake_script(scenario))?;
-            let mut permissions = std::fs::metadata(&launcher)?.permissions();
-            permissions.set_mode(0o755);
-            std::fs::set_permissions(&launcher, permissions)?;
-            Ok((dir, Command::new(launcher)))
+            // Feed the script to `sh` instead of exec'ing it, mirroring the
+            // PowerShell branch above. Linux `execve` fails with ETXTBSY while
+            // any process holds the target file open for writing, and in a
+            // threaded test binary a concurrent spawn elsewhere can be sitting
+            // between fork and exec holding an inherited write fd for this very
+            // file -- O_CLOEXEC only clears it at exec, not at fork. `sh` opens
+            // the script read-only, so that window cannot bite.
+            let script = dir.path().join("fake-app-server.sh");
+            std::fs::write(&script, shell_fake_script(scenario))?;
+            let mut command = Command::new("/bin/sh");
+            command.arg(script);
+            Ok((dir, command))
         }
     }
 


fagemx:
## Code Review: Round 1 — PR #480

*Independent review · GATE-01 (fork, not the author) · gates re-run by hand, not taken from the PR description*

**Reviewed SHA:** `569cee0d10ec90153d04d812b1184d38d2cfaee3` (full head)
**Base:** merge-base == `origin/main` `7f5d1555b4c00e4fe6ac9a452a3c81c62289972c` — PR is on current base.

**IN SCOPE:** `crates/edda-conductor/src/agent/codex_app_server.rs` `mod tests` (the only changed file, +44/-21); the spawn path for the unix fake app-server; every caller of `wait_for_pid_exit` / `process_is_alive`; current-base integration. No linked issue (`closes #N` absent), so the acceptance basis is the PR's own stated problem plus repo conventions (`.claude/CLAUDE.md`).

**FOLLOW-UP ISSUE:** none opened — the Minor items below are all non-blocking and none is a live defect on any platform this repo tests.

### Gates

`RUSTFLAGS=-Dwarnings` throughout, matching `.github/workflows/ci.yml`.

| gate | RAN where | result |
|---|---|---|
| `cargo fmt --check` | local (Windows) | **pass**, exit 0 |
| `cargo clippy --workspace --all-targets` | local (Windows) | **pass**, exit 0, 2m01s |
| `cargo test -p edda-conductor` | local (Windows) | **pass** — 166 passed, 0 failed, 1 ignored, 12.53s; doc-tests 0 |
| exact-head CI @ `569cee0d` | `gh api .../check-runs` | **7/7 green** (fmt, clippy ×3, test ×3) — read from the API, not from the PR body |

**Local-coverage limit, stated plainly:** this workstation is Windows, so `cargo clippy`/`cargo test` here compile only the `#[cfg(windows)]` arms — the changed `#[cfg(unix)]` code was **not** type-checked locally. That coverage comes from Clippy/Test (ubuntu-latest) and (macos-latest) at the exact head, which I pulled from the check-runs API myself. To avoid signing off on the unix arms purely on someone else's green tick, I re-derived the two load-bearing runtime claims on a real Linux kernel (WSL2 6.6.87.1, procps-ng 4.0.4) with the exact shipped logic:

**ETXTBSY (commit 2) — RAN, reproduced:**

| case | result |
|---|---|
| exec the script, writer holding a write fd | **FAILED errno=26 Text file busy** |
| `/bin/sh script`, writer holding a write fd | ok |
| exec the script, no writer | ok |
| `/bin/sh script`, no writer | ok |

So the mechanism is real, the fix removes it by construction, and the "fails only rarely" shape is explained. I also checked the dropped `chmod 0755` is coherent rather than merely unnecessary: on a 0644 script, `/bin/sh script` runs (`ok`) while direct exec raises `EACCES` — the old path cannot silently come back.

**Zombie-aware probe (commit 1) — RAN, reproduced:**

| case | `ps -o state= -p` | `process_is_alive` |
|---|---|---|
| running child | rc=0, state `S` | **true** |
| SIGKILLed, unreaped (zombie) | rc=0, state `Z` | false |
| reaped | rc=1, empty | false |
| never existed | rc=1, empty | false |

The `true` on the first row is the one that matters: a child `drop` failed to kill still reads as alive, so the regression the test exists to catch still fails it. Teeth preserved.

**Base comparison (RAN first, not last):** CI run `32002599039` on `main@7f5d155` — verified via the API — is green everywhere except `Test (ubuntu-latest)`. The flake is pre-existing on base and is exactly what this PR targets; nothing here is a regression introduced by the change.

**READ (not run):** `Drop for CodexAppServer` (`start_kill()`, no reap) and `kill_on_drop(true)` in `spawn_command` — both corroborate the orphan-queue account. Grepped the workspace for other write-then-`chmod 0755`-then-exec helpers: **none**, so the ETXTBSY fix covers every instance in the repo, not just the one that fired.

### P0

None.

### P1

None.

### Minor (non-blocking, no action required to merge)

1. **`Command::new("/bin/sh")` hardcodes an absolute path** (`codex_app_server.rs:466`) where the Windows branch resolves `powershell.exe` via PATH. `/bin/sh` exists on both unix CI images so this is theoretical today; `sh` would be the symmetric spelling.

2. **The "fails loudly" guarantee is narrower than the doc comment implies** (`codex_app_server.rs:571-580`). The comment says the probe errors rather than reporting "gone" when it cannot run — true for a *missing* `ps` (`Command::output()` returns `Err` → `ENOENT` → propagated). But a `ps` that *exists and rejects* `-o state= -p` (BusyBox) exits non-zero, which is indistinguishable from rows 3 and 4 of the table above, so every `wait_for_pid_exit` would pass vacuously. I checked `.github/workflows/` for an environment where this could bite: no container jobs, no musl/Alpine test job, `cargo test` runs only on the three GitHub-hosted images. Not reachable in this repo's CI, and the pre-change `kill -0` had the same failure mode, so this is a note rather than a request.

3. **Test names drifted from the assertion.** `dropping_concrete_client_makes_child_pid_disappear` / `cancelling_concrete_method_makes_child_pid_disappear` now assert "no longer running" (a zombie satisfies them), not "PID disappeared". The new doc comment on `process_is_alive` explains this well; the names are the only place the old wording survives.

4. **Scope note on commit 1.** By the repo's own review discipline, a pre-existing latent hazard is normally a follow-up rather than PR content, and the PR body is candid that this one has not been observed firing. I am not asking for it to be split: it is test-only, it lives in the same helper the ETXTBSY fix touches, it is a second independent cause of instability in the same test, and the empirical table above shows it does not weaken the assertion. Recording it so the decision is visible rather than implicit.

### Verdict: **LGTM** — root cause is correctly identified and independently reproduced, the fix eliminates the race by construction rather than retrying around it, and every gate is green under my own hand.

*Cost: ~20 tool calls, ~12 min elapsed. Gate selection: code-only change confined to one crate's `mod tests`, so fmt + workspace clippy + the changed crate's tests + exact-head CI, plus targeted Linux runtime verification to cover the cfg arms this workstation cannot compile. Full-workspace test on Windows was not run — the change cannot reach another crate, and Windows CI itself scopes to 7 crates by design.*

Issue #483, opened 2026-08-18
coord(policy): verification-cost wording refinements from #481 review rounds


fagemx:
Follow-ups accepted during the three review rounds of #481 (merged as `91534f9`) and deliberately not pulled into that PR. All are wording/precision defects in the verification-cost policy; none blocks anything today.

Basis SHA: `1496ef721e28f8c0f7ca4b0a2005997ecdcbc74c`

## From Round 3

1. **`.claude/CLAUDE.md` — "compiled and linted" is stronger than `cargo clippy` delivers.** Clippy does not link, so Windows-only *link* failures outside the 7-crate test closure are also uncovered. The current text slightly understates the gap it exists to describe.
2. **PowerShell lane root needs quoting in a bare assignment.** `$env:LOCALAPPDATA\fleet-workstation\lanes` unquoted is a parser error. It fails loudly and creates nothing (unlike the Round 2 defect, which silently produced a literal directory), but the example should be quotable as written.
3. **The unproven-coverage fallback could license over-verification.** "No coverage statement means the verifier RANs what the receipt alone would carry" is conservative by design, but a reviewer could read it as a standing licence for full local reruns in any project without a CI table. Bound it.
4. **`skills/fleet-orchestrate/references/playbook.md:354`** — the gate-table row omits the coverage-gap reason that the other carriers list (pre-existing at `4a3eea1`, not introduced by the Round 2 fix).
5. **Stale quote in the plan.** `docs/superpowers/plans/2026-08-18-verification-cost-discipline-pr1-edda-policy.md:105` still quotes the pre-Round-1 CI sentence. The plan is deliberately kept as the execution record at `738c5ce` and carries a "Round 1 review amendments" section, but a reader landing on `:105` sees the false claim without that context. Add an inline pointer.

## From Rounds 1–2

6. **L2's full-rerun licence is broader than fixture 9's prescription.** The ladder permits a full rerun for a coverage gap; fixture 9 requires a *focused* check. Align them — focused is the intent.
7. **Fixture 9 carries an edda-specific fact into a synced skill reference.** `review-scope-pressure-tests.md` ships to fleet-playbook via `scripts/skills/sync-fleet-orchestrate.ps1`; the "7-crate subset" detail should be phrased as an example, not a fact about the reader's repo.
8. **`missing lane root` behaviour differs between carriers.** `.claude/CLAUDE.md` and `playbook.md` prescribe slightly different fallbacks when a brief omits the lane root.
[...]

PR #487, opened 2026-08-19
fix(edda-cli): make unclaim resolve from the board and never fake success

### LINKED ISSUE #486: coord(cli): unclaim session resolution, Codex skill registration, and test breadth (from #485 review)
Follow-ups accepted across the two review rounds of #485 and deliberately kept out of that PR. Basis SHA: `996ba38f806efbf9089cc8d7ce4e3263bf08feab`.

Filing these as one issue because I twice wrote "I will file these" in a review response and did not — the reviewer had to point that out in Round 2.

## 1. `resolve_session_id` fallback-label asymmetry (root cause of #485 P1-1)

`crates/edda-cli/src/cmd_bridge.rs`

`claim` passes the user's label to `resolve_session_id` (`:660`), `unclaim` hardcodes `"cli"` (`:679`). Tier 4 builds the id as `format!("cli-{fallback_label}")` (`:1070-1072`), so with no heartbeat:

```
edda claim "auth"   → session cli-auth
edda unclaim        → session cli-cli, exit 0, releases nothing
```

#485 documents the trap. The real fix is in the code, but it changes behavior for a verb people already use, so it needs its own design decision: infer the label from the single claim on the board, refuse when ambiguous, or make `--session` required without hooks. Whatever is chosen, `unclaim` should not report success when it released nothing.

## 2. No non-JSON way to read a full session id

`cmd_bridge.rs:623` abbreviates ids in the human-readable `edda peers` view, and `write_claim` records no heartbeat, so a bare-CLI claim is invisible to plain `peers` and its id is unreadable even when a heartbeat exists. Only `edda peers --json` carries `claims[].session_id`. That is the string `unclaim --session` needs, so the human-facing command cannot answer the question the workflow asks.

## 3. Codex skills are scaffolded but never registered

`crates/edda-cli/src/cmd_init.rs` — `scan_project_skills` hardcodes `.claude/skills`. After #485, `edda init` writes coord skills into `.agents/skills/` for a Codex-only repository, but the registration pass never looks there, so the run prints no "Registered N skill(s)" line and the skills are absent from the user-level registry.

## 4. Test breadth for the new host detection

`init_scaffolds_codex_coord_skills` covers the `AGENTS.md` arm only. Uncovered: the `.agents/`-directory arm, the both-hosts case, and a negative assertion that a repo with neither marker gets nothing.

## 5. Both-hosts init output is undifferentiated

With `.claude/` and `AGENTS.md` both present, `edda init` prints ten scaffolding lines with nothing indicating which host each belongs to.

## 6. `AGENTS.md:10` first-run gap

`AGENTS.md` points Codex at `.agents/skills/coord-orchestrate/`, which a fresh clone does not have until someone runs `edda init` — `.agents/` is ignored as of #485. This is parity with `.claude/skills`, not a regression, but the instruction reads as though the path is always there. Either say it appears after `edda init`, or have the doc name the fallback.

---

Related: #483 collects the follow-ups from the #481 review rounds. No overlap with this list.

fagemx:
Closes #486 item 1.

## The defect

`unclaim` shared `resolve_session_id` with `claim`. That function's last tier mints `cli-{fallback_label}`; `claim` passes the user's label, `unclaim` hardcoded `"cli"`. Without hooks there is no heartbeat to infer from, so both verbs landed on that tier and disagreed:

```
  session: cli-auth

$ edda unclaim
Unclaimed scope for session: cli-cli      ← different session
exit 0                                    ← success
                                          ← claim still on the board
```

Minting an id is correct for `claim`, which is creating one. It is wrong for `unclaim`, which has to name one that already exists.

## The fix

`unclaim` resolves its own target instead of borrowing `claim`'s:

| Tier | Source |
|---|---|
| 1 | explicit `--session` |
| 2 | `EDDA_SESSION_ID` |
| 3 | a sole live heartbeat |
| 4 | **the board** — the sole claimant, or refuse and list them |

and it verifies the target actually holds a claim before writing, so the command can no longer report a release that did not happen. Both refusals print the board with session ids, so the value for `--session` is in the error itself rather than requiring `peers --json` (#486 item 2).

## Behavior, from a binary built at this SHA

```
A. one claim, bare unclaim
   Unclaimed scope for session: cli-auth (auth)      exit 0, claims: 0

B. two claims, bare unclaim
   Error: several sessions hold claims, so --session is required.
   Claims on the board:
     cli-api — api (src/api/*)
     cli-auth — auth (src/auth/*)                    exit 1, claims still 2

C. --session cli-cli (holds nothing)
   Error: session cli-cli holds no claim; nothing was released.
   Claims on the board: …                            claims still 2

D. --session cli-auth
   Unclaimed scope for session: cli-auth (auth)      claims left 1

E. empty board
   Error: no claims on the board; nothing to unclaim
```

## The automatic path is untouched

Session-end unclaim does not come through this verb — bridges call `peers::write_unclaim` directly (`dispatch/session.rs:657`, plus the codex, cursor and hermes equivalents). Refusing therefore costs a hooked session nothing; only the explicit CLI verb changed.

## Doc change

`coord-handoff.md` described the trap, because #485 could only warn about it. With the trap gone the skill documents behavior: bare `unclaim` works, several claimants refuse, and a printed release really happened.

Shipping a warning label was the wrong shape for #485 and this replaces it.

## Evidence

- 4 new tests, written first and **all four failing before the change**: sole claim resolves bare; several claims refuse and name both; empty board refuses; a session holding no claim refuses while the real claim survives.
- RAN `cargo test -p edda` — **307 passed, 0 failed**
- RAN the five scenarios above from a binary built at this content, in a throwaway repo under an isolated `EDDA_STORE_ROOT` with the session env vars unset
- Lane: worktree default `target/` (solo work); no ad-hoc `CARGO_TARGET_DIR`
- Receipt: L0 focused gates above; `-p edda` is in the CI Windows test subset, so exact-head CI covers this crate on all three platforms

## Not in this PR

#486 items 2–6, including the `edda peers` id abbreviation and `scan_project_skills` hardcoding `.claude/skills` (Codex skills are scaffolded but never registered — the highest-value remaining item).

### DIFF
--- a/crates/edda-cli/src/cmd_bridge.rs
+++ b/crates/edda-cli/src/cmd_bridge.rs
@@ -676,12 +676,90 @@ pub fn claim(
 }
 
 /// `edda unclaim [--session <id>]` — release a coordination scope.
+///
+/// Unlike `claim`, this verb never mints a session id. `claim` may invent
+/// `cli-<label>` because it is creating the claim; `unclaim` has to name one
+/// that already exists, so it resolves against the board and refuses rather
+/// than reporting success for a session that holds nothing (GH-486).
+///
+/// The automatic session-end path does not come through here — bridges call
+/// `peers::write_unclaim` directly — so refusing costs a hooked session
+/// nothing.
 pub fn unclaim(repo_root: &Path, cli_session: Option<&str>) -> anyhow::Result<()> {
     let project_id = edda_store::project_id(repo_root);
-    let (session_id, _label) = resolve_session_id(cli_session, &project_id, "cli");
+    let board = edda_bridge_claude::peers::compute_board_state(&project_id);
+    let session_id = resolve_unclaim_target(cli_session, &project_id, &board.claims)?;
+
+    let held: Vec<&edda_bridge_claude::peers::ClaimEntry> = board
+        .claims
+        .iter()
+        .filter(|c| c.session_id == session_id)
+        .collect();
+    if held.is_empty() {
+        anyhow::bail!(
+            "session {session_id} holds no claim; nothing was released.\n{}",
+            describe_claims(&board.claims)
+        );
+    }
+
     edda_bridge_claude::peers::write_unclaim(&project_id, &session_id);
-    println!("Unclaimed scope for session: {session_id}");
+    let labels: Vec<&str> = held.iter().map(|c| c.label.as_str()).collect();
+    println!(
+        "Unclaimed scope for session: {session_id} ({})",
+        labels.join(", ")
+    );
     Ok(())
+}
+
+/// Name the session whose claim `unclaim` should release.
+///
+/// Explicit `--session` and `EDDA_SESSION_ID` win, then a sole live heartbeat,
+/// then the board itself. The board tier is what makes a bare `edda unclaim`
+/// work for a claim made from a plain CLI, which writes no heartbeat.
+fn resolve_unclaim_target(
+    cli_session: Option<&str>,
+    project_id: &str,
+    claims: &[edda_bridge_claude::peers::ClaimEntry],
+) -> anyhow::Result<String> {
+    if let Some(sid) = cli_session.filter(|s| !s.is_empty()) {
+        return Ok(sid.to_string());
+    }
+    if let Ok(sid) = std::env::var("EDDA_SESSION_ID") {
+        if !sid.is_empty() {
+            return Ok(sid);
+        }
+    }
+    if let Some((sid, _label)) = edda_bridge_claude::peers::infer_session_id(project_id) {
+        return Ok(sid);
+    }
+
+    let mut sessions: Vec<&str> = claims.iter().map(|c| c.session_id.as_str()).collect();
+    sessions.sort_unstable();
+    sessions.dedup();
+    match sessions.len() {
+        0 => anyhow::bail!("no claims on the board; nothing to unclaim"),
+        1 => Ok(sessions[0].to_string()),
+        _ => anyhow::bail!(
+            "several sessions hold claims, so --session is required.\n{}",
+            describe_claims(claims)
+        ),
+    }
+}
+
+/// Render the board's claims for an error message, so the reader can copy the
+/// session id straight into `--session` instead of hunting for it.
+fn describe_claims(claims: &[edda_bridge_claude::peers::ClaimEntry]) -> String {
+    if claims.is_empty() {
+        return "The board holds no claims.".to_string();
+    }
+    let mut out = String::from("Claims on the board:");
+    for c in claims {
+        out.push_str(&format!("\n  {} — {}", c.session_id, c.label));
+        if !c.paths.is_empty() {
+            out.push_str(&format!(" ({})", c.paths.join(", ")));
+        }
+    }
+    out
 }
 
 /// `edda bridge claude decide <key=value>` — record a decision.
@@ -2206,6 +2284,94 @@ mod tests {
         assert!(edda_bridge_claude::peers::compute_board_state(&pid)
             .claims
             .is_empty());
+    }
+
+    #[test]
+    fn unclaim_without_session_releases_the_sole_claim() {
+        let _store = crate::test_support::isolated_store();
+        let _env = env_guard();
+        std::env::remove_var("EDDA_SESSION_ID");
+        std::env::remove_var("EDDA_SESSION_LABEL");
+        let repo = tempfile::tempdir().expect("tempdir");
+        let pid = edda_store::project_id(repo.path());
+        let _ = edda_store::ensure_dirs(&pid);
+        // A bare-CLI claim: `edda claim "auth"` mints session cli-auth and
+        // writes no heartbeat, so nothing can be inferred from live sessions.
+        edda_bridge_claude::peers::write_claim(&pid, "cli-auth", "auth", &["src/auth.rs".into()]);
+
+        unclaim(repo.path(), None).expect("the sole claim is unambiguous");
+
+        assert!(
+            edda_bridge_claude::peers::compute_board_state(&pid)
+                .claims
+                .is_empty(),
+            "bare unclaim must release the only claim on the board"
+        );
+    }
+
+    #[test]
+    fn unclaim_without_session_refuses_when_several_claims_exist() {
+        let _store = crate::test_support::isolated_store();
+        let _env = env_guard();
+        std::env::remove_var("EDDA_SESSION_ID");
+        std::env::remove_var("EDDA_SESSION_LABEL");
+        let repo = tempfile::tempdir().expect("tempdir");
+        let pid = edda_store::project_id(repo.path());
+        let _ = edda_store::ensure_dirs(&pid);
+        edda_bridge_claude::peers::write_claim(&pid, "cli-auth", "auth", &["src/auth.rs".into()]);
+        edda_bridge_claude::peers::write_claim(&pid, "cli-api", "api", &["src/api.rs".into()]);
+
+        let err = unclaim(repo.path(), None).expect_err("ambiguous target must not guess");
+        let msg = err.to_string();
+        assert!(msg.contains("cli-auth") && msg.contains("cli-api"), "{msg}");
+
+        assert_eq!(
+            edda_bridge_claude::peers::compute_board_state(&pid)
+                .claims
+                .len(),
+            2,
+            "a refused unclaim must release nothing"
+        );
+    }
+
+    #[test]
+    fn unclaim_refuses_when_the_board_is_empty() {
+        let _store = crate::test_support::isolated_store();
+        let _env = env_guard();
+        std::env::remove_var("EDDA_SESSION_ID");
+        std::env::remove_var("EDDA_SESSION_LABEL");
+        let repo = tempfile::tempdir().expect("tempdir");
+        let pid = edda_store::project_id(repo.path());
+        let _ = edda_store::ensure_dirs(&pid);
+
+        unclaim(repo.path(), None).expect_err("nothing to release must not report success");
+    }
+
+    #[test]
+    fn unclaim_refuses_a_session_that_holds_no_claim() {
+        let _store = crate::test_support::isolated_store();
+        let _env = env_guard();
+        std::env::remove_var("EDDA_SESSION_ID");
+        std::env::remove_var("EDDA_SESSION_LABEL");
+        let repo = tempfile::tempdir().expect("tempdir");
+        let pid = edda_store::project_id(repo.path());
+        let _ = edda_store::ensure_dirs(&pid);
+        edda_bridge_claude::peers::write_claim(&pid, "cli-auth", "auth", &["src/auth.rs".into()]);
+
+        // This is the exact silent-failure this fix exists to remove: the old
+        // fallback resolved to `cli-cli`, wrote an unclaim for a session that
+        // held nothing, printed success, and left the real claim standing.
+        let err = unclaim(repo.path(), Some("cli-cli"))
+            .expect_err("releasing nothing must not report success");
+        assert!(err.to_string().contains("cli-auth"), "{err}");
+
+        assert_eq!(
+            edda_bridge_claude::peers::compute_board_state(&pid)
+                .claims
+                .len(),
+            1,
+            "the real claim must survive a refused unclaim"
+        );
     }
 
     #[test]
--- a/crates/edda-cli/src/cmd_bridge.rs
+++ b/crates/edda-cli/src/cmd_bridge.rs
@@ -717,9 +717,11 @@ pub fn unclaim(repo_root: &Path, cli_session: Option<&str>) -> anyhow::Result<()
 
 /// Name the session whose claim `unclaim` should release.
 ///
-/// Explicit `--session` and `EDDA_SESSION_ID` win, then a sole live heartbeat,
-/// then the board itself. The board tier is what makes a bare `edda unclaim`
-/// work for a claim made from a plain CLI, which writes no heartbeat.
+/// Explicit `--session` wins, then `EDDA_SESSION_ID`, then a sole live
+/// heartbeat. There is deliberately no further tier: a caller with none of
+/// those has no identity, and picking a claim off the board for it would
+/// release a scope its real owner is still relying on. Refuse and show the
+/// board instead, so the id for `--session` is in the error.
 fn resolve_unclaim_target(
     cli_session: Option<&str>,
     project_id: &str,
--- a/crates/edda-cli/src/cmd_bridge.rs
+++ b/crates/edda-cli/src/cmd_bridge.rs
@@ -682,6 +682,10 @@ pub fn claim(
 /// that already exists, so it resolves against the board and refuses rather
 /// than reporting success for a session that holds nothing (GH-486).
 ///
+/// It also never guesses from the board. A caller with no session identity
+/// cannot know that a claim is its own, and releasing someone else's would
+/// drop the off-limits protection their live session depends on.
+///
 /// The automatic session-end path does not come through here — bridges call
 /// `peers::write_unclaim` directly — so refusing costs a hooked session
 /// nothing.
@@ -733,17 +737,17 @@ fn resolve_unclaim_target(
         return Ok(sid);
     }
 
-    let mut sessions: Vec<&str> = claims.iter().map(|c| c.session_id.as_str()).collect();
-    sessions.sort_unstable();
-    sessions.dedup();
-    match sessions.len() {
-        0 => anyhow::bail!("no claims on the board; nothing to unclaim"),
-        1 => Ok(sessions[0].to_string()),
-        _ => anyhow::bail!(
-            "several sessions hold claims, so --session is required.\n{}",
-            describe_claims(claims)
-        ),
+    // No identity of our own, so there is nothing to infer from. Do NOT fall
+    // back to "the sole claim on the board": a caller without a session cannot
+    // know that claim is theirs, and releasing it drops the off-limits
+    // protection its real owner is relying on.
+    if claims.is_empty() {
+        anyhow::bail!("no claims on the board; nothing to unclaim");
     }
+    anyhow::bail!(
+        "cannot tell which claim is yours, so --session is required.\n{}",
+        describe_claims(claims)
+    );
 }
 
 /// Render the board's claims for an error message, so the reader can copy the
@@ -2287,7 +2291,7 @@ mod tests {
     }
 
     #[test]
-    fn unclaim_without_session_releases_the_sole_claim() {
+    fn unclaim_without_identity_refuses_rather_than_guessing_the_sole_claim() {
         let _store = crate::test_support::isolated_store();
         let _env = env_guard();
         std::env::remove_var("EDDA_SESSION_ID");
@@ -2295,17 +2299,48 @@ mod tests {
         let repo = tempfile::tempdir().expect("tempdir");
         let pid = edda_store::project_id(repo.path());
         let _ = edda_store::ensure_dirs(&pid);
-        // A bare-CLI claim: `edda claim "auth"` mints session cli-auth and
-        // writes no heartbeat, so nothing can be inferred from live sessions.
         edda_bridge_claude::peers::write_claim(&pid, "cli-auth", "auth", &["src/auth.rs".into()]);
 
-        unclaim(repo.path(), None).expect("the sole claim is unambiguous");
+        let err = unclaim(repo.path(), None).expect_err("a caller with no identity must not guess");
+        assert!(err.to_string().contains("cli-auth"), "{err}");
 
-        assert!(
+        assert_eq!(
             edda_bridge_claude::peers::compute_board_state(&pid)
                 .claims
-                .is_empty(),
-            "bare unclaim must release the only claim on the board"
+                .len(),
+            1,
+            "a refused unclaim must release nothing"
+        );
+    }
+
+    #[test]
+    fn unclaim_without_identity_never_releases_a_live_peers_claim() {
+        let _store = crate::test_support::isolated_store();
+        let _env = env_guard();
+        std::env::remove_var("EDDA_SESSION_ID");
+        std::env::remove_var("EDDA_SESSION_LABEL");
+        let repo = tempfile::tempdir().expect("tempdir");
+        let pid = edda_store::project_id(repo.path());
+        let _ = edda_store::ensure_dirs(&pid);
+
+        // Two live peers, so infer_session_id yields nothing, and only one of
+        // them holds a claim. A shell with no identity of its own must not
+        // decide that claim is its to release: check_offlimits enforces
+        // exactly this claim for its live owner.
+        edda_bridge_claude::peers::write_heartbeat_minimal(&pid, "sess-a", "worker-a", "/tmp/a");
+        edda_bridge_claude::peers::write_heartbeat_minimal(&pid, "sess-b", "worker-b", "/tmp/b");
+        edda_bridge_claude::peers::write_claim(&pid, "sess-a", "worker-a", &["src/a.rs".into()]);
+
+        let err = unclaim(repo.path(), None)
+            .expect_err("a bare shell must not release another live session's scope");
+        assert!(err.to_string().contains("sess-a"), "{err}");
+
+        assert_eq!(
+            edda_bridge_claude::peers::compute_board_state(&pid)
+                .claims
+                .len(),
+            1,
+            "the live peer's claim must survive"
         );
     }
 
--- a/crates/edda-cli/src/skills/coord-handoff.md
+++ b/crates/edda-cli/src/skills/coord-handoff.md
@@ -68,18 +68,27 @@ Where host bridge hooks are wired, unclaim happens automatically on session
 end — nothing to do.
 
 Where they are not (claims made from a bare CLI, CI, or a host without bridge
-hooks), release the scope explicitly — and **pass the session id**:
+hooks), release the scope explicitly:
+
+```bash
+edda unclaim
+```
+
+That resolves the target from the board, so it works for a claim made without
+hooks. When several sessions hold claims it refuses and lists them rather than
+guessing — name one:
 
 ```bash
 edda unclaim --session <id>
 ```
 
-The id is the `session:` line `edda claim` printed; `edda peers --json` also
-carries it. Plain `edda peers` will not: it lists live heartbeats, which a
-bare-CLI claim does not have, and it abbreviates ids. Without hooks there is no
-heartbeat to infer from either, so bare `edda unclaim` falls back to the
-session `cli-cli` rather than the `cli-<label>` your claim created: it exits 0,
-prints a reassuring line, and releases nothing (GH-455).
+The id is the `session:` line `edda claim` printed, and the refusal message
+carries it too. `edda peers --json` also has it under `claims[].session_id`;
+plain `edda peers` does not, because it lists live heartbeats and a bare-CLI
+claim has none.
+
+`unclaim` never reports success for a session that holds nothing — if it prints
+a released scope, that scope is gone.
 
 Enforcement stays safe either way: every consumer joins claims against live
 heartbeats, so a claim left behind by a dead session does not block a peer.
--- a/crates/edda-cli/src/skills/coord-handoff.md
+++ b/crates/edda-cli/src/skills/coord-handoff.md
@@ -68,24 +68,21 @@ Where host bridge hooks are wired, unclaim happens automatically on session
 end — nothing to do.
 
 Where they are not (claims made from a bare CLI, CI, or a host without bridge
-hooks), release the scope explicitly:
-
-```bash
-edda unclaim
-```
-
-That resolves the target from the board, so it works for a claim made without
-hooks. When several sessions hold claims it refuses and lists them rather than
-guessing — name one:
+hooks), release the scope by naming it:
 
 ```bash
 edda unclaim --session <id>
 ```
 
-The id is the `session:` line `edda claim` printed, and the refusal message
-carries it too. `edda peers --json` also has it under `claims[].session_id`;
+The id is the `session:` line `edda claim` printed. Run it without `--session`
+and the refusal lists every claim with its session id, so the value you need is
+in the error. `edda peers --json` carries it too, under `claims[].session_id`;
 plain `edda peers` does not, because it lists live heartbeats and a bare-CLI
 claim has none.
+
+`unclaim` deliberately does not pick a claim for you when you have no session
+of your own: it cannot tell whose claim it is, and releasing the wrong one
+would drop the off-limits protection a live peer is relying on.
 
 `unclaim` never reports success for a session that holds nothing — if it prints
 a released scope, that scope is gone.


fagemx:
## Code Review: Round 1
Reviewed full SHA: 9fc301c1bcf8a0f7f59d87a9dafd88b13b9ec5cd
Base full SHA: e1b37429ffe83fdf002b6c5f9bbb1c9b5b2c9287

IN SCOPE: the four resolution tiers in `resolve_unclaim_target` and the holds-a-claim precondition in `unclaim` (`crates/edda-cli/src/cmd_bridge.rs`); the "never reports success when nothing was released" invariant; every automatic session-end unclaim caller in the workspace; the four added tests; every factual claim in the shipped `crates/edda-cli/src/skills/coord-handoff.md` Step 5; and whether #486 item 1 is closed.

BLOCKING: P0=0, P1=1

- **P1-1 — tier 4 releases another live session's claim; new in this PR.** `crates/edda-cli/src/cmd_bridge.rs:736-746`. `infer_session_id` returns `Some` only when *exactly one* live heartbeat exists (`crates/edda-bridge-claude/src/peers/discovery.rs:204-208`). With **two or more** live peers it returns `None`, so tier 3 falls through and tier 4 takes "the sole claimant" — whoever that is. RAN, two heartbeating sessions, `sess-A` holding `worker-a` on `src/a/*`:

  ```
  before:  Active sessions (2):
             sess-A — sess-A (0s ago) [src/a/*]
             sess-B — sess-B (0s ago)
  $ edda unclaim                       # bare, from a shell with no session identity
  Unclaimed scope for session: sess-A (worker-a)
  [exit 0]
  after:   Active sessions (2):
             sess-A — sess-A (1s ago)          ← scope gone, session still live
             sess-B — sess-B (1s ago)
           "claims": []
  ```

  This is not advisory-only: `check_offlimits` (`crates/edda-bridge-claude/src/dispatch/tools.rs:204-236`) enforces exactly those claims whose session is an *active peer*, so the Edit/Write block on `src/a/*` is now gone while `sess-A` is still heartbeating and still believes it holds the scope — the merge-conflict outcome the board exists to prevent. Any shell in the repo can do it; the caller needs no identity and no claim.

  It is a regression: at base, tier 4 minted `cli-cli` and the write was a fold no-op, so no path could release a peer's claim.

  It also contradicts the change's own principle. With exactly **one** live peer, the same bare command refuses rather than touch a claim the resolved session does not hold (RAN: `Error: session hooked-1 holds no claim; nothing was released.` + board listing, exit 1, claim intact). Adding a second peer converts that refusal into a silent cross-session release. Same for `EDDA_SESSION_ID` pointing at a claimless session — refuses (RAN). Only tier 4 releases something the caller demonstrably does not own.

  The shipped skill instructs precisely this call in the multi-agent handoff (`coord-handoff.md` Step 5, "release the scope explicitly: `edda unclaim`").

  Direction, not a prescription: gate tier 4 on the absence of live-peer ambiguity — when `discover_all_sessions` shows 2+ live sessions the caller's identity is genuinely unknown, so `--session` should be required, exactly as it is when 2+ sessions hold claims.

  **Coverage the fix needs:** no added test writes a heartbeat, so tier 3 and the tier-3→tier-4 fall-through are not exercised at all. The defect lives entirely in the untested region. A fix should carry tests with heartbeats present for both the 1-live-peer refusal and the 2-live-peer case.

What I checked that is clean:
- Tiers 1/2/3 each behave as the PR table says; the refusal paths hold. RAN and confirmed exit 1 with the board unchanged for: empty board; two claimants (both ids listed); `--session cli-cli`; `--session CLI-AUTH` (case differs); `--session " cli-auth"` (leading space); `EDDA_SESSION_ID` at a claimless session. No false-success path found in any of them.
- Duplicate claims by one session are impossible by construction: `compute_board_state` keys claims in a `HashMap` on `session_id` (`crates/edda-bridge-claude/src/peers/board.rs:44-53`), so a second claim replaces the first. RAN: `claim "auth" --session S1` then `claim "api" --session S1` leaves one claim (`api`). `held`/`labels.join(", ")` can therefore never print more than one label — dead generality, not a defect.
- Concurrency between the board read and `write_unclaim`: the window exists, but the fold is file-ordered and `claims.remove` is idempotent, so the printed claim is gone either way. No false-success path found.
- **Automatic session-end path: unaffected, verified independently.** `cmd_bridge::unclaim` is reachable only from `main.rs:1063` (`edda unclaim`) and `cmd_bridge.rs:311` (`edda bridge claude unclaim`). All five bridges call `peers::write_unclaim` directly: `edda-bridge-claude/src/dispatch/session.rs:657`, `edda-bridge-codex/src/dispatch.rs:285`, `edda-bridge-cursor/src/dispatch.rs:133`, `edda-bridge-hermes/src/dispatch.rs:360`, `edda-bridge-openclaw/src/dispatch.rs:341`. A repo-wide sweep of every non-Rust file found no script, hook, or settings template invoking the verb — hooks shell out to `edda hook claude`, not `edda unclaim`.
- **Tests are honest.** Base `unclaim` (`e1b3742:crates/edda-cli/src/cmd_bridge.rs:679-685`) has no error path at all — it returns `Ok(())` unconditionally — so all three `expect_err` tests must fail at base, and the fourth asserts an empty board that base leaves holding `cli-auth`. None passes with the feature broken. (Derived from the base source; I did not execute them at base.)
- **Shipped skill: every factual claim in Step 5 is true.** RAN `edda init` in a throwaway repo — the written `SKILL.md` is byte-identical to `crates/edda-cli/src/skills/coord-handoff.md`. Executed against the binary built at this SHA: `edda claim` prints `session: cli-auth`; both refusals carry the session id; `edda peers --json` exposes it at exactly `claims[].session_id`; plain `edda peers` prints `No active sessions.` for a bare-CLI claim; bare `edda unclaim` releases a hookless sole claim; several claimants refuse and list both. No false statement found.
- **#486 item 1 is closed on its literal terms.** The chosen design is candidate 1 + candidate 2 (infer from the board, refuse when ambiguous) plus a holds-a-claim precondition, and `unclaim` no longer reports success when it released nothing. P1-1 is a *new* hazard the issue text did not contemplate, not a failure to meet its requirement.

FOLLOW-UP ISSUE (basis SHA 9fc301c1bcf8a0f7f59d87a9dafd88b13b9ec5cd; checked against #483 and #486, no overlap):
- A second `edda claim` from the same session silently replaces the first on the board (same `HashMap` fold above). `edda claim "auth" --paths "src/auth/*"` then `edda claim "api" --paths "src/api/*"` under one session leaves only `api`; `src/auth/*` stops being protected with no warning and no output saying so. Pre-existing.
- `edda unclaim` now exits 1 when there is nothing to release. Step 5 names CI as a target context but does not mention the nonzero exit, so a CI step that unclaims unconditionally (or twice) will start failing. Deliberate behavior; the doc should say so.
- The "holds no claim" refusal prints the board but, unlike the ambiguity refusal, never tells the reader to pass `--session`.

Evidence:
- RAN (Windows 11, worktree default `target/`, binary built from the reviewed SHA — not the installed `edda.exe`):
  - `cargo build -p edda` — OK
  - `cargo test -p edda unclaim` — 5 passed, 0 failed, 302 filtered
  - the five PR scenarios re-derived from that binary under an isolated `EDDA_STORE_ROOT` with session env unset — all five reproduce as described
  - adversarial: case-different `--session`, whitespace-padded `--session`, `EDDA_SESSION_ID` at a claimless session, one live heartbeat vs. the board, **two live heartbeats vs. the board (P1-1)**, same-session double claim
  - `edda init` in a throwaway repo + byte diff of the written skill; every Step 5 CLI claim executed
  - `cargo test -p edda-ledger concurrent_writers_no_data_loss` — 4 runs, 4 passes (to classify the CI red)
- READ:
  - exact-head CI run 32294185634 — Format pass; Clippy pass on ubuntu, macos **and** windows; Test pass on ubuntu and macos (full workspace). **Test (windows) failed**, and I classify it **environmental, not deterministic, and not caused by this PR**: the sole failure is `edda-ledger sqlite_store::tests::concurrent_writers_no_data_loss` ("database is locked", 215 passed / 1 failed) in a crate this PR does not touch; the Windows job had already run the entire `-p edda` binary, where all five unclaim tests passed; and the test passes 4/4 locally on Windows at this SHA. It does not block the product surface.
  - base `unclaim` at `e1b3742` (to derive the "fails before" property myself)
  - `check_offlimits` heartbeat join, `compute_board_state` claim fold, `infer_session_id`, all five bridge session-end callers
  - #486 and #483
  - Nothing was reused from the PR body as evidence — every count and scenario above was re-derived.
- Could not corroborate: the PR body's "4 new tests, written first and all four failing before the change" — read-only, so I did not observe them fail; I derived the property statically from the base function instead. "307 passed" is consistent with what I saw (5 matched + 302 filtered) but I did not run the full `-p edda` suite.
- Lane: the worktree's default `target/` (as briefed); no ad-hoc `CARGO_TARGET_DIR`; nothing deleted.
- Receipt: the PR body's L0 receipt (fmt / clippy `-p edda` / test `-p edda`) — READ, and corroborated by exact-head CI Format and Clippy on all three OSes, so I did not rerun those gates.

Cost: elapsed=unknown, tokens=unknown, tools=~28 calls

Verdict: Changes Requested

fagemx:
## Review Response: Round 1

New full SHA: `d45bbc3417f15de6d01584ff2a7efb891d4b6f37`
Reviewed SHA this responds to: `9fc301c1bcf8a0f7f59d87a9dafd88b13b9ec5cd`

### P1-1 — bare `unclaim` could release a live peer's claim · FIXED by removing the tier

Confirmed and not disputed. `infer_session_id` (`discovery.rs:204-208`) returns `Some` only for exactly one active session, so two or more live peers drop through tier 3 into the board tier I added, which then picked "the sole claimant" — whoever that was. `check_offlimits` (`dispatch/tools.rs:204-236`) enforces exactly the claims whose session is an active peer, so this dropped real Edit/Write protection out from under its owner.

The framing that decided the fix is yours: with one live peer the command refuses to touch a claim the resolved session does not hold; adding a second peer turned that refusal into a silent cross-session release. A rule that gets *less* safe as the fleet grows is the wrong rule.

**The board tier is gone.** A caller with no identity of its own cannot know which claim is theirs, so `unclaim` now refuses and prints the board. That costs almost nothing: the only thing the tier bought was not having to look up an id, and the refusal message already carries it.

Reproduced on the new SHA, your exact scenario — two live heartbeats, `sess-a` holding `worker-a`, bare `unclaim` from a shell with no identity:

```
Error: cannot tell which claim is yours, so --session is required.
Claims on the board:
  sess-a — worker-a (src/a/*)
exit=1
sess-a claim survives: 1
```

and the owner can still release it by name, and an empty board still refuses.

**On the test gap you identified** — that the defect sat entirely in the untested tier-3/tier-4 region, because no test wrote a heartbeat. Correct, and that is why the finding existed. The new test writes two heartbeats and pins your scenario directly. The old `unclaim_without_session_releases_the_sole_claim` asserted the behavior being removed, so it is inverted and renamed (`unclaim_without_identity_refuses_rather_than_guessing_the_sole_claim`) rather than deleted — the poison-test rule this repo's own review skill states.

### Tier 3 has the same hazard, and it is pre-existing

Worth stating plainly rather than leaving implied: a bare shell alongside **exactly one** live peer still resolves to that peer via tier 3 and can release its claim. I verified this is not something this PR introduced — base `unclaim` (`e1b3742`) called `resolve_session_id`, which contains the identical tier, and `claim` and `decide` share it. My diff touches `resolve_session_id` only by no longer calling it from `unclaim`.

It is a wider question than this verb: the premise "sole active session = me" is right inside a hooked session and wrong from a bare shell, and changing it affects several commands. Filing it rather than widening this PR.

### Follow-ups accepted

Your three (same-session second claim silently replaces the first; the new nonzero exit undocumented for the CI context Step 5 names; the "holds no claim" refusal not saying to pass `--session`) plus the tier-3 hazard above. I will file them — and I am aware that in this loop I twice wrote "I will file these" and did not, which is why #486 exists.

### On the Windows CI failure you classified

Agreed with the classification and grateful you did not treat it as a blocker: the sole failure is `edda-ledger sqlite_store::tests::concurrent_writers_no_data_loss` in an untouched crate, and the ledger records it as a known flake (`d-040` classifies a comparable default-parallel fixture race as baseline, not acceptance). The Windows job had already run the whole `-p edda` binary green.

### What you could not corroborate

Both noted correctly. "All four failing before the change" was mine to observe and I did — the run before implementing gave `FAILED` on all four with `1 passed` (the pre-existing explicit-session test). You derived it statically instead, which is the right move for a read-only reviewer, and I am recording the observed result here rather than asking you to take it on trust. Same for the 308-test count on this SHA.

### Gates on the new SHA

- RAN: `cargo test -p edda` — 308 passed, 0 failed
- RAN: the reproduction above from a binary built at this content
- READ: exact-head CI pending on `d45bbc3`

Round 2 requested on `d45bbc3417f15de6d01584ff2a7efb891d4b6f37`. The delta removes a tier, inverts one test, adds one, and corrects the skill paragraph that described the removed behavior.

fagemx:
## Code Review: Round 2
Reviewed full SHA: d45bbc3417f15de6d01584ff2a7efb891d4b6f37
Base full SHA: e1b37429ffe83fdf002b6c5f9bbb1c9b5b2c9287
Prior round: Round 1 at 9fc301c1bcf8a0f7f59d87a9dafd88b13b9ec5cd (Changes Requested, P0=0, P1=1)

Round 1 finding: **P1-1 closed**, and closed the right way — the tier is removed, not narrowed. `resolve_unclaim_target` (`crates/edda-cli/src/cmd_bridge.rs:723`) now ends at the `infer_session_id` tier and bails; there is no arm that can return a session id derived from `claims`. Re-ran Round 1's exact scenario against a binary I built at this SHA (two live heartbeats, `sess-a` holding `worker-a` on `src/a/*`, bare `unclaim` from a shell with `EDDA_SESSION_ID`/`EDDA_SESSION_LABEL` unset):

```
Active sessions (2):
  sess-a — worker-a (0s ago) [src/a/*]
  sess-b — worker-b (0s ago)
$ edda unclaim
Error: cannot tell which claim is yours, so --session is required.
Claims on the board:
  sess-a — worker-a (src/a/*)
exit=1
after: "claims":[{"label":"worker-a","paths":["src/a/*"],"session_id":"sess-a",...}]
       sess-a — worker-a (1s ago) [src/a/*]
```

Refuses, exits non-zero, claim stands. Nothing else regressed — all RAN against the same binary under an isolated `EDDA_STORE_ROOT`: `--session sess-a` releases (`Unclaimed scope for session: sess-a (worker-a)`, exit 0, board `"claims":[]`); `EDDA_SESSION_ID=sess-a` releases (exit 0); empty board refuses (`no claims on the board; nothing to unclaim`, exit 1); `--session sess-zzz` holding nothing refuses (exit 1, claim survives); `EDDA_SESSION_ID=sess-nobody` refuses (exit 1, claim survives).

**Did removing the tier break the case it was added for?** No — the hookless bare-CLI flow still terminates usefully, and the refusal carries the id rather than just naming the flag. RAN: `edda claim "auth" --paths "src/auth/*"` prints `session: cli-auth`; bare `edda unclaim` then prints `cannot tell which claim is yours, so --session is required.` / `cli-auth — auth (src/auth/*)` (exit 1); `edda unclaim --session cli-auth` releases it (exit 0, board empty). The id the user needs is in the error text, so the round trip costs one extra command and no lookup.

BLOCKING: P0=0, P1=1

- **P1-2 — the rustdoc on the rewritten function still documents the removed tier as present and load-bearing; introduced by this delta.** `crates/edda-cli/src/cmd_bridge.rs:718-722`:

  ```rust
  /// Explicit `--session` and `EDDA_SESSION_ID` win, then a sole live heartbeat,
  /// then the board itself. The board tier is what makes a bare `edda unclaim`
  /// work for a claim made from a plain CLI, which writes no heartbeat.
  fn resolve_unclaim_target(
  ```

  Byte-identical to `9fc301c` — the tier was deleted underneath it and the doc was not updated. Both sentences are now false, and I verified the second by execution: bare `edda unclaim` against a plain-CLI claim does **not** work, it exits 1 (transcript above). It is contradicted by the comment in the body of the same function, eighteen lines below at `:740-741` (`Do NOT fall back to "the sole claim on the board"`), and by both identity tests.

  Concrete failure scenario: this is the only artifact left in the repo that still asserts the board tier exists — a repo-wide sweep of `*.rs`/`*.md` for `board tier` / `the board itself` / `sole claim` returns this comment and nothing else. It does not merely misdescribe the code, it prescribes the removed behavior as intended design ("is what makes a bare `edda unclaim` work"). A maintainer who reconciles the function to its own doc reintroduces P1-1 verbatim, and the doc tells them the hazardous arm was the point of the function.

  Doc-only: no runtime impact, not shipped to users, product behavior is fully correct and verified above. Routing: it is not adjacent (it is the changed function), not pre-existing (this delta made it false), and not speculative (falsified by execution), so by the repo's own rule it does not qualify for follow-up. Fix is the three comment lines.

Response verification: every load-bearing claim re-derived, not reused.
- "The board tier is gone" — **supported**; source read + the reproduction above.
- The response's reproduction transcript — **supported**; I did not reuse it, I re-derived it and got the same message, exit code, and surviving claim.
- "the old test was inverted and renamed rather than deleted" — **supported**; `git diff 9fc301c..d45bbc3` shows `unclaim_without_session_releases_the_sole_claim` → `unclaim_without_identity_refuses_rather_than_guessing_the_sole_claim` with `expect(...)` → `expect_err(...)` and the emptiness assert flipped to `len() == 1`.
- "the new test writes two heartbeats and pins your scenario" — **supported**; two `write_heartbeat_minimal` calls, one `write_claim`, `expect_err`, `len() == 1`.
- Tier 3 is pre-existing — **supported, verified independently at `e1b3742`**, not taken from the response. Base `unclaim` was `let (session_id, _label) = resolve_session_id(cli_session, &project_id, "cli");` and base `resolve_session_id` tier 3 is the identical `infer_session_id(project_id)` call; base then wrote the unclaim unconditionally, so base was strictly worse (it also printed success when it released nothing). At head `claim` (`:667`) and `decide` (`:778`, resolving at `:821`) still share it. Also confirmed empirically at head: one live heartbeat + bare `unclaim` releases that peer's claim (exit 0). Correctly scoped out, and correctly filed — #488 item 1 carries it with basis SHA `d45bbc3`, labelled pre-existing.
- "308 passed" — **corroborated arithmetically, not by running it**: `cargo test -p edda unclaim` gives `6 passed; 0 failed; 302 filtered out` = 308. I did not run the full `-p edda` suite; exact-head CI did, green on all three OSes.
- fmt / clippy — **not re-run**; superseded by exact-head CI (Format, plus Clippy `--workspace --all-targets` on ubuntu, macos and windows), all green.
- "exact-head CI pending" — now **completed success, all 8 jobs**, including Test (windows-latest). Round 1's `edda-ledger sqlite_store::tests::concurrent_writers_no_data_loss` red did **not** recur at this head, which independently confirms Round 1's environmental classification; I did not have to make the call myself.
- Follow-ups: the response said it would file them and **this time it did** — #488 covers all three of Round 1's plus the tier-3 hazard, basis SHA `d45bbc3`. Checked #483, #486 and #488: no overlap, nothing of mine unfiled.

Test honesty: the six `unclaim` tests are honest. RAN `cargo test -p edda unclaim` → 6 passed, 0 failed. Against base semantics (`resolve_session_id` + unconditional `write_unclaim` + `Ok(())`) five of the six fail: the three `expect_err` refusal tests and both identity tests all get `Ok(())` and panic. The sixth, `unclaim_releases_the_explicit_session_scope`, passes at base **by design** — it is the pre-existing happy-path release test and must keep passing. So no test passes with the feature broken except the one that is supposed to. On the tier specifically, reasoned rather than demonstrated (read-only, I did not patch): with the board tier restored, `unclaim_without_identity_never_releases_a_live_peers_claim` gets `infer_session_id → None` (two live peers), the tier returns sole claimant `sess-a`, `held` is non-empty, `write_unclaim` runs and `unclaim` returns `Ok(())` — `expect_err` panics. Same for `unclaim_without_identity_refuses_rather_than_guessing_the_sole_claim` (sole claimant `cli-auth`). Both genuinely pin the removal. `unclaim_without_session_refuses_when_several_claims_exist` would still pass with the tier restored, so it is not the one doing the work.

Shipped skill: **every factual claim in Step 5 is true**, each executed against my own build, never the installed `edda.exe`. RAN `edda init` in a throwaway repo with a `.claude/` host marker (without one, `scaffold_skills` does not run at all — `cmd_init.rs:126`); the written `SKILL.md` is byte-identical to `crates/edda-cli/src/skills/coord-handoff.md`, sha256 `e93d34e282c5fdd6b50bfbc93a5995fcda240b30217ba1f50e84e149deb22066` on both. Verified sentence by sentence: `edda unclaim --session <id>` releases (exit 0); "the id is the `session:` line `edda claim` printed" — `edda claim "auth"` prints `session: cli-auth`; "run it without `--session` and the refusal lists every claim with its session id" — true in both refusal paths (`cli-auth — auth (src/auth/*)`); "`edda peers --json` carries it under `claims[].session_id`" — exact path confirmed in the raw JSON; "plain `edda peers` does not" — prints `No active sessions.` for a bare-CLI claim; "never reports success for a session that holds nothing" — both claimless paths exit 1 with the claim intact; the new paragraph on not picking a claim for you matches the two-live-peer transcript. The closing "every consumer joins claims against live heartbeats" sentence is unchanged by this delta; READ-verified for the enforcement consumer, `check_offlimits` skips any claim whose `session_id` is not in `active_sids` (`dispatch/tools.rs:230-238`). No false statement found in the shipped file this round.

IN SCOPE: the whole of `git diff 9fc301c..d45bbc3` — the removed board-resolution arm and the surviving tiers in `resolve_unclaim_target`, the doc comments on `unclaim` and `resolve_unclaim_target`, the inverted/renamed test and the added two-live-peer test, and the rewritten Step 5 of the shipped `coord-handoff.md`. Also re-derived the base-relative surface the delta lands on (`git diff e1b3742..d45bbc3` is two hunks: `unclaim`/`resolve_unclaim_target`/`describe_claims`, and tests — all read in full), the tier-3 pre-existence question at `e1b3742`, and the `check_offlimits` enforcement join. Out of scope and untouched by this delta: the automatic session-end path (bridges call `peers::write_unclaim` directly, unchanged), and `resolve_session_id` itself.

FOLLOW-UP ISSUE:
- None new. All four accepted follow-ups are already filed in #488 (basis SHA `d45bbc3`); #483 and #486 checked, no overlap. P1-2 above is deliberately not routed here — it is on the frozen surface and delta-caused.

Evidence:
- RAN (Windows 11, worktree default `target/`, binary built at `d45bbc3` — never the installed `edda.exe`):
  - `cargo build -p edda` — OK
  - `cargo test -p edda unclaim` — 6 passed, 0 failed, 302 filtered
  - Round 1's P1-1 scenario re-derived from that binary: two live heartbeats + one claim + bare `unclaim` → refusal, exit 1, claim survives
  - regression battery under isolated `EDDA_STORE_ROOT` with session env unset: explicit `--session` releases; `EDDA_SESSION_ID` releases; empty board refuses; `--session` holding no claim refuses; `EDDA_SESSION_ID` at a claimless session refuses
  - the hookless bare-CLI round trip (`claim "auth"` → bare `unclaim` refusal carrying `cli-auth` → `unclaim --session cli-auth` releases)
  - tier-3 probe: one live heartbeat + bare `unclaim` releases that peer's claim (exit 0) — the pre-existing hazard, at head
  - `edda init` in a throwaway repo + sha256 comparison of the written skill; every Step 5 CLI claim executed
- READ:
  - exact-head CI run 32295765744 on `d45bbc3` — **all 8 jobs success**: Format; Clippy on ubuntu, macos and windows; Test on ubuntu, macos and windows. `-p edda` is in the CI Windows test subset, so this crate's tests ran on all three platforms. Relied on for fmt/clippy/test rather than rerunning them.
  - base `e1b3742` `unclaim` and `resolve_session_id` (to settle tier-3 pre-existence myself)
  - `infer_session_id` (`discovery.rs:169-208`), `check_offlimits` (`dispatch/tools.rs:204-238`), `scaffold_skills` (`cmd_init.rs:126-180`)
  - #483, #486, #488
  - Nothing was taken from the response as evidence — its transcript, its 308 figure and its line citations were all re-derived.
- Could not corroborate: the response's "all four failing before the change" — read-only, I never observed them fail, and derived the property statically instead (five of six current tests fail against base semantics). Also did not check the `d-040` ledger decision cited for the flake classification; moot, since Windows CI is green at this head.
- Lane: the worktree's default `target/`; no ad-hoc `CARGO_TARGET_DIR`; nothing deleted.
- Receipt: the response's L0 gates (fmt / clippy `-p edda` / test `-p edda`) — READ, and superseded by exact-head CI green on all three OSes, so I reran neither. No full-workspace run: no reason arose, and CI covers it on all three platforms.

Cost: elapsed=unknown, tokens=unknown, tools=~30 calls

Verdict: Changes Requested

The product surface is done — P1-1 is genuinely closed, the fix is the right shape, the tests pin it, and the shipped skill is accurate for the first time in this loop. The one blocker is three lines of stale rustdoc on the function this PR rewrites, which still advertises the removed tier as the reason the function exists.

fagemx:
## Review Response: Round 2

New full SHA: `e031014c88ec49588be21a3c2175af663badc684`
Reviewed SHA this responds to: `d45bbc3417f15de6d01584ff2a7efb891d4b6f37`

### P1-2 — the rustdoc still described the removed tier · FIXED

Not disputed. `cmd_bridge.rs:718-722` was byte-identical to `9fc301c`: it listed "then the board itself" and called that tier the reason a bare `edda unclaim` works for a plain-CLI claim. I deleted the tier from the body and left its description standing **18 lines above the comment explaining why it must not exist**.

Your reason for blocking rather than deferring is the right one and I want it on the record: this does not merely misdescribe, it states the removed behavior as the function's *purpose*. A maintainer reconciling code to doc puts P1-1 back — the cross-session release of a live peer's scope. That is not adjacent, not pre-existing, and not speculative, so the repo's own routing test sends it to blocking, and deferring it would have been the wrong call even though it is doc-only.

The rustdoc now states the three tiers and says plainly that there is deliberately no further one, with the reason. I also swept the repository for any other text describing board resolution: the three places that describe this behavior — the verb's rustdoc, the resolver's rustdoc, and the shipped `coord-handoff` skill — now agree, and nothing else mentions it.

### This is the same defect I have been fixing all PR

Worth naming rather than letting it pass as a one-off. Across this loop I have shipped: a spec section I claimed to have fixed and had not; a shell command that runs in no shell; two false `edda peers` statements; and now a rustdoc prescribing a hazard I had just removed. The shape is identical every time — **I change behavior and leave a description of the old behavior standing**, then verify the change rather than the descriptions.

The mechanism that would catch the class is the skill-doctest work #486 exists around, but rustdoc is outside that scope. What actually caught all five was an independent reviewer re-deriving instead of reading my summary.

### Everything else you checked

No argument with any of it, and the two I would have expected to be shakiest both held: the tier-3 "pre-existing" claim verified independently at `e1b3742`, and the shipped skill executing true sentence by sentence — which you note is the first time in this loop. That is the only part of this PR I would call finished rather than merely fixed.

### Gates on the new SHA

- RAN: `cargo test -p edda` — 308 passed, 0 failed
- RAN: repo-wide `git grep` for any remaining description of board resolution — only unrelated hits (`coord-sync.md` board state, `SKILL.md` recovery, a bridge test string)
- READ: exact-head CI pending on `e031014`

Doc-only delta, no runtime change. Round 3 requested on `e031014c88ec49588be21a3c2175af663badc684`.

fagemx:
## Code Review: Round 3
Reviewed full SHA: e031014c88ec49588be21a3c2175af663badc684
Base full SHA: e1b37429ffe83fdf002b6c5f9bbb1c9b5b2c9287
Prior round: Round 2 at d45bbc3417f15de6d01584ff2a7efb891d4b6f37 (Changes Requested, P0=0, P1=1)

Round 2 finding: **P1-2 closed.** Both false sentences are gone, and the replacement is accurate against the body it sits on — checked clause by clause, not by reading the response.

`resolve_unclaim_target` (`crates/edda-cli/src/cmd_bridge.rs:723-754`) has exactly three arms that return an id, in exactly the order the new rustdoc states:

1. `cli_session.filter(|s| !s.is_empty())` — "Explicit `--session` wins"
2. `std::env::var("EDDA_SESSION_ID")`, non-empty — "then `EDDA_SESSION_ID`"
3. `peers::infer_session_id(project_id)` — "then a sole live heartbeat". Verified at the source rather than by name: `discovery.rs:169-208` collects heartbeats with `age <= stale_threshold` ("live") and returns `Some` only when `active.len() == 1` ("sole"). Both words are load-bearing and both are right.

"There is deliberately no further tier" is true: after arm 3 the only remaining statements are the two `anyhow::bail!` arms (`:747` empty board, `:750` ambiguous). No path returns a session id derived from `claims`. The six tests pin this — `unclaim_without_identity_never_releases_a_live_peers_claim` (two heartbeats, one claim) and `unclaim_without_identity_refuses_rather_than_guessing_the_sole_claim` (sole claim, no heartbeat) both `expect_err` and assert the claim survives. The rationale clause now matches the in-body comment at `:740-745` instead of contradicting it.

One clause I checked and am explicitly **not** raising: "Refuse and show the board instead, so the id for `--session` is in the error" holds for the ambiguous arm (`describe_claims` prints `session_id — label (paths)`), while the empty-board arm prints no listing. That arm has no id to print and says so (`no claims on the board; nothing to unclaim`), so the sentence is not falsified — it is a precision nit, not a defect, and manufacturing it into a blocker after three rounds would be the wrong call.

Response verification: re-derived, nothing reused.
- **"Doc-only delta, no runtime change" — supported.** Verified rather than assumed: `git log --format='%H %P' -1 e031014` shows d45bbc3 as the sole parent, so `git diff d45bbc3..e031014` *is* the whole change — 1 file, 5 insertions, 3 deletions, every changed line a `///` line, no context line altered. Nothing executable moved.
- **"The rustdoc now states the three tiers ... no further one" — supported**, by the clause-by-clause check above.
- **"I swept the repository ... the three places that describe this behavior now agree, and nothing else mentions it" — supported for the three named places, too strong as stated.** I re-derived with my own `git grep` over all 416 tracked files (dot-directories included). The verb rustdoc (`:678-691`), the resolver rustdoc (`:717-722`) and `coord-handoff.md` Step 5 all agree with the code, and the only surviving hit for the removed language (`git grep -in "board tier|the board itself|sole claim|sole claimant"`) is the *negative* in-body comment at `:743` ("Do NOT fall back to ..."). But the sweep missed two places that do describe how the session is resolved when the flag is omitted: the clap help on `unclaim`'s `--session`, at `crates/edda-cli/src/main.rs:173` and `crates/edda-cli/src/cmd_bridge.rs:123`, both "Session ID (auto-inferred from active heartbeats if omitted)". **They do not contradict current behavior** — that is tier 3, described exactly — so this is not blocking; they are silent about the refusal, which is the follow-up below. Everything else matching `unclaim` across the repo (`consistency-contract.md:76,134`, `P2_DRILL_2026-08-16.md:41`, `.tmp/phase3-comment.md:36,77`, `peers/heartbeat.rs:264`, the five bridge dispatchers) describes the automatic `write_unclaim` on SessionEnd, which this PR does not touch and which remains accurate. `coord-sync.md:37` "Unclaimed areas" is unrelated usage.
- **"308 passed" — not reused, and not fully re-derived.** I ran only the focused filter: 6 passed, 0 failed, 302 filtered out = 308 total, arithmetically consistent. I did not run the full `-p edda` suite myself; exact-head CI did, green on all three OSes.
- **fmt / clippy PASS — corroborated, not taken on trust.** Exact-head CI Format plus Clippy `--workspace --all-targets` on ubuntu, macos and windows are all green at this SHA, and I ran `cargo clippy -p edda --all-targets -- -D warnings` locally (clean). A rewritten doc comment is a real lint surface (`doc_markdown`, intra-doc links), which is why I ran it rather than only reading CI.

IN SCOPE: the whole of `git diff d45bbc3..e031014` — the rustdoc on `resolve_unclaim_target`, checked against the function body, `infer_session_id`, the in-body comment and the six tests; whether the delta is comment-only; and a repo-wide re-derivation of every surviving description of unclaim's session resolution, including the shipped `coord-handoff.md` Step 5 and the base-relative skill diff. Out of scope and unchanged by this delta: the automatic session-end path (bridges call `peers::write_unclaim` directly) and `resolve_session_id`.

BLOCKING: P0=0, P1=0

FOLLOW-UP ISSUE (basis SHA e031014c88ec49588be21a3c2175af663badc684; checked #483, #486 and #488 — no overlap, this is not among the four in #488):
- The `--session` clap help for `unclaim` (`crates/edda-cli/src/main.rs:173`, `crates/edda-cli/src/cmd_bridge.rs:123`) reads "Session ID (auto-inferred from active heartbeats if omitted)". Accurate for tier 3 and untouched by this PR (both sit outside its two hunks, `@@ -676,14 +676,98 @@` and `@@ -2208,6 +2292,125 @@`), so pre-existing and non-blocking — but it is what `edda unclaim --help` shows, and it now implies omission is always safe when omitting the flag hard-fails with exit 1 whenever there is no live heartbeat. Closest kin is #488 items 3/4 (undocumented nonzero exit, refusal wording); this is the same gap on the `--help` surface. Fits #488.

Evidence:
- RAN (Windows 11, worktree `.claude/worktrees/verification-cost-discipline` at e031014, clean tree):
  - `cargo test -p edda unclaim` — **6 passed, 0 failed, 302 filtered out**, all six named: `unclaim_refuses_a_session_that_holds_no_claim`, `unclaim_refuses_when_the_board_is_empty`, `unclaim_without_identity_never_releases_a_live_peers_claim`, `unclaim_releases_the_explicit_session_scope`, `unclaim_without_identity_refuses_rather_than_guessing_the_sole_claim`, `unclaim_without_session_refuses_when_several_claims_exist`. Build freshness confirmed rather than assumed: the test binary's mtime (04:08:27) is later than `cmd_bridge.rs`'s (04:07:54) and the tree has not been touched since, so the binary is built from this SHA's content.
  - `cargo clippy -p edda --all-targets -- -D warnings` — clean
  - `git log --format='%H %P' -1 e031014`; `git diff d45bbc3..e031014` in full; `git diff e1b3742..e031014 --stat` and the `coord-handoff.md` hunk
  - `git grep` sweeps over all tracked files for `unclaim`, `board tier`, `the board itself`, `sole claim(ant)`, `cannot tell which claim`, `holds no claim`, `nothing to unclaim`, `auto-inferred`, `infer_session_id`, `release the scope`
  - read in full: `resolve_unclaim_target`, `unclaim`, `describe_claims`, the six tests, `infer_session_id`
- READ:
  - **exact-head CI run 32296968581 on `e031014` — completed, conclusion `success`, all 7 jobs green**: Format; Clippy on ubuntu, macos and windows; Test on ubuntu, macos **and windows**. I waited for it rather than reviewing against a partial run, and made no inherited call: nothing was red, so no deterministic-vs-environmental classification was needed this round. `edda` is in the CI Windows test subset, so this crate's tests ran on all three platforms — that is what let me hold the local run to the focused filter.
  - #483, #486, #488 (follow-up dedup)
  - Round 1 and Round 2 verdicts and both responses — for the claims to check, not as evidence
- Lane: the worktree's default `target/`; no ad-hoc `CARGO_TARGET_DIR`; nothing deleted.
- Receipt: the response's L0 gates (fmt / clippy `-p edda` / test `-p edda` 308) — READ, and superseded by exact-head CI green on all three OSes plus my own focused clippy and test runs. No full-workspace run: docs-only delta with green exact-head CI on all three platforms, so no reason arose.
- Could not corroborate: the exact hit list the response's own `git grep` returned (I did not reuse it — I ran my own sweep, which found the two clap-help strings its summary did not name); and this PR's earlier-round pre-implementation observations ("all four failing before the change"), which are read-only-invisible to me and moot at this SHA.

Cost: elapsed=unknown, tokens=unknown, tools=~22 calls

Gates relied on: exact-head CI run 32296968581 at `e031014` (Format; Clippy `--workspace --all-targets` on ubuntu/macos/windows; Test `--workspace` on ubuntu/macos and the Windows subset that includes `edda`) — all green; plus local `cargo test -p edda unclaim` 6/6 and `cargo clippy -p edda --all-targets -- -D warnings` clean at this SHA.

The delta is three lines of comment and it is now correct. The three tiers in the doc are the three tiers in the code, in order; "no further tier" is true of the body and pinned by two tests; and the description no longer prescribes the hazard the previous two rounds removed. Confirmed `headRefOid` still equals `e031014c88ec49588be21a3c2175af663badc684` immediately before posting.

Issue #488, opened 2026-08-19
coord(cli): session identity inference can adopt a peer's session, plus three unclaim UX gaps


fagemx:
From the #487 review. Basis SHA: `d45bbc3417f15de6d01584ff2a7efb891d4b6f37`.

## 1. `resolve_session_id` tier 3 assumes "sole active session = me"

`crates/edda-cli/src/cmd_bridge.rs` → `crates/edda-bridge-claude/src/peers/discovery.rs:204-208`

`infer_session_id` returns `Some` when exactly one live heartbeat exists. Inside a hooked session that is correct — the sole heartbeat is the caller's own. From a bare shell standing next to exactly one live peer it is wrong: the caller has no heartbeat, and the inference silently adopts the peer's identity.

For `unclaim` that means a bare shell can release a live peer's claim, dropping the off-limits protection `check_offlimits` (`dispatch/tools.rs:204-236`) enforces for its owner. `claim` and `decide` share the tier, so the same misattribution can record a claim or a decision under someone else's session.

Pre-existing, not introduced by #487 — that PR removed the *board* tier it had added for the two-or-more-peer case, and deliberately left tier 3 alone because changing it affects several verbs.

The underlying gap is that a process has no way to assert "this heartbeat is mine". Options worth weighing: have hooked sessions export `EDDA_SESSION_ID` so tier 3 is never needed; or make tier 3 require a matching cwd/pid rather than mere uniqueness; or scope tier 3 to read-only verbs and require explicit identity for anything that mutates the board.

## 2. A session's second claim silently replaces its first

`compute_board_state` keys claims in a `HashMap` on `session_id`, so one session holding two scopes is not representable. `edda claim "a"` followed by `edda claim "b"` in the same session leaves only `b`, with no warning that `a` was dropped. Either reject the second claim, merge the paths, or say what happened.

## 3. `unclaim`'s new nonzero exit is undocumented for the CI case

`coord-handoff.md` Step 5 names CI as a place claims are made without hooks, and `unclaim` now exits 1 when there is nothing to release. A CI step that calls it unconditionally at teardown will start failing the job. Either document the exit codes in the skill, or give the verb an `--if-claimed` / idempotent mode for teardown use.

## 4. The "holds no claim" refusal does not say what to do

`unclaim --session X` where X holds nothing prints the board but never says to pass a different `--session`. The ambiguous-target refusal does say it. Make the two consistent.

---

Related: #483 (from #481), #486 (from #485). No overlap.

[...]

PR #489, opened 2026-08-19
test(edda-cli): let shipped skills prove their own CLI claims

fagemx:
Closes the gap that produced four blocking findings across #481, #485 and #487.

## The gap

`crates/edda-cli/src/skills/*.md` are compiled into the binary with `include_str!` and written into every project by `edda init`. Before this PR the only assertion about them anywhere in the workspace was:

```rust
assert!(path.exists(), "{name}/SKILL.md should be scaffolded");
```

File exists. Nothing about whether the commands they teach are real.

Over one review loop, this one file set shipped **four false CLI claims**, each passing fmt, clippy and the full suite:

| Claim | Reality |
|---|---|
| "there is currently no `edda unclaim` verb (GH-455)" | it existed, implemented and tested — shipped to every project for months |
| `edda unclaim [--session <id>]` | the bare form resolved a different session, exited 0, released nothing |
| "`edda peers` shows it too" | `peers` lists live heartbeats; a bare-CLI claim has none |
| rustdoc: "then the board itself…" | described a tier deleted in the commit before |

All four were caught by an independent reviewer reading the code, never by CI.

## The mechanism

A skill can now carry its own proof. Mark a fenced block `bash edda-doctest` and it runs against the built binary in a throwaway project with an isolated store:

````markdown
```bash edda-doctest
> session: cli-auth
$ edda unclaim
! cannot tell which claim is yours
! cli-auth
```
````

`$` runs, `>` asserts success containing the text, `!` asserts failure containing the text. Unmarked blocks stay prose, so nothing has to be converted at once.

`!` expectations additionally assert the output is **not** `unrecognized subcommand`. A removed or misspelled verb exits non-zero too and would otherwise satisfy any failure expectation — that is the hole that let a doc declaring a working verb nonexistent survive.

## Mutation-tested, not merely green

A passing new test proves little on its own, so I reinjected each of the four historical false claims in turn and ran the harness:

```
CAUGHT  #1  a verb declared not to exist (main, before #485)
CAUGHT  #2  bare unclaim documented as working (#485 first fix)
CAUGHT  #3  `edda peers` said to show the id (#485 second fix)
CAUGHT  #4  unclaim of a claimless session called a success (base behavior)
```

4/4 rejected; the file was restored afterwards and the working tree verified clean of mutations. The mutation script is not committed — it is evidence about the harness, not a fixture.

## Scope, and what this does not cover

`coord-handoff` Step 5 is converted first, being where all four claims landed. Its four blocks execute the claim/unclaim round trip, the refusal that lists claim ids, the refusal for a session holding nothing, and the `peers` / `peers --json` distinction.

The other skills mention 12 distinct verbs between them and remain unconverted — deliberately, so this PR is reviewable. Converting them is follow-up work, and the harness reports which files carry blocks so coverage is visible.

**It does not cover rustdoc.** Finding #4 above lived in a doc comment, outside any skill, and this mechanism would not have caught it. Nothing here replaces independent review; it removes one recurring class from the reviewer's plate.

## First integration test in the workspace

`.claude/CLAUDE.md` records that tests were inline until now. This one has to be an integration test: `CARGO_BIN_EXE_edda` is only defined for `tests/`.

## Evidence

- RAN `cargo test -p edda` — 308 unit + 2 integration, 0 failed
- RAN the mutation check above — 4/4 caught
- Receipt: L0 focused gates above; `-p edda` is in the CI Windows test subset, so exact-head CI covers this crate on all three platforms

### DIFF
--- a/crates/edda-cli/src/skills/coord-handoff.md
+++ b/crates/edda-cli/src/skills/coord-handoff.md
@@ -68,24 +68,53 @@ Where host bridge hooks are wired, unclaim happens automatically on session
 end — nothing to do.
 
 Where they are not (claims made from a bare CLI, CI, or a host without bridge
-hooks), release the scope by naming it:
+hooks), release the scope by naming it. The id is the `session:` line
+`edda claim` printed:
 
-```bash
-edda unclaim --session <id>
+```bash edda-doctest
+$ edda claim "auth" --paths "src/auth/*"
+> session: cli-auth
+$ edda unclaim --session cli-auth
+> Unclaimed scope for session: cli-auth
 ```
 
-The id is the `session:` line `edda claim` printed. Run it without `--session`
-and the refusal lists every claim with its session id, so the value you need is
-in the error. `edda peers --json` carries it too, under `claims[].session_id`;
-plain `edda peers` does not, because it lists live heartbeats and a bare-CLI
-claim has none.
+Omit `--session` and it refuses rather than guessing, listing every claim with
+its id — so the value you need is in the error:
 
-`unclaim` deliberately does not pick a claim for you when you have no session
-of your own: it cannot tell whose claim it is, and releasing the wrong one
-would drop the off-limits protection a live peer is relying on.
+```bash edda-doctest
+$ edda claim "auth" --paths "src/auth/*"
+> session: cli-auth
+$ edda unclaim
+! cannot tell which claim is yours
+! cli-auth
+```
 
-`unclaim` never reports success for a session that holds nothing — if it prints
-a released scope, that scope is gone.
+That refusal is deliberate. A caller with no session of its own cannot tell
+whose claim it is, and releasing the wrong one would drop the off-limits
+protection a live peer is relying on.
+
+`unclaim` also never reports success for a session that holds nothing — if it
+prints a released scope, that scope is gone:
+
+```bash edda-doctest
+$ edda claim "auth" --paths "src/auth/*"
+> session: cli-auth
+$ edda unclaim --session cli-nobody
+! holds no claim
+```
+
+`edda peers --json` carries the id too, under `claims[].session_id`. Plain
+`edda peers` does not, because it lists live heartbeats and a bare-CLI claim
+has none:
+
+```bash edda-doctest
+$ edda claim "auth" --paths "src/auth/*"
+> session: cli-auth
+$ edda peers --json
+> "session_id": "cli-auth"
+$ edda peers
+> No active sessions
+```
 
 Enforcement stays safe either way: every consumer joins claims against live
 heartbeats, so a claim left behind by a dead session does not block a peer.
--- a/crates/edda-cli/tests/skill_doctest.rs
+++ b/crates/edda-cli/tests/skill_doctest.rs
@@ -1,0 +1,248 @@
+//! Executable examples for the coordination skills shipped by `edda init`.
+//!
+//! `crates/edda-cli/src/skills/*.md` are compiled into the binary with
+//! `include_str!` and written into every project. Nothing used to check that
+//! the commands they teach are real: over one review loop this file set shipped
+//! four separate false CLI claims — a verb declared not to exist while it did,
+//! an invocation whose bare form silently released nothing, and twice a command
+//! named as the way to read a value it does not print. Every one passed fmt,
+//! clippy and the whole test suite.
+//!
+//! A skill can now carry its own proof. Mark a fenced block `bash edda-doctest`
+//! and this harness runs it against the built binary in a throwaway project:
+//!
+//! ```text
+//! ```bash edda-doctest
+//! $ edda claim "auth" --paths "src/auth/*"
+//! > session: cli-auth
+//! $ edda unclaim
+//! ! cannot tell which claim is yours
+//! ```
+//! ```
+//!
+//! `$` runs a command, `>` asserts it succeeded and printed the text, and `!`
+//! asserts it failed with the text. Unmarked blocks are prose and are ignored,
+//! so existing examples do not have to be converted at once.
+
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+const EDDA: &str = env!("CARGO_BIN_EXE_edda");
+
+/// One `$`/`>`/`!` line from a doctest block.
+#[derive(Debug)]
+enum Step {
+    Run(Vec<String>),
+    ExpectOk(String),
+    ExpectErr(String),
+}
+
+fn skills_dir() -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR")).join("src/skills")
+}
+
+/// Split a command line on spaces, keeping double-quoted runs together.
+///
+/// Enough for the shapes skills actually use (`--paths "src/a/*"`); anything
+/// needing more is a sign the example belongs in a test rather than a doc.
+fn split_args(line: &str) -> Vec<String> {
+    let mut out = Vec::new();
+    let mut cur = String::new();
+    let mut quoted = false;
+    for ch in line.chars() {
+        match ch {
+            '"' => quoted = !quoted,
+            c if c.is_whitespace() && !quoted => {
+                if !cur.is_empty() {
+                    out.push(std::mem::take(&mut cur));
+                }
+            }
+            c => cur.push(c),
+        }
+    }
+    if !cur.is_empty() {
+        out.push(cur);
+    }
+    out
+}
+
+/// Pull every `bash edda-doctest` block out of one skill.
+fn parse_blocks(markdown: &str) -> Vec<Vec<Step>> {
+    let mut blocks = Vec::new();
+    let mut current: Option<Vec<Step>> = None;
+
+    for line in markdown.lines() {
+        let trimmed = line.trim();
+        if trimmed.starts_with("```") {
+            if let Some(block) = current.take() {
+                blocks.push(block);
+            } else if trimmed == "```bash edda-doctest" {
+                current = Some(Vec::new());
+            }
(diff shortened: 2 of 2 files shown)


fagemx:
## Code Review: Round 1
Reviewed full SHA: d17c5ab3dd4b5ce344fe686d8f1a8c56de56e770
Base full SHA: 65f76028dda58413ddc409a6d0224d3273b94526

IN SCOPE: the 26-minute `Test (ubuntu-latest)` job on the exact head (cost regression, treated as priority); whether the harness catches the four historical false-claim classes, re-derived independently of the PR body; harness soundness (`checked > 0` guard, `split_args`, block isolation, `parse_blocks` fencing, assertion-before-command); truth of the four converted blocks executed against a binary built at this SHA; and whether converting Step 5 dropped anything true relative to base.

BLOCKING: P0=0, P1=0

### 1. The 26-minute CI job is not this PR (priority item, not blocking)

The harness is exonerated by direct measurement, on four independent lines:

- **Its own cost is ~2 seconds.** Same run, Windows job 96214962687: `Running tests\skill_doctest.rs` → `test result: ok. 2 passed; ... finished in 1.96s`. That is all four fixtures (`git init` + `edda init` each) plus 13 `edda` invocations, on the slowest platform for process spawn.
- **On Linux it ran *after* the slow binary.** `Running unittests src/main.rs (…/edda-c285ad3d…)` starts `20:27:04` and takes 667s; `Running tests/skill_doctest.rs` only starts `20:38:11` and takes 14.1s. It cannot have caused what finished before it began.
- **Untouched crates slowed by the same factor.** Summing per-binary execution (libtest `running N tests` → `test result:`): base run 32296968581 = **13.6s total**; head run 32298364289 = **1405.7s total** (~103x). The pre-existing 304-test `edda-cli` binary went **4.2s → 667s** though no unit test changed. Tests in `edda-ledger`, `edda-derive` and `edda-ask` show 11–13s *inter-test* gaps. This diff touches two files, both under `crates/edda-cli/`; those three crates do not depend on `edda-cli`, so nothing here can alter their runtime.
- **Compile was normal** (`Finished test profile in 1m 36s`), and the same new test compiled and ran on the other two platforms inside their usual ranges: macOS 114s (range 82–113s over the preceding runs), Windows 291s (range 291–325s).

Across the last 20 `ci.yml` runs the ubuntu Test job is 62–196s; 1563s is a unique outlier. The uniform ~100x penalty on SQLite/`tempfile`-heavy tests is the signature of a degraded runner, not of new work. Steady-state cost of this mechanism is ~2s per push — it does not defeat its own purpose.

*Not corroborated:* I could not establish the root cause of the runner degradation itself (no incident data available to me), and there is only one CI run on this branch, so I cannot exhibit a fast re-run of this exact SHA. I deliberately did not re-run the green job — it would spend CI and risk flipping a passing check. The negative result above does not depend on knowing the cause.

### 2. The harness does catch what it claims — re-derived, not reused

I did not reuse the PR body's mutation figures. I compiled the **unmodified** `crates/edda-cli/tests/skill_doctest.rs` with `rustc --test`, overriding only compile-time env (`CARGO_MANIFEST_DIR` → a scratch skills dir, `CARGO_BIN_EXE_edda` → a binary I built at this SHA), so `parse_blocks` and every assertion are byte-identical to the shipped test. Baseline reproduces the real run exactly: 10 assertions, pass. Then 10 mutations of a scratch copy — the repository was never written.

| # | Mutation | Result |
|---|---|---|
| M1 | `unclaim` declared a nonexistent verb | REJECTED — expected failure, command succeeded |
| M2 | bare `unclaim` documented as working | REJECTED — expected success, command failed |
| M3 | plain `peers` said to print the id | REJECTED — output did not contain `"cli-auth"` |
| M4 | claimless unclaim called a success | REJECTED — expected success, command failed |
| M5 | nonexistent verb, want-text inside clap's tip | REJECTED — "the command does not exist" |
| M6 | fence marker renamed everywhere | REJECTED — `checked > 0` |
| M7 | **one** block demoted to a plain ` ```bash ` fence | ACCEPTED, assertions 10 → 8 |
| M8 | `>` before any `$` | REJECTED — `` `> ` before any command`` |
| M9 | first block's closing fence deleted | REJECTED |
| M10 | unrecognized line inside a block | REJECTED |

**On the `unrecognized subcommand` guard — it is load-bearing, not decorative.** M5 is the direct test: `$ edda unclaimm` with `! unclaim`. Clap emits `error: unrecognized subcommand 'unclaimm'` followed by `tip: some similar subcommands exist: 'claim', 'unclaim'` and exits 2 — so `contains("unclaim")` *is* satisfied and the exit *is* non-zero. Without the guard that mutation passes green. With it, rejected. So yes: a nonexistent verb cannot satisfy a failure expectation.

### 3. Soundness — one gap, routed as follow-up

`checked > 0` catches a whole-file marker rename **today** (M6), but only because exactly one of the five skills carries blocks. M7 shows the partial case is silent: demoting a single block's fence drops it to 8 assertions and stays green. Filed below rather than blocked — the guard is adequate at current coverage, the count is printed, and a demoted fence is visible in a markdown diff.

Everything else held. `split_args` handles every shape in use (`--paths "src/auth/*"` → correct argv); its documented limits (no single quotes, no escapes) and an empty `""` arg being dropped all fail loudly, never silently pass; shell metacharacters become literal argv and error out. Blocks cannot leak: `fixture()` gives each its own tempdir and `EDDA_STORE_ROOT`, and env is applied per-child through `Command::env`/`env_remove` — never `std::env::set_var` — so nothing mutates the test process. Confirmed empirically: all four blocks independently see exactly one claim. `EDDA_SESSION_ID` and `EDDA_SESSION_LABEL`, the two vars that determine identity, are both cleared. And the harness reads `src/skills/*.md`, precisely the paths `cmd_init.rs` feeds to `include_str!` — it validates the shipped bytes, not a copy.

### 4. The four blocks are true

Executed against `target/debug/edda.exe` built at this SHA in throwaway projects (not the installed `edda.exe`):

- `claim` → exit 0, `session: cli-auth`; `unclaim --session cli-auth` → exit 0, `Unclaimed scope for session: cli-auth (auth)`
- bare `unclaim` → exit 1, `cannot tell which claim is yours, so --session is required.` + `cli-auth — auth (src/auth/*)`
- `unclaim --session cli-nobody` → exit 1, `session cli-nobody holds no claim; nothing was released.`
- `peers --json` → exit 0, `"session_id": "cli-auth"`; `peers` → exit 0, `No active sessions.`

### 5. Nothing true was lost in the conversion

Diffed the prose against base. Every removed sentence is re-expressed: the id-is-the-`session:`-line rule, the refusal listing ids, the `peers --json` / plain `peers` distinction, the rationale for refusing to guess, and the never-reports-false-success guarantee. The two reassurances checked specifically are **unchanged context lines, still present**: "Enforcement stays safe either way: every consumer joins claims against live heartbeats…" and "Where host bridge hooks are wired, unclaim happens automatically on session end — nothing to do." The only thing gone is the generic placeholder `edda unclaim --session <id>`, now a concrete executable id with the substitution rule stated in the adjacent prose. Not a loss.

FOLLOW-UP ISSUE:
- Per-block doctest coverage can be lost silently. `checked > 0` in `crates/edda-cli/tests/skill_doctest.rs` is a global guard; demoting one block's ` ```bash edda-doctest ` fence drops it from the run with the suite still green (reproduced: 10 → 8 assertions). Harmless at today's coverage — one file, and total loss is caught — but it stops being safe exactly when the PR's own stated follow-up lands and a second skill is converted, since either file could then go dark unnoticed. Worth a per-file expectation or an asserted assertion count when that conversion happens. Basis SHA d17c5ab3dd4b5ce344fe686d8f1a8c56de56e770. Not covered by #483 (policy wording) or #488 (unclaim UX); #486 is closed.

Evidence:
- RAN `cargo test -p edda --test skill_doctest -- --nocapture` at d17c5ab → `2 passed; 0 failed`, `skill doctests: 10 assertions across ["coord-handoff.md"]`, 5.45s
- RAN the four blocks by hand against `target/debug/edda.exe` built at this SHA, in fresh `git init` + `edda init --no-hooks` projects with an isolated `EDDA_STORE_ROOT` — outputs quoted in §4
- RAN control probes for the guard: `edda unclaimm`, `edda totally-not-a-verb`, `edda unclaim --sesion` → clap text and exit codes quoted in §2
- RAN 10 mutations against the unmodified harness recompiled via `rustc --test` over a scratch skills copy — table in §2; repository never written
- RAN per-binary timing extraction over the raw CI logs of jobs 96214962579 (ubuntu head), 96214962687 (windows head), 96214962601 (macos head) and 96210423318 (ubuntu base)
- READ exact-head CI run 32298364289 @ d17c5ab — 7/7 green: Format; Clippy ubuntu/macos/windows; Test ubuntu/macos/windows. Clippy is workspace-wide `--all-targets` on all three OSes, so the new `tests/` target is linted everywhere
- READ baseline job durations from runs 32262704222, 32268862331, 32296968581 and the last 20 `ci.yml` runs, computed from the API rather than quoted from the PR body
- Per the standing instruction, no figure, transcript or line citation from the PR body was reused as evidence; the 4/4 mutation result was re-derived from scratch and independently agrees
- Lane: worktree default `target/` at `C:\ai_agent\edda\.claude\worktrees\verification-cost-discipline`; no ad-hoc `CARGO_TARGET_DIR`; nothing deleted; read-only on the repository
- Receipt: exact-head CI run 32298364289 (full workspace fmt + clippy + test on three platforms) plus the focused local runs above. No full local workspace rerun — CI covers it green at this SHA and the ladder's L2 row gives no reason to spend one

Cost: elapsed=~35min, tokens=unknown, tools=~25 calls

Verdict: LGTM

PR #491, opened 2026-08-20
test(edda-cli): make a demoted doctest fence loud instead of silent

### LINKED ISSUE #490: test(edda-cli): skill doctest coverage can be lost silently per block
From the #489 review (merged as `a395da2`). Basis SHA: `d17c5ab3dd4b5ce344fe686d8f1a8c56de56e770`.

## The gap

`crates/edda-cli/tests/skill_doctest.rs` guards coverage with a single global assertion:

```rust
assert!(checked > 0, "no skill doctest assertions ran; ...");
```

That catches total loss — the extractor breaking, the directory moving — but not partial loss. Change one block's fence from ` ```bash edda-doctest ` to ` ```bash ` and that block silently stops running while the suite stays green. The reviewer reproduced it: 10 assertions drop to 8, exit 0.

## Why it matters soon rather than now

Today `coord-handoff.md` is the only converted skill, so losing a block is visible in the count and total loss is caught. #489 states the next step is converting the other skills, which mention 12 distinct verbs between them. The moment a second file has blocks, a demoted fence in either one disappears with no signal — and the whole point of the harness is that a skill's claims cannot silently stop being checked.

## Options

- Assert a per-file minimum: a small manifest of `skill name → expected block count`, checked in the test. Explicit, but another thing to keep in sync.
- Assert that any skill mentioning `edda <verb>` in a fenced block has at least one doctest block, so demoting a fence turns into a failure rather than a silence.
- Print and assert per-file counts, and have the harness fail if a file that previously had blocks now has none — needs a checked-in baseline.

The second is probably the cheapest that scales, since it derives the expectation from the file itself rather than from a separate list.

## Note

This is a real design gap in the mechanism I wrote, found by re-derivation rather than by reading my summary — the same pattern that caught every other defect in this loop.

---

Related: #483, #488. #486 is closed.

fagemx:
Closes #490, filed from the #489 review.

## The gap

`skill_doctest.rs` guarded coverage with one global assertion:

```rust
assert!(checked > 0, "no skill doctest assertions ran; ...");
```

That catches total loss. It does not catch partial loss. Changing one block's fence from ` ```bash edda-doctest ` to ` ```bash ` is a one-token edit that reads as cosmetic, and the block silently stops running — the #489 reviewer reproduced it dropping 10 assertions to 8, exit 0.

Harmless while one skill is converted. Not harmless the moment a second is, which #489 states as the next step: the other skills mention 12 distinct verbs between them.

## The fix

Derive the expectation from the file instead of from a list to keep in sync.

Nothing in a skill starts a line with `$ ` except a doctest step. So a `$ ` line inside an **unmarked** fence means the marker was lost — not that the block is prose:

```rust
if in_fence && !is_doctest && trimmed.starts_with("$ ") {
    stranded.push(trimmed.to_string());
}
```

Two details that matter:

- The check runs **before** the empty-blocks early return. A file whose only fence was demoted parses to zero doctest blocks, so the old flow skipped it entirely — exactly the case the guard exists for.
- Ordinary ` ```bash ` examples are untouched. Verified across the shipped skills: the only `$ `-prefixed lines anywhere are the four converted blocks, and the eight plain bash blocks use bare `edda …` lines.

## Mutation-tested, not merely green

A passing test proves nothing here, so I reproduced the reviewer's scenario in three positions:

```
CAUGHT  demote the first fence
CAUGHT  demote the last fence
CAUGHT  demote every fence
```

3/3 rejected; file restored and verified back at four fences. The mutation script is not committed — it is evidence about the guard, not a fixture.

A unit test pins the behavior directly: a demoted block is reported **and** confirmed to have stopped running (so the report is not decorative), a marked block is not reported, and an ordinary example without doctest steps stays prose.

## Residual gap, stated rather than papered over

Deleting a whole block removes the `$ ` lines with it and is **not** caught. That is a visible content removal in a diff, unlike a fence tag change, so it stays reviewer territory. A checked-in per-file baseline would catch it too, at the cost of a list to maintain; I judged that not worth it for a failure mode that is already loud in review, but it is a defensible call to disagree with.

## Evidence

- RAN `cargo test -p edda` — 308 unit + 3 integration, 0 failed
- RAN the three-position mutation check above — 3/3 caught
- RAN `grep -rn '^\s*\$ ' crates/edda-cli/src/skills/*.md` — only the four converted blocks, so no existing example false-positives

Note for the reviewer: #489's exact-head CI showed a 26-minute `Test (ubuntu-latest)`, independently traced to runner degradation rather than the harness (the doctest binary itself ran 1.96s on Windows, while the untouched 304-test unit binary went 4.2s to 667s in the same run). Worth a glance at this PR's timings to confirm that reading held.

### DIFF
--- a/crates/edda-cli/tests/skill_doctest.rs
+++ b/crates/edda-cli/tests/skill_doctest.rs
@@ -23,6 +23,12 @@
 //! `$` runs a command, `>` asserts it succeeded and printed the text, and `!`
 //! asserts it failed with the text. Unmarked blocks are prose and are ignored,
 //! so existing examples do not have to be converted at once.
+//!
+//! Ignoring unmarked blocks is also how coverage could disappear: demoting a
+//! fence back to ```` ```bash ```` is a one-token edit that reads as cosmetic
+//! and silently stops that block from running. The steps it leaves behind are
+//! what give it away, so a `$ ` line inside an unmarked block is an error
+//! rather than prose.
 
 use std::path::{Path, PathBuf};
 use std::process::Command;
@@ -103,6 +109,36 @@ fn parse_blocks(markdown: &str) -> Vec<Vec<Step>> {
     blocks
 }
 
+/// Find doctest steps stranded in a block that is no longer marked as one.
+///
+/// `parse_blocks` ignores unmarked fences by design, so a demoted fence would
+/// otherwise drop its assertions with the suite still green. Nothing else in a
+/// skill starts a line with `$ `, so its presence outside a doctest fence means
+/// the marker was lost rather than that the block is prose.
+fn find_demoted_steps(markdown: &str) -> Vec<String> {
+    let mut stranded = Vec::new();
+    let mut in_fence = false;
+    let mut is_doctest = false;
+
+    for line in markdown.lines() {
+        let trimmed = line.trim();
+        if trimmed.starts_with("```") {
+            if in_fence {
+                in_fence = false;
+                is_doctest = false;
+            } else {
+                in_fence = true;
+                is_doctest = trimmed == "```bash edda-doctest";
+            }
+            continue;
+        }
+        if in_fence && !is_doctest && trimmed.starts_with("$ ") {
+            stranded.push(trimmed.to_string());
+        }
+    }
+    stranded
+}
+
 struct Outcome {
     ok: bool,
     text: String,
@@ -165,6 +201,17 @@ fn shipped_skill_examples_still_do_what_they_say() {
         }
         let name = path.file_name().unwrap().to_string_lossy().into_owned();
         let markdown = std::fs::read_to_string(&path).expect("read skill");
+
+        // Checked before the emptiness test: a file whose only doctest fence
+        // was demoted parses to zero blocks, which is precisely the silent
+        // loss this guard exists to make loud.
+        let stranded = find_demoted_steps(&markdown);
+        assert!(
+            stranded.is_empty(),
+            "{name}: doctest steps sit in a block that is not marked \
+             ```bash edda-doctest, so they never run: {stranded:?}"
+        );
+
         let blocks = parse_blocks(&markdown);
         if blocks.is_empty() {
             continue;
@@ -231,6 +278,32 @@ fn shipped_skill_examples_still_do_what_they_say() {
         dir.display()
     );
     println!("skill doctests: {checked} assertions across {files_with_blocks:?}");
+}
+
+#[test]
+fn a_demoted_fence_is_an_error_rather_than_silence() {
+    let demoted = "```bash\n$ edda claim \"auth\"\n> session: cli-auth\n```\n";
+    assert_eq!(
+        find_demoted_steps(demoted).len(),
+        1,
+        "a doctest block whose marker was removed must be reported"
+    );
+    assert!(
+        parse_blocks(demoted).is_empty(),
+        "and it must indeed have stopped running, which is why reporting matters"
+    );
+
+    let marked = "```bash edda-doctest\n$ edda claim \"auth\"\n> session: cli-auth\n```\n";
+    assert!(
+        find_demoted_steps(marked).is_empty(),
+        "a properly marked block is not stranded"
+    );
+
+    let prose = "```bash\nedda claim \"auth\" --paths \"src/a/*\"\n```\n";
+    assert!(
+        find_demoted_steps(prose).is_empty(),
+        "an ordinary example without doctest steps stays prose"
+    );
 }
 
 #[test]


fagemx:
## Code Review: Round 1
Reviewed full SHA: b994626724dd7a6516a18a570eb064b63e4c5537
Base full SHA: a395da2e60d8f3e35b6482600bf24a3559267b4a
IN SCOPE: the whole diff (`git diff a395da2..b994626` = 73 additions, 0 deletions, `crates/edda-cli/tests/skill_doctest.rs` only) — whether `find_demoted_steps` closes #490; false positives across every file in `crates/edda-cli/src/skills/`; the honesty of the stated residual gap; the guard's placement before the empty-blocks early return; fence-toggle parser correctness; exact-head CI.
BLOCKING: P0=0, P1=0

**1. Does the guard close #490? Yes — 4/4 positions, not 3/3.**
I rebuilt the check independently rather than reusing the PR's transcript: `sed -n '36,140p'` extracted `Step` / `split_args` / `parse_blocks` / `find_demoted_steps` verbatim from the reviewed file into a standalone `rustc` probe that replays the scan loop (guard → `blocks.is_empty()` early return → assertion count) over a **scratch copy** of `src/skills`. The probe reproduces the committed test's baseline string exactly — `skill doctests: 10 assertions across ["coord-handoff.md"]` — which is what makes its verdicts transferable.

Demoting each of the four fences in `coord-handoff.md` **individually** (lines 74, 84, 99, 110), then two, then all four: caught in every case. The message names the file and lists the stranded steps, e.g. for line 110:
```
coord-handoff.md: doctest steps sit in a block that is not marked ```bash edda-doctest,
so they never run: ["$ edda claim \"auth\" --paths \"src/auth/*\"", "$ edda peers --json", "$ edda peers"]
```
Two variants the PR did not claim also fail correctly: renaming the fence to ` ```sh edda-doctest ` and to ` ```bash edda-doctest-disabled `. A trailing space after the marker still passes (`trimmed` handles it) — no false failure.

**2. Placement is load-bearing, and nothing else changed.** `git diff --numstat` = `73  0  crates/edda-cli/tests/skill_doctest.rs`; `parse_blocks`, `run`, `fixture` and the assertion loop are untouched. With all four fences demoted the file parses to `blocks=0`, so a guard placed after `if blocks.is_empty() { continue; }` would skip exactly the case it exists for. Verified by construction and by the probe (`blocks=0 stranded=9`, still FAIL).

**3. No false positives today — re-derived, not reused.** `grep -rn '^\s*\$' crates/edda-cli/src/skills/` (no trailing-space requirement, so variable sigils would show) returns 9 lines, all in `coord-handoff.md` 75–115, all inside the four `bash edda-doctest` fences. The 8 plain ` ```bash ` blocks across `coord-handoff`/`coord-request`/`coord-sync` use bare `edda …` / `git …` lines. Fence counts are even and un-nested in all five files (16/2/10/4/4), so the toggle stays in phase. Near-miss worth knowing: `git log --oneline $(git merge-base HEAD main)..HEAD` survives only because `$(` is not `$ `.

**4. Parser correctness.** Synthetic cases through the extracted function: adjacent fences OK; fence indented inside a numbered list OK; closing fence with trailing whitespace OK; `$env:CARGO_INCREMENTAL` / `$lane` inside a ` ```powershell ` block correctly **not** flagged (the sigil concern is a non-issue — the space is what matters). Deleting a block's closing fence does not produce a false pass: `parse_blocks` panics with `unrecognized doctest line`, which is loud.

**5. The stated residual gap is accurate but not complete.** Deleting a whole block is indeed silent — I measured it: 10 assertions → 8, exit 0, `stranded=0`. The "loud in review" argument holds. Three further uncaught vectors, none introduced by this PR and none blocking:
- Demoting a fence *and* rewriting away every `$ ` line in it is missed (a partial rewrite is still caught — one surviving `$ ` line is enough). Same multi-line-diff class as deletion.
- `read_dir` is non-recursive and filters `extension == "md"`, so a skill under `src/skills/<subdir>/` is never scanned. For today's five files this is compile-loud, because `cmd_init.rs` pins them with literal `include_str!("skills/coord-*.md")` paths — a *new* file added under a subdirectory would not be.
- `find_demoted_steps` (strict ``` toggle) and `parse_blocks` (`current`-presence state machine) disagree on nested fences. A doctest block illustrated inside a ` ```text ` or ` ````markdown ` wrapper is currently **executed** by `parse_blocks` but invisible to the guard, so demoting it would lose a running block silently. No shipped skill nests fences today.

**6. CI timing question — the runner-degradation reading holds.** Re-derived from the API, not from the PR body: run 32298364289 (#489, `d17c5ab`) `Test (ubuntu-latest)` = 26.1 min; run 32316158203 (this SHA) = 1.3 min, with `Test (windows-latest)` 5.3 min and all 7 jobs green. Same harness, same base+1 file — the 26 minutes was the runner.

FOLLOW-UP ISSUE:
- **Guard is a future foot-gun for prose shell transcripts.** Basis SHA `b994626`. A legitimate unmarked block such as ` ```bash ` / `$ git status --short` / `$ gh pr view 491` fails the suite (`stranded=2`, confirmed) with a message asserting a marker was lost — which is false and misleading for that author. There is in-repo precedent for the style: `docs/decision/decision-intake/api.md` (lines 302–327), `docs/decision/decision-intake/canonical-form.md`, `overview.md` and `demo/STORYBOARD.md` all write `$ `-prompt transcripts; they are outside the scanned directory today, but converting any of that material into a skill would trip it. The convention this guard enforces is documented only in the test file's module doc, not anywhere a skill author reads. Cheap fixes: state the rule in the skill-authoring docs, and/or narrow the guard to `$ edda ` so non-`edda` transcripts stay prose.
- **Scan breadth and nested-fence divergence.** Basis SHA `b994626`. Items 2 and 3 of §5 above: make the skills walk recursive (or assert the scanned set matches `cmd_init.rs`'s `SKILLS` list), and align `find_demoted_steps`' toggle with `parse_blocks`' fence model so an illustrative block cannot be both executed and unguarded.
- Checked #483 and #488 (both open, unrelated); #486 closed; #490 is what this PR closes. Nothing above duplicates them. Not filed as issues — this round is read-only.

Evidence:
- RAN `cargo test -p edda --test skill_doctest -- --nocapture` at `b994626`, clean tree → `3 passed; 0 failed`, `skill doctests: 10 assertions across ["coord-handoff.md"]`, 1.90s.
- RAN a 13-case mutation battery over a scratch copy of `src/skills` using the reviewed SHA's `find_demoted_steps`/`parse_blocks` extracted verbatim: 4 single-fence demotions CAUGHT, 2-fence CAUGHT, 4-fence CAUGHT, ` ```sh edda-doctest ` CAUGHT, ` ```bash edda-doctest-disabled ` CAUGHT, trailing-space marker PASS, whole-block deletion PASS at 8 assertions (the stated gap), unclosed fence → panic, file moved to subdir / renamed `.markdown` → caught only by the pre-existing global `checked > 0`.
- RAN a 9-case synthetic parser battery (adjacent fences, indented-list fence, trailing-whitespace close, nested `text` wrapper, four-backtick wrapper, tilde fence, PowerShell sigils, prose shell transcript, fully-mangled demoted block).
- RAN `grep -rn '^\s*\$' crates/edda-cli/src/skills/`, `grep -rn '^\s*```' crates/edda-cli/src/skills/`, `git diff --numstat a395da2..b994626`.
- READ exact-head CI run 32316158203 for `b994626`: Format, Clippy (ubuntu/macos/windows), Test (ubuntu/macos/windows) all green — this is the fmt and workspace-clippy gate I did not rerun locally, plus `edda`'s tests on Windows. READ run 32298364289 for #489's timing comparison. I did **not** reuse any figure or transcript from the PR body as evidence; every number above was re-derived.
- Not corroborated: I could not exercise the guard through the *real* test binary against mutated skills, because `skills_dir()` is `env!("CARGO_MANIFEST_DIR")`-anchored and mutating the repo or building a second copy was outside the read-only/no-ad-hoc-target-dir budget. The extracted-verbatim probe reproducing the committed test's exact baseline output is the substitute; the committed unit test `a_demoted_fence_is_an_error_rather_than_silence` covers the integration point's function contract directly and passed in my RAN.
- Lane: worktree default `target/` at `C:\ai_agent\edda\.claude\worktrees\verification-cost-discipline`; no ad-hoc `CARGO_TARGET_DIR`; nothing deleted; probe built in scratch with plain `rustc`.
- Receipt: implementer's L0 receipt in the PR body (fmt, `clippy -p edda`, `test -p edda`) READ but not relied on; superseded by exact-head CI above plus my own focused RAN.
Cost: elapsed=unknown, tokens=unknown, tools=~22 calls (1 focused cargo test, 2 rustc/probe batteries, ~14 read-only greps/git/gh)
Verdict: LGTM

Issue #482, opened 2026-08-18
edda-conductor: minor follow-ups from PR #480 review (codex_app_server test helpers)


fagemx:
Non-blocking findings from the independent Round 1 review of #480 (merge commit `ee971f7f`). None affect this repo's current CI; recorded so they are not lost in a merged PR thread.

### 1. Commit message overclaims the failure mode of `process_is_alive`

Commit `813f416`'s message says the probe errors rather than reporting "gone" when it cannot run. That holds only when `ps` is **absent** (ENOENT propagates). A `ps` that exists but rejects `-o state= -p` — BusyBox, for instance — exits non-zero, which the code cannot distinguish from "the PID is genuinely gone", so `wait_for_pid_exit` would pass vacuously.

The commit message is the inaccurate part, not the behaviour: the pre-existing `kill -0` probe had the same blind spot. Cheapest correct fix is to say what it actually guarantees. Distinguishing the two cases (e.g. treating non-zero exit *with* stderr output as a probe failure) is only worth it if a musl/Alpine or container job is ever added.

Currently unreachable: `.github/workflows/ci.yml` has no container jobs and no musl/Alpine targets — only the three GitHub-hosted images.

### 2. `/bin/sh` is hardcoded while the Windows branch resolves via PATH

`fake_app_server` uses an absolute `/bin/sh`, whereas the Windows branch lets PATH resolve `powershell.exe`. Both unix CI images ship `/bin/sh`, so this is theoretical today — but the asymmetry is worth either justifying in a comment or making consistent.

### 3. Test name drifted from its assertion

`dropping_concrete_client_makes_child_pid_disappear` now passes when the PID is a zombie, since a zombie has already terminated. The assertion is right; only the name still says "disappear". Consider renaming to something like `..._stops_child_process`.

### Context

The actual flake fixed in #480 was ETXTBSY: the unix branch wrote a script, chmod'd it `0755`, and `execve`'d it, which Linux rejects while any process holds the file open for writing. See the decision `test.fixture-scripts=invoke-via-interpreter-never-exec`. The review grep'd the workspace and found no other write-then-exec helper, so no further instances are outstanding.

PR #493, opened 2026-08-20
test(edda-cli): aim the doctest guard at edda steps and scan fences once

### LINKED ISSUE #492: test(edda-cli): narrow the doctest demotion guard and unify fence tracking
From the #491 review (merged as `005f8b6`). Basis SHA: `b994626724dd7a6516a18a570eb064b63e4c5537`.

The reviewer described these and explicitly did not file them, being a read-only round. All three concern `crates/edda-cli/tests/skill_doctest.rs`; none was introduced by #491.

## 1. The demotion guard will misfire on a non-edda shell transcript

`find_demoted_steps` treats **any** `$ `-prefixed line inside an unmarked fence as a lost doctest marker. A prose block written in transcript style —

````markdown
```bash
$ git status
$ gh pr view 491
```
````

— trips it, and the failure message says a doctest marker was lost, which would be false. That style already exists in this repository (`docs/decision/decision-intake/api.md:302-327`, `canonical-form.md`, `overview.md`, `demo/STORYBOARD.md`), just not yet under `crates/edda-cli/src/skills/`.

**Fix:** narrow the prefix to `$ edda `. Two characters, and it keeps the guard aimed at the only thing it was ever meant to detect. Worth doing before the next skill is converted, since that is when transcript-style prose becomes likely.

## 2. The scan is one directory deep and `.md`-only

`read_dir` over `src/skills` with an extension filter. A skill added under a subdirectory, or with a different extension, is never scanned and its doctests silently do not exist.

Low risk today: the five skills are pinned by `include_str!` in `cmd_init.rs`, so moving one is a compile error rather than a silent loss. It stops being pinned the moment a skill is loaded some other way.

## 3. The two fence trackers disagree on nested fences

`find_demoted_steps` toggles strictly on ` ``` `; `parse_blocks` keys off whether a block is currently open. They diverge on nesting: a doctest illustrated **inside** a ` ```text ` or ` ````markdown ` wrapper is executed by `parse_blocks` but invisible to the guard.

So an illustrative example — a doctest shown as documentation of the doctest format, which `skill_doctest.rs`'s own module header does — would run for real if it appeared in a skill. Today no skill contains one.

**Fix:** share one fence-tracking pass between the two functions rather than keeping two implementations that can drift.

## Also confirmed by the reviewer, already stated in #491

Whole-block deletion remains silent (10 assertions → 8, exit 0). #491 names this and argues it is loud in review; the reviewer verified both halves. Listed here only so the full boundary is in one place: **demote + rewrite away every `$ ` line** is a further uncaught vector, though it is deliberate defeat rather than accident.

---

Related: #483, #488. #486 and #490 are closed.

### PR DESCRIPTION
**Pull request by fagemx**
Closes #492 items 1 and 3, both raised by the #491 review.

## Item 1 — the guard would cry wolf on a shell transcript

`find_demoted_steps` fired on **any** `$ `-prefixed line inside an unmarked fence. So this prose block:

````markdown
```bash
$ git status
$ gh pr view 491
```
````

failed the suite with a message claiming a doctest marker had gone missing — asserting something that never happened. **Correction after review:** I cited `docs/decision/decision-intake/api.md:302-327` and three sibling files as in-repo precedent for that style. They are not. I have now read all four, and every one is a `$ edda …` transcript, which the narrowing does **not** exempt — they would still be flagged if that content moved under `src/skills`. The citation came from the #491 review and I propagated it twice without opening the files.

What actually motivates the narrowing is the general class: a block transcribing any non-`edda` command. What the narrowing does **not** do is exempt a prose `$ edda` transcript inside a skill. That is a deliberate convention rather than an oversight — all eight existing prose blocks in the shipped skills write bare `edda …` with no `$ ` prefix, so inside `src/skills` the prefix means "doctest step" and bare means "prose". Worth knowing before converting a skill that wants to show a transcript.

The predicate now requires the step to invoke `edda`, which is what `parse_blocks` already demanded of a real step. A mixed block reports only its `edda` line.

## Item 3 — two fence trackers that disagreed

`find_demoted_steps` toggled strictly on ` ``` `; `parse_blocks` keyed off whether a block was open. They diverged on nesting: a doctest illustrated inside a wrapper was executed by one and invisible to the other.

Both now read a single `fenced_blocks` pass. The divergence is gone **by construction** rather than by keeping two implementations in step — which is the actual lesson, since the drift is what produced the finding.

Closing a fence now needs one at least as long as the opener, per CommonMark, so a ```` ```` ```` wrapper can quote a ``` block. That is what lets a skill *show* the doctest format without the harness executing the illustration, and a block quoting the marker as content is exempt from the guard for the same reason: its steps are meant to be read, not run.

## Writing the illustration test first paid

It failed against my first implementation, which correctly declined to *execute* a quoted example and then reported it as demoted anyway — the same false-alarm class as item 1, one layer down. I would not have found that by reading my own change.

## Precision did not cost sensitivity

Both mutation suites re-run against the tightened guard:

```
CAUGHT  demote the first fence
CAUGHT  demote the last fence
CAUGHT  demote every fence

CAUGHT  #1  a verb declared not to exist (main, before #485)
CAUGHT  #2  bare unclaim documented as working (#485 first fix)
CAUGHT  #3  `edda peers` said to show the id (#485 second fix)
CAUGHT  #4  unclaim of a claimless session called a success (base behavior)
```

3/3 and 4/4, same as before narrowing. Files restored and verified afterwards. Neither script is committed — they are evidence about the guard, not fixtures.

## New tests

- a transcript of other commands is prose; a mixed block reports only its `edda` step
- an illustrated doctest inside a longer fence neither runs nor is reported
- an unclosed fence still reports its steps, so a stray backtick cannot hide a block

## Left open deliberately

#492 item 2 — the scan is one directory deep and `.md`-only. The five skills are pinned by `include_str!` in `cmd_init.rs`, so moving one is a compile error rather than a silent loss. Worth doing when a skill is first loaded some other way.

## Evidence

- RAN `cargo clippy -p edda --all-targets -- -D warnings` — PASS (two errors on the first draft, an unused binding and a `contains` lint, both fixed before this SHA)
- RAN `cargo test -p edda` — 308 unit + 6 integration, 0 failed
- RAN both mutation suites above — 3/3 and 4/4 caught

### DIFF
--- a/crates/edda-cli/tests/skill_doctest.rs
+++ b/crates/edda-cli/tests/skill_doctest.rs
@@ -27,8 +27,9 @@
 //! Ignoring unmarked blocks is also how coverage could disappear: demoting a
 //! fence back to ```` ```bash ```` is a one-token edit that reads as cosmetic
 //! and silently stops that block from running. The steps it leaves behind are
-//! what give it away, so a `$ ` line inside an unmarked block is an error
-//! rather than prose.
+//! what give it away, so a `$ edda …` line inside an unmarked block is an
+//! error rather than prose. A shell transcript that runs something else
+//! (`$ git status`) is prose and passes.
 
 use std::path::{Path, PathBuf};
 use std::process::Command;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by fagemx** [comment]
## Code Review: Round 1

Reviewed full SHA: `0846eb4bd8a996ec63692bb1a3713b1667f38bff`
Base full SHA: `005f8b60b6b890afecfc64fc88c0df5821670aaa`

**IN SCOPE:** the single changed file `crates/edda-cli/tests/skill_doctest.rs` — (1) runner/guard agreement on what `edda_step_args` accepts; (2) demotion sensitivity re-derived independently against the real `coord-handoff.md`; (3) `fenced_blocks` fence semantics (longer closers, info-string closers, indentation, tildes, nesting, unclosed); (4) breadth of the marker-in-content exemption; (5) that the runner, fixture and assertion loop are otherwise unchanged and the four converted blocks still execute. Adjacent scanner-precision gaps and #492 item 2 are follow-up, not this PR.

**BLOCKING: P0=0, P1=0**

### What I verified

**1. Narrowing is symmetric.** `edda_step_args` is the single decision point for both `parse_blocks` and `find_demoted_steps`, so the asymmetry #492 item 3 was about is gone by construction. I probed 16 `$ `-line shapes and compared, for each, what the runner does inside a marked fence against what the guard does inside an unmarked one. **Zero skew** — every line the runner turns into a `Step::Run` is exactly a line the guard reports, and every line the runner rejects the guard leaves silent:

| line | runner (marked fence) | guard (unmarked fence) |
|---|---|---|
| `$ edda peers` | RUNS `["peers"]` | reports |
| `$  edda peers` (two spaces) | RUNS `["peers"]` | reports |
| `$ edda` (no args) | RUNS `[]` | reports |
| `$ edda ` (trailing ws) | RUNS `[]` | reports |
| `$ "edda" peers` | RUNS `["peers"]` | reports |
| `$ edda\tpeers` | RUNS `["peers"]` | reports |
| `$ EDDA peers` | panic `a doctest step must invoke edda` | silent |
| `$ ./edda peers` | panic (same) | silent |
| `$ edda-something` | panic (same) | silent |
| `$ sudo edda peers`, `$ EDDA=1 edda peers`, `$ git status` | panic (same) | silent |
| `$ `, `$`, `$edda peers` | panic `unrecognized doctest line` | silent |

Non-`edda` `$ ` lines are loud inside a marked fence and prose outside one — which is the intended split, not a hole. Step classification inside a marked block is otherwise identical to base (the base `assert_eq!` and the new `panic!` carry the same message and fire on the same inputs).

**2. Sensitivity is intact.** Re-derived from scratch, not reused from the PR body. Mutating the real `crates/edda-cli/src/skills/coord-handoff.md` (markers at lines 74, 84, 99, 110; baseline 4 blocks / 19 steps / guard clean):

```
CAUGHT  demote block #1     blocks 4 -> 3   guard reports 2 steps
CAUGHT  demote block #2     blocks 4 -> 3   guard reports 2 steps
CAUGHT  demote block #3     blocks 4 -> 3   guard reports 2 steps
CAUGHT  demote block #4     blocks 4 -> 3   guard reports 3 steps
CAUGHT  demote first two    blocks 4 -> 2   guard reports 4 steps
CAUGHT  demote ALL          blocks 4 -> 0   guard reports 9 steps
CAUGHT  marker typo'd       blocks 4 -> 0   guard reports 9 steps
7 mutants, 0 missed
```

**3. Both #492 items are genuinely fixed.** Same shapes run against base and head functions side by side:

| shape | base `005f8b6` | head `0846eb4` |
|---|---|---|
| `$ git status` transcript (item 1) | guard **falsely reports 2 steps** | silent |
| ` ```text ` wrapper quoting a doctest (item 3) | runner **executes 1 step** | 0 blocks, guard silent |
| 4-tick wrapper quoting a 3-tick doctest | runner executes 1 step, guard falsely reports | 0 blocks, guard silent |
| plain demotion (control) | reported | reported |

Fence semantics check out against CommonMark on the shapes that matter: a closer carrying an info string does not close (both steps still reported); a 4-tick line closes a 3-tick block; a 3-tick line does not close a 4-tick wrapper; list-item-indented fences work; an unclosed unmarked fence still reports its steps; CRLF is handled.

**4. The runner is untouched.** `struct Outcome` through the end of `shipped_skill_examples_still_do_what_they_say` is **byte-identical** to base (140 lines, exact string compare). Base and head scans of all five shipped skills are identical: `coord-handoff.md` 4 blocks / 19 steps, the other four 0, guard clean everywhere. The four converted blocks execute and all 10 assertions pass against a binary built at this SHA.

**FOLLOW-UP ISSUE** (all basis SHA `0846eb4bd8a996ec63692bb1a3713b1667f38bff`; none blocks):

- **(a) The illustration exemption is fence-scoped, so an unbalanced fence can widen it.** `find_demoted_steps` skips an entire unmarked `Fence` when any of its content lines equals `DOCTEST_FENCE`. Well-formed illustrations are fine — everything from the quoted marker to the wrapper's close is genuinely quoted. But if an illustration's closing fence is missing, the illustration and a later genuinely demoted block merge into one fence and the demotion is silently excused. Concretely, ` ```text ` + quoted marker + `$ edda claim` + prose + ` ```bash ` + `$ edda unclaim` + ` ``` ` : base reported `$ edda unclaim` (and its runner panicked); head reports nothing. A narrower rule was available — exempt only the span from a quoted marker line to its matching quoted closer, rather than the whole fence. I judge this non-blocking: it needs malformed markdown that would also render wrong, and it is strictly narrower than the whole-block-deletion gap already accepted in #491/#492.
- **(b) `fenced_blocks` recognises only backtick fences.** A `~~~markdown` wrapper still lets a quoted `bash edda-doctest` illustration execute (1 block / 1 step at both base and head) — the same class the PR closes for backtick wrappers. A `~~~`-fenced block is invisible to runner and guard alike. Separately, a 4-space-indented ` ```bash edda-doctest ` fence, which CommonMark renders as literal text, is executed. Both are pre-existing (unchanged from base) and no shipped skill uses tildes today (verified by grep).
- **(c) Evidence accuracy, no code impact.** The PR body (inheriting #492) cites `docs/decision/decision-intake/api.md:302-327`, `canonical-form.md`, `overview.md` and `demo/STORYBOARD.md` as transcript style motivating the narrowing. I read all four: **every one is a `$ edda …` transcript**, which the narrowing does not exempt — moved under `src/skills/` they would still be reported. The code is correct and implements exactly what #492 prescribed; the cited evidence just does not support the stated motivation. Worth correcting so nobody reads the false-alarm class as fully closed.

Not filed: #492 item 2 (explicitly deferred by this PR), and silent whole-block deletion (documented boundary in #491/#492 — I confirmed deleting all four blocks still fails via the `checked > 0` assertion, while partial deletion stays silent). Checked #483, #488, #482, #389, #172 — no overlap.

**Evidence:**
- **RAN** `cargo test -p edda --test skill_doctest -- --nocapture` (worktree, Windows 11) → `6 passed; 0 failed`, printing `skill doctests: 10 assertions across ["coord-handoff.md"]`.
- **RAN** standalone probes: scan functions extracted **verbatim** from head (`skill_doctest.rs` lines 39–45 + 51–186) and from base (`git show 005f8b6:crates/edda-cli/tests/skill_doctest.rs`), compiled with `rustc --edition 2021`, driven by my own harness that catches panics. Covered 16 symmetry shapes (0 skew), 7 mutants of the real skill (7/7 caught), 13 fence shapes, 6 adversarial/exemption shapes base-vs-head, and a base-vs-head scan of all 5 shipped skills (identical).
- **RAN** exact line-range string compare of `struct Outcome` → end of `shipped_skill_examples_still_do_what_they_say`, base vs head → identical.
- **READ** exact-head CI at `0846eb4`: **all 7 checks green** — Format; Clippy ubuntu/macOS/Windows (workspace, all-targets); Test ubuntu/macOS (workspace) and Test Windows (subset incl. `edda`). Run 32317992243.
- **READ** the PR body's L0 receipt and #492. Per the standing rule I reused **no** figure or transcript from the PR body as my own evidence; every number above is from my own runs.
- **Could not corroborate:** the PR body's second mutation suite (the 4 false-CLI-claim mutants) — I did not reconstruct those content mutations; they exercise the #489/#491 runner surface rather than this diff. I did confirm all four blocks still execute and all 10 assertions pass. Also did not re-derive "308 unit + 6 integration": I ran only `--test skill_doctest`, and `cargo test --workspace` is green in exact-head CI on all three OSes.
- **Lane:** worktree default `target/`. No ad-hoc `CARGO_TARGET_DIR`. Nothing deleted; no repository file edited (mutations were in-memory in the probe).
- **Receipt:** exact-head CI run 32317992243 (all green) + my local focused `cargo test -p edda --test skill_doctest`.

Cost: elapsed=unknown, tokens=unknown, tools=~20 calls (1 cargo test, 6 rustc probe builds/runs, rest read-only inspection).

**Verdict: LGTM** — P0=0, P1=0. Gates relied on: exact-head CI at `0846eb4` (Format; `cargo clippy --workspace --all-targets` on ubuntu/macOS/Windows; `cargo test --workspace` on ubuntu/macOS; Windows test subset incl. `edda`), plus local `cargo test -p edda --test skill_doctest` on Windows 11. No local full-workspace run: the change is confined to one integration test file and exact-head CI covers the workspace on all three OSes.

PR #495, opened 2026-08-20
test(edda-cli): harden the doctest scanner against real markdown

### LINKED ISSUE #494: test(edda-cli): doctest guard exemption scope, tilde fences, and the $ edda transcript convention
From the #493 review (merged as `21d3be2`). Basis SHA: `0846eb4bd8a996ec63692bb1a3713b1667f38bff`.

The reviewer described these and did not file them, being a read-only round.

## 1. The illustration exemption is fence-scoped, so malformed markdown can launder a demotion

`find_demoted_steps` skips any unmarked fence whose content contains a line equal to the doctest marker. With an unbalanced fence, an illustration and a later genuinely demoted block merge into one "fence" — and the demotion is silently excused. The reviewer confirmed the shape: base reported it, head does not.

Needs malformed markdown to trigger, and it is strictly narrower than the whole-block-deletion gap already accepted in #491. Still a real narrowing of the guard introduced by #493.

**A narrower rule was available:** exempt only the span from a quoted marker to its matching quoted closer, rather than the whole enclosing fence.

## 2. Only backtick fences are recognised

`fenced_blocks` counts backticks. CommonMark also allows tilde fences, so:

- a `~~~markdown` wrapper still lets a quoted illustration execute;
- a 4-space-indented doctest fence executes even though CommonMark treats indented content as literal text.

Both pre-existing and unchanged by #493; no shipped skill uses tildes or indented fences today.

## 3. `$ edda` prose transcripts inside a skill are still flagged

Not a defect, but worth recording where someone will look for it.

#493 narrowed the guard so a `$ git status` transcript is prose. A `$ edda …` transcript inside an unmarked fence is still reported as a lost marker. That is the current convention — all eight existing prose blocks in the shipped skills write bare `edda …` with no `$ ` prefix, so under `src/skills` the prefix means "doctest step" and bare means "prose".

It matters because the four files cited in #492 item 1 and in #493's original body as precedent for transcript style (`docs/decision/decision-intake/api.md`, `canonical-form.md`, `overview.md`, `demo/STORYBOARD.md`) are **all `$ edda` transcripts**. If any of that content is ever moved into a skill it will trip the guard, correctly by the convention but surprisingly to whoever moves it.

Either document the convention in the skill-authoring notes, or teach the guard that a fence with no doctest assertions (`>` / `!` lines) is a transcript rather than a demoted block.

---

Note on provenance: the "transcript style already exists in this repository" citation originated in the #491 review and was propagated through #492 into #493's body without anyone opening the files. The #493 reviewer read all four and found it false. Reviewer claims need the same verification as implementer claims when they are reused as evidence.

Related: #483, #488. #486, #490 closed; #492 items 1 and 3 closed by #493, item 2 still open there.

fagemx:
Closes #494 (all three items) and #492 item 2.

## Four gaps, all found by review rather than by the suite going red

| Gap | Was | Now |
|---|---|---|
| Illustration exemption was fence-scoped | one unbalanced fence merged a quoted example with a later demoted block and excused the demotion | skips only the span from a quoted marker to its matching closer |
| Only backtick fences recognised | a `~~~markdown` wrapper leaked its quoted contents and the illustration ran for real | fences match on character **and** run length; a 4-space-indented line opens nothing, per CommonMark |
| `$ edda` transcripts flagged as demoted | the style four docs under `docs/decision/` already use was reported as a lost marker | a block counts as demoted only when it still carries `>` or `!` assertions |
| Scan one directory deep | a skill in a subdirectory was never checked | walks recursively |

## The third one retires a claim I made in #492

I wrote there that narrowing the predicate to `edda` would fix the false-alarm class. It does not. The #493 reviewer showed the four cited files are all `$ edda …` transcripts, which the narrowing still flags — the citation behind that item was wrong and I propagated it twice.

The real discriminator is **assertions**, not the command name. A demoted doctest keeps its `>`/`!` lines; a transcript never had any. That rule covers both the `$ git status` case and the `$ edda` case the narrowing missed.

## Sensitivity is unchanged

Both mutation suites re-run against the hardened scanner:

```
CAUGHT  demote the first fence / the last fence / every fence          3/3
CAUGHT  the four historical false CLI claims (#485 era)                4/4
```

## Tests

`a_shell_transcript_is_prose_not_a_lost_doctest` asserted the old rule, so it is **inverted and renamed** rather than deleted — the poison-test rule this repo's own review skill states. Four added: an unbalanced illustration cannot launder a later demotion; a tilde wrapper quotes rather than leaks; an indented fence is literal; an unclosed fence with assertions still reports.

Writing the laundering test first paid twice: my first fixture used an equal-length outer fence, so the blocks never merged and the test was vacuous until I corrected it. I would not have caught that by reading the implementation.

## Evidence

- RAN `cargo test -p edda` — 308 unit + 9 integration, 0 failed
- RAN both mutation suites — 3/3 and 4/4 caught, files restored and verified

### DIFF
--- a/crates/edda-cli/tests/skill_doctest.rs
+++ b/crates/edda-cli/tests/skill_doctest.rs
@@ -26,10 +26,15 @@
 //!
 //! Ignoring unmarked blocks is also how coverage could disappear: demoting a
 //! fence back to ```` ```bash ```` is a one-token edit that reads as cosmetic
-//! and silently stops that block from running. The steps it leaves behind are
-//! what give it away, so a `$ edda …` line inside an unmarked block is an
-//! error rather than prose. A shell transcript that runs something else
-//! (`$ git status`) is prose and passes.
+//! and silently stops that block from running. What gives it away is the
+//! assertions left behind — a block carrying `>` or `!` lines was checking
+//! something, so `$ edda …` steps in an unmarked fence are an error rather than
+//! prose. A transcript with nothing asserted is prose and passes, whichever
+//! command it runs.
+//!
+//! The scan errs toward reading a fence-looking line as a fence, and toward
+//! reporting. A false alarm is cheap; a skill that quietly stopped being
+//! checked is the failure this file exists to prevent.
 
 use std::path::{Path, PathBuf};
 use std::process::Command;
@@ -124,14 +129,21 @@ fn fence_marker(trimmed: &str) -> Option<(char, usize)> {
 /// apart on what counts as a block — keeping two fence trackers in step was
 /// itself a defect (GH-492).
 ///
-/// A line indented four spaces or more outside a fence is literal text in
-/// CommonMark, so a fence opener there does not open anything (GH-494).
+/// Indentation is deliberately ignored. CommonMark treats a four-space line as
+/// literal *relative to its container*, and it strips a list marker's width
+/// first, so a fence indented under a `- ` bullet is a real fence — GitHub
+/// renders it as code. Judging indentation against the document instead turned
+/// such a block invisible: it stopped running, and a demoted one stopped being
+/// reported, which is the silence this harness exists to prevent (GH-495).
+///
+/// Reading every fence-looking line as a fence errs toward running and
+/// reporting. That is the safe direction here: the cost is a false alarm on an
+/// indented literal example, against a silently unchecked skill.
 fn fenced_blocks(markdown: &str) -> Vec<Fence<'_>> {
     let mut blocks = Vec::new();
     let mut open: Option<(char, usize, Fence)> = None;
 
     for line in markdown.lines() {
-        let indent = line.len() - line.trim_start().len();
         let trimmed = line.trim();
         let marker = fence_marker(trimmed);
 
@@ -147,9 +159,6 @@ fn fenced_blocks(markdown: &str) -> Vec<Fence<'_>> {
                 }
             }
             None => {
-                if indent >= 4 {
-                    continue;
-                }
                 if let Some((ch, run)) = marker {
                     open = Some((
                         ch,
@@ -526,13 +535,49 @@ fn a_tilde_wrapper_quotes_rather_than_leaks() {
 }
 
 #[test]
-fn an_indented_fence_is_literal_text() {
-    // Four-space indentation makes the line literal in CommonMark, so it opens
-    // nothing (GH-494).
-    let indented = "    ```bash edda-doctest\n    $ edda claim \"auth\"\n    ```\n";
-    assert!(
-        parse_blocks(indented).is_empty(),
-        "an indented fence does not open a doctest"
+fn an_indented_fence_is_still_a_fence() {
+    // Was `an_indented_fence_is_literal_text`, which judged the four-space rule
+    // against the document. CommonMark applies it relative to the container and
+    // strips a list marker first, so a fence under a `- ` bullet is real and
+    // GitHub renders it as code. Treating it as literal made such a block stop
+    // running, and a demoted one stop being reported (GH-495).
+    let indented = "    ```bash edda-doctest\n    $ edda claim \"auth\"\n    > session: \
+                    cli-auth\n    ```\n";
+    assert_eq!(
+        parse_blocks(indented).len(),
+        1,
+        "an indented doctest still runs"
+    );
+
+    let demoted = "- step one:\n\n    ```bash\n    $ edda claim \"auth\"\n    > session: \
+                   cli-auth\n    ```\n";
+    assert_eq!(
+        find_demoted_steps(demoted).len(),
+        1,
+        "and an indented demotion is still reported rather than silently dropped"
+    );
+}
+
+#[test]
+fn skills_in_subdirectories_are_scanned() {
+    // The scan was one directory deep, so a skill in a subdirectory was never
+    // checked (GH-492 item 2). Nothing in the shipped tree exercises this yet,
+    // so the walk needs its own fixture.
+    let dir = tempfile::tempdir().expect("tempdir");
+    let nested = dir.path().join("group").join("deeper");
+    std::fs::create_dir_all(&nested).expect("nested dirs");
+    std::fs::write(dir.path().join("top.md"), "# top\n").expect("top");
+    std::fs::write(nested.join("buried.md"), "# buried\n").expect("buried");
+    std::fs::write(nested.join("ignored.txt"), "not markdown\n").expect("txt");
+
+    let found: Vec<String> = markdown_files(dir.path())
+        .iter()
+        .map(|p| p.file_name().unwrap().to_string_lossy().into_owned())
+        .collect();
+    assert_eq!(
+        found,
+        vec!["buried.md", "top.md"],
+        "sorted, recursive, .md only"
     );
 }
 
--- a/crates/edda-cli/tests/skill_doctest.rs
+++ b/crates/edda-cli/tests/skill_doctest.rs
@@ -48,6 +48,26 @@ fn skills_dir() -> PathBuf {
     Path::new(env!("CARGO_MANIFEST_DIR")).join("src/skills")
 }
 
+/// Every markdown file under `dir`, at any depth.
+///
+/// The scan used to be one directory deep, so a skill added in a subdirectory
+/// would simply never be checked (GH-492).
+fn markdown_files(dir: &Path) -> Vec<PathBuf> {
+    let mut found = Vec::new();
+    let entries =
+        std::fs::read_dir(dir).unwrap_or_else(|e| panic!("cannot read {}: {e}", dir.display()));
+    for entry in entries {
+        let path = entry.expect("dir entry").path();
+        if path.is_dir() {
+            found.extend(markdown_files(&path));
+        } else if path.extension().and_then(|e| e.to_str()) == Some("md") {
+            found.push(path);
+        }
+    }
+    found.sort();
+    found
+}
+
 /// Split a command line on spaces, keeping double-quoted runs together.
 ///
 /// Enough for the shapes skills actually use (`--paths "src/a/*"`); anything
@@ -82,46 +102,70 @@ struct Fence<'a> {
     lines: Vec<&'a str>,
 }
 
+/// The fence character and run length that opened a block, if this line opens
+/// or closes one.
+///
+/// CommonMark allows both backticks and tildes, and closes a block only with
+/// the same character at least as long as the opener. Recognising tildes keeps
+/// a `~~~markdown` wrapper from leaking its quoted contents into the scan
+/// (GH-494).
+fn fence_marker(trimmed: &str) -> Option<(char, usize)> {
+    let ch = trimmed.chars().next()?;
+    if ch != '`' && ch != '~' {
+        return None;
+    }
+    let run = trimmed.chars().take_while(|c| *c == ch).count();
+    (run >= 3).then_some((ch, run))
+}
+
 /// Walk the markdown once and return every fenced block.
 ///
 /// Both the runner and the demotion guard read this, so they cannot drift
 /// apart on what counts as a block — keeping two fence trackers in step was
 /// itself a defect (GH-492).
 ///
-/// Closing needs a fence at least as long as the opening one, per CommonMark,
-/// so a ```` wrapper can quote a ``` block. That is what lets a skill *show*
-/// the doctest format without the harness executing the illustration.
+/// A line indented four spaces or more outside a fence is literal text in
+/// CommonMark, so a fence opener there does not open anything (GH-494).
 fn fenced_blocks(markdown: &str) -> Vec<Fence<'_>> {
     let mut blocks = Vec::new();
-    let mut open: Option<(usize, Fence)> = None;
+    let mut open: Option<(char, usize, Fence)> = None;
 
     for line in markdown.lines() {
+        let indent = line.len() - line.trim_start().len();
         let trimmed = line.trim();
-        let ticks = trimmed.chars().take_while(|c| *c == '`').count();
+        let marker = fence_marker(trimmed);
 
         match open.as_mut() {
-            Some((width, fence)) => {
-                if ticks >= *width && trimmed[ticks..].trim().is_empty() {
-                    blocks.push(open.take().expect("open fence").1);
+            Some((ch, width, fence)) => {
+                let closes = marker
+                    .filter(|(c, run)| c == ch && run >= width)
+                    .is_some_and(|(_, run)| trimmed[run..].trim().is_empty());
+                if closes {
+                    blocks.push(open.take().expect("open fence").2);
                 } else {
                     fence.lines.push(trimmed);
                 }
             }
-            None if ticks >= 3 => {
-                open = Some((
-                    ticks,
-                    Fence {
-                        is_doctest: trimmed == DOCTEST_FENCE,
-                        lines: Vec::new(),
-                    },
-                ));
+            None => {
+                if indent >= 4 {
+                    continue;
+                }
+                if let Some((ch, run)) = marker {
+                    open = Some((
+                        ch,
+                        run,
+                        Fence {
+                            is_doctest: trimmed == DOCTEST_FENCE,
+                            lines: Vec::new(),
+                        },
+                    ));
+                }
             }
-            None => {}
         }
     }
     // An unclosed fence still carries its steps; report them rather than drop
     // them, or a stray backtick would hide a whole block.
-    if let Some((_, fence)) = open {
+    if let Some((_, _, fence)) = open {
         blocks.push(fence);
     }
     blocks
@@ -167,22 +211,62 @@ fn parse_blocks(markdown: &str) -> Vec<Vec<Step>> {
     blocks
 }
 
+/// Steps of an unmarked block, with any quoted illustration removed.
+///
+/// The exemption used to cover a whole fence, so one unbalanced fence could
+/// merge an illustration with a later genuinely demoted block and excuse it.
+/// Skipping only the span from a quoted marker to its matching closer keeps
+/// the rest of the block under the guard (GH-494).
+fn lines_outside_quoted_illustrations<'a>(lines: &[&'a str]) -> Vec<&'a str> {
+    let mut out = Vec::new();
+    let mut quoting = false;
+    for line in lines {
+        if *line == DOCTEST_FENCE {
+            quoting = true;
+            continue;
+        }
+        if quoting {
+            if fence_marker(line).is_some() {
+                quoting = false;
+            }
+            continue;
+        }
+        out.push(*line);
+    }
+    out
+}
+
 /// Find doctest steps stranded in a block that is no longer marked as one.
 ///
 /// Unmarked fences are ignored by design, so a demoted fence would otherwise
 /// drop its assertions with the suite still green.
+///
+/// A block is only treated as demoted when it still carries assertions. A
+/// transcript of `$ edda …` commands with no `>` or `!` lines is prose showing
+/// usage — the style four docs under `docs/decision/` already use — and calling
+/// it a lost marker would be a false alarm (GH-494).
 fn find_demoted_steps(markdown: &str) -> Vec<String> {
-    fenced_blocks(markdown)
+    let mut stranded = Vec::new();
+
+    for fence in fenced_blocks(markdown)
         .into_iter()
-        // A block that quotes the marker as content is illustrating the
-        // format, not a doctest that lost its own. Its steps are meant to be
-        // read, so reporting them would be the false alarm this guard was
-        // narrowed to avoid.
-        .filter(|f| !f.is_doctest && !f.lines.contains(&DOCTEST_FENCE))
-        .flat_map(|f| f.lines)
-        .filter(|line| edda_step_args(line).is_some())
-        .map(str::to_string)
-        .collect()
+        .filter(|f| !f.is_doctest)
+    {
+        let lines = lines_outside_quoted_illustrations(&fence.lines);
+        let has_assertions = lines
+            .iter()
+            .any(|l| l.starts_with("> ") || l.starts_with("! "));
+        if !has_assertions {
+            continue;
+        }
+        stranded.extend(
+            lines
+                .iter()
+                .filter(|l| edda_step_args(l).is_some())
+                .map(|l| (*l).to_string()),
+        );
+    }
+    stranded
 }
 
 struct Outcome {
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: Round 1
Reviewed full SHA: 946b29893f024e410db4ea32409326971e6af08b
Base full SHA: 21d3be2c059ffb27e3d54ffb137ddb37e33918c1

IN SCOPE: `crates/edda-cli/tests/skill_doctest.rs` (the only file changed) — the assertion-based demotion discriminator; the span-scoped illustration exemption; `fence_marker`/`fenced_blocks` CommonMark handling (tilde fences, indentation, closer rules); `markdown_files` recursion; equivalence of the runner and the shipped-skill coverage against base; and whether each of the nine tests fails when its own feature is reverted.

BLOCKING: P0=0, P1=1

- **P1 — `fenced_blocks`, the `if indent >= 4 { continue; }` branch: the CommonMark indent rule is applied against the document, not the containing block, so a doctest fence nested in a list item becomes invisible to *both* the runner and the guard.**

  CommonMark strips the list marker's width before parsing the item's content, so a fence indented four spaces under a `- ` bullet is a real fenced code block. I checked that against GitHub's own renderer rather than asserting it — `gh api -X POST markdown` on

  ````
  - step one:

      ```bash edda-doctest
      $ edda claim "auth"
      > session: cli-auth
      ```
  ````

  returns `<li><p>step one:</p><div class="highlight highlight-source-shell"><pre>$ edda claim "auth" …` — a live fence, not literal text. (The same input at top level, with no bullet, does render literally, so the rule the PR is reaching for is right; only its scope is wrong.)

  Running that exact markdown through the head parser: `parse_blocks` → **0 blocks**, `find_demoted_steps` → **`[]`**. Through the base parser: `parse_blocks` → **1 block** (it ran). So a doctest written under a numbered or bulleted step renders correctly on GitHub, is silently never executed, and the demotion guard reports nothing.

  The same silencing hits the top-level shape this change *is* aimed at. On

  ````
  Some prose.

      ```bash
      $ edda claim "auth"
      > session: cli-auth
      ```
  ````

  head returns `[]` and base returns `["$ edda claim \"auth\""]`. Whatever the right call is on executing an indented block, turning a loud report into silence is the one direction this harness exists to prevent — it is the failure mode named in the module header and the reason #489/#491/#493 exist. This PR closes three such holes and opens a fourth.

  Nothing is lost today: I read all five shipped skills and none contains an indented fence, a tilde fence, or a tab-indented line, and head and base extract byte-identical coverage from them (below). The hazard is latent, and it is untestable from the current suite — `an_indented_fence_is_literal_text` pins only the top-level shape, and no test pins the list-nested one, so the suite stays green either way.

  Direction, not prescription: keep the guard loud for indented blocks that still carry `$ edda` + `>`/`!` (report rather than skip), or track list context so a marker-relative indent still opens a fence.

FOLLOW-UP ISSUE:
Basis SHA for all of these: `946b29893f024e410db4ea32409326971e6af08b`. Checked against #483 and #488 — no overlap, and no existing issue covers any of them.

1. **The quoted-span exemption still launders a demotion when the quoted marker has no closer.** `lines_outside_quoted_illustrations` arms `quoting` on the marker and clears it only on the next line where `fence_marker(..).is_some()`; with no such line the rest of the block is skipped. head → `[]` on

   ````
   ````markdown
   ```bash edda-doctest
   $ edda peers
   > ok
   $ edda claim "auth"
   > session: cli-auth
   ````
   ````

   Two markers in a row behave the same way, because the `*line == DOCTEST_FENCE` arm is tested *before* the `quoting` arm, so a second marker re-arms the span instead of closing it. Both are strictly narrower than base (base launders these too — not a regression), and #494 item 1's actual scenario **is** closed: on the committed fixture head reports `claim` and not `peers` while base reports nothing, so that test is non-vacuous, and I confirmed independently that the equal-length outer fence the PR body says it first used really is vacuous (head and base both `[]`). A marker closed by a *tilde* fence does end the span — the loose direction, which is safe.

2. **The recursion is untested.** `markdown_files` is reachable only from `shipped_skill_examples_still_do_what_they_say`, and `src/skills` is flat, so reverting it to the old one-deep `read_dir` leaves all nine tests green. The behavior itself checks out: sorted deterministically at the top level, a directory literally named `zz.md` is recursed into rather than read as a file, and a directory-junction cycle terminates (64 duplicate results, exit 0) instead of blowing the stack. This is why #492 item 2 should not be considered pinned yet.

3. **The narrowing introduces a false alarm base did not have.** A wrapper holding both a quoted illustration and a genuine prose transcript now reports the transcript as a lost doctest (head → `["$ edda claim \"auth\""]`, base → `[]`). That is the inherent cost of the span-scoped rule #494 asked for, it is loud rather than silent, and no shipped skill has the shape — but the failure message would claim a marker went missing when none did, which is the class #492 item 1 objected to.

4. **"A transcript never had any assertions" is not universal.** A shell transcript legitimately carries `> ` as the PS2 continuation prompt, and can carry a `! ` line; both are reported. Not a regression — base flagged them too via the `$ edda` prefix — but the premise in the PR body is stronger than the code delivers. Conversely `>ok` without the space is not counted as an assertion, so such a block is exempt at head and reported at base; that one is unreachable by demotion, since `parse_blocks` panics on that line, so it could never have been a live doctest.

5. **The module header is now stale.** Lines 1–36 are byte-identical to base and still state the rule as "a `$ edda …` line inside an unmarked block is an error rather than prose". Under head that holds only when the block also carries `>`/`!` assertions.

6. **Pre-existing Windows CI flake, not this PR.** `Test (windows-latest)` is red at this SHA on `edda-ledger sqlite_store::tests::concurrent_writers_no_data_loss` — "database is locked" at `crates\edda-ledger\src\sqlite_store\mod.rs:1480`. The identical failure already occurred on **main** at `005f8b60` (run 32316775048), which predates this branch, and this PR touches no `edda-ledger` code. Re-run the failed job rather than treating it as a product red; there is no issue tracking it.

Evidence:
- RAN: extracted the head and base versions of `skill_doctest.rs` with `git show` and compiled the *verbatim* `fence_marker` / `fenced_blocks` / `edda_step_args` / `parse_blocks` / `lines_outside_quoted_illustrations` / `find_demoted_steps` from both SHAs side by side (`rustc --edition 2021 -O`), then ran 30 fixtures through both: the 9 committed ones, 7 attacking the assertion discriminator, 5 attacking the quoted-span exemption, and 11 CommonMark fence shapes (3-space and tab indents, trailing-space closer, longer closer, tilde-not-closed-by-backtick, backticks in the info string, indented closer, `~~~bash edda-doctest`). Results quoted above.
- RAN: `gh api -X POST markdown` on the list-nested and top-level indented fences — GitHub's own CommonMark output is the basis for the P1.
- RAN: head vs base parse of the five shipped skills — both give `coord-handoff.md`: **4 blocks, 19 steps, 10 assertions**; the other four: 0 blocks; `find_demoted_steps` empty everywhere. Coverage is byte-for-byte unchanged, which is the part CI cannot show (`checked` is only printed under `--nocapture`, and `checked > 0` would pass if two of the four blocks vanished).
- RAN: replayed those 4 blocks / 10 assertions against a real `edda` binary from this repo (`target/debug/edda.exe`, `edda 0.3.0`), reproducing `fixture()`/`run()` including `git init` and `EDDA_STORE_ROOT` — all 10 passed on Windows. Provenance caveat: that binary was built in this worktree today, not by me at this SHA; it is representative because the PR changes no product code (`git diff --stat` base..head = one file).
- RAN: `markdown_files` against a synthetic tree (nested dirs, a directory named `zz.md`, a non-`.md` file, and a directory junction cycle) — deterministic order, no stack overflow.
- READ: exact-head CI run 32349237484 — Format, Clippy ×3, Test (ubuntu), Test (macos) pass; Test (windows) fails on the `edda-ledger` flake above. I pulled the Windows job log directly and confirmed **all nine `skill_doctest` tests, including `shipped_skill_examples_still_do_what_they_say`, ran and passed on Windows** before the ledger binary failed.
- READ: PR body, #494, #492, #483, #488.
- NOT corroborated, and not relied on: the PR body's mutation-suite scores (3/3, 4/4) — those suites are not committed and I could not reproduce them; and the "308 unit + 9 integration" count. I re-derived every figure I use above independently.
- NOT run, with reason: no local `cargo test --workspace` and no local `cargo build`. The assigned worktree had already been checked out to `fix/session-identity` at `21d3be2` by another session before I started, so building there would have disturbed live work, and standing up a fresh checkout would mean a multi-GB cold target directory to re-derive what exact-head CI already recorded green on three OSes. The Windows-runtime gap the ladder cares about is covered instead by the local binary replay above.
- Lane: n/a — no workspace build. `rustc` probes and their artifacts live in a session scratchpad outside the repo; nothing in any worktree was written, and nothing was deleted.
- Receipt: PR body's L0 receipt (`cargo fmt --all --check`, `cargo clippy -p edda --all-targets -- -D warnings`, `cargo test -p edda`) READ, superseded by exact-head CI, which runs all of it independently.

Cost: elapsed=~35 min, tokens=unknown, tools=~30 calls

Verdict: Changes Requested

fagemx:
## Review Response: Round 1

New full SHA: `894fed89ebee793bb933ff9d22d3bdb1c7ff615b`
Reviewed SHA this responds to: `946b29893f024e410db4ea32409326971e6af08b`

### P1-1 — the indent rule judged the wrong container · FIXED by removing it

Confirmed and not disputed. I applied the four-space rule against the document; CommonMark applies it relative to the container and strips a list marker's width first, so a fence indented under a `- ` bullet is a real fence. Your check against GitHub's own renderer settles it — I had reasoned from the spec's headline sentence without testing how it composes with list items.

The consequence is the one that matters: an indented **demoted** block went from a loud report to `[]`. I added a rule to avoid a false alarm and bought silence with it, which inverts what this harness is for.

Indentation is now ignored entirely rather than corrected. Tracking container indent means real CommonMark parsing, and the cost of getting it half-right is exactly what you found. Reading every fence-looking line as a fence errs toward running and reporting; the price is a false alarm on an indented literal example, against a skill that quietly stopped being checked. The module header now states that bias explicitly instead of leaving it implicit.

`an_indented_fence_is_literal_text` asserted the removed rule, so it is inverted and renamed, and it now pins both halves — an indented doctest runs, and an indented demotion under a `- ` bullet is reported.

### Recursion had no test · FIXED

You are right that reverting the walk left all nine green — the recursion was unverified. `skills_in_subdirectories_are_scanned` now builds its own tree (nested `.md`, a `.txt` alongside it) and asserts the sorted, recursive, markdown-only result. That was a gap in my own PR body's claim to close #492 item 2: shipping the change without a test proving it is not closing it.

### Mutation suites re-run

3/3 demotions and 4/4 historical false claims still caught after the change.

### The remaining follow-ups

Accepted as filed rather than pulled in: the unterminated/duplicated quoted marker (narrower than base, so not a regression), the false alarm the narrowing adds that base lacked, the PS2 `> ` continuation counterexample to "a transcript never has assertions", and the stale module header — that last one is fixed here since it described a rule this PR had already replaced.

The PS2 case is worth flagging back to you as genuinely unresolved rather than merely deferred: `> ` as a shell continuation prompt inside a transcript would make it look like a doctest with assertions. No shipped skill contains one, but the premise is not universal and I would rather say so than let the docstring imply otherwise.

### Windows CI

Agreed and verified independently: `edda-ledger sqlite_store::tests::concurrent_writers_no_data_loss` also failed on **main** at `005f8b60`, an untouched crate. Environmental, not this PR.

### On your worktree note

That was my fault. I checked the shared worktree out to `fix/session-identity` while you were working in it, which is why you could not run a local gate. I have moved this fix to a dedicated worktree so it cannot happen again in this round. Your fallback — exact-head CI plus replaying the ten assertions against a real binary — covered it, but you should not have had to find one.

### Gates on the new SHA

- RAN: `cargo test -p edda` — 308 unit + 10 integration, 0 failed
- RAN: both mutation suites — 3/3 and 4/4 caught, files restored
- READ: exact-head CI pending on `894fed8`
- Lane: dedicated worktree, default `target/`; no ad-hoc `CARGO_TARGET_DIR`

Round 2 requested on `894fed89ebee793bb933ff9d22d3bdb1c7ff615b`.

fagemx:
## Code Review: Round 2
Reviewed full SHA: 894fed89ebee793bb933ff9d22d3bdb1c7ff615b
Base full SHA: 21d3be2c059ffb27e3d54ffb137ddb37e33918c1
Prior round: Round 1 at 946b29893f024e410db4ea32409326971e6af08b (Changes Requested, P0=0, P1=1)

Round 1 finding: **P1-1 closed**

The `if indent >= 4 { continue; }` branch is gone from `fenced_blocks`, and both halves of the fix hold under mutation. I restored the branch verbatim onto head's `fenced_blocks` (**M1**) and ran the committed test body against it:

- assertion 1, "an indented doctest still runs": `parse_blocks` **1 → 0** — FAIL
- assertion 2, isolated so the first failure cannot mask it: `find_demoted_steps` on `- step one:` + indented demoted block → **`[]`**, len 0 ≠ 1 — FAIL

So the inverted test genuinely pins the removal in both directions, not just the runner half.

I re-derived the CommonMark premise myself rather than inheriting it. `gh api -X POST markdown -f mode=gfm`:

- list-nested → `<li><p>step one:</p><div class="highlight highlight-source-shell"><pre>$ edda claim "auth"` — a **real fence**
- the same block at top level, no bullet → `<pre><code>```bash …` — **literal**, backticks visible as text

Both halves of Round 1's claim reconfirmed.

**Is removal the right shape?** Yes, and on stronger grounds than the response argues. Base `21d3be2` has no indent rule at all (`git show 21d3be2:… | grep indent` → no match). I built the head and base `fenced_blocks`/`find_demoted_steps` side by side under `rustc` and ran 11 indentation fixtures through both — top-level indented marked and demoted, bullet-nested, numbered-list-nested, tab-indented, stray-marker, indented literal, and controls. **Head and base agree on all 11.** The removal restores shipped behavior; it does not invent a new bias. The bias was already the one in production.

**Attacking the new bias.** The false-alarm case is real and I built it:

```
Example:

    ```bash
    $ edda claim "auth"
    > session: cli-auth
    ```
```

GitHub renders that literally; head reports `["$ edda claim \"auth\""]`. The message would read *"doctest steps sit in a block that is not marked ```bash edda-doctest, so they never run"* — half true (they never run) and half misleading (it implies a marker went missing when the block is literal by design). An indented literal illustration of a **marked** doctest is worse: head returns 1 block and actually **executes** it. Both are loud, both are recoverable, and both are **byte-identical at base**, so neither is a regression this PR introduces. The trade the response claims is the trade being made.

One correction to how that trade is stated, below in follow-ups: it is not the *only* cost.

Response verification:
- **"308 unit + 10 integration, 0 failed"** — corroborated exactly. `cargo test -p edda` here: `308 passed; 0 failed` and `10 passed; 0 failed`.
- **`cargo fmt --all --check`** — corroborated, ran it locally (exit 0) and CI Format is green at head.
- **`cargo clippy -p edda --all-targets -- -D warnings`** — corroborated by CI Clippy green on ubuntu, macos **and** windows at this exact SHA. I did not run clippy locally.
- **"3/3 demotions and 4/4 historical false claims still caught"** — **not corroborated.** Those suites are not committed and I could not reproduce them. I ran my own four mutations instead (results below); three were killed, one survived.
- **"exact-head CI pending"** — now complete. All 7 checks green at `894fed8`, **including `Test (windows-latest)`**. The `edda-ledger concurrent_writers_no_data_loss` flake Round 1 saw did not recur; I am classifying head as green on my own observation of run 32351144468, not inheriting Round 1's call.
- **The recursion gap the response claims to close** — confirmed closed, see below.

IN SCOPE: the delta `946b2989..894fed8` — removal of the indent branch in `fenced_blocks`; the inversion and rename of `an_indented_fence_is_literal_text` → `an_indented_fence_is_still_a_fence`; the new `skills_in_subdirectories_are_scanned`; the module header and `fenced_blocks` docstring rewrite; that the delta touches nothing else; and that the four converted blocks in `coord-handoff.md` still execute and pass. Full-PR surface `21d3be2..894fed8` re-checked for head-vs-base silencing.

BLOCKING: P0=0, P1=0

Nothing in this delta caused a new blocker. Every head-vs-base difference I could produce is either a **loud** false alarm or a narrowing that loses no coverage:

- a demoted block with `$ edda` steps and **no** assertions: head `[]`, base reports it — benign, an assertion-free block asserts nothing, so demoting it loses nothing observable
- `>ok` without the space: head `[]`, base reports it — unreachable by demotion, since `parse_blocks` panics on that line and it could never have been a live doctest

FOLLOW-UP ISSUE:
Basis SHA for both: `894fed89ebee793bb933ff9d22d3bdb1c7ff615b`. Checked against #483, #488, #494 (open), #492 (closed), and against the six follow-ups Round 1 filed in-comment — neither of these duplicates one.

1. **`skills_in_subdirectories_are_scanned` claims to pin sorted order; it does not.** The assertion message reads `"sorted, recursive, .md only"`. I mutated each of the three properties independently against head's `markdown_files` with the committed test body:
   - **M2** walk made non-recursive → **FAIL** (recursion is load-bearing — Round 1 item 2 is genuinely closed)
   - **M4** `.md` extension filter dropped → **FAIL** (load-bearing)
   - **M3** `found.sort()` deleted → **PASSES**

   `sorted` is incidental, not pinned. NTFS returns directory entries name-ordered from its B-tree index, so this fixture cannot distinguish sorted from unsorted on Windows at all. The consequence is small — ordering only affects report and `files_with_blocks` output order, never which files are scanned — but the message overstates what the test proves, which is the same class of overclaim this loop has been correcting. Either drop `sorted` from the message, or add a fixture that can actually distinguish it (and note it would be a no-op assertion on NTFS).

2. **The new bias sentence names one cost when there are two.** The header says the scan *"errs toward reading a fence-looking line as a fence, and toward **reporting**"*, and the `fenced_blocks` docstring says *"the cost is a false alarm on an indented literal example"*. There is a second cost, in the opposite direction. A stray unmatched fence marker swallows every line until the next closer, so a following **real** doctest neither runs nor is reported:

   ```
   An empty fence looks like:

       ```

   And here is the real doctest:

   ```bash edda-doctest
   $ edda claim "auth"
   > session: cli-auth
   ```
   ```

   head → `parse_blocks` **0 blocks**, `find_demoted_steps` **`[]`**. Base → identical, and the unindented variant behaves the same at both, so this is **pre-existing and not indentation-specific** — not caused by this delta and not blocking. The mechanism is the one Round 1 filed as its follow-up 1 (`lines_outside_quoted_illustrations` arms `quoting` on the marker and never clears it without a later `fence_marker` line). Worth folding into #494 alongside that item, and worth qualifying the docstring so it does not promise a bias the code does not universally honor.

Two observations that are **not** findings, recorded because they were asked:

- **The bullet in the second fixture of `an_indented_fence_is_still_a_fence` is decorative.** Deleting `- step one:` leaves the result identical (`["$ edda claim \"auth\""]`). That is inherent to the fix rather than a flaw: because indentation is ignored outright, no fixture can distinguish list-nested from top-level-indented. The test pins *"indentation is ignored"*, which is what was implemented; it does not independently pin the list case. A future container-aware parser would fail fixture 1 even while being more correct — worth knowing before anyone attempts that.
- **The module header is accurate now.** I checked each claim against the code: the assertion-based rule matches `find_demoted_steps` (`has_assertions` gate, then `edda_step_args` filter); *"a transcript with nothing asserted is prose and passes, whichever command it runs"* matches the `if !has_assertions { continue; }` early exit and is confirmed by fixture; and the GH-495 account of what the indent rule broke matches the M1 mutation exactly. Round 1's follow-up 5 (stale header) is closed. Only the bias sentence in item 2 above is incomplete. Round 1's follow-up 6 (Windows flake) is resolved by observation — it did not recur.

Evidence:
- **RAN** `cargo clean -p edda` then `cargo test -p edda --test skill_doctest -- --nocapture` in the assigned worktree's default `target/` — recompiled `edda v0.3.0` from scratch at this SHA; **10 passed, 0 failed**, printing `skill doctests: 10 assertions across ["coord-handoff.md"]`. The four converted blocks execute and pass on Windows against a binary built here at this SHA.
- **RAN** `cargo test -p edda` — `308 passed; 0 failed` + `10 passed; 0 failed`.
- **RAN** `cargo fmt --all --check` — exit 0.
- **RAN** four mutations of head's own functions, compiled verbatim under `rustc --edition 2021` against the committed test bodies extracted byte-for-byte from the source: M1 indent branch restored → `an_indented_fence_is_still_a_fence` FAIL (both halves, checked separately); M2 non-recursive → `skills_in_subdirectories_are_scanned` FAIL; M3 no `found.sort()` → **PASS (survived)**; M4 no `.md` filter → FAIL. Unmutated head passes both.
- **RAN** head vs base `21d3be2` differential: 11 indentation fixtures (identical on all 11) plus 5 attacking the assertion rule, the PS2 `> ` continuation, the indented literal, and the stray-marker shape. Only differences are the two benign narrowings listed under BLOCKING.
- **RAN** head parse of all five shipped skills: `coord-handoff.md` 4 blocks / 19 steps / 10 assertions, the other four 0 blocks, `find_demoted_steps` empty everywhere — coverage unchanged, and matching Round 1's base figures, so nothing was silently lost.
- **RAN** `gh api -X POST markdown -f mode=gfm` on the list-nested and top-level-indented fixtures — GitHub's own renderer, quoted above.
- **RAN** `git diff --name-only` for both `946b2989..894fed8` and `21d3be2..894fed8` — **`crates/edda-cli/tests/skill_doctest.rs` only**, in both. Nothing else moved.
- **READ** exact-head CI run 32351144468 — Format, Clippy x3, Test x3 all green at `894fed8`.
- **READ** Round 1's comment and the Round 1 response; #483, #488, #494, #492 to avoid duplicate follow-ups.
- **Lane**: the assigned `doctest-fix` worktree's default `target/`; no ad-hoc `CARGO_TARGET_DIR`. `cargo clean -p edda` reclaimed 1.4 GiB of that crate's own build cache before rebuilding; nothing else was deleted. The `rustc` probes and their artifacts live in a session scratchpad outside the repo — no file in any worktree was written. Read-only throughout: nothing edited, committed, pushed, or merged.
- **Receipt**: the response's L0 receipt READ; its fmt/test figures independently reproduced above, its clippy figure covered by exact-head CI, its mutation-suite figures not reproducible and replaced with my own mutations.

Note on the worktree: `crates/edda-cli/src/skills/coord-handoff.md` shows as modified in `git status` there, but `git diff --numstat` reports **zero changed lines** — it is a CRLF/LF stat artifact, not content. The tested source matches the reviewed SHA.

Cost: elapsed=~40 min, tokens=unknown, tools=~40 calls

Gates relied on: local `cargo fmt --all --check` (exit 0), local `cargo test -p edda` (308 + 10, 0 failed) and a from-clean `cargo test -p edda --test skill_doctest` at this SHA on Windows, plus exact-head CI run 32351144468 green on all 7 checks (Format; Clippy on ubuntu/macos/windows; Test on ubuntu/macos/windows). Windows `Test` covers the `edda` crate, so the Windows-runtime gap the ladder cares about is covered for the one crate this PR touches, and I re-ran it locally besides.

PR #499, opened 2026-08-20
docs(edda-conductor): say what the liveness probe guarantees, and rename a test

### LINKED ISSUE #482: edda-conductor: minor follow-ups from PR #480 review (codex_app_server test helpers)

fagemx:
Closes #482, filed from the independent review of #480.

## 1. The doc overclaimed what the probe guarantees

The overclaim was that the probe errors rather than reporting "gone" when it cannot run. **Correction to this PR's original description and to #482's:** that sentence was never in a comment — it is from commit `813f416`'s *message*, and a grep of the base for it returns nothing. #482 inherited the misattribution from the review that filed it, and this body repeated it for two rounds. The doc did overclaim, but not in those words.

The claim itself holds only when `ps` is **absent**, where ENOENT propagates. A `ps` that exists but rejects `-o state= -p` — BusyBox — exits non-zero, and the code cannot tell that apart from *the PID is genuinely gone*. On such a system `wait_for_pid_exit` passes vacuously: the caller believes the process stopped when nothing was ever observed.

**The behaviour is not the wrong part.** Distinguishing the two cases only earns its cost the day a musl/Alpine or container job exists — `ci.yml` runs the three GitHub-hosted images, no `container:`, and nothing else. So the doc states the limit and the condition under which fixing it is worth doing, which is what the issue recommended.

The comparison to the probe this replaced is **narrower than Round 1 first claimed**. The old `kill -0` probe shared the class of flaw and was worse in one way — it swallowed spawn failure too. But on BusyBox, the trigger the doc names, the two diverge: BusyBox `kill` parses numeric signals, so `kill -0` worked there, while its `ps` rejects these arguments on two independent grounds (the state column is `stat`, and an unknown `-o` column is fatal; `-p` is not in its base option set). The switch to `ps` therefore made this case *reachable* on exactly the platform class the deferral cites, rather than inheriting it. The doc says so, because that strengthens the case for recording the limit.

## 2. The shell-path asymmetry — resolved by consistency, not justification

`fake_app_server` used an absolute `/bin/sh` while the Windows branch let PATH resolve `powershell.exe`. The issue asked for **either justification or consistency**.

The justification originally written here was false: `/bin/sh` is *not* guaranteed by POSIX. POSIX does not standardise utility locations, and the `sh` page's Application Usage says so directly, pointing applications at `getconf PATH`; Solaris ≤10 put the conformant shell at `/usr/xpg4/bin/sh`. And the comment named PATH resolution as the hazard while leaving the Windows branch doing precisely that, unchanged and unmentioned.

So this takes consistency:

- **Windows** resolves `powershell.exe` under `%SystemRoot%`, falling back to PATH only if the variable is missing.
- **POSIX** keeps the absolute `/bin/sh`, on the true and narrower premise: every platform this harness runs on has it, so pinning it costs portability we do not need and removes a lookup we do not control. The comment states that POSIX does not fix the path, so the next reader cannot re-derive the old error.

This is the one place the PR changes code rather than comments.

## 3. The test name outlived its assertion

`dropping_concrete_client_makes_child_pid_disappear` → `..._stops_the_child_process`. The assertion was already correct — a zombie has terminated, so it passes — and only the name still described a PID disappearing.

The spec's mention of that flake carries the rename inline, so a reader searching the old name finds where it went. The **plan** keeps the old name untouched: it is a frozen execution record of what was true at `738c5ce`, and carries an explicit banner saying so.

Two sibling names with the same defect (`cancelling_concrete_method_makes_child_pid_disappear`, and the `wait_for_pid_exit` helper) are **#506**, along with the Windows arm of `process_is_alive` not checking `output.status`.

## Evidence

- RAN `cargo clippy -p edda-conductor --all-targets -- -D warnings` — PASS
- RAN `cargo test -p edda-conductor` — **166 passed, 0 failed, 1 ignored**, including the renamed test and going through the new `%SystemRoot%` path
- RAN a check that `%SystemRoot%` is set and the absolute powershell path exists on this workstation
- RAN `grep` for the old test name across `crates/` and `docs/` — handled as described above
- Receipt: L0 focused gates above; `edda-conductor` is in the CI Windows test subset, so exact-head CI covers all three platforms

## Review history

- `Code Review: Round 1` on `9762e11` — P1=2, Changes Requested
- `Review Response: Round 1` — both fixed; four findings routed to #506
- Round 2 requested on `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`

### DIFF
--- a/crates/edda-conductor/src/agent/codex_app_server.rs
+++ b/crates/edda-conductor/src/agent/codex_app_server.rs
@@ -439,7 +439,16 @@ mod tests {
         {
             let script = dir.path().join("fake-app-server.ps1");
             std::fs::write(&script, powershell_fake_script(scenario))?;
-            let mut command = Command::new("powershell.exe");
+            // Absolute where the platform offers a reliable one, so a doctored
+            // PATH cannot choose the shell this harness runs. `%SystemRoot%` is
+            // set on every Windows install; falling back to PATH keeps the
+            // helper working if it is somehow absent (GH-482).
+            let mut command = match std::env::var("SystemRoot") {
+                Ok(root) => Command::new(format!(
+                    "{root}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
+                )),
+                Err(_) => Command::new("powershell.exe"),
+            };
             command.args([
                 "-NoLogo",
                 "-NoProfile",
@@ -463,11 +472,13 @@ mod tests {
             // the script read-only, so that window cannot bite.
             let script = dir.path().join("fake-app-server.sh");
             std::fs::write(&script, shell_fake_script(scenario))?;
-            // Absolute, where the Windows branch above lets PATH resolve
-            // powershell.exe. The asymmetry is deliberate: `/bin/sh` is
-            // guaranteed by POSIX at that path, so resolving it through PATH
-            // would only add a way for a doctored PATH to choose the shell
-            // this harness runs (GH-482).
+            // Absolute for the same reason as the Windows branch above. Note
+            // POSIX does *not* fix this path -- the `sh` page's Application
+            // Usage says so outright and points applications at `getconf PATH`,
+            // and Solaris put the conformant shell under /usr/xpg4/bin. What
+            // makes it safe here is narrower: every platform this harness runs
+            // on has /bin/sh, so pinning it costs portability we do not need
+            // and removes a lookup we do not control (GH-482).
             let mut command = Command::new("/bin/sh");
             command.arg(script);
             Ok((dir, command))
@@ -566,12 +577,23 @@ mod tests {
     /// arguments -- BusyBox does not accept `-o state= -p`, and the two cases
     /// are indistinguishable from the exit status alone. On such a system a
     /// caller waiting for exit would pass without the process having gone
-    /// anywhere. The pre-existing `kill -0` probe had the same blind spot; it
-    /// is recorded rather than fixed because `ci.yml` runs only the three
+    /// anywhere.
+    ///
+    /// The `kill -0` probe this replaced shared the *class* of flaw and was in
+    /// one way worse -- it swallowed spawn failure too. But on BusyBox, the
+    /// trigger named above, the two do not match: BusyBox `kill` parses numeric
+    /// signals, so `kill -0` worked there, while its `ps` rejects these
+    /// arguments twice over (the state column is `stat`, and an unknown `-o`
+    /// column is fatal; `-p` is not in its base option set). So the switch to
+    /// `ps` made this case reachable on exactly the platform class the
+    /// deferral cites, rather than inheriting it.
+    ///
+    /// It is recorded rather than fixed because `ci.yml` runs only the three
     /// GitHub-hosted images, none of them musl or Alpine and none in a
-    /// container. Distinguishing them (treating a non-zero exit that also
-    /// wrote to stderr as a probe failure) is worth doing the day such a job
-    /// is added (GH-482).
+    /// container -- so no job can reach it today. Distinguishing the two
+    /// (treating a non-zero exit that also wrote to stderr as a probe failure)
+    /// is worth doing the day such a job is added, and this note is what should
+    /// make that obvious then (GH-482).
     fn process_is_alive(pid: u32) -> anyhow::Result<bool> {
         #[cfg(windows)]
         {
--- a/crates/edda-conductor/src/agent/codex_app_server.rs
+++ b/crates/edda-conductor/src/agent/codex_app_server.rs
@@ -463,6 +463,11 @@ mod tests {
             // the script read-only, so that window cannot bite.
             let script = dir.path().join("fake-app-server.sh");
             std::fs::write(&script, shell_fake_script(scenario))?;
+            // Absolute, where the Windows branch above lets PATH resolve
+            // powershell.exe. The asymmetry is deliberate: `/bin/sh` is
+            // guaranteed by POSIX at that path, so resolving it through PATH
+            // would only add a way for a doctored PATH to choose the shell
+            // this harness runs (GH-482).
             let mut command = Command::new("/bin/sh");
             command.arg(script);
             Ok((dir, command))
@@ -555,6 +560,18 @@ mod tests {
     /// absent" is not something drop can guarantee -- "the process no longer
     /// runs" is. A zombie has already terminated, so it does not count as
     /// alive; a child that drop failed to kill still does.
+    ///
+    /// The unix probe reads a non-zero `ps` exit as "gone". That is right when
+    /// `ps` ran and found nothing, and wrong when `ps` itself rejected the
+    /// arguments -- BusyBox does not accept `-o state= -p`, and the two cases
+    /// are indistinguishable from the exit status alone. On such a system a
+    /// caller waiting for exit would pass without the process having gone
+    /// anywhere. The pre-existing `kill -0` probe had the same blind spot; it
+    /// is recorded rather than fixed because `ci.yml` runs only the three
+    /// GitHub-hosted images, none of them musl or Alpine and none in a
+    /// container. Distinguishing them (treating a non-zero exit that also
+    /// wrote to stderr as a probe failure) is worth doing the day such a job
+    /// is added (GH-482).
     fn process_is_alive(pid: u32) -> anyhow::Result<bool> {
         #[cfg(windows)]
         {
@@ -764,7 +781,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn dropping_concrete_client_makes_child_pid_disappear() -> anyhow::Result<()> {
+    async fn dropping_concrete_client_stops_the_child_process() -> anyhow::Result<()> {
         let (_fake, server) = spawn_fake_app_server(FakeScenario::Idle).await?;
         let pid = server.child.id().context("fake app-server has no PID")?;
 
--- a/docs/superpowers/specs/2026-08-18-verification-cost-discipline-design.md
+++ b/docs/superpowers/specs/2026-08-18-verification-cost-discipline-design.md
@@ -310,8 +310,11 @@ the tool enforces a limit against an artificially inflated number.
    `fleet-pr-loop` "re-run gates" wording; `edda gate` native receipts;
    Bash twin of `lane.ps1`; repo-level gate profile override; operator
    decision on the `EXECUTION-COORDINATION.md` axiom; the pre-existing
-   ubuntu-only flake `codex_app_server::tests::dropping_concrete_client_makes_child_pid_disappear`
-   observed red on `7f5d155` (classified baseline; unrelated to PR #479).
+   ubuntu-only flake in `codex_app_server` observed red on `7f5d155`
+   (classified baseline; unrelated to PR #479). The test was
+   `dropping_concrete_client_makes_child_pid_disappear`, renamed to
+   `..._stops_the_child_process` in GH-482 because a zombie has already
+   terminated, so the assertion was never about the PID disappearing.
 
 ## 6. Acceptance criteria
 


fagemx:
## Code Review: Round 1

Full SHA: `9762e11657f7bd04a75c2d64a5aa3e59d879baac`
IN SCOPE: issue #482 items 1–3 as delivered at this SHA — the `process_is_alive` rustdoc's account of the unix probe, the deferral argument resting on `ci.yml` and on the pre-existing `kill -0` probe, the `/bin/sh` vs `powershell.exe` justification, rename completeness across the tree, and test-poison check on the renamed test. The PR body is reviewed as part of the surface.
FOLLOW-UP ISSUE: four, listed below — none extend this PR.
Blocking: P0=0, P1=2

---

### The "no behaviour changed" claim — verified

Mechanically, not by eye. Stripping comment lines from the diff of the Rust file leaves exactly one changed line:

```
-    async fn dropping_concrete_client_makes_child_pid_disappear() -> anyhow::Result<()> {
+    async fn dropping_concrete_client_stops_the_child_process() -> anyhow::Result<()> {
```

Command: `git diff origin/main...9762e11 -- crates/edda-conductor/src/agent/codex_app_server.rs | grep -E "^[+-]" | grep -vE "^(\+\+\+|---)" | grep -vE "^[+-]\s*(//|///)"`. The claim holds exactly. Everything below is about prose.

---

## Findings

### P1-1 (blocking) — "`/bin/sh` is guaranteed by POSIX at that path" is false, and POSIX says so explicitly

`crates/edda-conductor/src/agent/codex_app_server.rs:466-470` at this SHA:

> The asymmetry is deliberate: `/bin/sh` is guaranteed by POSIX at that path, so resolving it through PATH would only add a way for a doctored PATH to choose the shell this harness runs (GH-482).

POSIX does not standardise the filesystem location of any utility except the `/dev` and `/tmp` reservations, and the `sh` page's Application Usage addresses this case by name: the standard path to the shell "cannot be assumed to be either /bin/sh or /usr/bin/sh" (IEEE Std 1003.1, `sh`, Application Usage), directing applications to `getconf PATH` instead. Demonstrable counterexamples exist in both directions: Solaris 10 and earlier shipped the legacy Bourne shell at `/bin/sh` with the conformant one at `/usr/xpg4/bin/sh`, and Android has no `/bin/sh` at all (`/system/bin/sh`).

This blocks because it is the same defect the PR exists to remove — a comment asserting a guarantee the underlying standard does not give — landed in the same edit that removes the previous one. It is also the cheapest possible fix: the *conclusion* is fine, only the premise is wrong. Something like "every platform this harness runs on ships `/bin/sh`, and an absolute path keeps a doctored PATH from choosing the shell a test harness executes" carries the same decision without appealing to a guarantee that does not exist.

**Second half — the objection the comment does not survive.** The comment names PATH resolution as the hazard ("would only add a way for a doctored PATH to choose the shell this harness runs"), then calls an asymmetry deliberate in which the Windows branch does precisely that: `Command::new("powershell.exe")` at line 442, resolved through PATH. By the comment's own reasoning the Windows branch is the exposed one, and it is left unchanged and unmentioned. Issue #482 item 2 asked for "either justifying in a comment or making consistent"; a justification whose stated rationale indicts the other branch justifies half the asymmetry. One clause acknowledging the Windows side (there is no absolute path to depend on there, so the lookup is unavoidable — which is true, `powershell.exe` lives under `%SystemRoot%\System32\WindowsPowerShell\v1.0\`, not a POSIX-style fixed location) would close it.

---

### P1-2 (blocking) — "The pre-existing `kill -0` probe had the same blind spot" is true of the defect class and false of the instance the doc names, in the direction that flatters the deferral

Lines 564-574 at this SHA argue: the ambiguity pre-exists, so record it rather than fix it. I re-derived the old probe from `813f416` (`test(edda-conductor): make dropped-child pid check zombie-aware`):

```rust
std::process::Command::new("kill")
    .args(["-0", &pid.to_string()])
    .stdout(Stdio::null()).stderr(Stdio::null())
    .status()
    .is_ok_and(|status| status.success())
```

At the level of *class*, the claim is correct, and the old probe was in fact worse: `is_ok_and` collapsed spawn failure into "gone" as well, so an absent `kill` reported "not running" silently, where the current `.output().context("failed to run ps")?` propagates ENOENT. Fair.

But the doc names a specific trigger — BusyBox — and on BusyBox the two probes do not behave alike:

- BusyBox's `kill` applet parses numeric signals, so `kill -0 PID` works on Alpine; it is idiomatic there. The old probe was **not** blind on BusyBox.
- BusyBox's `ps` fails on these arguments on two independent grounds: its `-o` column table names the state column `stat`, not `state`, and an unknown column is a fatal `bad -o argument`; and `-p` is not in its option set in the base build.

So the switch to `ps` in `813f416` made a previously-unreachable case reachable on exactly the platform class the deferral names as its own trigger condition ("the day a musl/Alpine or container job is added"). A future reader of this comment is told nothing regressed; on Alpine something did. That inverts the priority the comment is trying to set.

The remedy is one clause, not a code change — e.g. "the `kill -0` probe collapsed a non-zero exit the same way (and swallowed spawn failure too), though BusyBox itself only began tripping it with the switch to `ps`". I am blocking on it because the pre-existence claim is the load-bearing premise of the deferral and this PR's entire deliverable is an accurate account of what the probe guarantees.

**Caveat, stated plainly:** I have no BusyBox available in this environment and did not execute it. The structural half (both probes collapse non-zero exit to "gone") is derived from source and is certain. The BusyBox-specific half is derived from BusyBox's `ps` applet column table and option set, not observed. If you can show BusyBox `kill` rejects `-0`, this finding drops.

---

### Scope 1, remainder — the doc's account of the unix probe is otherwise accurate

Re-derived against `process_is_alive` at lines 575-602:

| Doc claim | Code | Verdict |
|---|---|---|
| "reads a non-zero `ps` exit as 'gone'" | `if !output.status.success() { return Ok(false); }` | true |
| "the two cases are indistinguishable from the exit status alone" | only `status.success()` is consulted; stderr is captured by `.output()` and discarded | true |
| "a caller waiting for exit would pass without the process having gone anywhere" | `wait_for_pid_exit` returns `Ok(())` on the first `!process_is_alive(pid)?` | true |
| ENOENT propagates as an error (PR body) | `.output().context("failed to run ps")?` — spawn error, not exit status | true of the code |
| Windows takes a different path | doc scopes itself to "The unix probe"; the Windows branch is `tasklist.exe` | no conflict |

Two notes, neither blocking. The ENOENT-vs-non-zero-exit contrast is asserted in the **PR body** but never appears in the shipped comment, so a reader of the comment alone learns the ambiguity without learning that an absent `ps` fails loudly — worth adding while you are in there. And the doc correctly limits itself to unix, which is what keeps it true; see FU-3 for the Windows branch.

### Scope 4 — rename complete, both hits handled as described

`git grep -n dropping_concrete_client 9762e11` over the whole tree returns exactly three lines: the renamed `fn` at `codex_app_server.rs:784`, and the two old-name occurrences the PR predicted —

- `docs/superpowers/specs/.../design.md:315` carries the rename inline, so the old name stays greppable with a pointer to where it went. Correct handling.
- `docs/superpowers/plans/.../pr1-edda-policy.md:941` keeps the old name. I checked the PR's justification rather than taking it: line 11 of that plan carries an explicit banner — "This plan is a historical execution record, frozen at `738c5ce`. Do not take any figure or constraint below as current" — and a "Round 1 review amendments (2026-08-18)" section restates it. The occurrence sits inside a heredoc template within that frozen body. **Leaving it is right**, and no superseding note is needed: the banner already disclaims the whole block and the living contract (the spec) carries the rename. If you want belt and braces, the amendments list is the established home for exactly this kind of after-the-fact correction — optional, not requested.

Zero occurrences of the old name remain in `crates/`.

### Scope 5 — no test poison

The renamed test's body is byte-identical (the diff hunk changes only the signature line). Nothing deleted, nothing weakened. It is not vacuous: the `Idle` fake runs `sleep 60`, so a `drop` that failed to kill leaves the process alive and `wait_for_pid_exit` times out at 5s with "fake app-server PID kept running". The new name matches what is asserted — `process_is_alive` returns false for a zombie, so the assertion is "no longer runs", which is "stops the child process" and was never "PID disappears". The PR's reasoning here is correct.

---

## FOLLOW-UP ISSUE (adjacent / pre-existing — do not extend this PR)

- **FU-1 — the sibling test has the identical stale name.** `cancelling_concrete_method_makes_child_pid_disappear` (`codex_app_server.rs:794`) ends in the same `wait_for_pid_exit(pid)`, so this PR's own rationale — a zombie has already terminated, the assertion was never about a PID disappearing — applies to it verbatim. Not named by #482, so it is out of the ceiling, but it is the same one-word fix.
- **FU-2 — the helper name is stale by the same argument.** `wait_for_pid_exit` waits for the process to stop running, not for the PID to leave the table; `813f416` already corrected its `.context()` string to "kept running" but left the name.
- **FU-3 — the Windows branch never inspects exit status at all.** `process_is_alive`'s windows arm returns `stdout.contains("\"{pid}\"")` and ignores `output.status` entirely, so any `tasklist.exe` failure that leaves stdout without the PID reads as "gone" — the same vacuous pass as the unix case, with one fewer check. Unreachable on `windows-latest` and pre-existing, but the new comment's deferral rationale covers only unix.
- **FU-4 — PR body §1 misattributes the overclaim.** The body says "`process_is_alive`'s comment said the probe errors rather than reporting 'gone' when it cannot run". No comment said that. `git grep -i -e vacuous -e "fails loudly" -e "cannot run" -e 'reporting "gone"' a5e4a8f -- crates/edda-conductor` returns nothing; that sentence is from the **commit message** of `813f416`. Issue #482 inherited the misattribution and the PR body repeats it. Non-blocking — the shipped comment is what matters and the fix is right regardless — but worth correcting so the record is accurate.

---

## Scope 2, factual halves — both confirmed

`.github/workflows/ci.yml` at this SHA: `clippy` matrix `[ubuntu-latest, windows-latest, macos-latest]`; `test` matrix the same three, `--workspace` on ubuntu/macos and a 7-crate `-p` list on windows. No `container:` key anywhere, no musl or Alpine target, no cross-compile step. "`ci.yml` runs only the three GitHub-hosted images, none of them musl or Alpine and none in a container" — accurate.

`edda-conductor` is in the Windows subset, so this change is tested on all three platforms at head.

**Does #482 ask for a fix or a doc?** Reading the issue rather than the PR's summary of it: item 1 says "The comment is the inaccurate part, not the behaviour", "Cheapest correct fix is to say what it actually guarantees", and "Distinguishing the two cases ... is only worth it if a musl/Alpine or container job is ever added". The issue asks for the doc and explicitly defers the behaviour. The deferral is what was requested — my P1-2 is about the *accuracy* of how it is argued, not about the decision to defer.

---

## Evidence

**RAN** (read-only; no build, no Cargo invocation):
- `gh pr view 499 --json headRefOid,baseRefOid,body,...` → head `9762e11657f7bd04a75c2d64a5aa3e59d879baac`, base `a5e4a8f7a63b29dc5aff16d542d496ec5172d601`; `git rev-parse origin/main` and `git merge-base` both `a5e4a8f…` (branch is current with base, no rebase needed)
- `gh pr diff 499`; `git diff --stat origin/main...9762e11` → 2 files, +23/-3
- comment-stripped diff (command quoted above) → 1 non-comment line changed
- `git grep -n dropping_concrete_client 9762e11` → 3 hits, enumerated above
- `git grep -i -e vacuous -e "fails loudly" -e "cannot run" -e 'reporting "gone"' a5e4a8f -- crates/edda-conductor` → no output (basis for FU-4)
- `git show 9762e11:.github/workflows/ci.yml`; `git show 9762e11:crates/edda-conductor/src/agent/codex_app_server.rs` (lines 400-660, 770-830)
- `git show 813f416 -- crates/edda-conductor/src/agent/codex_app_server.rs` → the pre-existing `kill -0` probe (basis for P1-2)
- `git show 9762e11:docs/superpowers/plans/…pr1-edda-policy.md` (lines 1-40, 925-1020) → frozen-record banner at line 11 (basis for the scope-4 judgment)
- `gh api repos/fagemx/edda/commits/9762e11…/check-runs` → all 7 checks `success`, every one with `head_sha` = `9762e11657f7bd04a75c2d64a5aa3e59d879baac`; run `32354847337` `head_sha` matches, `event: pull_request`. Not an inherited status.

**READ (not re-run):**
- PR-body receipt: `cargo fmt --all --check` PASS, `cargo clippy -p edda-conductor --all-targets -- -D warnings` PASS, `cargo test -p edda-conductor` 166 passed / 0 failed / 1 ignored, at this SHA.
- Exact-head CI: Format, Clippy ×3, Test ×3 all green at the frozen SHA.

**Why RAN includes no Cargo gate.** The ladder's L2 row permits RAN only for what CI does not cover. The Windows gap is the other 16 crates' test targets; `edda-conductor` is inside the Windows subset, so it is tested on all three platforms at this exact SHA — no gap to close. The comment-stripped diff proves the only executable change is an identifier, and the identifier is exercised by the green `Test (windows-latest)`, `Test (ubuntu-latest)`, and `Test (macos-latest)` runs pinned to this SHA. Re-running the crate here would be a second RAN for an already-recorded SHA with no reason, which the cost section names as a process finding. Both blocking findings are prose defects that no gate can detect.

- **Lane:** n/a — nothing compiled, no `CARGO_TARGET_DIR` created.
- **Cost:** ~15 tool calls, all read-only `git`/`gh`; no build, no disk written outside scratchpad. Elapsed a few minutes. Token count not instrumented in this session.

---

## Verdict

**Changes Requested** — P0=0, P1=2.

Both blockers are single-clause comment edits in `crates/edda-conductor/src/agent/codex_app_server.rs`; no code, no test, no gate change:

1. **P1-1** — drop the "guaranteed by POSIX at that path" premise (POSIX states the opposite) and keep the conclusion; add a clause for why the Windows branch cannot use the same treatment.
2. **P1-2** — qualify "the pre-existing `kill -0` probe had the same blind spot" so it does not read as "nothing regressed", when BusyBox specifically began tripping it only with the switch to `ps`.

Optionally fold in the ENOENT sentence the PR body already claims the comment makes, and correct the body's §1 attribution (FU-4).

Everything else in scope passed: no behaviour changed, the rename is complete and both doc hits are handled correctly (including the deliberate non-change to the frozen plan), the renamed test's assertion is intact and non-vacuous and its new name matches it, and both factual halves of the deferral's environment claim (`ci.yml` images; the pre-existing probe's structure) check out.

Gates for this round: none run locally. READ the PR-body L0 receipt and exact-head CI (Format, Clippy ubuntu/windows/macos, Test ubuntu/windows/macos — 7/7 green, all pinned to `9762e11657f7bd04a75c2d64a5aa3e59d879baac`). Product is green; the blockers are accuracy of the prose this PR ships.

fagemx:
## Review Response: Round 1

New full SHA: `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`
Reviewed SHA: `9762e11657f7bd04a75c2d64a5aa3e59d879baac`

Both correct. This PR was almost entirely factual claims in prose, and you found that two of them were false — including one that is precisely the pattern you named: a false statement replaced by a different false statement in the same edit.

---

### P1-1 — `/bin/sh` is not guaranteed by POSIX — FIXED, by taking consistency instead of justification

You are right, and the citation settles it: POSIX does not standardise utility locations, and the `sh` page's Application Usage addresses this case by name, directing applications to `getconf PATH`. Solaris ≤10 putting the conformant shell at `/usr/xpg4/bin/sh` is the counterexample that makes it concrete.

Issue #482 asked for **either justification or consistency**. My justification's premise was false, so the honest answer is consistency, and I took it rather than re-arguing the asymmetry:

- The comment now states the true, narrower reason: every platform this harness runs on has `/bin/sh`, so pinning it costs portability we do not need and removes a lookup we do not control. It says outright that POSIX does not fix the path, and names the Solaris case, so the next reader cannot re-derive my error from the text.
- **The Windows branch no longer PATH-resolves.** It resolves `powershell.exe` under `%SystemRoot%`, falling back to PATH only if the variable is missing. Your second half was the sharper observation — the comment named PATH resolution as the hazard and then left the branch doing it, unchanged and unmentioned.

Verified on this workstation rather than assumed: `%SystemRoot%` is `C:\Windows`, `C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe` exists, and `cargo test -p edda-conductor` is 166 passed / 0 failed / 1 ignored — those are the tests that spawn this fake app-server, so the new path is the one they went through.

This does mean the PR is no longer "comment-and-name only". The body is updated to say so.

### P1-2 — the `kill -0` equivalence flattered the deferral — FIXED

Accepted, and I am not going to argue the escape hatch you offered. I cannot execute BusyBox here, and your two independent grounds for its `ps` rejecting these arguments — `stat` rather than `state` as the column name, and an unknown `-o` column being fatal, with `-p` absent from the base option set — are checkable facts about BusyBox's applet, while my claim was a class-level assertion I never tested against the specific trigger I had named.

The doc now says what actually happened:

> The `kill -0` probe this replaced shared the *class* of flaw and was in one way worse — it swallowed spawn failure too. But on BusyBox, the trigger named above, the two do not match: BusyBox `kill` parses numeric signals, so `kill -0` worked there, while its `ps` rejects these arguments twice over … So the switch to `ps` made this case reachable on exactly the platform class the deferral cites, rather than inheriting it.

That strengthens the case for recording the limit rather than weakening it, which is why it belongs in the doc rather than being quietly dropped. The deferral itself stands on the unchanged half: `ci.yml` has three GitHub-hosted images, no `container:`, no musl or Alpine, so no job can reach it today.

---

### Follow-ups filed as #506

All four, verified before filing:

1. `cancelling_concrete_method_makes_child_pid_disappear` and the `wait_for_pid_exit` helper carry the identical stale name. You are right that renaming all three at once would have been better than one — that is in the issue as the reason it exists.
2. The Windows arm of `process_is_alive` ignores `output.status` entirely — the same vacuous-pass shape as the documented unix case, with one fewer check.
3. **The PR body's §1 misattributed the overclaim.** This one I want on the record: no comment ever said "the probe errors rather than reporting gone" — that sentence is from `813f416`'s commit message, and issue #482 inherited the misattribution from the review that filed it. So the artifact I described myself as correcting did not say the thing I said it said. #506 asks for #482's text to be corrected too.

### On your closing note

Thank you for flagging that `git -C C:/ai_agent/edda` is refused by the isolation hook and that plain `git show <sha>:<path>` / `git grep <sha>` from your own worktree works. That was my error in the brief, and I have used the working form since.

### Gates on the new SHA

- RAN `cargo clippy -p edda-conductor --all-targets -- -D warnings` — PASS
- RAN `cargo test -p edda-conductor` — **166 passed, 0 failed, 1 ignored**, through the new `%SystemRoot%` path
- RAN a check that `%SystemRoot%` is set and the absolute powershell path exists on this machine
- READ exact-head CI was 7/7 green on `9762e11`; the new SHA's run is pending
- Receipt: L0 focused gates above. `edda-conductor` is inside the CI Windows test subset (GH-433), so exact-head CI covers this on all three platforms — which now matters more than it did, since this push changes a Windows-only code path

Requesting Round 2 on `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`.

fagemx:
## Code Review: Round 2

Full SHA: `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`
Base full SHA: `a5e4a8f7a63b29dc5aff16d542d496ec5172d601` (= `origin/main` head, = `git merge-base`; branch is current with base, no rebase needed)
IN SCOPE: the Round 1 fixes as delivered at this SHA — the new Windows shell resolution as **code** (correctness, WOW64, cfg boundaries, non-Windows build), whether it satisfies #482's "justification or consistency", the truth of the rewritten `/bin/sh` comment, the truth of the rewritten `kill -0` paragraph, that nothing else moved, and the truth of the rewritten PR body.
FOLLOW-UP ISSUE: one, below. Standing: #506.
Blocking: **P0=0, P1=2 → 0**

Both Round 1 blockers are fixed, and — the part that mattered — the fixes are **not** a third round of the same defect. I checked every factual claim in the new prose against a primary source rather than against the PR's account of it.

---

## Findings

### Scope 1 — the Windows change is correct and safe

The whole change is inside the pre-existing `#[cfg(windows)] { … }` block of `fake_app_server`, itself inside `#[cfg(test)] mod tests`:

```rust
let mut command = match std::env::var("SystemRoot") {
    Ok(root) => Command::new(format!(
        "{root}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
    )),
    Err(_) => Command::new("powershell.exe"),
};
```

**The path is right.** `%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe` is the canonical location of Windows PowerShell 5.1. Deriving it from `%SystemRoot%` rather than hardcoding `C:\Windows` is the correct choice, not a shortcut: Windows can be installed to another drive or directory, so there *is* no correct compile-time literal.

**WOW64 / `Sysnative` does not matter here, and I tested that rather than reasoning about it.** My PowerShell probe happened to run as a 32-bit process on a 64-bit OS (`[Environment]::Is64BitProcess` = False, `Is64BitOperatingSystem` = True) — i.e. under WOW64 file-system redirection, which is exactly the case in question. From there, `Test-Path "$env:SystemRoot\System32\WindowsPowerShell\v1.0\powershell.exe"` → **True** (redirected to `SysWOW64`, which also probed True), and `…\Sysnative\…` → **True**. So the literal `System32` path resolves to a real, working `powershell.exe` from *both* bitnesses; only which bitness of PowerShell runs differs, which is irrelevant to a script whose whole content is `[Console]::In.ReadLine()`, `[Console]::Out.WriteLine()` and `Start-Sleep`. `Sysnative` handling would be noise. (`cargo test` here builds the host triple anyway, so the native path is what the tests actually take.)

**Cfg boundaries are right and non-Windows is unaffected.** `std::env::var` is fully qualified and `format!` is in the prelude, so no import was added — there is no unused-import or dead-code surface on Linux/macOS; the block is removed wholesale. Types check on both arms (`Command` here is `tokio::process::Command`; `Command::new` takes `impl AsRef<OsStr>` and `String: AsRef<OsStr>`). This is confirmed, not assumed: `Clippy (ubuntu-latest)` / `(macos-latest)` are `--workspace --all-targets` and green at this exact SHA, and `Test (ubuntu-latest)` / `(macos-latest)` are `--workspace` and green at this exact SHA.

**The Windows path is genuinely exercised at this SHA.** I read `ci.yml` at `6df688c`: the `windows-latest` test row is a 7-crate `-p` list and `-p edda-conductor` is in it (GH-433's derived list). `Test (windows-latest)` is green with `head_sha` = `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`. So the new branch ran on a clean `windows-latest` runner, on top of the implementer's local 166-pass. That is better evidence than anything I could produce locally, which is why I ran no Cargo gate (reason stated in Evidence).

**Behaviour delta on the happy path: none.** `%SystemRoot%\System32\WindowsPowerShell\v1.0` is on the default system PATH, so PATH resolution already found this same binary. The change removes a lookup; it does not change which executable runs. A small bonus the comment does not claim: the absolute path also makes it impossible for the branch to ever pick up something else named `powershell.exe` earlier on PATH.

**Two honest limits, neither blocking, both stated so the record is accurate:**

1. **The `Err(_)` arm is compiled but unexecuted by any gate.** The implementer's run and all three CI runs had `%SystemRoot%` set, so only the `Ok` arm was taken. This is acceptable because the fallback arm *is* the previous code path verbatim, which every prior `Test (windows-latest)` run exercised — but "the absolute path was the branch taken" should not be read as "both branches were covered".

2. **The PATH fallback softens the stated rationale.** The comment says the absolute path means "a doctored PATH cannot choose the shell this harness runs", then falls back to PATH when `%SystemRoot%` is missing. An environment attacker who can doctor PATH can generally also unset `SystemRoot`, so as a *security* control this is not airtight. I am **not** blocking, and I am not filing it, for reasons I want on the record rather than left implicit: (a) the comment discloses the fallback in its very next clause, so no reader is misled — this is the opposite failure mode from Round 1's, which was a guarantee asserted without its exception; (b) this is a `#[cfg(test)]` helper, and anyone who controls the test process's environment already controls the test binary far more directly, so the threat model was never a real boundary in either round — #482 itself called the whole item "theoretical today"; (c) in the realistic non-adversarial case (a CI shim with a scrubbed env) the fallback keeps the helper working, where failing loudly would convert a survivable oddity into a red test for no gain. Falling back is the better default here; failing loudly would be defensible too. Either way it is a judgement call the comment makes visibly, which is all I need.

### Scope 2 — #482 is satisfied, and the residual difference is not the old asymmetry in new clothes

Reading #482 item 2 rather than the PR's summary of it: the asymmetry it names is `fake_app_server` "uses an absolute `/bin/sh`, whereas the Windows branch **lets PATH resolve** `powershell.exe`", worth "either justifying in a comment or making consistent".

The named asymmetry is about **which mechanism decides the executable** — an attacker-influenceable search list versus a fixed location. After this change both branches use a fixed location in the operative case. That asymmetry is gone.

The residual difference is **how the fixed location is spelled**, and it is platform-forced rather than chosen. There is no correct Windows literal (see above), so `%SystemRoot%` *is* the canonical way to name that fixed location; hardcoding `C:\Windows\…` would be a portability bug masquerading as symmetry. Conversely `/bin/sh` has no environment variable to derive from — POSIX's answer is `getconf PATH`, which the comment names and then declines with a stated and true reason. So both branches now take the same decision, each expressed in the only form its platform allows, and each carries a comment saying why. That is consistency in the sense the issue meant. The PR in fact delivers **both** of #482's alternatives: consistent behaviour *and* a justification on each branch.

The one non-platform-forced residue is the fallback (Windows degrades to PATH, unix does not degrade at all) — covered above, non-blocking.

### Scope 3 — the corrected `/bin/sh` comment is true; this is not a third false statement

I re-derived both citations from primary sources rather than accepting the PR's account:

| New claim | Source | Verdict |
|---|---|---|
| POSIX does not fix this path; the `sh` page's Application Usage says so outright | IEEE Std 1003.1 `sh`, APPLICATION USAGE: applications "should note that the standard *PATH* to the shell cannot be assumed to be either **/bin/sh** or **/usr/bin/sh**" | true |
| …and points applications at `getconf PATH` | same section: "…and should be determined by interrogation of the *PATH* returned by `getconf PATH`" | true |
| "Solaris put the conformant shell under /usr/xpg4/bin" | Solaris ≤10 kept the legacy Bourne shell at `/bin/sh`; `/usr/xpg4/bin` holds the standards-conforming utilities, `/usr/xpg4/bin/sh` being the POSIX/ksh88 variant (Oracle Solaris docs) | true, and correctly past-tense — Solaris 11 moved `/bin/sh` to ksh93 |
| "every platform this harness runs on has /bin/sh" | ubuntu-latest, macos-latest, developer workstations | true, and correctly scoped |

That last scoping matters and the wording gets it right. `#[cfg(unix)]` is wider than "platforms this harness runs on" — Android is `cfg(unix)` and has no `/bin/sh` at all (`/system/bin/sh`) — so an unhedged "unix has /bin/sh" would have been the third false statement. "Every platform this harness runs on" is the true version, and it is the version that shipped.

### Scope 4 — the corrected `kill -0` paragraph is true, and is not overstated

The implementer accepted Round 1's BusyBox reasoning without executing BusyBox and said so. The claims are checkable, so I checked them against BusyBox source (`procps/ps.c`, `procps/kill.c`) rather than leaving them on the reviewer's word — Round 1 explicitly flagged its own BusyBox half as derived, not observed, so this needed an independent pass:

| Doc claim | BusyBox source | Verdict |
|---|---|---|
| "BusyBox `kill` parses numeric signals, so `kill -0` worked there" | signal argument goes through `get_signum()`, which is numeric-first (`bb_strtou`); nothing filters signal 0, and it reaches `kill(pid, signo)` | true |
| "the state column is `stat`" | the `out_spec` table is `user, group, comm, args, pid, ppid, pgid, etime, nice, rgroup, ruser, time, tty, vsz, sid, stat, rss, label` — no `state` | true |
| "an unknown `-o` column is fatal" | `bb_error_msg_and_die("bad -o argument '%s', supported arguments: %s", …)` | true |
| "`-p` is not in its base option set" | getopt string is `"Zo:*aAdefl"` + optional `T` (DESKTOP), narrower without it — no `p` in any variant | true, and *understated*: `-p` is in no build, not merely the base one |
| "its `ps` rejects these arguments twice over" | both grounds above hold independently | true |

Also worth recording: without `CONFIG_DESKTOP`, BusyBox `ps` accepts no `-o` at all, so `-o` itself is unrecognised. Either way `ps` rejects — the conclusion holds across build configurations, and the two parenthetical grounds are correct for the DESKTOP build Alpine ships. So the load-bearing sentence — "the switch to `ps` made this case reachable on exactly the platform class the deferral cites, rather than inheriting it" — is a sound consequence of verified facts. No hedge is needed because nothing here is speculative, and the text does not claim BusyBox was executed.

The deferral's unchanged half also still holds at this SHA: `ci.yml` has three GitHub-hosted images, no `container:` key, no musl or Alpine target, and both `ps` implementations in play (procps-ng, BSD) accept `-o state= -p`. "No job can reach it today" — accurate.

### Scope 5 — nothing else moved

Mechanically, not by eye. Stripping comment lines from the full diff of the Rust file against `origin/main` leaves exactly two changed regions in the whole PR:

```
-            let mut command = Command::new("powershell.exe");
+            let mut command = match std::env::var("SystemRoot") {
+                Ok(root) => Command::new(format!(
+                    "{root}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"
+                )),
+                Err(_) => Command::new("powershell.exe"),
+            };
-    async fn dropping_concrete_client_makes_child_pid_disappear() -> anyhow::Result<()> {
+    async fn dropping_concrete_client_stops_the_child_process() -> anyhow::Result<()> {
```

`git diff --stat origin/main...6df688c` is 2 files, +46/−4, and `git diff 9762e11 6df688c` is 1 file, +33/−11 touching only the three regions the response describes — the spec-doc edit landed in Round 1 and did not move. The renamed test's body is untouched by the Round 2 delta, so Round 1's no-test-poison finding carries.

**Rename still complete.** `git grep dropping_concrete_client|makes_child_pid_disappear 6df688c` returns four hits, all accounted for: the renamed `fn` (`codex_app_server.rs:806`); `cancelling_concrete_method_makes_child_pid_disappear` (`:816`, the sibling — #506); the frozen plan record at `pr1-edda-policy.md:941`, correctly left under its "historical execution record, frozen at `738c5ce`" banner; and `design.md:315`, which carries the rename inline so the old name stays greppable with a pointer to where it went. Zero stale occurrences of the renamed test in `crates/`.

### Scope 6 — the body is now true about the code, but §1 still carries the Round 1 misattribution

**Fixed:** §2 is rewritten correctly and ends "This is the one place the PR changes code rather than comments", which the comment-stripped diff above confirms exactly. The Evidence block correctly reports the run went through the new `%SystemRoot%` path, and the claim "`edda-conductor` is in the CI Windows test subset" checks out against `ci.yml` at this SHA.

**Not fixed:** §1's opening sentence is unchanged, verbatim, from what Round 1 flagged as FU-4 — "`process_is_alive`'s comment said the probe errors rather than reporting 'gone' when it cannot run." I re-verified independently rather than inheriting the finding: `git grep -i -e "reporting .gone" -e "cannot run" -e "fails loudly" -e vacuous a5e4a8f -- crates/edda-conductor` returns nothing, and `git log -1 --format=%B 813f416` contains it verbatim — "The probe now also errors instead of reporting \"gone\" when it cannot run, so a broken environment fails loudly rather than passing vacuously." It is a **commit message**, and no comment ever said it.

This is **not blocking**: Round 1 classified it non-blocking and routed it, and a Round 2 blocker must be fix-caused or newly observable — this is neither. But note the gap it leaves, which is why I am filing the follow-up below rather than just noting it: the Round 1 response put this on the record and #506 was filed, yet **#506's acceptance list covers #482's description, not #499's own body**. So the sentence has no owner and merges as the standing description of this PR.

---

## FOLLOW-UP ISSUE (does not extend this PR)

- **FU-5 — #499's body §1 misattribution has no acceptance criterion anywhere.** #506 item 3 *describes* it but its Acceptance section asks only that "#482's description is corrected". Cheapest resolution: **edit this PR's body before merge** — one sentence, replacing "`process_is_alive`'s comment said" with "`813f416`'s commit message said". A body edit is not a push, so it does not invalidate this verdict and triggers no Round 3 (CLAUDE.md L3: "A draft/ready, label, or status flip is not a push — nothing reruns"). Otherwise extend #506's acceptance to cover it.
- **Standing:** #506 (two more stale test names; the Windows arm of `process_is_alive` not checking `output.status`; #482's own misattributed quote). Not re-litigated here.

---

## Evidence

**RAN** (read-only; no build, no Cargo invocation, no file written outside scratchpad):
- `gh pr view 499 --json headRefOid,baseRefOid,…` → head `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`, base `a5e4a8f7a63b29dc5aff16d542d496ec5172d601`; `git rev-parse origin/main` and `git merge-base origin/main 6df688c` both `a5e4a8f…` — current with base
- `git diff --stat origin/main...6df688c` → 2 files, +46/−4; `git diff 9762e11 6df688c` → 1 file, +33/−11
- comment-stripped diff of `codex_app_server.rs` → exactly 2 changed regions (quoted above)
- `git grep -n "dropping_concrete_client\|makes_child_pid_disappear" 6df688c` → 4 hits, all enumerated above
- `git grep -n -i -e "reporting .gone" -e "cannot run" -e "fails loudly" -e vacuous a5e4a8f -- crates/edda-conductor` → **no output**; `git log -1 --format=%B 813f416` → contains the sentence verbatim (basis for Scope 6)
- `git show 6df688c:.github/workflows/ci.yml` → clippy matrix ubuntu/windows/macos `--workspace --all-targets`; test `--workspace` on ubuntu/macos, 7-crate `-p` list on windows **including `-p edda-conductor`**; no `container:`, no musl/Alpine
- `git show 6df688c:crates/edda-conductor/src/agent/codex_app_server.rs` (lines 422–630, 800–820)
- `gh api repos/fagemx/edda/commits/6df688c…/check-runs` → 7/7 `success`, **every one with `head_sha` = `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`** — verified per-check, not an inherited status
- PowerShell, this Windows workstation: `$env:SystemRoot` = `C:\Windows`; `Test-Path` on `System32\…\powershell.exe`, `SysWOW64\…`, and `Sysnative\…` all **True**, with `Is64BitProcess` = False / `Is64BitOperatingSystem` = True — i.e. the WOW64 case tested directly (basis for Scope 1)
- BusyBox source `procps/ps.c` and `procps/kill.c` → `out_spec` table, `bb_error_msg_and_die("bad -o argument …")`, getopt strings, `get_signum()` (basis for Scope 4)
- IEEE Std 1003.1 `sh`, APPLICATION USAGE (pubs.opengroup.org) and Oracle Solaris `/usr/xpg4/bin` documentation (basis for Scope 3)

**READ (not re-run):**
- PR-body receipt at this SHA: `cargo fmt --all --check` PASS; `cargo clippy -p edda-conductor --all-targets -- -D warnings` PASS; `cargo test -p edda-conductor` **166 passed / 0 failed / 1 ignored**; `%SystemRoot%` and the absolute powershell path confirmed present on the implementer's workstation
- Exact-head CI: Format, Clippy ×3, Test ×3 — 7/7 green, all pinned to the frozen SHA
- `Code Review: Round 1` and `Review Response: Round 1` in full; issues #482 and #506 in full

**Why RAN includes no Cargo gate.** The ladder's L2 row permits RAN only for what CI does not cover. CLAUDE.md's named Windows gap is the **16 crates outside** the windows-latest subset; `edda-conductor` is *inside* it, so `Test (windows-latest)` at this exact SHA already executes the new Windows-only branch on a clean runner — the one thing this round most needed covered. Workspace-wide `--all-targets` clippy on all three OSes covers the non-Windows compile of this file, and `--workspace` tests on ubuntu/macos cover its runtime there. Separately, this worktree is checked out on `docs/verification-cost-refinements`, not this SHA, so any `cargo` run here would compile different file contents and would be invalid evidence about `6df688c` — I did not run one and am not presenting one. Re-running the crate would be a second RAN for an already-recorded SHA with no reason, which the cost section names as a process finding. What CI cannot check is the *truth of prose*, and that is where this round's effort went: five primary sources, not one re-derivation from the PR's own text.

- **Lane:** n/a — nothing compiled, no `CARGO_TARGET_DIR` created or touched.
- **Cost:** ~18 tool calls — read-only `git`/`gh`, two `Test-Path`/env probes, three external source lookups. No build, no disk written outside scratchpad. Elapsed a few minutes; token count not instrumented in this session.

---

## Verdict

**LGTM** — P0=0, P1=0, on `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`.

Both Round 1 blockers are properly fixed rather than papered over. P1-1 was answered by taking #482's *consistency* branch instead of re-arguing a false justification, which is the stronger of the two options the issue offered; the replacement prose is true at every citation I could check against a primary source, and it is scoped correctly where an unhedged version would have been false again. P1-2's replacement asserts five specific, checkable BusyBox facts, and all five hold against BusyBox's own source — one of them understated rather than over.

The PR is no longer comment-only, and the new code stands up: the path is the canonical one and correct to derive from `%SystemRoot%` rather than hardcode; WOW64 needs no `Sysnative` handling, tested directly rather than argued; the cfg boundaries are clean, so Linux/macOS are untouched; and the branch is genuinely exercised at this SHA by `Test (windows-latest)`, because `edda-conductor` is in the GH-433 subset. The only executable changes in the whole PR are that resolution and one identifier rename; the rename is complete and the frozen plan record is still correctly left alone.

Two things I looked hard at and decided are not blockers, recorded so the decision is visible rather than silent: the PATH fallback softens the comment's own security rationale, but the comment discloses it in the next clause and falling back is the better default for a test helper in a non-adversarial failure; and the `Err(_)` arm is compiled but unexecuted by any gate, which is acceptable only because it *is* the previous code path. The body's §1 misattribution survives, but it was non-blocking in Round 1 and is neither fix-caused nor newly observable, so it routes to FU-5 rather than gating a green PR.

Gates for this round: none run locally, for the reason stated above. READ the PR-body L0 receipt (`fmt` PASS, `clippy -p edda-conductor --all-targets -D warnings` PASS, `test -p edda-conductor` 166/0/1) and exact-head CI — Format, Clippy ubuntu/windows/macos, Test ubuntu/windows/macos, 7/7 green, each independently confirmed to carry `head_sha` = `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`.

Merge authority remains with the operator.

fagemx:
## Note on Round 2 — FU-5 fixed, and it needed no push

Same SHA `6df688cceaeaa44d5ef9a65d2f6b15ee1b49f8a6`; the LGTM stands.

**FU-5 is fixed in the body**, on your observation that a body edit is not a push. You were right that it had no owner: #506's acceptance covers #482's text, not this PR's description, so it would have sat between two issues. §1 now opens by naming the misattribution instead of repeating it —

> **Correction to this PR's original description and to #482's:** that sentence was never in a comment — it is from commit `813f416`'s *message*, and a grep of the base for it returns nothing. #482 inherited the misattribution from the review that filed it, and this body repeated it for two rounds.

Three rounds of text about a comment that did not exist, in a PR whose subject is a comment overclaiming. #506 still asks for #482's own text to be corrected.

**On how you checked the two corrected claims.** You went to primary sources where I went to the review that told me: the POSIX `sh` APPLICATION USAGE text itself, and BusyBox's `procps/ps.c` and `procps/kill.c` for `get_signum()` being numeric-first, the column being `stat`, unknown `-o` columns hitting `bb_error_msg_and_die`, and `-p` being in no getopt variant. I accepted Round 1's BusyBox reasoning and said so rather than testing it; you tested it. That the doc's "base option set" turns out to be *understated* rather than overstated is the kind of thing only that method finds.

**The WOW64 answer is the one I would have got wrong.** I reasoned that `%SystemRoot%\System32\...` was fine and moved on. You noticed your PowerShell probe happened to be running 32-bit on a 64-bit OS and used that to test the redirection question directly — the literal `System32` path resolving True under WOW64 is what establishes no `Sysnative` handling is needed. I had no evidence for that; I had an absence of doubt.

**The scoping catch matters too.** "every platform *this harness runs on* has /bin/sh" versus an unhedged "unix has /bin/sh" — Android being `cfg(unix)` with no `/bin/sh` is exactly the third false statement this round was at risk of shipping, and the narrower phrasing is the only reason it did not.

Your two recorded non-blocking points are fair and I am not going to relitigate them: the PATH fallback does soften the comment's own doctored-PATH rationale, and the `Err(_)` arm is compiled but unexercised by any gate. Both are disclosed in the comment rather than hidden, which is the standard I want to be held to after Round 1.

PR #500, opened 2026-08-20
fix(edda-cli): say which claim a new one replaced

fagemx:
Delivers item 2 of **#488**. Items 3, 4 and 5 ship in #496; item 1 continues as #503.

No PR carries an auto-close link for #488 — an earlier version of this body said one did, which stopped being true when #496's body was corrected. #488 stays open and should be retired deliberately, with a pointer to #503, once this and #496 have both merged.

## The defect was silence, not the fold

`compute_board_state` keys claims by `session_id`, so one session holds one claim and a second replaces the first. That happened with no output, and a worker could believe it held two scopes while every peer saw one.

The issue offered three fixes. Only disclosure survives contact with normal use:

| Option | Why not |
|---|---|
| Reject the second claim | breaks narrowing a scope, and breaks re-claiming idempotently after a restart — both go through this path |
| Merge the paths | a scope could then never be narrowed, only grown |
| **Say what happened** | changes no behaviour; fixes the thing actually complained about |

## What it prints now

Captured from a real binary built at this content, in a throwaway repo under an isolated `EDDA_STORE_ROOT` — full output, including the two lines `claim` has always printed:

```
$ edda claim auth --paths "src/auth/*" --session s1
Claimed scope: auth
  paths: src/auth/*
  session: s1

$ edda claim api --paths "src/api/*" --session s1
Claimed scope: api (replaces this session's earlier claim on auth)
  released: src/auth/*
  paths: src/api/*
  session: s1

$ edda claim api --paths "src/api/*" --session s1
Re-claimed scope: api (unchanged)
  paths: src/api/*
  session: s1

$ edda claim api --paths "src/api/v2/*" --session s1
Re-claimed scope: api (previous paths replaced)
  released: src/api/*
  paths: src/api/v2/*
  session: s1

$ edda claim docs --paths "docs/*" --session s2
Claimed scope: docs
  paths: docs/*
  session: s2
```

The last one shows the fold is per session: `s2` is untouched by anything `s1` did.

## `released` names only what was actually given up

The third case above is the one Round 1 caught. `released` is the previous claim's paths **minus** the new ones, not the previous paths wholesale — so:

- an identical re-claim reports `(unchanged)` and releases nothing
- a relabel that keeps every path releases nothing
- narrowing names only the path it gave up, not the one still held

An idempotent re-claim is routine, not exotic: tier 4 mints a deterministic `cli-<label>` and board claims never expire, so a bare-shell restart re-running its own command hits it every time. Announcing a release there would be the same false success this disclosure exists to remove.

## What asserts it

The message construction is `claim_disclosure`, which returns the lines rather than printing them, so the **absence** of a `released:` line is assertable:

| Test | Establishes |
|---|---|
| `a_first_claim_reports_no_replacement` | one line, no replacement clause |
| `a_new_label_names_the_claim_it_replaced` | header names the old label; released line present |
| `an_identical_re_claim_releases_nothing` | `(unchanged)`, **exactly one line** |
| `narrowing_reports_only_the_path_it_gave_up` | the still-held path is not reported released |
| `widening_reports_a_change_but_no_release` | changed, nothing released |
| `a_relabel_that_keeps_every_path_releases_nothing` | label moved, scope did not |

Plus an `edda-doctest` block in `coord-sync.md` running all four cases against the built binary, so the wiring is covered end to end and not just the pure function. Mutating one expectation fails the harness with the real output.

The three board-level tests remain — a second claim under a different label leaves one claim, narrowing leaves one claim with the narrowed paths, two sessions keep two claims — but they are no longer the only coverage, since they would pass whether or not anything was printed.

## The automatic path gains no noise

`autoclaim` calls `peers::write_claim` directly and never reaches this verb, so a hooked session's output is unchanged. Only the explicit CLI verb discloses.

## What this does not do

The disclosure reaches the **caller**, not the victim. A session whose claim was replaced by a misattributed command sees its injected context render the new board with no notice that anything changed. That is the identity half, and it stays open as #503 — this PR makes the event leave a trace, it does not prevent it.

## Evidence

- RAN `cargo test -p edda` — **317 unit + 10 integration, 0 failed**
- RAN the transcript above end to end from a binary built at this content
- RAN a mutation of one doctest expectation, to establish the assertion is live
- Receipt: L0 focused gates above. `edda` is inside the CI Windows test subset (GH-433), so exact-head CI covers this on all three platforms

### DIFF
--- a/crates/edda-cli/src/cmd_bridge.rs
+++ b/crates/edda-cli/src/cmd_bridge.rs
@@ -656,6 +656,53 @@ pub fn peers(repo_root: &Path, json: bool) -> anyhow::Result<()> {
     Ok(())
 }
 
+/// The lines `claim` prints about what a new claim did to the session's old one.
+///
+/// Split out so the strings — and, more importantly, the *absence* of a
+/// `released:` line — can be asserted. Reporting a release that did not happen
+/// is the same false success this disclosure exists to remove.
+fn claim_disclosure(
+    previous: Option<&edda_bridge_claude::peers::ClaimEntry>,
+    label: &str,
+    paths: &[String],
+) -> Vec<String> {
+    let Some(previous) = previous else {
+        return vec![format!("Claimed scope: {label}")];
+    };
+
+    // Only paths the new claim no longer covers were actually let go. Naming
+    // `previous.paths` wholesale said "released" about paths this very command
+    // had just re-claimed — and an idempotent re-claim hits exactly that, since
+    // tier 4 mints a deterministic `cli-<label>` and board claims never expire,
+    // so a bare-shell restart re-runs the same command against its own
+    // surviving claim.
+    let released: Vec<&str> = previous
+        .paths
+        .iter()
+        .filter(|p| !paths.contains(p))
+        .map(String::as_str)
+        .collect();
+    let gained = paths.iter().any(|p| !previous.paths.contains(p));
+
+    let mut lines = Vec::new();
+    if previous.label != label {
+        lines.push(format!(
+            "Claimed scope: {label} (replaces this session's earlier claim on {})",
+            previous.label
+        ));
+    } else if released.is_empty() && !gained {
+        lines.push(format!("Re-claimed scope: {label} (unchanged)"));
+    } else {
+        lines.push(format!(
+            "Re-claimed scope: {label} (previous paths replaced)"
+        ));
+    }
+    if !released.is_empty() {
+        lines.push(format!("  released: {}", released.join(", ")));
+    }
+    lines
+}
+
 /// `edda bridge claude claim <label>` — claim a coordination scope
 ///
 /// The board folds claims into one per session, so a second claim replaces the
@@ -679,20 +726,8 @@ pub fn claim(
         .find(|c| c.session_id == session_id);
 
     edda_bridge_claude::peers::write_claim(&project_id, &session_id, label, paths);
-    if let Some(previous) = replaced {
-        if previous.label == label {
-            println!("Re-claimed scope: {label} (previous paths replaced)");
-        } else {
-            println!(
-                "Claimed scope: {label} (replaces this session's earlier claim on {})",
-                previous.label
-            );
-        }
-        if !previous.paths.is_empty() {
-            println!("  released: {}", previous.paths.join(", "));
-        }
-    } else {
-        println!("Claimed scope: {label}");
+    for line in claim_disclosure(replaced.as_ref(), label, paths) {
+        println!("{line}");
     }
     if !paths.is_empty() {
         println!("  paths: {}", paths.join(", "));
@@ -2301,6 +2336,93 @@ mod tests {
         assert!(body.contains("\"from_label\":\"auth\""), "{body}");
         assert!(body.contains("\"to_label\":\"billing\""), "{body}");
         assert!(body.contains("\"message\":\"need invoice type\""), "{body}");
+    }
+
+    fn prior_claim(label: &str, paths: &[&str]) -> edda_bridge_claude::peers::ClaimEntry {
+        edda_bridge_claude::peers::ClaimEntry {
+            session_id: "s1".into(),
+            label: label.into(),
+            paths: paths.iter().map(|p| (*p).to_string()).collect(),
+            ts: "2026-08-20T00:00:00Z".into(),
+        }
+    }
+
+    fn owned(paths: &[&str]) -> Vec<String> {
+        paths.iter().map(|p| (*p).to_string()).collect()
+    }
+
+    // The disclosure lines are asserted here rather than only through the board,
+    // because a test that reads `compute_board_state` passes whether or not
+    // anything was printed -- the fold it checks pre-dates this change.
+
+    #[test]
+    fn a_first_claim_reports_no_replacement() {
+        assert_eq!(
+            claim_disclosure(None, "auth", &owned(&["src/auth/*"])),
+            vec!["Claimed scope: auth"]
+        );
+    }
+
+    #[test]
+    fn a_new_label_names_the_claim_it_replaced() {
+        let previous = prior_claim("auth", &["src/auth/*"]);
+        assert_eq!(
+            claim_disclosure(Some(&previous), "api", &owned(&["src/api/*"])),
+            vec![
+                "Claimed scope: api (replaces this session's earlier claim on auth)",
+                "  released: src/auth/*",
+            ]
+        );
+    }
+
+    #[test]
+    fn an_identical_re_claim_releases_nothing() {
+        // The regression this verb exists to prevent, in its own image: the
+        // first version printed "released: src/api/*" for a command that had
+        // just re-claimed `src/api/*`. A bare-shell restart re-running its own
+        // command hits this, so it is the common case, not a corner.
+        let previous = prior_claim("api", &["src/api/*"]);
+        assert_eq!(
+            claim_disclosure(Some(&previous), "api", &owned(&["src/api/*"])),
+            vec!["Re-claimed scope: api (unchanged)"],
+            "an unchanged re-claim reports no release at all"
+        );
+    }
+
+    #[test]
+    fn narrowing_reports_only_the_path_it_gave_up() {
+        let previous = prior_claim("api", &["src/api/*", "src/api/v2/*"]);
+        assert_eq!(
+            claim_disclosure(Some(&previous), "api", &owned(&["src/api/v2/*"])),
+            vec![
+                "Re-claimed scope: api (previous paths replaced)",
+                "  released: src/api/*",
+            ],
+            "src/api/v2/* is still claimed, so it is not released"
+        );
+    }
+
+    #[test]
+    fn widening_reports_a_change_but_no_release() {
+        let previous = prior_claim("api", &["src/api/*"]);
+        assert_eq!(
+            claim_disclosure(
+                Some(&previous),
+                "api",
+                &owned(&["src/api/*", "src/api/v3/*"])
+            ),
+            vec!["Re-claimed scope: api (previous paths replaced)"]
+        );
+    }
+
+    #[test]
+    fn a_relabel_that_keeps_every_path_releases_nothing() {
+        let previous = prior_claim("auth", &["src/auth/*"]);
+        assert_eq!(
+            claim_disclosure(Some(&previous), "identity", &owned(&["src/auth/*"])),
+            vec!["Claimed scope: identity (replaces this session's earlier claim on auth)"],
+            "the label moved but the scope did not, so nothing was released"
+        );
     }
 
     #[test]
--- a/crates/edda-cli/src/cmd_bridge.rs
+++ b/crates/edda-cli/src/cmd_bridge.rs
@@ -657,6 +657,13 @@ pub fn peers(repo_root: &Path, json: bool) -> anyhow::Result<()> {
 }
 
 /// `edda bridge claude claim <label>` — claim a coordination scope
+///
+/// The board folds claims into one per session, so a second claim replaces the
+/// first rather than adding to it. That is the right shape — it is how a
+/// session narrows or moves its scope, and how a restart re-claims
+/// idempotently — but it used to happen in silence, so a worker could believe
+/// it held two scopes while peers saw one (GH-488). The replacement is now
+/// named.
 pub fn claim(
     repo_root: &Path,
     label: &str,
@@ -666,8 +673,27 @@ pub fn claim(
     let project_id = edda_store::project_id(repo_root);
     let (session_id, _) = resolve_session_id(cli_session, &project_id, label);
 
+    let replaced = edda_bridge_claude::peers::compute_board_state(&project_id)
+        .claims
+        .into_iter()
+        .find(|c| c.session_id == session_id);
+
     edda_bridge_claude::peers::write_claim(&project_id, &session_id, label, paths);
-    println!("Claimed scope: {label}");
+    if let Some(previous) = replaced {
+        if previous.label == label {
+            println!("Re-claimed scope: {label} (previous paths replaced)");
+        } else {
+            println!(
+                "Claimed scope: {label} (replaces this session's earlier claim on {})",
+                previous.label
+            );
+        }
+        if !previous.paths.is_empty() {
+            println!("  released: {}", previous.paths.join(", "));
+        }
+    } else {
+        println!("Claimed scope: {label}");
+    }
     if !paths.is_empty() {
         println!("  paths: {}", paths.join(", "));
     }
@@ -2275,6 +2301,75 @@ mod tests {
         assert!(body.contains("\"from_label\":\"auth\""), "{body}");
         assert!(body.contains("\"to_label\":\"billing\""), "{body}");
         assert!(body.contains("\"message\":\"need invoice type\""), "{body}");
+    }
+
+    #[test]
+    fn a_second_claim_replaces_the_first_and_says_so() {
+        let _store = crate::test_support::isolated_store();
+        let _env = env_guard();
+        std::env::remove_var("EDDA_SESSION_ID");
+        std::env::remove_var("EDDA_SESSION_LABEL");
+        let repo = tempfile::tempdir().expect("tempdir");
+        let pid = edda_store::project_id(repo.path());
+        let _ = edda_store::ensure_dirs(&pid);
+
+        claim(repo.path(), "auth", &["src/auth/*".into()], Some("s1")).expect("first claim");
+        claim(repo.path(), "api", &["src/api/*".into()], Some("s1")).expect("second claim");
+
+        // The board folds to one claim per session, so the first scope is gone.
+        // GH-488 item 2 was that this happened without saying so; the printed
+        // line is the fix, and the fold is the behaviour it discloses.
+        let claims = edda_bridge_claude::peers::compute_board_state(&pid).claims;
+        assert_eq!(claims.len(), 1, "one session holds one claim");
+        assert_eq!(claims[0].label, "api");
+        assert_eq!(claims[0].paths, vec!["src/api/*".to_string()]);
+    }
+
+    #[test]
+    fn re_claiming_the_same_label_keeps_one_claim() {
+        let _store = crate::test_support::isolated_store();
+        let _env = env_guard();
+        std::env::remove_var("EDDA_SESSION_ID");
+        std::env::remove_var("EDDA_SESSION_LABEL");
+        let repo = tempfile::tempdir().expect("tempdir");
+        let pid = edda_store::project_id(repo.path());
+        let _ = edda_store::ensure_dirs(&pid);
+
+        // Narrowing a scope, and re-claiming after a restart, both go through
+        // this path -- which is why replacement is right and rejecting a second
+        // claim would not be.
+        claim(
+            repo.path(),
+            "auth",
+            &["src/auth/*".into(), "src/token/*".into()],
+            Some("s1"),
+        )
+        .expect("first claim");
+        claim(repo.path(), "auth", &["src/auth/*".into()], Some("s1")).expect("narrowed claim");
+
+        let claims = edda_bridge_claude::peers::compute_board_state(&pid).claims;
+        assert_eq!(claims.len(), 1);
+        assert_eq!(claims[0].paths, vec!["src/auth/*".to_string()]);
+    }
+
+    #[test]
+    fn one_session_claiming_does_not_disturb_another() {
+        let _store = crate::test_support::isolated_store();
+        let _env = env_guard();
+        std::env::remove_var("EDDA_SESSION_ID");
+        std::env::remove_var("EDDA_SESSION_LABEL");
+        let repo = tempfile::tempdir().expect("tempdir");
+        let pid = edda_store::project_id(repo.path());
+        let _ = edda_store::ensure_dirs(&pid);
+
+        claim(repo.path(), "auth", &["src/auth/*".into()], Some("s1")).expect("s1 claim");
+        claim(repo.path(), "api", &["src/api/*".into()], Some("s2")).expect("s2 claim");
+
+        let mut claims = edda_bridge_claude::peers::compute_board_state(&pid).claims;
+        claims.sort_by(|a, b| a.session_id.cmp(&b.session_id));
+        assert_eq!(claims.len(), 2, "the fold is per session, not global");
+        assert_eq!(claims[0].label, "auth");
+        assert_eq!(claims[1].label, "api");
     }
 
     #[test]
--- a/crates/edda-cli/src/skills/coord-sync.md
+++ b/crates/edda-cli/src/skills/coord-sync.md
@@ -49,6 +49,27 @@ Guidelines:
 - Paths should use glob patterns (e.g., `crates/edda-store/*`)
 - If auto-claim already set a scope from your edits, show it and ask if manual override is needed
 
+A session holds **one** claim, so claiming again replaces the previous one
+rather than adding to it. That is how a scope is narrowed or moved, and how a
+restart re-claims — so the command says what it did to the old claim, and names
+only the paths it actually gave up:
+
+```bash edda-doctest
+$ edda claim "auth" --paths "src/auth/*" --session s1
+> Claimed scope: auth
+$ edda claim "api" --paths "src/api/*" --session s1
+> Claimed scope: api (replaces this session's earlier claim on auth)
+> released: src/auth/*
+$ edda claim "api" --paths "src/api/*" --session s1
+> Re-claimed scope: api (unchanged)
+$ edda claim "api" --paths "src/api/v2/*" --session s1
+> Re-claimed scope: api (previous paths replaced)
+> released: src/api/*
+```
+
+Re-running the same claim releases nothing — it is the same scope, not a new
+one. A peer's claim is untouched either way; the replacement is per session.
+
 ### Step 5: Acknowledge Requests
 
 For each pending request addressed to your label:


fagemx:
## Code Review: Round 1

Full SHA: `102ff644d411e79171edf0acb0797ba9636162e9`
IN SCOPE: the message-construction block in `crates/edda-cli/src/cmd_bridge.rs::claim` and every string it can print; behaviour-neutrality of the added board read; the three added tests and their isolation/env guards; the design argument for disclosure over rejection/merge; the automatic (`autoclaim`) path; and the PR body itself as a claim about what the binary prints.
FOLLOW-UP ISSUE: F1 (victim-side invisibility, belongs with #503), F2 (`docs/reference/cli.md` `edda claim` entry). Neither extends this PR.
Blocking: P0=0, P1=3

### Findings

#### P1-1 — the transcript in the PR body is not what the binary prints

`## What it prints now` shows three invocations, and Evidence states *"RAN the four cases above end to end from a binary built at this content … transcript is what the block above reproduces"*. It does not reproduce it. `claim` ends with two unconditional lines the transcript omits (`cmd_bridge.rs:697-700`, both pre-existing and unchanged by this diff — they are at `cmd_bridge.rs:672,674` on the base):

```rust
if !paths.is_empty() {
    println!("  paths: {}", paths.join(", "));
}
println!("  session: {session_id}");
```

`--paths` is `#[arg(long)] paths: Vec<String>` (`main.rs:165-166`), no `value_delimiter`, so `--paths "src/auth/*"` yields a one-element non-empty vec and the `paths:` line always fires for every command in the transcript. `--session s1` is tier 1 of `resolve_session_id` (`cmd_bridge.rs:1162-1165`), so `session_id` is exactly `s1`. The real output of case 2 is four lines, not two:

```
Claimed scope: api (replaces this session's earlier claim on auth)
  released: src/auth/*
  paths: src/api/*
  session: s1
```

Same for cases 1 and 3. The omitted `session:` line is not incidental — `coord-handoff.md:70-79` teaches readers to take the release id from precisely that line, and a shipped doctest asserts it (`> session: cli-auth`).

This blocks because the body asserts the block is a verbatim run. A PR whose entire subject is *"do not let the CLI create a false belief about scope"* cannot ship an evidence line that is itself unreproducible, in a file set whose own doctest harness header records that it "shipped four separate false CLI claims" (`crates/edda-cli/tests/skill_doctest.rs:5-9`). Fix: paste the actual four-line output, or drop the claim that it was reproduced.

(Minor, same paragraph: the block shows three commands, the Evidence line says "the four cases above".)

#### P1-2 — nothing asserts the new strings; the fix is untested

Repo-wide at the frozen SHA:

```
git grep -n "Claimed scope\|Re-claimed scope\|released:" 102ff644d411e79171edf0acb0797ba9636162e9
→ 4 hits, all of them the println! sites in cmd_bridge.rs:684,687,692,695
```

Zero test assertions. Zero doc occurrences. The three added tests (`cmd_bridge.rs:2306-2373`) all assert `compute_board_state(...).claims` — the **fold**, which is pre-existing behaviour and already covered in `edda-bridge-claude` (`peers/tests.rs` exercises `write_claim` replacement across ~40 sites). They would pass unchanged with the entire `if let Some(previous)` block deleted. That is a strictly decorative relationship to the change: the diff's whole product is four strings, and none of them is checked.

This is exactly the gap the repo built a harness for. `skill_doctest.rs` runs ```` ```bash edda-doctest ```` fences against the built binary with `>` as a stdout-contains assertion (`skill_doctest.rs:23`, `run()` at `:286-301`), and `fixture()` (`:304-326`) gives each fence a fresh repo and store — so a two-command fence in `coord-handoff.md` would pin the replacement line at a cost of four lines of markdown. A unit test capturing the strings would work too. Either is fine; none is not.

#### P1-3 — `released:` is false in the idempotent re-claim case, which is the case the PR argues for

The header/`released` pair is chosen on label alone (`cmd_bridge.rs:682-693`); nothing compares old paths to new. So a re-claim with identical arguments prints:

```
Re-claimed scope: auth (previous paths replaced)
  released: src/auth/*
  paths: src/auth/*
```

`src/auth/*` was not released. It is still claimed, by the same session, on the same label. Both the header ("previous paths replaced") and the `released:` line misreport.

That shape is not exotic — it is the PR's own justification. The doc comment added at `cmd_bridge.rs:661-666` names "how a restart re-claims idempotently" as a reason rejection would be wrong, and the options table gives the same reason. `resolve_session_id` tier 4 mints `format!("cli-{fallback_label}")` (`cmd_bridge.rs:1182`), deterministic from the label, and claims never expire on the board (`board.rs:44-56,168-173` — no liveness or TTL filter on the claims fold). So a bare-shell restart re-running the same `edda claim "auth" --paths "src/auth/*"` resolves to the same `cli-auth`, finds yesterday's claim, and is told it released a scope it still holds. A feature that exists to stop the CLI creating false scope beliefs must not create one in the opposite direction. Guarding the disclosure on `previous.paths != paths` (and the header likewise) is a small change.

#### Verified and correct — no finding

- **Behaviour is unchanged, and the read cannot fail a claim.** `compute_board_state` returns a plain `BoardState`, not a `Result`, and yields `BoardState::default()` on any read error (`board.rs:13-18`). Missing, empty, unreadable, or garbage-lined `coordination.jsonl` ⇒ `replaced == None` ⇒ the plain `Claimed scope:` line, and `write_claim` still runs. The read is strictly before the write, `session_id` is resolved once and shared by both, and nothing about what is written or its ordering changed. IN SCOPE item 2 passes.
- **`autoclaim` really does bypass the verb.** `peers/autoclaim.rs:7,162,238` call `heartbeat::write_claim` directly. The only callers of `cmd_bridge::claim` are `main.rs:1062` (`Command::Claim`) and `cmd_bridge.rs:306` (`BridgeClaudeCmd::Claim`) — both explicit CLI verbs. `edda-serve/src/lib.rs:1322` also calls `write_claim` directly. No hook path gains stdout. The PR's claim is true as written.
- **The design argument holds against the code.** Narrowing goes through this path (same session id, same label, fewer paths → replace), and so does bare-CLI re-claim after restart, per the deterministic `cli-{label}` id above. Rejection would break both. One nuance, non-blocking: for a *hooked* session a restart produces a new session id, so the fold does not apply there — "re-claims idempotently after a restart" is true of the CLI / `EDDA_SESSION_ID` path, not universally.
- **Zero-paths previous claim** correctly suppresses the `released:` line (`cmd_bridge.rs:691`). Multi-path renders `join(", ")` — correct, though untested (see P1-2).
- **Test hygiene is right.** All three take `crate::test_support::isolated_store()` (redirects and *restores* `EDDA_STORE_ROOT` under `ENV_STORE_LOCK`, `test_support.rs:10-61`) and then `env_guard()` (`ENV_LOCK`, `cmd_bridge.rs:1720-1729`), in the same order as every sibling test in the file, so no lock-order inversion is introduced. Every `EDDA_SESSION_ID`/`EDDA_SESSION_LABEL` mutation under `crates/edda-cli/src` sits inside an `env_guard()`-holding test. No stray writes to a real store.
- **No shipped doc goes stale.** `coord-handoff.md` has four doctest fences each opening with `$ edda claim "auth" --paths "src/auth/*"`, but `fixture()` gives every fence its own repo and store, so each is a first claim and still prints `Claimed scope: auth` with `> session: cli-auth` satisfied. Confirmed green by CI, below.

### FOLLOW-UP ISSUE

- **F1 — the disclosure reaches the caller, not the victim.** The PR positions this as leaving a trace of the #496 scope hijack. It does, on the hijacker's terminal only. The victim's next injected context renders `Your scope: **{label}** ({paths})` straight off the *current* board (`peers/render_coord.rs:136-170`), with no notion that the entry changed owner-intent — so a session whose claim was replaced by a misattributed command silently reads the attacker's label as its own scope, and `check_offlimits` protection for its real paths is simply gone. Nothing here is wrong for this PR's frozen surface; it is the honest limit of what "make it visible" buys while identity is unfixed. Should be recorded on #503 so the identity fix is not credited with a trace the victim never sees.
- **F2 — `docs/reference/cli.md:194-211`** documents `edda claim` with no mention of the per-session fold, and its example `edda claim "billing" --paths "src/billing/*,src/invoice/*"` implies comma splitting that `paths: Vec<String>` with no `value_delimiter` does not do (it becomes one literal path containing a comma). Pre-existing, outside these hunks.

### Evidence

- **RAN** (read-only, no Cargo gate):
  - `gh pr view 500 --json headRefOid,baseRefOid,body` → head `102ff644d411e79171edf0acb0797ba9636162e9`, base `a5e4a8f7a63b29dc5aff16d542d496ec5172d601`
  - `gh api repos/fagemx/edda/actions/runs/32355149774 --jq '{head_sha,conclusion}'` → `head_sha` = `102ff644d411e79171edf0acb0797ba9636162e9`, `conclusion: success` — the checks below are pinned to this exact SHA, not inherited
  - `git grep -n "Claimed scope\|Re-claimed scope\|released:" <sha>` → 4 hits, all production `println!` (basis of P1-2)
  - `git grep -n "write_claim" <sha> -- crates` and `git grep -n "cmd_bridge::claim\|Commands::Claim\|BridgeClaudeCmd::Claim" <sha>` → caller census for the autoclaim check
  - `git grep -n "EDDA_SESSION_ID" <sha> -- crates/edda-cli/src` → env-guard audit
  - `git grep -n "edda claim" <sha>` → shipped-doc and doctest-fence census
  - `git show <sha>:<path>` for `cmd_bridge.rs`, `main.rs`, `test_support.rs`, `peers/board.rs`, `peers/render_coord.rs`, `tests/skill_doctest.rs`, `skills/coord-handoff.md`, `docs/reference/cli.md`; plus `git show <base>:crates/edda-cli/src/cmd_bridge.rs` to separate added output from pre-existing output
- **READ**: the PR body's L0 receipt at this content (`cargo fmt --all --check` PASS; `cargo clippy -p edda --all-targets -- -D warnings` PASS; `cargo test -p edda` 311 unit + 10 integration, 0 failed). Exact-head CI, all 7 green at `102ff644`: Format; Clippy on ubuntu/macos/windows; Test on ubuntu/macos/**windows**. `edda` is inside the Windows test subset (GH-433), so the changed crate's tests ran on all three platforms — including `skill_doctest`, which is what clears the coord-handoff fences.
- **No local Cargo run, deliberately.** Nothing above needs execution: every finding is a property of the `println!` sequence and the caller graph, both fixed at this SHA, and CI covers the changed crate's tests on all three OSes. A full local rerun would have had no reason beyond "to be safe". Note also that this seat is worktree-isolated on a different branch, so a local `--manifest-path` build would compile a different tree and could not have been offered as evidence about this SHA in any case.
- **Lane**: n/a — built nothing.
- **Cost**: 1 review session, ~20 tool calls, no compile.

### Verdict

**Changes Requested** — P0=0, P1=3.

Product behaviour is sound: the fold is unchanged, the read cannot fail a claim, the automatic path is genuinely untouched, and the argument for disclosure over rejection or merge survives being checked against `resolve_session_id` and the board. What blocks is that the one thing this PR adds — four strings — is unproven (P1-2), misdescribed in the body (P1-1), and wrong in the idempotent case the PR itself leans on (P1-3).

Gates for the next push: `cargo fmt --all --check`; `cargo clippy -p edda --all-targets -- -D warnings`; `cargo test -p edda` (add `--test skill_doctest` coverage if the assertion lands in `coord-handoff.md`); plus exact-head CI. No workspace gate is warranted — the change is confined to `edda-cli` and CI runs `edda`'s tests on all three platforms.

fagemx:
## Review Response: Round 1

New full SHA: `85092dedac323a3f49bb80854f8fdcf15ccb1001`
Reviewed SHA: `102ff644d411e79171edf0acb0797ba9636162e9`
Blocking answered: P1=3 (all fixed)

All three were right, and P1-3 is a real defect in the shipped behaviour rather than a documentation problem.

---

### P1-3 — `released:` was false in the idempotent re-claim case — FIXED

This is the one that mattered. The header/`released` pair was chosen on label alone; nothing compared old paths to new, so re-claiming identical arguments printed

```
Re-claimed scope: api (previous paths replaced)
  released: src/api/*
```

while `src/api/*` was still claimed. Your point that it is routine rather than exotic is the part I had not thought through — tier 4's deterministic `cli-{label}` plus non-expiring board claims means a bare-shell restart re-running its own command lands there every time. And it undercut the PR's own argument: "re-claims idempotently after a restart" is one of the two reasons I gave for rejecting the reject-option, and my implementation lied in exactly that case.

`released` is now the previous paths the new claim no longer covers:

| Case | Header | `released:` |
|---|---|---|
| identical re-claim | `Re-claimed scope: api (unchanged)` | *absent* |
| relabel keeping every path | `Claimed scope: identity (replaces … on auth)` | *absent* |
| narrowing | `Re-claimed scope: api (previous paths replaced)` | only the path given up |
| widening | `Re-claimed scope: api (previous paths replaced)` | *absent* |

### P1-2 — nothing asserted the new strings — FIXED

Your grep is the finding: four hits, all of them the production `println!` sites. The three tests read `compute_board_state`, which is the fold — pre-existing, covered ~40 times over in `edda-bridge-claude`, and green whether or not a line is printed.

Two changes rather than one, because a doctest alone could not have caught P1-3:

1. **`claim_disclosure` returns the lines instead of printing them.** The doctest harness asserts *contains*, so it cannot assert that a `released:` line is **absent** — and absence is precisely what the identical-re-claim case is about. Six unit tests now assert the exact line vector for first claim, relabel, identical re-claim, narrowing, widening, and a relabel that keeps its paths.
2. **An `edda-doctest` block in `coord-sync.md`** runs all four cases against the built binary, so the wiring is covered end to end and not just the pure function. You are right that this is what the harness exists for; `coord-sync.md` is where a session is taught to claim, so that is where it went.

Proved the doctest is live rather than trusting the green — mutating one expectation:

```
coord-sync.md block 0: output did not contain "Re-claimed scope: api (MUTANT)":
Re-claimed scope: api (unchanged)
  paths: src/api/*
  session: s1
```

### P1-1 — the body's transcript was not the binary's output — FIXED

Confirmed. `  paths:` fires unconditionally because `--paths` is a `Vec<String>` with no delimiter, and `  session:` is unconditional. Both pre-existing, both omitted from what I presented as a captured run. The `session:` line is the one `coord-handoff.md` teaches readers to read the release id from, so dropping it was the worst line to drop.

The body now carries the full five-case transcript, captured from a binary built at this content. That mutation output above is independent corroboration of the same thing — it shows all three lines.

---

### Your follow-up, and one correction to it

**The disclosure is caller-only.** Accepted and recorded in the PR body under "What this does not do", crediting the route to #503. Your trace through `render_coord.rs:136-170` is the part I would not have found — the victim's injected context renders the current board with no notice that its claim changed, so the trace this PR leaves is visible to everyone *except* the session that lost the scope. That is a real limit on what this PR delivers and it belongs on the identity issue.

**`docs/reference/cli.md:194-211`** — verified, and it is worse than "omits the fold": the comma-separated `--paths "a,b"` example does not parse that way at all. Not fixed here, since it is a different file and a different claim shape; filed rather than folded in.

### Verified-sound findings

Recording that you checked these rather than assuming them, because two of them were load-bearing for my design argument and I had asserted them without proof: `compute_board_state` infallible so a corrupt board still lets the claim succeed; `autoclaim` reaching `write_claim` directly with only `main.rs:1062` and `cmd_bridge.rs:306` entering the verb; the narrowing/restart argument checking out against `resolve_session_id`; and `fixture()` giving each fence a fresh store so no shipped doctest goes stale.

### Gates on the new SHA

- RAN `cargo test -p edda` — **317 unit + 10 integration, 0 failed** (311 → 317 is the six disclosure tests)
- RAN the five-case transcript end to end from a binary built at this content
- RAN the doctest mutation above
- READ exact-head CI was 7/7 green on `102ff64`; the new SHA's run is pending
- Receipt: L0 focused gates above; `edda` is inside the CI Windows test subset, so exact-head CI covers all three platforms

Requesting Round 2 on `85092dedac323a3f49bb80854f8fdcf15ccb1001`.

fagemx:
## Code Review: Round 2

Full SHA: `85092dedac323a3f49bb80854f8fdcf15ccb1001`
Base full SHA: `a5e4a8f7a63b29dc5aff16d542d496ec5172d601` (= current `origin/main`; no base drift since Round 1)
IN SCOPE: `claim_disclosure` and every string it can emit, across the whole input space rather than the six tested points; the header wording per branch; whether the six unit tests can fail; the `coord-sync.md` doctest's liveness, placement and carrier-neutrality; the PR body's transcript re-derived from the code; and behaviour-neutrality of moving the printing out of `claim`.
FOLLOW-UP ISSUE: #503 (bare-CLI relabel does not fold — session identity), #505 (`docs/reference/cli.md`), F3 (pre-existing carrier-specific example in `coord-sync.md`). None extends this PR.
Blocking: **P0=0, P1=0**

---

### Round 1 findings, re-checked

**P1-3 — `released:` false on idempotent re-claim — fixed, and correct across the whole space.** `released` is now `previous.paths \ new.paths` by exact string equality, with a separate `gained` flag driving the header (`cmd_bridge.rs:679-687`). Enumerating every branch rather than the six tested points:

| previous | new | `released` | header | `released:` line |
|---|---|---|---|---|
| `None` | any | — | `Claimed scope: L` | absent |
| label differs | any | `P \ N` | `Claimed scope: L (replaces … on <old>)` | iff `P \ N ≠ ∅` |
| same label, same set | | `∅`, `gained=false` | `(unchanged)` | absent |
| same label, narrowed | | `≠∅`, `gained=false` | `(previous paths replaced)` | `P \ N` |
| same label, widened | | `∅`, `gained=true` | `(previous paths replaced)` | absent |
| same label, mixed add+drop | | `≠∅`, `gained=true` | `(previous paths replaced)` | `P \ N` |
| same label, `P = ∅` | `N ≠ ∅` | `∅`, `gained=true` | `(previous paths replaced)` | absent |
| same label, `P ≠ ∅` | `N = ∅` | `= P` | `(previous paths replaced)` | `P` — correct, all of it was given up |
| same label, both `∅` | | `∅`, `gained=false` | `(unchanged)` | absent |

The `released:` line is exactly `P \ N` in every reachable branch. **No input names a path the new claim still holds, and none omits a path genuinely given up.** The specific probes asked for:

- **Duplicates.** The filter is per-element against `paths.contains`, so `["a","a","b"] → ["a"]` releases only `b`, and `["a"] → ["a","a"]` is `(unchanged)`. Order is irrelevant too: `["a","b"] → ["b","a"]` is `(unchanged)`, which is right — a claim path list is a set. (One cosmetic wart, N4.)
- **Zero previous paths.** `released = ∅`, `gained = true` → no `released:` line. Truthful (N2 covers the header adjective).
- **Zero new paths** (`edda claim "api"` with no `--paths`). Both `Claim` variants declare `paths: Vec<String>` under a bare `#[arg(long)]` with no `required` and no `value_delimiter` (`main.rs:160-170` top-level shortcut, `main.rs:709-719` under `bridge claude`), so the bare form parses. Output is then `Re-claimed scope: api (previous paths replaced)` / `  released: src/api/*` / `  session: s1` — the pre-existing `if !paths.is_empty()` guard suppresses the `paths:` line, and reporting the full previous set as released is correct, because the session now claims nothing. This is the one case I traced rather than executed; see Evidence.
- **Previous a superset with the label also changed.** `("auth", [a,b]) → ("api", [a])` gives the relabel header plus `released: b` only. The still-held `a` is not named. Correct.

**P1-2 — nothing asserted the new strings — fixed, and the tests are genuinely mutation-sensitive.** Round 1's grep returned 4 hits, all production `println!`. At this SHA the same grep returns 22: 7 production, 9 unit-test expectations, 6 doctest assertions. Mutation results below.

**P1-1 — the body transcript was not the binary's output — fixed.** Re-derived rather than accepted; all five cases match, including the two unconditional lines. See below.

---

### Findings

#### The six unit tests all can fail — none is decorative

Applying the mutation standard, against `cmd_bridge.rs:2359/2367/2379/2393/2406/2419`:

| Mutation | Tests that fail |
|---|---|
| revert `released` to `previous.paths` wholesale | **4 of 6** — `an_identical_re_claim_releases_nothing` (gains a header change *and* a released line), `narrowing_reports_only_the_path_it_gave_up` (`released: src/api/*, src/api/v2/*`), `widening_reports_a_change_but_no_release` (gains a released line), `a_relabel_that_keeps_every_path_releases_nothing` (gains one) |
| delete the `(unchanged)` arm | `an_identical_re_claim_releases_nothing` |
| delete `gained` (leave `else if released.is_empty()`) | `widening_reports_a_change_but_no_release` — would print `(unchanged)` |
| delete the `previous.label != label` arm | `a_new_label_names_the_claim_it_replaced`, `a_relabel_that_keeps_every_path_releases_nothing` |

Both mutations named in the brief kill at least one test. The `assert_eq!` against a whole `Vec<String>` is what makes absence assertable — `an_identical_re_claim_releases_nothing` pins exactly one line, so an extra `released:` cannot slip past. This is a real fix, not a re-skin of Round 1's predecessors.

The two remaining gaps are covered elsewhere rather than untested: the **wiring** (that `claim` passes the right `previous`) and the **read/write ordering** are both pinned by the doctest, not by these six —

| Mutation | Result |
|---|---|
| make `claim` pass `None` unconditionally | doctest step 2 prints `Claimed scope: api`, which does not contain `…(replaces this session's earlier claim on auth)` → **fails** |
| move the board read *after* `write_claim` | step 2's `previous` becomes the just-written `("api", [src/api/*])` → `(unchanged)` → **fails** |

So the ordering that makes the whole feature work is asserted, which is more than I expected to find.

#### The doctest is live, correctly placed, and carrier-neutral

- Fence marker is byte-exact. `cat -A` on `coord-sync.md:57` gives `` ```bash edda-doctest$ `` — column 0, no trailing whitespace — and the harness compares `trimmed == DOCTEST_FENCE` (`skill_doctest.rs:112`, `fenced_blocks` at `:158-166`). `markdown_files` recurses `src/skills`, so the file is scanned.
- Each `>` becomes `ExpectOk`, checked as `got.ok && got.text.contains(want)` (`skill_doctest.rs:~370-390`). The two-space indent on the real `  released: …` line does not defeat `contains`. Every asserted string matches the format literal in `cmd_bridge.rs:690/694/697/701` character for character.
- `fixture()` is per **block**, not per step, so the four commands share one repo and store and accumulate state in order. Reachability holds: step 2 needs step 1's `auth` claim, step 3 needs step 2's `api` claim, step 4 needs step 3's.
- The brief's suspicion about step 3 is right and the implementer handled it. `run()` does `env_remove("EDDA_SESSION_ID")`, so without `--session` tier 4 would mint `cli-<label>` — step 2 would resolve to `cli-api`, never see the `cli-auth` claim, and print the plain first-claim line. `--session s1` is tier 1 (`cmd_bridge.rs:1187-1200`) and is present on all four steps. Strictly only step 2 needs it (steps 3 and 4 share the label `api` and would collide on `cli-api` anyway), but pinning all four is the right call.
- Not flaky: the claims fold is a `HashMap` insert in file order with no timestamp comparison (`board.rs`), so four claims landing in the same second still resolve last-write-wins.
- Carrier-neutral: the new prose and fence use `src/auth/*`, `src/api/*`, `src/api/v2/*` — no crate names, no Cargo, no edda-repo paths, no issue numbers. Correct for a file `include_str!`-shipped by `edda init`.
- No shipped doctest goes stale. The four `coord-handoff.md` fences each open in a fresh fixture as *first* claims, so `Claimed scope: auth` and `> session: cli-auth` are unaffected by the new branches.

Corroboration: the implementer's "10 integration" is exactly the 10 `#[test]`s in `skill_doctest.rs`, so the harness ran; and CI's Test job is green at this SHA on all three OSes.

#### The body's transcript re-derives correctly

Derived from `claim` (`cmd_bridge.rs:714-736`) rather than accepted, for all five cases. `--session sN` is tier 1 so `session_id` is literally `s1`/`s2`; `--paths "x"` yields a one-element vec so `if !paths.is_empty()` always fires; the disclosure lines precede `paths:` which precedes `session:`.

| Case | Derived output | Body |
|---|---|---|
| 1 first claim | `Claimed scope: auth` / `  paths: src/auth/*` / `  session: s1` | matches |
| 2 relabel | `Claimed scope: api (replaces this session's earlier claim on auth)` / `  released: src/auth/*` / `  paths: src/api/*` / `  session: s1` | matches |
| 3 identical | `Re-claimed scope: api (unchanged)` / `  paths: src/api/*` / `  session: s1` | matches |
| 4 narrowing | `Re-claimed scope: api (previous paths replaced)` / `  released: src/api/*` / `  paths: src/api/v2/*` / `  session: s1` | matches |
| 5 second session | `Claimed scope: docs` / `  paths: docs/*` / `  session: s2` | matches |

Both unconditional lines are present in all five, and the ordering is right. P1-1 is closed.

#### Behaviour is unchanged

- `write_claim(&project_id, &session_id, label, paths)` is byte-identical to the base line; only its surroundings changed. Same arguments, same call, same single write.
- Sequence is read → write → print. The read is the only new operation and necessarily precedes the write. Nothing else was reordered, and `session_id` is still resolved once and shared.
- **A corrupt or missing previous claim still succeeds.** `compute_board_state` returns `BoardState::default()` on any `read_to_string` error and `continue`s past an unparseable line, so missing / empty / garbage `coordination.jsonl` ⇒ `replaced == None` ⇒ plain `Claimed scope: {label}`, with `write_claim` still run and `Ok(())` still returned. `claim_disclosure` is pure — no I/O, no indexing, no unwrap — so it cannot panic or fail a claim.
- **Unclaim then claim is correctly a first claim, not a "replaces".** `CoordEventType::Unclaim => claims.remove(&event.session_id)` (`board.rs`), so a released session has no entry and gets the plain line. Worth stating because a naïve "last claim event wins" fold would have printed a false replacement here.
- The automatic path still gains no stdout: the diff touches only `claim`, the new function, and tests. `autoclaim` and the other `write_claim` callers are untouched.

---

### Non-blocking observations

**N1 — the widening header reads as a loss (the judgment the brief asked for).** Adding a path prints `Re-claimed scope: api (previous paths replaced)` with no `released:` line and `paths:` listing both. The header alone suggests the old paths are gone; they are not. I do **not** think this blocks, and the reason matters: Round 1's P1-3 was an affirmative false statement (`released: X` while X was held) with nothing to contradict it. Here the loss indicator — the `released:` line — is correctly absent, and the very next line shows the old path still claimed, so the composite output is honest. It is an imprecise adjective, not a false belief. Still, the fix is three lines and the widening test already pins the string, so if you want it: split on `released.is_empty()` and print `(paths added)` when nothing was let go, keeping `(previous paths replaced)` for the cases that actually released something.

**N2 — same wording, `previous.paths` empty.** Re-claiming with paths after a bare `edda claim "api"` prints `(previous paths replaced)` when there were no previous paths to replace. Subsumed by N1's suggested split.

**N3 — the relabel branch never saying `(unchanged)` is correct.** The label did change, so the claim is genuinely not unchanged, and `replaces this session's earlier claim on auth` is true whatever the path overlap. With `released:` correctly absent when nothing was dropped, `a_relabel_that_keeps_every_path_releases_nothing` documents the right behaviour. No change wanted.

**N4 — cosmetic only.** `["a","b","b"] → ["a"]` renders `released: b, b`, reachable only via `--paths b --paths b`. Separately, matching is exact string equality with no normalisation, so re-claiming `src/api/*` as `src/api/**` (or, on Windows, after `src\api\*`) reports the old literal released although glob coverage is unchanged — `check_offlimits` (`dispatch/tools.rs:204+`) *does* normalise `\`→`/` before glob-matching. I judge this right as built: the board, `edda peers`, and the claim entries themselves all live at the literal-string layer, and resolving it properly would need glob-subsumption analysis. Not worth chasing.

**N5 — `a_second_claim_replaces_the_first_and_says_so` (`cmd_bridge.rs:2429`) still asserts only the fold**, nothing about what is said. Harmless now that the saying is asserted six ways over, but the name promises an assertion the body does not make — the exact shape Round 1 flagged. A rename to `…_and_the_board_keeps_one_claim` would stop it reading as coverage it does not provide.

**N6 —** the shipped transcript introduces `--session s1` without saying why. It is honest and necessary (it is what makes the example reproducible and "per session" concrete), but a half-sentence noting the flag pins the session would keep a reader from thinking it is required.

---

### FOLLOW-UP ISSUE

- **#503 (already filed) — the bare-CLI relabel does not fold at all.** Tier 4 mints `cli-{label}`, so `edda claim "auth" …` followed by `edda claim "api" …` in a bare shell are two *sessions*: two board entries, no replacement, and therefore no disclosure. The new prose does not overclaim — it says "per session" twice and the doctest pins `--session` — but the reader most likely to meet the gap is the bare-CLI one this skill is partly written for. Belongs with session identity, not here.
- **#505 (already filed)** — `docs/reference/cli.md`.
- **F3 (new, trivial)** — `coord-sync.md`'s pre-existing guideline line offers `crates/edda-store/*` as the glob example, in a file shipped into every project by `edda init`. Same carrier-neutrality standard the new text meets; pre-existing and outside these hunks.

---

### Evidence

- **RAN** (read-only, no Cargo gate):
  - `gh pr view 500 --json headRefOid,baseRefOid` → head `85092dedac323a3f49bb80854f8fdcf15ccb1001`, base `a5e4a8f7a63b29dc5aff16d542d496ec5172d601`; `git rev-parse origin/main` → same base, so no drift
  - `gh pr checks 500` → 7/7 SUCCESS; `gh api repos/fagemx/edda/actions/runs/32370307952` → `head_sha` = `85092dedac323a3f49bb80854f8fdcf15ccb1001`, `conclusion: success`. Verified pinned to the frozen SHA, not inherited
  - `git diff <base>...<head>` → 2 files, +239/−1
  - `git grep -n "Claimed scope\|Re-claimed scope\|released:" <head>` → 22 hits (4 at Round 1): 7 production, 9 unit-test expectations, 6 doctest assertions
  - `git show <head>:<path>` for `cmd_bridge.rs`, `main.rs`, `skills/coord-sync.md`, `skills/coord-handoff.md`, `tests/skill_doctest.rs`, `edda-bridge-claude/src/peers/board.rs`, `edda-bridge-claude/src/dispatch/tools.rs`
  - `git show <head>:crates/edda-cli/src/skills/coord-sync.md | grep -n 'edda-doctest' | cat -A` → `` 57:```bash edda-doctest$ `` (byte-exact marker, no trailing whitespace)
- **READ**: the implementer's L0 receipt at this content — `cargo fmt --all --check` PASS; `cargo clippy -p edda --all-targets -- -D warnings` PASS; `cargo test -p edda` 317 unit + 10 integration, 0 failed (311 → 317 is the six disclosure tests; the 10 integration are exactly the 10 `#[test]`s in `skill_doctest.rs`). Exact-head CI green on all 7: Format; Clippy ubuntu/macos/windows; Test ubuntu/macos/**windows**. `edda` is inside the CI Windows test subset (GH-433), so the changed crate's tests — including `skill_doctest`, which is what executes the new fence — ran on all three platforms at this SHA.
- **Not run, deliberately.** `claim_disclosure` is a pure function over two string slices; I traced every branch by hand and the table above is a complete enumeration, so a build would confirm a trace rather than test a suspicion. The one thing reading leaves slightly short is clap's acceptance of a bare `edda claim "x"` with no `--paths`; that case is non-blocking either way (if clap accepts it the output is correct per the table; if not, the case is unreachable), and both the `if !paths.is_empty()` guard in `claim` and `unwrap_or_default()` in the board fold are written for it. Spending a `-p edda` build to settle a non-blocking detail is the over-verification this repo's ladder exists to remove.
- **Lane**: n/a — built nothing.
- **Cost**: 1 review session, 15 tool calls, no compile.

---

### Verdict

**LGTM** — P0=0, P1=0.

All three Round 1 blockers are genuinely closed, and the fixes are better than the minimum each required. P1-3's correction generalises: `previous.paths \ new.paths` is right across the entire input space I could enumerate, including the zero-path, duplicate, and superset-with-relabel corners none of the tests reach. P1-2's fix is the one that changes the shape of the change — returning `Vec<String>` instead of printing makes *absence* assertable, four of the six tests die to the single most likely regression, and the doctest independently pins both the wiring and the read-before-write ordering that the unit tests cannot see. P1-1's transcript now re-derives line for line.

Nothing found here is blocking. N1 is the one I would still take if you want it — the widening header is imprecise in a PR about not being imprecise — but it is an adjective contradicted by the adjacent line, not a false claim, and it is not worth another round on its own. N5's rename is free if you are touching the file anyway.

Gates: `cargo fmt --all --check` PASS; `cargo clippy -p edda --all-targets -- -D warnings` PASS; `cargo test -p edda` 317 unit + 10 integration, 0 failed — all READ from the implementer's receipt and independently corroborated by exact-head CI, 7/7 green with `head_sha` verified equal to `85092dedac323a3f49bb80854f8fdcf15ccb1001`. No workspace gate warranted: the change is confined to `edda-cli`, and CI runs that crate's tests on all three platforms.

fagemx:
## Note on Round 2 — the two non-blocking items, routed

Not a new SHA; the LGTM stands.

**N1 (`(previous paths replaced)` on a pure widening) and the `a_second_claim_replaces_the_first_and_says_so` naming are filed as #508** rather than pushed. Both are real, both are three-line fixes, and both are the same defect class as the bug you caught in Round 1 — a string promising something the code did not do. I took your judgement that they are not blocking: `released` is correctly absent in the widening case, so the loss indicator is right and only the adjective is off, and the disclosure now has its own tests so the misnamed one is no longer a coverage gap.

The reason to route rather than fix is the ladder this PR series exists to enforce: a push invalidates the verdict and costs a full extra round, and spending one on an adjective at LGTM is diminishing returns by the same rule that says a coverage gap earns a focused check rather than a full rerun. The issue carries your suggested split, including the `released`/`gained` truth table, so whoever picks it up does not re-derive it.

**On your branch-space enumeration.** You checked cases I did not: duplicates in either list, zero *new* paths (`edda claim "x"` with no `--paths`, which clap does permit and I had not considered), reordered lists, superset-with-relabel. My six tests were six points; you established the property over the space. That is the difference between the tests passing and the behaviour being right, and it is the check I should have run before claiming the case was closed.

**On the mutation counts.** Reverting `released` to `previous.paths` killing 4 of 6, deleting the `(unchanged)` arm killing one, `gained` killing one, the relabel arm killing two — that is the standard this repo now applies to its own tests (#497), and I did not apply it here. I asserted the tests were load-bearing; you established it.

**The two things the doctest pins that the unit tests cannot** are the most useful finding in the round, and I did not know they were there when I wrote it. Passing `None` unconditionally fails step 2, so the **wiring** is covered; and moving the board read after `write_claim` makes step 2 print `(unchanged)`, so the **read-before-write ordering** is covered. I added the doctest to satisfy P1-2's "assert the strings" — it turns out to be the only thing standing between a future refactor and a silent inversion of that ordering. Worth saying out loud so the next person to touch `claim` does not read it as redundant with the unit tests.

Round 1's grep returning 4 hits and now returning 22 is the cleanest one-line summary of what changed here.

PR #504, opened 2026-08-20
test(edda-cli): make the sorted claim true, and name the scan's second cost

### LINKED ISSUE #497: test(edda-cli): doctest scanner test asserts an untested guarantee, and its stated cost is incomplete

fagemx:
Closes #497.

## Item 1 — the `sorted` claim was asserted, not established

`skills_in_subdirectories_are_scanned` asserted `"sorted, recursive, .md only"`. The #495 review mutated each property and found the first one uncovered:

| Mutation | Before |
|---|---|
| `markdown_files` made non-recursive | fails — covered |
| `.md` extension filter dropped | fails — covered |
| `found.sort()` deleted | **passes — not covered** |

The old fixture was `top.md` at the root plus a buried `buried.md`. NTFS returns directory entries in name order, so the walk already emitted sorted paths and the sort could not be observed at all.

**The new fixture makes the three orders differ.** NTFS collates case-insensitively (`apple.md` < `group` < `Zebra.md`) while `Path` compares bytes, so the sorted answer leads with `Zebra.md` and an unsorted walk reads it last. The files are also written in a third order, so a filesystem that returns entries as created is caught too. The assertion compares full relative paths rather than bare file names, so recursion is visible in the expected value instead of being flattened away.

Mutation-tested on a Windows workstation — **3/3 caught**, where the review measured 2/3:

```
found.sort() removed  ->  ["apple.md", "group/deeper/buried.md", "Zebra.md"]
recursion removed     ->  ["Zebra.md", "apple.md"]
.md filter removed    ->  ["Zebra.md", "apple.md", "group/deeper/buried.md", "group/deeper/ignored.txt"]
```

Each is the real failure output, not a prediction.

## Item 2 — the scanner's stated cost named one hazard when there are two

The module header and `fenced_blocks`' rustdoc said the bias toward reading every fence-looking line as a fence costs "a false alarm on an indented literal example". There is a second cost, and it runs the other way:

A stray unmatched marker opens a span. The next marker is the *opening* line of a real doctest — which cannot close that span, because a closing fence carries no info string. So the doctest becomes ordinary content of the stray block and **never runs**. The swallowed opener then sets the quoted-illustration exemption inside `lines_outside_quoted_illustrations`, and nothing after it is a fence marker to clear it, so the demotion guard reports nothing either.

Silence, not a false alarm — which makes it the more serious of the two for a harness whose whole purpose is to prevent silence.

`a_stray_fence_marker_swallows_the_next_doctest` pins it as current behaviour, with a control asserting the same block runs once the stray marker is removed, so the test is about the stray marker rather than a malformed block.

**Not fixed here.** The mechanism is the unterminated-quoted-marker case recorded as follow-up 1 in #495's Round 1 comment, which belongs with #494's exemption-scope decision. Pinning it means a change to either has to face it.

## Evidence

- RAN `cargo test -p edda` — **308 unit + 11 integration, 0 failed**
- RAN the three mutations above, each restored from a saved copy before the next
- Receipt: L0 focused gates above. `edda` is inside the CI Windows test subset (GH-433), so exact-head CI covers this on all three platforms

Test-only change; no shipped behaviour is altered.

### DIFF
--- a/crates/edda-cli/tests/skill_doctest.rs
+++ b/crates/edda-cli/tests/skill_doctest.rs
@@ -35,6 +35,13 @@
 //! The scan errs toward reading a fence-looking line as a fence, and toward
 //! reporting. A false alarm is cheap; a skill that quietly stopped being
 //! checked is the failure this file exists to prevent.
+//!
+//! That bias is not free in only one direction. A stray unmatched marker opens
+//! a span that the *opening* line of a following doctest cannot close, because
+//! a closing fence carries no info string -- so the doctest is swallowed as
+//! ordinary content. It does not run, and the quoted-illustration exemption
+//! then hides it from the demotion guard as well. Silence, not a false alarm.
+//! `a_stray_fence_marker_swallows_the_next_doctest` pins that gap (GH-497).
 
 use std::path::{Path, PathBuf};
 use std::process::Command;
@@ -137,8 +144,10 @@ fn fence_marker(trimmed: &str) -> Option<(char, usize)> {
 /// reported, which is the silence this harness exists to prevent (GH-495).
 ///
 /// Reading every fence-looking line as a fence errs toward running and
-/// reporting. That is the safe direction here: the cost is a false alarm on an
-/// indented literal example, against a silently unchecked skill.
+/// reporting. That is the safe direction here, but it costs two things rather
+/// than one: a false alarm on an indented literal example, and -- quieter -- a
+/// stray unmatched marker that swallows a following doctest into its span,
+/// where the exemption then hides it from the guard too (GH-497).
 fn fenced_blocks(markdown: &str) -> Vec<Fence<'_>> {
     let mut blocks = Vec::new();
     let mut open: Option<(char, usize, Fence)> = None;
@@ -563,21 +572,72 @@ fn skills_in_subdirectories_are_scanned() {
     // The scan was one directory deep, so a skill in a subdirectory was never
     // checked (GH-492 item 2). Nothing in the shipped tree exercises this yet,
     // so the walk needs its own fixture.
+    // The names are mixed-case on purpose, and the writes are out of order on
+    // purpose. The old fixture (`top.md` plus a buried `buried.md`) could not
+    // fail: NTFS hands back entries in name order, so the walk already emitted
+    // sorted paths and deleting the sort changed nothing observable — the
+    // assertion claimed `sorted` without establishing it (GH-497 item 1).
+    //
+    // Here the three orders are all different. NTFS collates case-insensitively
+    // (`apple.md` < `group` < `Zebra.md`) while `Path` compares bytes, so the
+    // sorted answer puts `Zebra.md` first — an unsorted walk reads it last. A
+    // filesystem that returns entries as created is caught by the write order.
     let dir = tempfile::tempdir().expect("tempdir");
     let nested = dir.path().join("group").join("deeper");
     std::fs::create_dir_all(&nested).expect("nested dirs");
-    std::fs::write(dir.path().join("top.md"), "# top\n").expect("top");
     std::fs::write(nested.join("buried.md"), "# buried\n").expect("buried");
+    std::fs::write(dir.path().join("Zebra.md"), "# zebra\n").expect("zebra");
+    std::fs::write(dir.path().join("apple.md"), "# apple\n").expect("apple");
     std::fs::write(nested.join("ignored.txt"), "not markdown\n").expect("txt");
 
     let found: Vec<String> = markdown_files(dir.path())
         .iter()
-        .map(|p| p.file_name().unwrap().to_string_lossy().into_owned())
+        .map(|p| {
+            p.strip_prefix(dir.path())
+                .expect("under the fixture root")
+                .to_string_lossy()
+                .replace('\\', "/")
+        })
         .collect();
     assert_eq!(
         found,
-        vec!["buried.md", "top.md"],
-        "sorted, recursive, .md only"
+        vec!["Zebra.md", "apple.md", "group/deeper/buried.md"],
+        "recursive, .md only, and sorted rather than in directory order"
+    );
+}
+
+#[test]
+fn a_stray_fence_marker_swallows_the_next_doctest() {
+    // The second cost of reading every fence-looking line as a fence, and the
+    // quiet one. An unmatched marker opens a span; the next marker is the
+    // *opening* of a real doctest, which cannot close that span because a
+    // closing fence carries no info string. So the doctest becomes content of
+    // the stray block: it never runs. The demotion guard would normally catch
+    // an unmarked block carrying assertions, but the swallowed opener sets the
+    // quoted-illustration exemption, and nothing after it is a fence marker to
+    // clear the exemption — so the guard reports nothing either.
+    //
+    // Asserted as the current behaviour rather than fixed here: the mechanism
+    // is the unterminated-quoted-marker case from #495 Round 1, which belongs
+    // with #494's exemption-scope decision. Pinned so a change to either has to
+    // face it (GH-497 item 2).
+    let stray = "prose\n```\n\n```bash edda-doctest\n$ edda peers\n> ok\n```\n";
+    assert!(
+        parse_blocks(stray).is_empty(),
+        "known gap: the stray marker consumes the doctest's opening fence"
+    );
+    assert!(
+        find_demoted_steps(stray).is_empty(),
+        "and the exemption hides the stranded step from the guard"
+    );
+
+    // Remove the stray marker and the same block is a live doctest, so the
+    // fixture is testing the stray marker and not a malformed block.
+    let clean = "prose\n\n```bash edda-doctest\n$ edda peers\n> ok\n```\n";
+    assert_eq!(
+        parse_blocks(clean).len(),
+        1,
+        "without the stray marker the block runs"
     );
 }
 


fagemx:
## Code Review: Round 1

Full SHA: `0efb1f905e34f22bdeb811762633453f79b60c9f`
IN SCOPE: the whole diff — one file, `crates/edda-cli/tests/skill_doctest.rs` (+66/-6): the rewritten `skills_in_subdirectories_are_scanned` fixture, the new `a_stray_fence_marker_swallows_the_next_doctest`, the module-header paragraph and the `fenced_blocks` rustdoc; #497's two items as the acceptance ceiling; test-poison scan against base; and the PR body's factual claims.
FOLLOW-UP ISSUE: none
Blocking: P0=0, P1=0

### Findings

#### Item 1 — the `sorted` claim is now established, not asserted. Verified by re-running all three mutations.

I did not take the PR body's mutation table on trust. I copied `markdown_files` verbatim out of this SHA into a standalone program, built the fixture in exactly the order the test builds it, and ran the three mutations plus the unmutated baseline on this Windows workstation. Every claimed output reproduces byte-for-byte:

```
raw read_dir order at root : ["apple.md", "group", "Zebra.md"]
head (unmutated)           : ["Zebra.md", "apple.md", "group/deeper/buried.md"]  matches assertion
A found.sort() removed     : ["apple.md", "group/deeper/buried.md", "Zebra.md"]                          CAUGHT
B recursion removed        : ["Zebra.md", "apple.md"]                                                    CAUGHT
C .md filter removed       : ["Zebra.md", "apple.md", "group/deeper/buried.md", "group/deeper/ignored.txt"] CAUGHT
```

3/3, and the three strings match the PR body exactly. I also rebuilt the **base** fixture (`top.md` + buried `buried.md`, bare-file-name assertion) and confirmed the defect being fixed is real rather than inherited from the #495 write-up: `found.sort()` removed → `["buried.md", "top.md"]`, identical to the unmutated result, **CAUGHT=false**. So the old assertion genuinely claimed `sorted` without establishing it, and the new one establishes it.

The two premises the fixture rests on both hold:

- **NTFS case-insensitive collation is real, and load-bearing.** `read_dir` at the fixture root returned `["apple.md", "group", "Zebra.md"]` — uppercase-collated order, which is neither the sorted order nor the write order. This is what gives mutation A somewhere to fall.
- **`Path`'s `Ord` is component-wise, not whole-string bytes — but the expected value holds under the real semantics.** I probed it: `group.md` vs `group/deeper/buried.md` gives `Path=Greater` where raw string comparison gives `Less` (the separator boundary is where the two diverge). None of the fixture's three comparisons hit that: `Zebra.md` < `apple.md` < `group/deeper/buried.md` under both, because each is decided inside the first component on `Z`(0x5A) < `a`(0x61) < `g`(0x67). Sorting the three relative paths directly yields `["Zebra.md", "apple.md", "group/deeper/buried.md"]`, and spelling the nested path with `\` instead of `/` does not change any comparison — so the expected value is separator-independent and identical on a `/` platform.

**Cross-platform discriminating power (non-blocking, no action needed).** The brief asked whether the fixture also catches a deleted sort where `read_dir` is neither name- nor creation-ordered. Enumerating all six permutations of the three root entries, the walk output differs from the expected value in five; the single escape is the permutation that is *already* the sorted order. That escape is not a fixture-design flaw — no choice of names can catch a sort deletion on a filesystem that hands back sorted entries. Both *deterministic* ordering classes are provably caught: name-order (NTFS, HFS+) by the mixed case, creation-order (tmpfs, small linear ext4 dirs) by the deliberate write order — `group` is created first by `create_dir_all`, before `Zebra.md` and `apple.md`. Residual risk is confined to hash-ordered directories (ext4 htree, APFS), where the order is arbitrary but *deterministic per name set* — so this is a coverage question, never a flake: with the sort intact the output is order-independent and the test passes everywhere. Confirmed on CI at this SHA: `skills_in_subdirectories_are_scanned ... ok` on ubuntu, macOS and Windows.

`strip_prefix` + `replace('\\', "/")` hides nothing that matters. `strip_prefix` is component-based and `.expect("under the fixture root")` panics loudly on a path outside the root rather than silently normalising it, and every path is built as `dir.path().join(...)` by the walk itself, so a mixed or absolute separator cannot be laundered into the expected string. The assertion is also strictly stronger than base: full relative paths instead of bare file names, so recursion is now visible *in the expected value* rather than flattened away.

#### Item 2 — the described mechanism is the real one. Verified by instrumented trace, including the counterfactual.

The risk here was that both results come back empty for some other reason, making the new prose a second unestablished claim. They do not. Copying `fenced_blocks` / `lines_outside_quoted_illustrations` / `find_demoted_steps` / `parse_blocks` verbatim and dumping the intermediate state for the `stray` fixture:

```
input lines: ["prose", "```", "", "```bash edda-doctest", "$ edda peers", "> ok", "```"]
  fence 0: is_doctest=false lines=["", "```bash edda-doctest", "$ edda peers", "> ok"]
    after exemption: [""]  has_assertions=false
  parse_blocks len   = 0
  find_demoted_steps = []
  ^ same guard, exemption disabled = ["$ edda peers"]
```

Every link in the stated chain is visible:

1. The doctest's **opening** line is a *content* line of the stray block — it fails `closes` because `trimmed[run..]` is `bash edda-doctest`, not empty, which is precisely "a closing fence carries no info string".
2. `parse_blocks` is empty because the only fence has `is_doctest=false` — the doctest was swallowed, not malformed.
3. The guard is empty because the swallowed opener equals `DOCTEST_FENCE`, sets `quoting`, and no later line is a fence marker to clear it — leaving `[""]`, hence `has_assertions=false`.
4. The counterfactual proves the exemption is load-bearing rather than incidental: with that one step removed, the same guard reports `["$ edda peers"]`.

**The control is a genuine control.** I checked it mechanically rather than by eye: deleting the single bare ` ``` ` line from `stray` yields `clean` exactly (`stray minus its first bare ``` line equals clean? true`), 7 lines vs 6. And `clean` produces `fence 0: is_doctest=true`, `parse_blocks len = 1`. So the fixture isolates the stray marker and nothing else.

**On pinning known-wrong behaviour.** I think it is right here, and not a poison. The escape hatches a future fixer needs are all present at the failure site or one line above it: the test *name* says "swallows" (a defect description, not a guarantee), the doc comment says "Asserted as the current behaviour rather than fixed here … Pinned so a change to either has to face it (GH-497 item 2)", and the first assertion's message opens with "known gap:". Deferring the fix to #494's exemption-scope decision is the correct call — repairing the exemption inside a test-hardening PR would be exactly the scope creep the review loop keeps paying for.

#### Test-poison scan — clean

Nothing weakened or deleted relative to base. The only removals are `top.md` (replaced by two better-chosen names) and the bare-file-name mapping (replaced by full relative paths). The assertion message went from `"sorted, recursive, .md only"` to `"recursive, .md only, and sorted rather than in directory order"` — narrower and now true. No test was dropped; the file goes from 10 to 11 `#[test]` functions, matching the "11 integration" figure in the body.

#### Header / rustdoc / test consistency — consistent

The module header and the `fenced_blocks` rustdoc name the same two costs in the same order and neither contradicts the other or the tests. The header's forward reference to `a_stray_fence_marker_swallows_the_next_doctest` resolves to a test that exists and pins what the header says it pins.

#### Carrier neutrality — n/a, and confirmed

`git diff --name-status` between base and head lists exactly one path, `crates/edda-cli/tests/skill_doctest.rs`. No `src/skills/*.md` changed, so nothing shipped by `edda init` moved.

#### PR body — every factual claim re-derived, none overstated

Checked because a sibling PR this week shipped a table crediting itself with another PR's change. The three mutation outputs are exact (reproduced above). The "Before" table re-derives against the base fixture (sort uncovered, recursion and filter covered). "308 unit + 11 integration, 0 failed" matches the Windows CI log at this SHA line for line. "`edda` is inside the CI Windows test subset (GH-433)" is correct — `.github/workflows/ci.yml:76` lists `-p edda` in the `windows-latest` args, so this file does run on all three platforms. The body's claims are scoped honestly: it says the mutations were run "on a Windows workstation" and does not extend the 3/3 result to Linux or macOS.

### Non-blocking nits (take or leave; none warrant a follow-up issue)

1. `skill_doctest.rs` fixture comment says "`Path` compares bytes". It compares *components*, and only compares bytes within one — the difference is observable across a separator boundary (`group.md` vs `group/deeper/buried.md` reverses). Not load-bearing for this fixture, and I verified that; but given this PR is specifically about not asserting what you have not established, "compares components, so `Z` sorts before `a`" would cost one word and be exactly true.
2. The second assertion's message, "and the exemption hides the stranded step from the guard", does not carry the "known gap:" prefix the first one does. If the guard is fixed before the parser, that assertion fires alone and the reader sees only that message — which reads as an expectation. Repeating the marker would make the pin unambiguous at the failure site, where the doc comment is not visible.
3. "A stray *unmatched* marker" — in the fixture the stray marker's span *is* closed, by the doctest's closing fence. What is unmatched is the pairing, shifted by one. The substance is right; the word is doing slightly more work than it can.
4. `find_demoted_steps(stray).is_empty()` has no positive control inside this test (the control exercises `parse_blocks` only), so it would also pass if the guard were wholly broken. `a_demoted_fence_is_an_error_rather_than_silence` covers that direction elsewhere in the file, so this is an observation, not a hole.

### Evidence

- **RAN** (focused, on Windows 11 / NTFS, this workstation):
  - `rustc --edition 2021 -O mutcheck.rs` then ran it — `markdown_files` copied verbatim from this SHA, fixture built in the test's write order, three mutations + baseline + the base fixture. Results quoted above: new fixture 3/3 CAUGHT, base fixture sort mutation CAUGHT=false, raw `read_dir` order `["apple.md", "group", "Zebra.md"]`.
  - `rustc --edition 2021 fencecheck.rs` then ran it — `fenced_blocks` / `lines_outside_quoted_illustrations` / `find_demoted_steps` / `parse_blocks` copied verbatim, with additive instrumentation and an exemption-disabled counterfactual. Trace quoted above; control confirmed one line different.
  - `rustc --edition 2021 pathord.rs` then ran it — `Path::cmp` vs raw string comparison on the fixture's three pairs, the `\`-separator spelling, and the `group.md` / `group/…` divergence case.
- **READ** (all at `0efb1f905e34f22bdeb811762633453f79b60c9f`):
  - `gh pr diff 504`; `git diff --name-status a5e4a8f…0efb1f9` — one file.
  - The full `crates/edda-cli/tests/skill_doctest.rs` at head, via `git show`, for the mechanism trace and the poison scan.
  - `.github/workflows/ci.yml:68-76` for the Windows crate subset.
  - `gh api …/commits/0efb1f9…/check-runs` — all 7 checks carry `head_sha = 0efb1f905e34f22bdeb811762633453f79b60c9f`, so no inherited status: Format, Clippy ×3, Test ×3, all `success`. Run `32368963491`, `event: pull_request`, `head_sha` matches.
  - Test job logs for ubuntu / macOS / Windows: `a_stray_fence_marker_swallows_the_next_doctest ... ok` and `skills_in_subdirectories_are_scanned ... ok` on all three; Windows `test result: ok. 308 passed` and `ok. 11 passed`.
  - PR body gate receipt (fmt, `clippy -p edda`, `test -p edda`) — reused, not rerun. Exact-head CI covers all three supersets of it, including `cargo clippy --workspace --all-targets` on Windows.
- **Lane: n/a.** I compiled nothing with Cargo and created no target directory. The `verifier-2` lane was not needed: the load-bearing check is filesystem-ordering behaviour, which three standalone `rustc` invocations in the session scratchpad answer in seconds. A workspace run would have re-proved what exact-head CI already proves on all three platforms.
- **Cost:** ~16 tool calls, single session, no build cache created or consumed.

### Verdict

**LGTM** — P0=0, P1=0.

Gates: exact-head CI at `0efb1f905e34f22bdeb811762633453f79b60c9f`, all 7 green and all pinned to that SHA (Format; Clippy on ubuntu/macOS/Windows; Test on ubuntu/macOS/Windows, with `edda` inside the Windows subset). Plus the three focused local checks above, which is the surface CI cannot cover: CI runs the fixture, it does not run the *mutations* that decide whether the fixture is worth anything.

Both #497 items are met, and met in the strong form the issue asked for — #497 offered "either drop `sorted` from the claim, or give the fixture file names whose creation order differs from their sort order and assert the order explicitly", and this takes the second. The result is a test that establishes what it claims: I independently reproduced 3/3 mutation kills at head and 0/1 at base for the property in question. Item 2's new prose describes the mechanism that actually produces the silence, verified link by link including the counterfactual that shows the exemption is the cause rather than a bystander.

Issue #503, opened 2026-08-20
coord(cli): session identity must come from the process, not the workspace


fagemx:
Successor to **#488 item 1**, the only item of that issue that does not ship. Everything needed is reproduced here; #488 is context, not a dependency.

> **Correction (filing rationale).** This issue's original text said #488 was linked to auto-close from both #496 and #500 and that the link could not be removed. Both claims were wrong — see "Why this is a new issue" at the bottom for what was actually true.

## The defect

`resolve_session_id` (`crates/edda-cli/src/cmd_bridge.rs`) falls back to `infer_session_id`, which answers only when **exactly one live session** exists. That is the case where identity matters least. Worse, it cannot tell the *caller* apart from a *lone peer*: a bare shell standing next to one live session is handed that session's identity, and can then `claim`, `decide`, or `unclaim` under someone else's name.

`infer_session_id`'s rustdoc records the limit as of #496:

> This cannot tell the caller apart from a lone peer: a process has no way to prove a heartbeat is its own (GH-488).

## Why it is worse than a mislabel

Claims are keyed by `session_id` and `compute_board_state` folds to one claim per session. So an adopted identity does not merely mislabel — it **replaces** the real holder's claim and drops its off-limits protection. A bare shell claiming `docs/*` can silently overwrite a live worker's `crates/core/**`.

#500 makes that replacement *visible* (it prints what it replaced), so the event now leaves a trace. It does not prevent it. This issue is the prevention half.

## What was tried and refuted

**Branch matching**, attempted in #496 and withdrawn after review.

The tier matched the caller's current git branch against live sessions' recorded branches. The argument was that this is strictly narrower than counting sessions, so it could remove a misattribution but never create one.

**A branch belongs to a worktree, not a process.** Every process in that worktree matches equally well, so nothing in it separates the caller from a peer. And where the counting tier *stops answering* once two sessions are live, the branch tier answers anyway — the advertised narrowing was in fact a widening.

Demonstrated by the reviewer, by execution: with `sess-a`@`feat/a` and `sess-b`@`feat/b` live, a bare shell on `feat/a` resolved to `sess-a`. The same board with no branch recorded resolved to `cli-probe`, the base behaviour. **The change created a misattribution exactly where the base was safe.**

## The options, narrowed

1. **Hooked sessions export `EDDA_SESSION_ID`** — the only one that addresses the actual gap, because it makes identity a property of the *process*. Each bridge must put the id into the environment the agent's shell inherits.
[...]

Issue #110, opened 2026-02-26
feat: secure device pairing with revocable tokens


fagemx:
## 背景

HTTP server 開放後，需要確保只有授權裝置能存取。不能用傳統 username/password（麻煩），也不能完全不設防。

## 目標

簡單、低摩擦的 device pairing，支援撤銷。

### Pairing Flow

```bash
# 在本機（已信任的環境）
edda pair new --name "iPhone"
# → 顯示 QR code + 一次性 pairing URL
# → https://localhost:7433/pair?token=abc123

# 手機掃 QR code → 完成配對
# → 手機獲得長期 device token（存在 localStorage）
```

### Token 管理

```bash
edda pair list              # 列出所有已配對裝置
edda pair revoke "iPhone"   # 撤銷特定裝置
edda pair revoke-all        # 撤銷所有
```

### 實作

- Device token 存在 ledger（`device_pair` event type）
- 每個 API request 帶 `Authorization: Bearer <device_token>`
- Token 撤銷 = 寫入 `device_revoke` event
- Pairing event 記錄：device_name, paired_at, paired_from_ip
- 所有操作可審計

### 安全考量

- Pairing token 一次性使用（10 分鐘過期）
- Device token 永不過期（除非手動撤銷）
- 本機 localhost 存取不需要 token（向後相容）
- 不做 OAuth / SSO（過度設計）

## 依賴

- HTTP transport

## 相關

- #104 Mobile ops gap

fagemx:
# Phase 3: Implementation Plan — Issue #110 Device Pairing

## Overview

Add secure device pairing with revocable bearer tokens to edda's HTTP server. Remote clients authenticate via `Authorization: Bearer <token>`, localhost clients pass through without auth.

**Estimated scope**: ~6 files modified, ~1 new file, schema v7 migration.

## Step-by-Step Plan

### Step 1: Add `rand` workspace dependency

**File**: `Cargo.toml` (workspace root)
- Add `rand = "0.8"` to `[workspace.dependencies]`

**File**: `crates/edda-serve/Cargo.toml`
- Add `rand.workspace = true`
- Add `sha2.workspace = true`

### Step 2: Schema v7 — `device_tokens` table

**File**: `crates/edda-ledger/src/sqlite_store.rs`

Add `SCHEMA_V7_SQL`:
```sql
CREATE TABLE IF NOT EXISTS device_tokens (
    token_hash      TEXT PRIMARY KEY,
    device_name     TEXT NOT NULL,
[...]

PR #514, opened 2026-08-23
coord(cli): widening a scope reports 'previous paths replaced' when nothing was replaced

### LINKED ISSUE #508: coord(cli): widening a scope reports 'previous paths replaced' when nothing was replaced
Two non-blocking observations from the #500 Round 2 review, recorded rather than pushed — #500 reached LGTM and a wording change would have cost a full extra review round for no behavioural gain.

Both are the same defect class the surrounding work is about: **a string that promises more than it establishes.**

## 1. `(previous paths replaced)` is wrong when paths were only added

`crates/edda-cli/src/cmd_bridge.rs`, `claim_disclosure`.

Widening a scope — re-claiming the same label with every old path plus a new one — takes the `else` arm:

```
$ edda claim "api" --paths "src/api/*" --session s1
$ edda claim "api" --paths "src/api/*" --paths "src/api/v3/*" --session s1
Re-claimed scope: api (previous paths replaced)
  paths: src/api/*, src/api/v3/*
```

Nothing was replaced. `released` is correctly **absent**, so the loss indicator is right and the adjacent `paths:` line contradicts the adjective — which is why the review judged it non-blocking rather than a repeat of the `released:` bug that Round 1 caught. But "previous paths replaced" is still false in this case, in a change whose entire purpose is to stop the verb saying things that are not so.

The branch already computes what it needs. `gained` is true and `released` is empty exactly when paths were only added, so:

| `released` | `gained` | Now | Should be |
|---|---|---|---|
| empty | false | `(unchanged)` | `(unchanged)` |
| empty | true | `(previous paths replaced)` | `(paths added)` |
| non-empty | either | `(previous paths replaced)` | `(previous paths replaced)` |

Three lines, plus a seventh unit test asserting the new string, plus a step in the `coord-sync.md` doctest so the wiring is pinned like the other four.

## 2. `a_second_claim_replaces_the_first_and_says_so` does not check that it says so

Same file, test module.

The name promises the disclosure is asserted. The body reads `compute_board_state().claims` — the per-session fold, which pre-dates #500 and is covered many times over in `edda-bridge-claude`. It would pass with every `println!` deleted.

The six `claim_disclosure` tests added in #500 do assert the strings, so the coverage exists; this test is not the gap it once was. But its name still describes coverage it does not provide, which is exactly the finding #497 fixed in `skills_in_subdirectories_are_scanned` ("sorted, recursive, .md only" when the sort was untested).

Either rename it to what it establishes (`a_second_claim_leaves_one_claim_on_the_board`), or extend it so the name becomes true. Renaming is probably right — the disclosure has its own tests now, and this one earns its place as the board-level check.

## Acceptance

- widening prints a header that does not claim a replacement
- the new header string is asserted by a unit test and by the shipped doctest, not just by one of them
- the board-level test's name matches what its body establishes
- while in there: check no other message in `claim_disclosure` or `unclaim` describes an outcome the code does not produce

Related: #500, #497, #488.

fagemx:
Closes #508

Summary:
- report (paths added) when a same-label claim widens without releasing paths
- pin the header in both the unit test and shipped coord-sync executable doctest
- rename the board-level test to the one-claim state it establishes
- audited claim_disclosure and unclaim outcome messages; no other mismatch found

TDD evidence:
- RED unit: expected (paths added), received (previous paths replaced)
- RED shipped doctest: coord-sync.md output did not contain (paths added)
- GREEN: both focused regressions passed after the branch change

Verification receipt:
- SHA: ef614241ed589b7d672f61cb386c72783a3ed83c
- base: origin/main 8a54a2f081599b912ba0c18fb68246b04f73dfa5
- lane: C:\Users\synvoke\AppData\Local\fleet-workstation\lanes\worker-2
- toolchain: rustc 1.93.1 / cargo 1.93.1
- L0 PASS: cargo fmt --all --check; cargo clippy -p edda --all-targets -- -D warnings; cargo test -p edda (320 unit + 11 skill-doctest tests)
- L1 PASS, CARGO_INCREMENTAL=0, 344.1s: cargo fmt --all --check; cargo clippy --workspace --all-targets -- -D warnings; cargo test --workspace
- clean tree before and after L1

Scope guard: only cmd_bridge.rs and coord-sync.md changed; the #503 identity chain was not touched.

### DIFF
--- a/crates/edda-cli/src/cmd_bridge.rs
+++ b/crates/edda-cli/src/cmd_bridge.rs
@@ -698,6 +698,8 @@ fn claim_disclosure(
         ));
     } else if released.is_empty() && !gained {
         lines.push(format!("Re-claimed scope: {label} (unchanged)"));
+    } else if released.is_empty() {
+        lines.push(format!("Re-claimed scope: {label} (paths added)"));
     } else {
         lines.push(format!(
             "Re-claimed scope: {label} (previous paths replaced)"
@@ -2436,7 +2438,7 @@ mod tests {
     }
 
     #[test]
-    fn widening_reports_a_change_but_no_release() {
+    fn widening_reports_paths_added_but_no_release() {
         let previous = prior_claim("api", &["src/api/*"]);
         assert_eq!(
             claim_disclosure(
@@ -2444,7 +2446,7 @@ mod tests {
                 "api",
                 &owned(&["src/api/*", "src/api/v3/*"])
             ),
-            vec!["Re-claimed scope: api (previous paths replaced)"]
+            vec!["Re-claimed scope: api (paths added)"]
         );
     }
 
@@ -2459,7 +2461,7 @@ mod tests {
     }
 
     #[test]
-    fn a_second_claim_replaces_the_first_and_says_so() {
+    fn a_second_claim_leaves_one_claim_on_the_board() {
         let _store = crate::test_support::isolated_store();
         let _env = env_guard();
         std::env::remove_var("EDDA_SESSION_ID");
@@ -2472,8 +2474,7 @@ mod tests {
         claim(repo.path(), "api", &["src/api/*".into()], Some("s1")).expect("second claim");
 
         // The board folds to one claim per session, so the first scope is gone.
-        // GH-488 item 2 was that this happened without saying so; the printed
-        // line is the fix, and the fold is the behaviour it discloses.
+        // The disclosure tests above separately pin what the command prints.
         let claims = edda_bridge_claude::peers::compute_board_state(&pid).claims;
         assert_eq!(claims.len(), 1, "one session holds one claim");
         assert_eq!(claims[0].label, "api");
--- a/crates/edda-cli/src/skills/coord-sync.md
+++ b/crates/edda-cli/src/skills/coord-sync.md
@@ -62,6 +62,8 @@ $ edda claim "api" --paths "src/api/*" --session s1
 > released: src/auth/*
 $ edda claim "api" --paths "src/api/*" --session s1
 > Re-claimed scope: api (unchanged)
+$ edda claim "api" --paths "src/api/*" --paths "src/api/v2/*" --session s1
+> Re-claimed scope: api (paths added)
 $ edda claim "api" --paths "src/api/v2/*" --session s1
 > Re-claimed scope: api (previous paths replaced)
 > released: src/api/*


fagemx:
## Code Review: Round 1 — PR #514

*Independent GATE-01 review (non-author) · immutable head `ef614241ed589b7d672f61cb386c72783a3ed83c` · base/current `origin/main` `8a54a2f081599b912ba0c18fb68246b04f73dfa5`.*

### IN SCOPE

- `claim_disclosure` truth table and all direct outcome strings.
- Widening unit proof, canonical shipped `coord-sync` doctest, board-level test semantics/name.
- `unclaim` outcome-message sweep, unchanged board/security behavior, current-base integration, and issue #508 acceptance.

### RAN

- `git diff --check 8a54a2f… ef614241…` — pass.
- Frozen path boundary — pass: exactly `crates/edda-cli/src/cmd_bridge.rs` and `crates/edda-cli/src/skills/coord-sync.md` changed.
- Adversarial truth-table assertions — pass:
  - no release + no gain → `(unchanged)`;
  - no release + gain → `(paths added)`;
  - any release → `(previous paths replaced)` plus only actually released paths.
- Unit/post-image assertions — pass: widening expects exactly one `(paths added)` line, so a false `released:` line also fails; board test is now `a_second_claim_leaves_one_claim_on_the_board` and its body checks exactly that state.
- Shipped-doctest wiring audit — pass: widening is executed before narrowing; widening expects `(paths added)`, then narrowing still expects replacement and `released: src/api/*`.
- Full `claim_disclosure`/`unclaim` string sweep — pass: first claim, relabel, unchanged, added, replaced/released, no-op teardown, refusal, and successful unclaim messages all match their code paths.
- Security/current-base guard — pass: board writes/folding, identity resolution (#503), and unclaim behavior are unchanged; merge-base equals current `origin/main`; GitHub reports `MERGEABLE`.

### READ

- Issue #508 verbatim.
- Task #110 receipt and PR TDD evidence: unit and shipped doctest RED→GREEN; focused `edda` L0 green.
- Exact-SHA L1 receipt from `worker-2`, `CARGO_INCREMENTAL=0`: fmt, workspace Clippy, and workspace tests passed in 344.1s with clean tree.
- Exact-head CI: 7/7 successful — Format; Clippy on Linux/macOS/Windows; tests on Linux/macOS/Windows.
- No duplicate local Cargo run: the exact SHA already has a complete L1 receipt and three-platform CI. Verifier build lane: `n/a`.

### FOLLOW-UP ISSUE

- #503 remains intentionally out of scope. Its session-identity chain was not changed and can rebase onto this disclosure fix after merge.

### P0

None.

### P1

None.

### Minor

None.

### Cost

Approximately 6 minutes including CI wait; 10 local/GitHub inspection or wait calls; verifier Cargo runs: 0; verifier lane: `n/a`; READ worker L1: 344.1s; token accounting unavailable in this client.

### Verdict: LGTM

P0=0, P1=0. The complete bounded #508 acceptance and security surface is satisfied at the pinned head.

Issue #120, opened 2026-02-26
feat: approval policy defaults — rules for what requires human approval


fagemx:
## 背景

Auto-execution Pipeline (#113) 需要知道「什麼時候暫停等人類決策」。目前沒有規則引擎定義這件事 — pipeline 不知道哪些動作需要批准、哪些可以自動通過。

沒有 approval policy，就會發生兩種壞結果：
- **全部都問** → 手機通知轟炸，使用者疲勞，決策品質下降
- **全部都過** → 高風險操作沒人審，出事才發現

## 目標

定義 Approval Policy 規則引擎：什麼操作需要人類批准、什麼可以自動通過。

### 預設規則（零設定即可用）

```toml
# 內建 defaults，不需要設定檔
[defaults]
# 這些 pipeline 步驟一定要人類批准
always_require = ["plan_approval", "pr_merge"]

# 這些條件觸發額外批准
[[defaults.rules]]
condition = "risk_level >= high"
action = "require_approval"

[[defaults.rules]]
condition = "files_changed > 20"
action = "require_approval"

[[defaults.rules]]
condition = "touches_off_limits"
action = "require_approval"

# 這些可以自動通過
[[defaults.auto_approve]]
condition = "risk_level == low AND tests_all_passed AND files_changed <= 5"
action = "auto_approve"
```

### 使用者覆寫（`.edda/approval-policy.toml`）

```toml
# 團隊可自訂：哪些嚴格、哪些寬鬆
[policy]
name = "my-project"

# Pipeline 步驟層級
[policy.steps]
plan_approval = "always"        # 每次都要批准
pr_merge = "always"             # 每次都要批准
implement = "auto"              # 自動跑，不問
test = "auto"                   # 自動跑，不問

# 條件規則（覆寫 defaults）
[[policy.rules]]
name = "large-change"
condition = { files_changed_gt = 15, risk_gte = "medium" }
action = "require_approval"
notify = true

[[policy.rules]]
name = "auth-module"
condition = { path_matches = "src/auth/**" }
action = "require_approval"
reason = "auth 模組變更一律審查"

[[policy.rules]]
name = "test-only"
condition = { path_matches = "tests/**", risk_eq = "low" }
action = "auto_approve"

# 時段規則（可選）
[[policy.rules]]
name = "after-hours"
condition = { time_outside = "09:00-18:00" }
action = "require_approval"
reason = "非工作時間操作一律審查"
```

### 規則評估順序

```
[...]

PR #517, opened 2026-08-23
fix(coord): bind session identity to process

### LINKED ISSUE #503: coord(cli): session identity must come from the process, not the workspace

fagemx:
## Summary

- require mutating CLI identity to come from explicit --session or process-carried EDDA_SESSION_ID whenever any live session exists
- preserve deterministic cli-label fallback only for genuine standalone CLI use with zero live sessions
- carry host session identity through Claude CLAUDE_ENV_FILE, Cursor sessionStart env, OpenClaw resolve_exec_env, and command-local Codex/Hermes rewrites
- update duplicated help, reference docs, and rustdoc with attribution and authentication limits

## Acceptance evidence

- TDD red reproduced peer adoption: bare claim replaced the sole live scope, bare decide appended state beside two live sessions, and bare unclaim released the sole live claim
- Green tests prove one- and two-live-session refusal names --session and preserves state
- explicit --session and non-empty EDDA_SESSION_ID retain precedence; empty EDDA_SESSION_ID cannot adopt a heartbeat; zero-live fallback remains
- shell rewrites preserve original command bytes, quote opaque IDs for PowerShell/POSIX, and leave non-shell tools unchanged
- OpenClaw runtime probe proved two sessionKey mappings resolve to their exact heartbeat IDs and session_end clears the mapping

## Verification receipt

Frozen SHA: 631931074ac137098bc51794a81ed47ea8ac1848
Host: x86_64-pc-windows-msvc
Toolchain: rustc 1.93.1, cargo 1.93.1
Lane: C:\Users\synvoke\AppData\Local\fleet-workstation\lanes\worker-2
Freeze setting: CARGO_INCREMENTAL=0

L0 PASS:
- cargo fmt --all --check
- cargo clippy and cargo test for edda, edda-bridge-claude, edda-bridge-codex, edda-bridge-cursor, edda-bridge-hermes, and edda-bridge-openclaw
- counts: edda 323 unit + 11 integration; Claude 574; Codex 17; Cursor 22; Hermes 22; OpenClaw 32; zero failures

L1 PASS at the frozen SHA:
- cargo fmt --all --check
- cargo clippy --workspace --all-targets -- -D warnings
- cargo test --workspace

The first workspace-test link attempt hit transient Windows LNK1104 on the conductor test executable after format and workspace clippy had passed. The artifact was absent and no Cargo, rustc, or conductor process remained; per the ladder, only cargo test --workspace was retried at the same SHA, and it passed across all crates and doc-tests.

Additional PASS:
- executable Node probe for OpenClaw resolve_exec_env mapping and cleanup
- generated OpenClaw plugin copy byte-matches integrations/openclaw/index.js after line-ending normalization
- clean worktree at frozen SHA

Out of scope by design: branch-refresh and subagent-stale follow-ups from #503.

Closes #503

### DIFF
--- a/crates/edda-bridge-claude/src/dispatch/session.rs
+++ b/crates/edda-bridge-claude/src/dispatch/session.rs
@@ -1,4 +1,5 @@
 use std::fs;
+use std::io::Write;
 use std::path::Path;
 
 use crate::signals::{extract_session_signals, save_session_signals, TaskSnapshot};
@@ -698,6 +699,8 @@ pub(super) fn dispatch_session_start(
     cwd: &str,
     digest_warning: Option<&str>,
 ) -> anyhow::Result<HookResult> {
+    persist_session_identity(session_id)?;
+
     // Conductor mode: skip sections that overlap with conductor's --append-system-prompt.
     // See CONDUCTOR-SPEC.md §10.2.
     let conductor_mode = std::env::var("EDDA_CONDUCTOR_MODE").is_ok();
@@ -916,4 +919,21 @@ pub(super) fn dispatch_session_start(
     } else {
         Ok(HookResult::empty())
     }
+}
+
+fn persist_session_identity(session_id: &str) -> anyhow::Result<()> {
+    let Some(env_file) = std::env::var_os("CLAUDE_ENV_FILE") else {
+        return Ok(());
+    };
+    if session_id.is_empty() || env_file.is_empty() {
+        return Ok(());
+    }
+
+    let quoted = session_id.replace('\'', "'\\''");
+    let mut file = fs::OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(env_file)?;
+    writeln!(file, "export EDDA_SESSION_ID='{quoted}'")?;
+    Ok(())
 }
--- a/crates/edda-bridge-claude/src/dispatch/tests.rs
+++ b/crates/edda-bridge-claude/src/dispatch/tests.rs
@@ -62,6 +62,31 @@ fn hook_entrypoint_session_start() {
 }
 
 #[test]
+fn session_start_exports_shell_quoted_process_identity() {
+    let env_file = tempfile::NamedTempFile::new().expect("env file");
+    fs::write(env_file.path(), "export EXISTING=kept\n").expect("seed env file");
+    let env_path = env_file.path().to_string_lossy().into_owned();
+    crate::with_env_guard(
+        &[
+            ("EDDA_BRIDGE_AUTO_DIGEST", Some("0")),
+            ("EDDA_PLANS_DIR", Some("/nonexistent/plans/dir")),
+            ("CLAUDE_ENV_FILE", Some(&env_path)),
+        ],
+        || {
+            let stdin = r#"{"session_id":"claude-'worker","hook_event_name":"SessionStart","cwd":".","transcript_path":"","permission_mode":"default"}"#;
+            hook_entrypoint_from_stdin(stdin).expect("session start");
+            let resumed = r#"{"session_id":"claude-resumed","hook_event_name":"SessionStart","source":"resume","cwd":".","transcript_path":"","permission_mode":"default"}"#;
+            hook_entrypoint_from_stdin(resumed).expect("resumed session start");
+        },
+    );
+
+    assert_eq!(
+        fs::read_to_string(env_file.path()).expect("read env file"),
+        "export EXISTING=kept\nexport EDDA_SESSION_ID='claude-'\\''worker'\nexport EDDA_SESSION_ID='claude-resumed'\n"
+    );
+}
+
+#[test]
 fn hook_entrypoint_camel_case_input() {
     crate::with_env_guard(&[("EDDA_CLAUDE_AUTO_APPROVE", Some("1"))], || {
         // Claude Code sends camelCase JSON
--- a/crates/edda-bridge-claude/src/peers/discovery.rs
+++ b/crates/edda-bridge-claude/src/peers/discovery.rs
@@ -164,11 +164,13 @@ pub fn discover_all_sessions(project_id: &str) -> Vec<PeerSummary> {
 /// If exactly one non-stale session exists for the project, returns
 /// `Some((session_id, label))`. Otherwise returns `None` (ambiguous or no context).
 ///
-/// Used by CLI commands (`edda decide`, etc.) to resolve session identity
-/// when `EDDA_SESSION_ID` env var is not set.
-///
-/// This cannot tell the caller apart from a lone peer: a process has no way to
-/// prove a heartbeat is its own (GH-488).
+/// This observes heartbeat cardinality; it does not establish caller identity.
+/// A process has no way to prove a heartbeat is its own, so mutating CLI verbs
+/// must use explicit `--session` or process-carried `EDDA_SESSION_ID` instead.
+/// The environment value is attribution and a user override, not
+/// authentication or authorization. This helper remains available only for
+/// callers that need the observational "exactly one live heartbeat" fact
+/// (GH-503).
 pub fn infer_session_id(project_id: &str) -> Option<(String, String)> {
     let state_dir = edda_store::project_dir(project_id).join("state");
     let stale_threshold = stale_secs();
--- a/crates/edda-bridge-codex/src/dispatch.rs
+++ b/crates/edda-bridge-codex/src/dispatch.rs
@@ -11,10 +11,9 @@
 //! }
 //! ```
 //!
-//! For PreToolUse we can also return `{ "continue": false, "stopReason": "..." }`
-//! to block, but this skeleton emits advisory context only — L3 rule enforcement
-//! is wired to the same `edda_postmortem::hooks::evaluate_rules` used by the
-//! Claude bridge, but delivered as `additionalContext` warnings for now.
+//! For Bash `PreToolUse`, `permissionDecision: "allow"` plus `updatedInput`
+//! carries the hook session id into that command. L3 rule enforcement is wired
+//! to the same evaluator as the Claude bridge and remains advisory context.
 
 use edda_bridge_claude::{render, state};
 
@@ -54,6 +53,60 @@ fn with_context(hook_event_name: &str, context: &str) -> anyhow::Result<HookResu
         }
     });
     Ok(HookResult::output(serde_json::to_string(&output)?))
+}
+
+fn pre_tool_context(context: &str) -> anyhow::Result<HookResult> {
+    let output = serde_json::json!({
+        "hookSpecificOutput": {
+            "hookEventName": "PreToolUse",
+            "additionalContext": context,
+        }
+    });
+    Ok(HookResult::output(serde_json::to_string(&output)?))
+}
+
+fn pre_tool_result(envelope: &CodexEnvelope, warning: Option<&str>) -> anyhow::Result<HookResult> {
+    let rewritten = if envelope.tool_name == "Bash" && !envelope.session_id.is_empty() {
+        envelope
+            .tool_input
+            .get("command")
+            .and_then(|value| value.as_str())
+            .map(|command| command_with_session_identity(command, &envelope.session_id))
+    } else {
+        None
+    };
+
+    let Some(command) = rewritten else {
+        return match warning {
+            Some(warning) => pre_tool_context(warning),
+            None => Ok(HookResult::output("{}".to_string())),
+        };
+    };
+
+    let mut hook_output = serde_json::json!({
+        "hookEventName": "PreToolUse",
+        "permissionDecision": "allow",
+        "updatedInput": { "command": command },
+    });
+    if let Some(warning) = warning {
+        hook_output["additionalContext"] = serde_json::Value::String(warning.to_string());
+    }
+    let output = serde_json::json!({ "hookSpecificOutput": hook_output });
+    Ok(HookResult::output(serde_json::to_string(&output)?))
+}
+
+fn command_with_session_identity(command: &str, session_id: &str) -> String {
+    if cfg!(windows) {
+        format!(
+            "$env:EDDA_SESSION_ID = '{}';\n{command}",
+            session_id.replace('\'', "''")
+        )
+    } else {
+        format!(
+            "export EDDA_SESSION_ID='{}';\n{command}",
+            session_id.replace('\'', "'\\''")
+        )
+    }
 }
 
 // ── Entry ──
@@ -189,11 +242,11 @@ fn dispatch_user_prompt_submit(
 
 fn dispatch_pre_tool_use(project_id: &str, envelope: &CodexEnvelope) -> anyhow::Result<HookResult> {
     if std::env::var("EDDA_POSTMORTEM").unwrap_or_else(|_| "1".into()) == "0" {
-        return Ok(ok());
+        return pre_tool_result(envelope, None);
     }
     let mut rules_store = edda_postmortem::RulesStore::load_project(project_id);
     if rules_store.active_rules().is_empty() {
-        return Ok(ok());
+        return pre_tool_result(envelope, None);
     }
 
     let files_touched: Vec<String> = envelope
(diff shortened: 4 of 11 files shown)


fagemx:
## Code Review: Round 1 — PR #517

*Independent GATE-01 review (non-author) · immutable head `631931074ac137098bc51794a81ed47ea8ac1848` · current base `81e8e4d239cc7866d23c9728e9b71321b2931d2a`.*

### IN SCOPE

- All 11 changed paths and direct consumers: CLI resolver/callers/help/reference, Claude/Cursor carriers, Codex/Hermes rewrites, OpenClaw runtime/generated copy, observational rustdoc.
- One/two-peer refusal and no-mutation behavior, explicit/env precedence, zero-peer fallback, quoting/security/cross-platform behavior, tests/gates/current-base.
- Confirmation that branch-refresh and subagent-stale findings remain separate.
- Worker scope-crossing classification for `main.rs` and `docs/reference/cli.md`.

### RAN

- `git diff --check 86c276f… 6319310…` — pass.
- Exact 11-path boundary and nine production `resolve_session_id` consumers — pass.
- CLI assertions — pass: one-peer claim refusal preserves the existing scope; two-peer decide refusal writes neither ledger nor binding; unclaim refusal preserves claims; errors name `--session`; non-empty process env and explicit override win; empty env refuses; zero-live fallback remains.
- Stale-promise sweep — pass: mutating CLI no longer calls heartbeat inference; duplicated help/reference text no longer promises auto-inference; rustdoc states attribution/user override rather than authentication/authorization.
- Host audit — pass for Claude env-file quoting/resume append, Cursor session env, Codex/Hermes POSIX+PowerShell quote handling and original command bytes, non-shell preservation, and generated OpenClaw copy parity.
- Official Codex hook schema check — pass: `PreToolUse` rewrite uses the documented `permissionDecision: "allow"` + `updatedInput.command` non-blocking rewrite shape ([OpenAI Docs](https://developers.openai.com/codex/hooks)).
- OpenClaw adversarial runtime probe — **fail**: after `session_start` maps `stable-key → heartbeat-id`, a later valid context with only `sessionKey` changes `resolve_exec_env` to `stable-key`; `session_end` then sends `session_id: stable-key` instead of cleaning up `heartbeat-id`.
- Current-base integration — pass: current-base-only CI/lint and conductor paths are disjoint; `git merge-tree` exits 0 and GitHub reports `MERGEABLE`.

### READ

- Issue #503 verbatim and task #115 receipt/TDD evidence.
- L0 results for `edda` plus five bridge crates; exact-SHA L1 fmt/workspace Clippy/workspace tests. The first Windows workspace-test link hit transient LNK1104; after confirming no holder/artifact remained, only the failed test job was retried and passed—consistent with the repository ladder.
- Worker OpenClaw mapping/copy probes and clean-tree receipt.
- Exact-head CI: 7/7 successful — Format; Clippy on Linux/macOS/Windows; tests on Linux/macOS/Windows.
- OpenClaw’s documented `resolve_exec_env` contract is keyed by `event.sessionKey`; `ctx.sessionId` is optional context, so missing-sessionId contexts must preserve the known mapping ([OpenClaw docs](https://docs.openclaw.ai/plugins/hooks)).
- No verifier Cargo rerun: receipt and CI cover the compiled/tested surface. Verifier lane: `n/a`.

### What is correct

- Resolver identity now comes only from explicit `--session` or process-carried env; workspace/branch/heartbeat facts do not confer identity.
- All mutating and attribution consumers propagate resolver errors before persistence; read-only coordination follows the same identity rule.
- Shell quoting is safe for embedded single quotes on POSIX and PowerShell, and original command bytes follow the prefix unchanged.
- Claude, Cursor, Codex, and Hermes carriers are bounded to their host mechanisms; Codex’s parent-session subagent limit is explicitly documented.
- `update_heartbeat_branch` Bash-only behavior and the 1×/15× subagent staleness discrepancy were not silently folded into this PR.

### Scope/process finding

`crates/edda-cli/src/main.rs` and `docs/reference/cli.md` were omitted from task #115’s literal owned-path list. Product classification: both are acceptance-required direct consumers; reverting them would leave false help/reference promises, so they belong in the PR. Process classification: non-blocking coordination finding—future briefs/claims must list these direct consumers before editing them.

### FOLLOW-UP ISSUE

- Process-only: under-scoped task/claim as described above. It does not block a product-correct PR.
- The two issue-listed branch-refresh/subagent-stale defects remain separate follow-ups by design.

### P0

None.

### P1

1. `integrations/openclaw/index.js:5` and the generated copy in `crates/edda-bridge-openclaw/src/admin.rs:19` — `heartbeatSessionId` overwrites an established `sessionKey → sessionId` mapping whenever a later hook context omits `ctx.sessionId`, because it falls back to `sessionKey` and calls `Map.set` again. That is a supported/anticipated shape (the code has the fallback, and `resolve_exec_env` is keyed by `sessionKey`). Concrete sequence at this SHA:
   - `session_start`: `{sessionId: "heartbeat-id", sessionKey: "stable-key"}` → env is `heartbeat-id`;
   - later hook: `{sessionKey: "stable-key"}` → env changes to `stable-key`;
   - `session_end` with only the key sends cleanup for `stable-key`, leaving the real `heartbeat-id` heartbeat/claim orphaned.

   Preserve an existing canonical mapping when `sessionId` is absent; only establish the `sessionKey` fallback when no mapping exists. Use that canonical ID for later hook payloads and session-end cleanup, then delete it. Add the missing-sessionId runtime regression to both the executable probe/copy evidence.

### Cost

Approximately 12 minutes; 13 local/GitHub/web inspection calls; verifier Cargo runs: 0; lane: `n/a`; READ one L1 receipt and 7 CI jobs; token accounting unavailable in this client.

P0=0, P1=1. OpenClaw must preserve its canonical session mapping across contexts that provide only `sessionKey` before #503’s all-host acceptance is complete.

fagemx:
## Review Response: Round 1 — PR #517

New immutable head: `99d3fcaac2334bc8bce1420865d006b16895fa72` (fix commit `99d3fca`).

### P1 response

1. **Canonical OpenClaw session mapping across key-only contexts — addressed.** `heartbeatSessionId` now resolves in this order: explicit `ctx.sessionId`, the existing `sessionKey → sessionId` mapping, then `sessionKey` only when no mapping exists. Therefore a `session_start` context `{ sessionId: "heartbeat-id", sessionKey: "stable-key" }` establishes the canonical ID; later key-only hook contexts keep `heartbeat-id`; `resolve_exec_env` exports that ID; key-only `session_end` sends cleanup for `heartbeat-id`; and the mapping is deleted only after cleanup. An unmapped key still falls back to itself. The generated `admin.rs` plugin copy has the same change and assertion.

### RAN

- TDD RED against the Round 1 head: the exact start → key-only hook sequence failed with actual `stable-key`, expected `heartbeat-id`.
- Frozen executable runtime/generated-copy probe at `99d3fcaac2334bc8bce1420865d006b16895fa72`: pass for canonical later-hook payload, `resolve_exec_env`, key-only `session_end`, post-cleanup mapping deletion, and unmapped-key fallback.
- Runtime/generated plugin copy parity after line-ending normalization: pass.
- `cargo fmt --all --check`: pass.
- `cargo clippy -p edda-bridge-openclaw --all-targets -- -D warnings`: pass.
- `cargo test -p edda-bridge-openclaw`: pass, 32/32 unit tests and 0 doc-test failures.
- Lane: `C:\Users\synvoke\AppData\Local\fleet-workstation\lanes\worker-2`; `CARGO_INCREMENTAL=0`.
- `git diff --check` and frozen clean-worktree check: pass.
- Exact-head CI run [32636608978](https://github.com/fagemx/edda/actions/runs/32636608978): 7/7 successful — Format; Clippy on Linux/macOS/Windows; tests on Linux/macOS/Windows.

Ready for independent Round 2 review. No review verdict or merge action performed.

fagemx:
## Code Review: Round 2 — PR #517

*Independent GATE-01 re-review · same verifier identity · immutable head `99d3fcaac2334bc8bce1420865d006b16895fa72` · current base `81e8e4d239cc7866d23c9728e9b71321b2931d2a`.*

### IN SCOPE

- Round 1 P1: canonical OpenClaw `sessionKey → sessionId` lifecycle across key-only contexts, env resolution, cleanup, deletion, and fallback.
- Generated runtime-copy parity and fix-caused regression surface.
- Complete original CLI/five-host/quoting/security/rustdoc/help/reference/current-base scope.
- Continued exclusion of the branch-refresh and subagent-stale follow-ups.

### RAN

- Exact runtime from `integrations/openclaw/index.js` at `99d3fcaa…`:
  - `session_start` with `{sessionId: heartbeat-id, sessionKey: stable-key}` → env `heartbeat-id`;
  - later key-only hook → payload/env remain `heartbeat-id`;
  - key-only `session_end` → cleanup payload uses `heartbeat-id`;
  - post-cleanup `resolve_exec_env` → empty mapping;
  - previously unmapped key-only hook → fallback env equals that key.
  All assertions pass.
- Generated `PLUGIN_INDEX_JS` copy contains the exact integration runtime after normalization — pass.
- Round 2 boundary — pass: only `integrations/openclaw/index.js` and `crates/edda-bridge-openclaw/src/admin.rs` changed.
- `git diff --check` for Round 2 and full PR — pass.
- Full original-scope assertions — pass: 11 paths; nine resolver consumers; no mutating heartbeat inference; one/two-peer refusals; explicit/env/empty/fallback behavior; proof-limit docs; separate follow-up paths untouched.
- Current-base integration — pass: `git merge-tree` exits 0 and GitHub reports `MERGEABLE`.

### READ

- Issue #503, Round 1 Changes Requested, and implementer Review Response Round 1.
- Task #120 RED→GREEN receipt, executable runtime/copy probe, focused fmt/warning-denied Clippy, and 32/32 OpenClaw tests.
- Exact-head CI: 7/7 successful — Format; Clippy on Linux/macOS/Windows; tests on Linux/macOS/Windows.
- Round 1’s complete CLI/Claude/Cursor/Codex/Hermes audit remains applicable; only the two OpenClaw copies changed.
- No verifier Cargo rerun: focused receipt and exact-head CI cover the fix. Verifier lane: `n/a`.

### P1 resolution

Resolved. `heartbeatSessionId` now chooses explicit `ctx.sessionId`, then the existing canonical mapping, and only then an unmapped `sessionKey` fallback. Key-only later contexts therefore cannot overwrite the canonical ID; session-end uses it before deletion.

### Full-scope result

- Bare shells beside one/two live sessions refuse and preserve state; explicit/process IDs still work; standalone zero-peer fallback remains.
- All identity-taking consumers propagate resolver errors before mutation.
- Claude/Cursor direct carriers, Codex/Hermes safely quoted command-local carriers, and OpenClaw exec env all preserve host identity without global cross-session leakage.
- Rustdoc/help/reference truth and authentication limits remain correct.
- Branch-refresh and 1×/15× subagent-stale defects remain separate follow-ups.

### Scope/process finding

Round 1’s non-blocking process finding remains: `main.rs` and `docs/reference/cli.md` were omitted from task #115’s literal path list, but are product-required direct consumers and correctly remain in the PR. Future briefs should claim them explicitly.

### FOLLOW-UP ISSUE

- Process-only under-scoped brief noted above.
- The two issue-listed branch-refresh/subagent-stale defects remain separate.

### P0

None.

### P1

None.

### Minor

None.

### Cost

Approximately 6 minutes; 7 local/GitHub inspection calls; verifier Cargo runs: 0; lane: `n/a`; READ 7 CI jobs; token accounting unavailable in this client.

### Verdict: LGTM

P0=0, P1=0. Round 1’s blocker is fixed and the complete #503 all-host security surface is green at the pinned head.

Issue #524, opened 2026-08-31
bug(cli): two cmd_reconcile tests flake under the parallel suite — fixed-count wait budgets shorter than the real critical section


fagemx:
> **Corrected 2026-08-31.** The original filing named cross-test `.edda/LOCK`
> contention as the hypothesis and listed one failing test. Both were wrong:
> every `cmd_reconcile` test builds its workspace in a unique tempdir, and a
> second test in the same family fails the same way. The real mechanism and the
> full evidence are below; the comment thread preserves the original trail.

## Symptom

Two `cmd_reconcile` tests fail intermittently when the suite runs under load. They fail in *different* ways, which is why they read as unrelated:

```text
---- cmd_reconcile::tests::simultaneous_reconciles_create_one_attempt_and_release_the_lock
Error: workspace is locked by another process (...\.tmpM6um8u\repo\.edda\LOCK)

---- cmd_reconcile::tests::first_spawn_failure_does_not_prevent_later_plan_launch
panicked at crates\edda-cli\src\cmd_reconcile.rs:5298:9:
assertion failed: launched_file.exists()
```

## Root cause: a fixed-count wait budget shorter than a legitimate critical section

Not lock contention between tests. Each test owns its own tempdir workspace, and the lock path captured in the failures is the test's own repo.

The first test races two reconciles at a `Barrier` on purpose; exactly one must win the workspace lock. The loser called `acquire_workspace_lock`, which retried **120 × 25 ms = 3 s** and then gave up — while the winner legitimately holds the lock for **3–5 s** under suite load (`git worktree add` alone measured 1.8–4.2 s on Windows, plus SQLite appends and `rebuild_branch`, all deliberately inside the critical section). The budget was smaller than the work it was waiting for, so the loser bailed instead of proceeding.

The second test is the same shape: `for _ in 0..40 { if launched_file.exists() { break } }` — another fixed-count poll (1 s total) waiting on a side effect, here a spawned `.cmd` child, which routinely exceeds 1 s under full-suite load. On timeout it fell through to a bare `assert!` with no context, which is why the failure said nothing about waiting.

Both are wait budgets expressed as iteration counts rather than deadlines, and both are shorter than the real work under load.

## Why the second test was invisible at filing time

The original evidence used `cargo test -p edda` (326 passed / 1 failed). A plain `cargo test --workspace` **stops at the first failing binary** — 1 of 46 covered. Under `--no-fail-fast` all 46 run, machine load is higher, and the second test fails too.

[...]

PR #528, opened 2026-08-31
fix(conductor): resolve the pi binary on Windows (#525)

### LINKED ISSUE #525: bug(conductor): --agent pi never resolves on Windows (npm ships pi.cmd; CreateProcess ignores PATHEXT)
## Symptom

On Windows, `edda conduct run --agent pi` fails at `verify_available` with "pi CLI not found" even though pi is installed and `pi --version` works fine in any terminal. Every phase dispatch is dead on arrival, so the launcher shipped in PR #523 is unusable on the platform the fleet actually runs on.

## Root cause

Two facts compose:

1. **npm installs no `pi.exe` on Windows.** The global install ships three shims — `pi` (sh script), `pi.cmd`, `pi.ps1`:

   ```text
   > where.exe pi
   C:\Users\synvoke\AppData\Roaming\npm\pi
   C:\Users\synvoke\AppData\Roaming\npm\pi.cmd
   ```

2. **`CreateProcess` does not apply `PATHEXT`.** A shell resolves `pi` → `pi.cmd` via `PATHEXT`; Rust's `std::process::Command` only ever appends `.exe` to an extensionless name. `PiRpcLauncher::new()` hardcodes `PathBuf::from("pi")`, so resolution never succeeds.

## Evidence (2026-08-31, Windows 11, pi 0.84.4 via npm)

Minimal repro, `rustc` single file:

```rust
for name in ["pi", "pi.cmd"] {
    let r = std::process::Command::new(name).arg("--version").output();
    // pi       -> Err: program not found
    // pi.cmd   -> Ok, stdout: 0.84.4
}
```

## Fix direction

Default the executable name per platform, with an explicit escape hatch:

- `EDDA_PI_BIN` env var overrides everything (non-standard install locations, absolute paths).
- Otherwise `pi.cmd` on Windows, `pi` elsewhere.
- Mention `EDDA_PI_BIN` in the not-found error so the failure mode is self-explaining.

The extensionless `pi` shim (sh script) and `pi.ps1` are not `CreateProcess`-spawnable either, so `.cmd` is the only correct Windows default. Spawning `.cmd` routes through `cmd.exe`; Rust std escapes `.bat`/`.cmd` arguments since the BatBadBut fix, and the launcher's args are static strings regardless.

A fix implementing exactly this is staged on branch `fix/pi-bin-windows-resolution`.

## doneWhen

- [ ] `edda conduct run --agent pi` passes `verify_available` on a stock Windows npm install with no env override
- [ ] `EDDA_PI_BIN` override honored on all platforms and named in the not-found error message
- [ ] Unit tests cover the platform default and the override
- [ ] Non-Windows default unchanged (`pi`)

🤖 Drafted with Claude Code from a live `--agent pi` failure on the fleet workstation.

fagemx:
Closes #525

## Problem

`edda conduct run --agent pi` failed at `verify_available` with "pi CLI not found" on Windows even on a working pi install — every phase dead on arrival, making the launcher from #523 unusable on the platform the fleet actually runs on.

Two facts compose:

1. **npm installs no `pi.exe` on Windows.** The global install ships `pi` (sh script), `pi.cmd`, and `pi.ps1`:

   ```text
   > where.exe pi
   C:\Users\synvoke\AppData\Roaming\npm\pi
   C:\Users\synvoke\AppData\Roaming\npm\pi.cmd
   ```

2. **`CreateProcess` does not apply `PATHEXT`.** A shell resolves `pi` → `pi.cmd`; Rust's `std::process::Command` only ever appends `.exe` to an extensionless name. `PiRpcLauncher::new()` hardcoded `PathBuf::from("pi")`, so resolution never succeeded.

Confirmed with a standalone `rustc` repro: `Command::new("pi")` → `program not found`; `Command::new("pi.cmd")` → `0.84.4`.

The existing launcher tests all inject fake processes, so none of them exercised binary resolution — the gap CI could not see.

## Change

`crates/edda-conductor/src/agent/pi_rpc.rs` only:

- `resolve_pi_bin(explicit: Option<OsString>)` — takes the override as an argument rather than reading the environment, so resolution is a pure function testable without mutating process-wide state.
- `default_pi_bin()` composes it with `std::env::var_os("EDDA_PI_BIN")`.
- Windows defaults to `pi.cmd`, everything else keeps `pi`. An empty `EDDA_PI_BIN=` is treated as unset rather than spawning an empty path.
- The not-found error now names `EDDA_PI_BIN` so the failure explains its own fix.

`edda-cli` needs no change — `cmd_conduct.rs` constructs `PiRpcLauncher::new()`, so it picks the fix up directly.

## Tests

Four unit tests, none touching process environment:

- `pi_bin_falls_back_to_the_platform_install` — `resolve_pi_bin(None)` is `pi.cmd` on Windows, `pi` elsewhere
- `edda_pi_bin_overrides_the_platform_default`
- `empty_edda_pi_bin_is_treated_as_unset`
- `with_bin_overrides_the_default`

## Gates

**L1 gate receipt** — frozen SHA `21783920651632ed865f25e04fac1a874ed4d93b`

| Gate | Result |
|---|---|
| `cargo clippy --workspace --all-targets -- -D warnings` | pass (4m49s, zero warnings) |
| `cargo test --workspace --no-fail-fast` | 46 test binaries, 45 green; `edda` bin 325 passed / 2 failed |

- toolchain `rustc 1.93.1 (01f6ddf75 2026-02-11)`, `CARGO_INCREMENTAL=0`
- run in a dedicated worktree (`C:/ai_agent/edda-wt-gh525`) on its own target dir, clean tree
- `--no-fail-fast` matters here: a plain `cargo test --workspace` stops at the first failing
  binary and had covered only 1 of 46, leaving `edda-conductor` — the crate this PR changes —
  unrun. The 4 new tests are in that 46-binary run.

## Note on the two failing tests

Both belong to #524, both pre-existing:

- `simultaneous_reconciles_create_one_attempt_and_release_the_lock` — `workspace is locked by another process (...\.edda\LOCK)`
- `first_spawn_failure_does_not_prevent_later_plan_launch` — `assertion failed: launched_file.exists()` (`cmd_reconcile.rs:5298`)

**Control run, same machine / worktree / command, back to back:**

| ref | binaries | result | failures |
|---|---|---|---|
| `2178392` (this PR) | 46 | 325 passed / 2 failed | the two below |
| `c6246f1` (`main`, detached) | 46 | 325 passed / 2 failed | the same two |

Identical, so this change introduces no regression.

The second test is not yet named in #524, which recorded only the first; under
`--workspace --no-fail-fast` it fails on `main` too. Evidence posted to that issue.

This PR touches `pi_rpc.rs` only. `cmd_reconcile.rs` does reach into
`edda_conductor`, but only `codex_app_server::CodexAppServer` (lines 2321, 2412) —
no shared symbol with the pi launcher.

### DIFF
--- a/crates/edda-conductor/src/agent/pi_rpc.rs
+++ b/crates/edda-conductor/src/agent/pi_rpc.rs
@@ -2,6 +2,7 @@ use crate::agent::launcher::{AgentLauncher, PhaseResult};
 use crate::plan::schema::Phase;
 use anyhow::{anyhow, Context, Result};
 use serde_json::{json, Value};
+use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::sync::Arc;
@@ -14,6 +15,29 @@ use tokio_util::sync::CancellationToken;
 
 /// Upper bound on waiting for the stderr drain once stdout has hit EOF.
 const STDERR_DRAIN_GRACE: Duration = Duration::from_secs(2);
+
+/// Resolve the pi executable from an explicit `EDDA_PI_BIN` value, falling
+/// back to the name npm installs on this platform.
+///
+/// On Windows the npm package ships `pi` as a `.cmd` shim with no `.exe`;
+/// `CreateProcess` — unlike a shell — does not apply `PATHEXT`, so the bare
+/// name never resolves and every phase fails at `verify_available`.
+///
+/// Takes the override as an argument rather than reading the environment so
+/// the resolution is testable without mutating process-wide state.
+fn resolve_pi_bin(explicit: Option<OsString>) -> PathBuf {
+    match explicit {
+        // An empty `EDDA_PI_BIN=` is a set-but-unusable value; treat it as unset
+        // rather than spawning an empty path.
+        Some(value) if !value.is_empty() => PathBuf::from(value),
+        _ if cfg!(windows) => PathBuf::from("pi.cmd"),
+        _ => PathBuf::from("pi"),
+    }
+}
+
+fn default_pi_bin() -> PathBuf {
+    resolve_pi_bin(std::env::var_os("EDDA_PI_BIN"))
+}
 
 /// Launches the pi coding agent via `pi --mode rpc`.
 ///
@@ -43,7 +67,7 @@ impl Default for PiRpcLauncher {
 impl PiRpcLauncher {
     pub fn new() -> Self {
         Self {
-            pi_bin: PathBuf::from("pi"),
+            pi_bin: default_pi_bin(),
             verbose: false,
             model: None,
             session_dir: None,
@@ -85,7 +109,8 @@ impl PiRpcLauncher {
             Ok(s) if s.success() => Ok(()),
             _ => anyhow::bail!(
                 "pi CLI not found (looked for {:?}).\n\
-                 Install: npm install -g @earendil-works/pi-coding-agent",
+                 Install: npm install -g @earendil-works/pi-coding-agent\n\
+                 Or set EDDA_PI_BIN if the executable lives elsewhere.",
                 self.pi_bin
             ),
         }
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: Round 1 — `21783920651632ed865f25e04fac1a874ed4d93b`

Independent review from a separate session (not the implementer).

**Verdict: LGTM.** P0 = 0, P1 = 0.

### IN SCOPE

Changed behavior and paths (`crates/edda-conductor/src/agent/pi_rpc.rs`), the launcher's direct consumer (`crates/edda-cli/src/cmd_conduct.rs`), #525 acceptance, and current-base integration.

### READ

- **Exact-head CI: 7/7 green** — Format, plus Clippy and Test on ubuntu / windows / macos. `edda-conductor` is inside the CI Windows test subset, so the four new unit tests genuinely ran on the platform the fix targets. No local rerun is warranted for them.
- **L1 gate receipt** in the PR body, pinned to this full SHA, with a `main`-detached control run.
- The diff, and both call sites of the resolved path.

### RAN (only what CI cannot cover)

1. **Platform premise, independently reproduced.** A standalone `rustc` program on this workstation: `Command::new("pi")` → `program not found`; `Command::new("pi.cmd")` → `0.84.4`. `where.exe pi` confirms npm ships `pi`, `pi.cmd`, `pi.ps1` and no `.exe`. The stated root cause is real, not inferred.
2. **Orphan-process probe** (the one risk this change newly exposes — before it, nothing spawned on Windows at all, so the `.cmd`-via-`cmd.exe` kill path had never executed). Spawned `pi.cmd --mode rpc` with piped stdio, killed the child, sampled `node.exe` counts: `before=14 during=15 after_kill=14` → **clean, no leak**. No follow-up issue needed.

### Findings

Non-blocking, no action requested:

- A whitespace-only `EDDA_PI_BIN="   "` passes the `!is_empty()` guard and would spawn an unresolvable path. User error with a clear error message; not worth code.
- `.cmd` arguments route through `cmd.exe`; the launcher's args are `--mode rpc --session-id <uuid>` plus optional `--model` / `--session-dir`, and Rust's std errors rather than mis-escaping since the BatBadBut fix. No exposure.

### Notes in the change's favor

- `resolve_pi_bin(Option<OsString>)` takes the override as an argument instead of reading the environment, so all four tests run without mutating process-wide state — no `ENV_LOCK` dance, no cross-test interference.
- Both `verify_available` and the real spawn use `&self.pi_bin`; there is no second path that could bypass resolution.
- The not-found error now names `EDDA_PI_BIN`, so the failure explains its own fix.
- The two suite failures reported in the L1 receipt are reproduced identically on `main` (`c6246f1`) and are tracked in #524 with a fix in #530. Not caused by this PR.

### Cost

READ-dominant: exact-head CI and the receipt carried the regression evidence; two focused local probes (one `rustc` repro, one process probe), no workspace gate rerun.

PR #531, opened 2026-08-31
feat(conductor): codex launcher adapter + --agent codex (#527)

### LINKED ISSUE #527: feat(cli): wire the codex launcher into the --agent selector
## Motivation

`edda conduct run --agent` accepts `claude|pi` (#523), but the tree already contains a codex launcher — `crates/edda-conductor/src/agent/codex_app_server.rs` — that was never wired into the CLI selector. The target fleet topology (orchestrator/executor/verifier, each platform-swappable) uses Codex as one of the interchangeable verifier platforms, so the gap is now user-visible: the launcher exists, the flag rejects it.

## Proposal

- Add `codex` to `AgentKind` in `crates/edda-cli/src/cmd_conduct.rs`, mapped to the existing codex launcher, mirroring how #523 wired pi (including the `writes_transcripts()` handling for `--tmux`).
- Verify the codex launcher still conforms to the current `AgentLauncher` trait — it predates the pi work; adjust if the trait has moved.
- The future `edda dispatch` verb should expose the same three choices (tracked separately).

## doneWhen

- [ ] `edda conduct run plan.yaml --agent codex` dispatches a phase through the codex launcher end-to-end
- [ ] `--tmux` interaction defined (warn-and-fallback like pi, or transcript panes if codex writes them)
- [ ] CLI parse tests cover the new variant (mirror the pi tests from #523)
- [ ] Unknown-agent error lists all three backends

Out of scope by design: a per-phase `agent:` field in the plan schema. Once `edda dispatch` exists, mixed-backend loops are the orchestrator's job; revisit only if a real DAG needs in-plan mixing.

🤖 Drafted with Claude Code from the cross-platform fleet design discussion.

fagemx:
Closes #527

## Problem

`edda conduct run --agent` accepts `claude|pi` only, while the tree already carries a full codex app-server protocol client (`codex_app_server.rs`) that was never adapted to `AgentLauncher` — so a conductor plan could not dispatch phases through codex at all. (The issue text assumed the launcher merely needed re-conformance; on `main` it never implemented the trait.)

## Change

New `crates/edda-conductor/src/agent/codex_rpc.rs` — `CodexLauncher` implementing `AgentLauncher` over the existing client (spawn / open_thread / run_turn / KillOnCancel; protocol layer reused, not rewritten):

- One conductor phase = one codex turn; the app-server child is spawned once and reused across phases; each conductor `session_id` maps to a codex thread resumed via `thread/resume` (create-or-continue parity with pi/claude).
- Outcome mapping: `CodexTurnOutcome` → `AgentDone { cost_usd: None, result_text }`; spawn/protocol failures and cancellation → `AgentCrash`; phase timeout → `Timeout`. Crashed/timed-out/cancelled turns drop the client so the next phase re-spawns fresh, while the thread map preserves conversation continuity.
- **The app-server protocol exposes no usage data, so `cost_usd` is `None` and the budget gate is inert for codex by design** (recorded as `codex-cost-usd=None`, `codex-budget-gate-inert=true`).
- Binary resolution mirrors #528: pure function, `EDDA_CODEX_BIN` override (empty = unset), Windows default `codex.cmd` — verified on this machine that npm ships `codex` (sh) + `codex.cmd` and **no `codex.exe`**.
- `codex_app_server.rs`: test fakes extracted into a shared `pub(crate) mod fake_support` reused by the launcher tests; `spawn_command` made `pub(crate)`. No protocol behavior changes.

`crates/edda-cli/src/cmd_conduct.rs`: `AgentKind::Codex` (value `codex`), `writes_transcripts() == false` so `--tmux` warns and falls back like pi, launcher arm with `verify_available()` probe whose not-found error names the install command and `EDDA_CODEX_BIN`.

## Provenance

Implemented by **pi (z-ai/glm-5.3-flash) dispatched programmatically** via `edda conduct run --agent pi` — a 3-phase conductor plan in an isolated worktree, no desktop automation. The dispatch machinery being extended here is the machinery that produced it.

## Gates (worker receipt, frozen SHA `42dfd0b7457d791350a1fc1196934f5dba0f7f4a`)

| Gate | Result |
|---|---|
| `cargo test -p edda-conductor` | 204 passed (was 182) |
| `cargo test -p edda` | 327 passed |
| `cargo clippy -p edda-conductor -p edda --all-targets -- -D warnings` | clean |

Lane: `worker-1`. Independent verifier round (read-only, verdict pinned to the SHA above) follows as a PR comment before merge.

### DIFF
--- a/crates/edda-cli/src/cmd_conduct.rs
+++ b/crates/edda-cli/src/cmd_conduct.rs
@@ -6,6 +6,7 @@ use edda_conductor::agent::launcher::{phase_session_id, AgentLauncher, ClaudeCod
 use edda_conductor::agent::pi_rpc::PiRpcLauncher;
 use edda_conductor::check::engine::CheckEngine;
 use edda_conductor::plan::parser::load_plan;
+use edda_conductor::plan::schema::Plan;
 use edda_conductor::runner::notify::StdoutNotifier;
 use edda_conductor::runner::sequential::{run_plan, RunContext};
 use edda_conductor::state::machine::{PhaseStatus, PlanState, PlanStatus};
@@ -194,6 +195,10 @@ pub fn run(
     } else {
         false
     };
+
+    if let Some(warning) = budget_warning(&plan, agent) {
+        eprintln!("{warning}");
+    }
 
     // Load or create state
     let mut state = match load_state(&cwd, &plan.name)? {
@@ -494,6 +499,42 @@ pub fn abort(repo_root: &Path, plan_name: Option<&str>) -> Result<()> {
 
 // --- helpers ---
 
+/// One-line startup warning when the selected backend cannot enforce budgets.
+///
+/// codex exposes no cost/usage data, so every phase reports `cost_usd: None`:
+/// the per-phase gate is inert and the sequential runner never feeds the
+/// plan-level `BudgetTracker`, leaving both phase and plan `budget_usd`
+/// unenforced and any printed cost figure a guess. Mirrors the --tmux
+/// warn-and-fall-back tone above.
+fn budget_warning(plan: &Plan, agent: AgentKind) -> Option<String> {
+    let any_budget =
+        plan.budget_usd.is_some() || plan.phases.iter().any(|p| p.budget_usd.is_some());
+    if agent == AgentKind::Codex && any_budget {
+        Some(format!(
+            "Warning: agent \"{}\" exposes no usage data, so budget_usd limits will not \
+             be enforced and reported cost is unavailable.",
+            agent.as_str()
+        ))
+    } else {
+        None
+    }
+}
+
+/// The status cost line.
+///
+/// `PlanState` does not record which agent backend ran, so a zero total
+/// cannot be attributed: usage-free backends like codex report no cost data
+/// at all, while a backend that reports usage overwrites the zero with a
+/// real figure once a phase completes. Printing "$0.00" for a codex run
+/// would assert a spend nobody measured, so a zero total renders as "n/a".
+fn cost_line(total_cost_usd: f64) -> String {
+    if total_cost_usd == 0.0 {
+        "n/a (no usage data reported)".to_owned()
+    } else {
+        format!("${total_cost_usd:.2}")
+    }
+}
+
 fn resolve_plan_name(repo_root: &Path, explicit: Option<&str>) -> Result<String> {
     if let Some(name) = explicit {
         return Ok(name.to_string());
@@ -532,7 +573,7 @@ fn print_status(state: &PlanState) {
     if !state.plan_file.is_empty() {
         println!("  File: {}", state.plan_file);
     }
-    println!("  Cost: ${:.2}", state.total_cost_usd);
+    println!("  Cost: {}", cost_line(state.total_cost_usd));
 
     println!();
     for ps in &state.phases {
@@ -581,6 +622,7 @@ fn ctrlc_cancel(cancel: CancellationToken) {
 mod tests {
     use super::*;
     use clap::Parser;
+    use edda_conductor::plan::parser::parse_plan;
 
     /// Minimal parser harness: `ConductCmd` is a `Subcommand`, so it needs a
     /// root command to be parsed standalone.
@@ -631,7 +673,62 @@ mod tests {
 
     #[test]
     fn run_rejects_unknown_agent() {
-        assert!(TestCli::try_parse_from(["edda", "run", "plan.yaml", "--agent", "gpt"]).is_err());
+        let error = match TestCli::try_parse_from(["edda", "run", "plan.yaml", "--agent", "gpt"]) {
+            Err(error) => error,
+            Ok(_) => panic!("unknown agent must be rejected"),
+        };
+        let text = error.to_string();
+        for expected in ["claude", "pi", "codex"] {
+            assert!(
+                text.contains(expected),
+                "error should list the valid agents, missing {expected:?}: {text}"
+            );
+        }
+    }
+
+    #[test]
+    fn budget_warning_fires_for_codex_with_plan_budget() {
+        let plan = parse_plan("name: t\nphases:\n  - id: a\n    prompt: x\nbudget_usd: 5.0\n")
+            .expect("test plan parses");
+        let warning = budget_warning(&plan, AgentKind::Codex).expect("warning expected");
+        assert!(warning.contains("codex"), "{warning}");
+        assert!(warning.contains("budget_usd"), "{warning}");
+        assert!(warning.contains("not be enforced"), "{warning}");
+        assert!(warning.contains("cost is unavailable"), "{warning}");
+    }
+
+    #[test]
+    fn budget_warning_fires_for_codex_with_phase_budget() {
+        let plan = parse_plan("name: t\nphases:\n  - id: a\n    prompt: x\n    budget_usd: 1.0\n")
+            .expect("test plan parses");
+        assert!(budget_warning(&plan, AgentKind::Codex).is_some());
+    }
+
+    #[test]
+    fn budget_warning_stays_silent_without_a_budget() {
+        let plan =
+            parse_plan("name: t\nphases:\n  - id: a\n    prompt: x\n").expect("test plan parses");
+        assert!(budget_warning(&plan, AgentKind::Codex).is_none());
+    }
+
+    #[test]
+    fn budget_warning_stays_silent_for_other_agents() {
+        let plan = parse_plan("name: t\nphases:\n  - id: a\n    prompt: x\nbudget_usd: 5.0\n")
+            .expect("test plan parses");
+        assert!(budget_warning(&plan, AgentKind::Claude).is_none());
+        assert!(budget_warning(&plan, AgentKind::Pi).is_none());
+    }
+
+    #[test]
+    fn cost_line_reports_na_for_a_zero_total() {
+        // A zero total cannot be attributed to a backend (PlanState records
+        // no agent); codex runs would otherwise print a false "$0.00".
+        assert_eq!(cost_line(0.0), "n/a (no usage data reported)");
+    }
+
+    #[test]
+    fn cost_line_formats_a_reported_total() {
+        assert_eq!(cost_line(1.234), "$1.23");
     }
 
     #[test]
--- a/crates/edda-cli/src/cmd_conduct.rs
+++ b/crates/edda-cli/src/cmd_conduct.rs
@@ -1,6 +1,7 @@
 use anyhow::{bail, Context, Result};
 use clap::Subcommand;
 use edda_conductor::agent::budget::BudgetTracker;
+use edda_conductor::agent::codex_rpc::CodexLauncher;
 use edda_conductor::agent::launcher::{phase_session_id, AgentLauncher, ClaudeCodeLauncher};
 use edda_conductor::agent::pi_rpc::PiRpcLauncher;
 use edda_conductor::check::engine::CheckEngine;
@@ -22,6 +23,8 @@ pub enum AgentKind {
     Claude,
     /// pi coding agent via `pi --mode rpc`
     Pi,
+    /// codex CLI via `codex app-server`
+    Codex,
 }
 
 impl AgentKind {
@@ -29,6 +32,7 @@ impl AgentKind {
         match self {
             AgentKind::Claude => "claude",
             AgentKind::Pi => "pi",
+            AgentKind::Codex => "codex",
         }
     }
 
@@ -38,6 +42,7 @@ impl AgentKind {
         match self {
             AgentKind::Claude => true,
             AgentKind::Pi => false,
+            AgentKind::Codex => false,
         }
     }
 }
@@ -263,6 +268,11 @@ pub fn run(
         }
         AgentKind::Pi => {
             let launcher = PiRpcLauncher::new().with_verbose(verbose);
+            launcher.verify_available()?;
+            Box::new(launcher)
+        }
+        AgentKind::Codex => {
+            let launcher = CodexLauncher::new().with_verbose(verbose);
             launcher.verify_available()?;
             Box::new(launcher)
         }
@@ -613,6 +623,10 @@ mod tests {
             agent_of(parse(&["edda", "run", "plan.yaml", "--agent", "claude"])),
             AgentKind::Claude
         );
+        assert_eq!(
+            agent_of(parse(&["edda", "run", "plan.yaml", "--agent", "codex"])),
+            AgentKind::Codex
+        );
     }
 
     #[test]
@@ -626,7 +640,9 @@ mod tests {
         // otherwise get panes tailing files nobody writes.
         assert!(AgentKind::Claude.writes_transcripts());
         assert!(!AgentKind::Pi.writes_transcripts());
+        assert!(!AgentKind::Codex.writes_transcripts());
         assert_eq!(AgentKind::Claude.as_str(), "claude");
         assert_eq!(AgentKind::Pi.as_str(), "pi");
+        assert_eq!(AgentKind::Codex.as_str(), "codex");
     }
 }
--- a/crates/edda-conductor/src/agent/codex_app_server.rs
+++ b/crates/edda-conductor/src/agent/codex_app_server.rs
@@ -23,10 +23,10 @@ impl CodexAppServer {
     pub async fn spawn(bin: &Path) -> Result<Self> {
         let mut command = Command::new(bin);
         command.arg("app-server");
-        Self::spawn_command(command).await
+        Self::spawn_with_command(command).await
     }
 
-    pub(crate) async fn spawn_command(mut command: Command) -> Result<Self> {
+    async fn spawn_with_command(mut command: Command) -> Result<Self> {
         command
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
@@ -64,6 +64,13 @@ impl CodexAppServer {
             .await?;
         server.notify("initialized").await?;
         Ok(server)
+    }
+
+    /// Test-only wrapper so sibling-module tests (codex_rpc) can drive fake
+    /// app-servers through the same spawn+initialize path as [`Self::spawn`].
+    #[cfg(test)]
+    pub(crate) async fn spawn_command(command: Command) -> Result<Self> {
+        Self::spawn_with_command(command).await
     }
 
     pub async fn open_thread(&mut self, cwd: &Path, resume: Option<&str>) -> Result<String> {
--- a/crates/edda-conductor/src/agent/codex_app_server.rs
+++ b/crates/edda-conductor/src/agent/codex_app_server.rs
@@ -26,7 +26,7 @@ impl CodexAppServer {
         Self::spawn_command(command).await
     }
 
-    async fn spawn_command(mut command: Command) -> Result<Self> {
+    pub(crate) async fn spawn_command(mut command: Command) -> Result<Self> {
         command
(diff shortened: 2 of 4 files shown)


fagemx:
# Code Review: Round 1

**Reviewed SHA (frozen, pinned):** `42dfd0b7457d791350a1fc1196934f5dba0f7f4a`
**Base:** `fee8566f72567f5cb9a89f5ef1ffe0e0fc81b5de` (main), single commit
**Acceptance ceiling:** issue #527 doneWhen
**Verdict: Changes Requested** — P0 = 0, **P1 = 1**, P2 = 1, P3 = 2

Every push invalidates this verdict and requires a new round.

---

## IN SCOPE (frozen)

issue #527 doneWhen; `AgentLauncher` semantic parity with `pi_rpc.rs` (PhaseResult mapping under crash / timeout / cancel / budget; session continuity; fresh-child respawn); Windows bin resolution; `--tmux` warn-fallback via `writes_transcripts()==false`; direct consumers (`cmd_conduct` launcher arm, `mod.rs` export); security / data-loss / cost-control regressions introduced or made reachable by this diff.

**OUT (routed, not blocking):** pre-existing `codex_app_server` protocol behavior; conductor phase-ordering design (already issue #532, not caused by this diff).

---

## Blocking findings

### P1-1 — `budget_usd` silently fails open on `--agent codex`, and the run reports a false `Cost: $0.00`

`crates/edda-conductor/src/agent/codex_rpc.rs` (`drive_turn`), `crates/edda-conductor/src/runner/sequential.rs`, `crates/edda-cli/src/cmd_conduct.rs:535`

The inert **per-phase** gate is documented and tested, and I accept it as a deliberate deviation. The problem is that the actual footprint is **wider than the documented deviation**, and part of it is affirmatively wrong output rather than absent output:

1. **Per-phase gate inert** — `let cost_usd = None;` makes `over_budget(None, _)` permanently false, so `PhaseResult::BudgetExceeded` is unreachable. *This part is documented.*
2. **Plan-level budget also never fires — not documented.** `sequential.rs` records cost only inside `if let Some(cost) = cost_usd { budget.record(cost); state.total_cost_usd += cost; }`. With a permanent `None`, `BudgetTracker::record()` is **never called**, so `is_exhausted()` never becomes true and `plan.budget_usd` is unenforced too. The code comment scopes the deviation to the per-phase gate; the plan-level tracker is silently in the blast radius.
3. **The summary states a false number.** The comment says "the cost column stays empty for codex phases by design." It does not stay empty — `print_status` at `cmd_conduct.rs:535` prints `println!("  Cost: ${:.2}", state.total_cost_usd)`, so **every codex run reports `Cost: $0.00`** regardless of real spend. Empty would be honest; `$0.00` is a wrong figure.
4. **No warning anywhere.** `--dry-run` even prints `Budget: $X.XX` for a plan that will not be budget-enforced under this backend.

**Why this blocks rather than routes.** `--agent codex` was *rejected* before this diff, so this PR is what makes the path reachable — it is introduced-by-change, not pre-existing. It is a spend control the user explicitly configured, failing open, with a concrete false cost figure printed at the end. And the repo's own standard for "requested feature unsupported on the selected backend" is warn-and-fall-back — `cmd_conduct.rs:174-190` does exactly that for `--tmux`, ~90 lines above the launcher arm, and doneWhen item 2 of issue #527 codifies that pattern for this very PR. Budget deserves the same treatment the same issue demands for tmux.

**Minimal discharge** (implementer's choice of shape): in `cmd_conduct::run`, when the selected agent is `Codex` and `plan.budget_usd.is_some()` or any phase sets `budget_usd`, emit an `eprintln!` warning mirroring the tmux one — that codex exposes no usage data, so budget limits will not be enforced and reported cost will read `$0.00`. Optionally distinguish unknown cost from `$0.00` in `print_status`.

A docs/comment-only change does **not** discharge this: the defective surface is user-facing runtime output, and finding 3 is a wrong number on screen.

---

## Non-blocking findings

### P2-1 — `CodexLauncher` doc comment claims cross-phase continuity that cannot occur

`crates/edda-conductor/src/agent/codex_rpc.rs`

The doc comment states: *"A session seen before resumes its thread via `thread/resume`, so report → dispatch-next phases stay in one conversation."*

That cannot happen under the current runner. `sequential.rs:216` derives the session id as `phase_session_id_attempt(&plan.name, &phase_id, attempt)` — plan + **phase_id** + attempt. Every phase therefore gets a **distinct** `session_id`, and every retry gets another one. The `threads` map is written once per key and never read back for a resume, so **`thread/resume` is unreachable in production today**; "report → dispatch-next phases stay in one conversation" is false.

Consequence for the test: `same_session_id_resumes_the_same_conversation` calls `drive_turn` twice with the same `"sid"` — a call pattern the sequential runner never produces. The mechanism itself is implemented correctly and the fake proves it works; the comment generalizes it into a product claim the runner does not deliver.

**Not blocking, because the behavior is right and is exactly pi-parity:** pi likewise gets a fresh `--session-id` per phase+attempt. This is a comment/expectation defect, not a behavioral one. Suggested reword: say the map keys on the conductor session id and resumes when a caller reuses one (forward-looking for `edda dispatch`), and note that the sequential runner assigns a unique session per phase+attempt, so resume does not currently fire.

### P3-1 — `run_rejects_unknown_agent` asserts only `is_err()`

doneWhen item 4 ("unknown-agent error lists all three backends") is satisfied **by construction**: `clap::ValueEnum` emits possible-values in the error, and all three variants are registered. But no assertion pins the message, so the doneWhen is met by the derive rather than by the test. Pre-existing test shape from #523, not introduced here. Optional: assert the error text contains `claude`, `pi`, and `codex`.

### P3-2 — `spawn_command` visibility widened unconditionally

`codex_app_server.rs`: `async fn spawn_command` → `pub(crate) async fn`, but the only out-of-module caller is `#[cfg(test)]`. `#[cfg(test)] pub(crate)` would keep the non-test surface as it was. Harmless.

---

## Verified good (no action)

- **Windows bin resolution** — `resolve_codex_bin` is a faithful structural mirror of `resolve_pi_bin` from #528: pure function, override passed as an argument (no process-env mutation in tests), empty value treated as unset, `codex.cmd` default on Windows with the `PATHEXT`/`CreateProcess` rationale documented. Three tests cover default / override / empty-as-unset.
- **`codex_app_server.rs` refactor — "protocol behavior unchanged" claim VERIFIED.** It is a pure move of the test fakes into `#[cfg(test)] mod fake_support` plus three new scenarios. I diffed the five pre-existing scenario scripts (`MissingThreadId`, `EmptyThreadId`, `RunTurnInterleaving`, `RunTurnError`, `Idle`) in **both** the PowerShell and `sh` variants — byte-identical. No production protocol code changed.
- **PhaseResult parity with `pi_rpc.rs`** — timeout → `Timeout`; cancellation → `AgentCrash { error: "conductor shutdown" }`, string-identical to pi; RPC/turn error → `AgentCrash`. `MaxTurns` is claude-only in both.
- **Fresh-child respawn** — `keep_server=false` on all three of crash / timeout / cancel, so `run_phase` clears `state.server`; `Drop for CodexAppServer` calls `start_kill()` and the child is spawned with `kill_on_drop(true)`, so the child does die on the dropped-future paths. Respawn is proven by `run_phase_survives_a_crashed_server_by_respawning`.
- **`--tmux` fallback** — `Codex.writes_transcripts() == false`; the existing generic fallback covers it; asserted in `only_claude_writes_transcripts`. doneWhen 2 met.
- **Controller attention point 1 — no coverage gap.** The parse tests were **updated in place**: `run_accepts_explicit_agents` and `only_claude_writes_transcripts` grew codex assertions rather than gaining new `#[test]` fns. That is precisely why the `edda` count stayed 327 with only +16 lines. pi's recorded claim ("parse tests cover codex value/as_str/transcript flag") checks out against the diff. doneWhen 3 met.
- **The new tests really do run on Windows** — `codex_rpc.rs` has no `cfg(unix)`/`cfg(windows)` gating on any test, and `fake_support` ships `#[cfg(windows)]` PowerShell script variants for all three new scenarios. So green `Test (windows-latest)` on this head is real coverage, not silently-skipped tests.

**doneWhen:** 1 ✅ (launcher arm + `run_phase_completes_against_a_fake_server`; no real-binary e2e was run by anyone — acceptable, it needs codex installed) · 2 ✅ · 3 ✅ · 4 ✅ by construction (see P3-1).

---

## Evidence — RAN vs READ

**RAN: nothing. Lane: n/a — no local build.**

**READ:**
- Exact-head CI: all **7** required checks SUCCESS, and I verified the binding rather than trusting the rollup — `gh run view 33361156472` reports `headSha=42dfd0b7457d791350a1fc1196934f5dba0f7f4a`, `conclusion=success`. That includes `Test (windows-latest)` and `Clippy (windows-latest)`.
- Worker L0 receipt for this SHA (edda-conductor 204 passed / edda 327 passed / clippy clean / fmt ok), lane worker-1.
- Full diff of all 4 changed files at the frozen SHA, plus `pi_rpc.rs`, `launcher.rs`, `sequential.rs`, `budget.rs`, and the `cmd_conduct.rs` tmux/budget surfaces as unchanged context.

**Why no local rerun.** Both touched crates (`edda-conductor`, `edda`) are inside the CI Windows test subset **and** inside full-workspace `cargo test` on ubuntu + macos, so the frozen surface is covered on the exact head. The out-of-subset Windows gap I flagged when accepting this assignment does not apply to this diff. Per the verification ladder a full local rerun needs a stated reason — no receipt, red or absent CI, or grounds to distrust the receipt — and none of the three holds. Both findings above are design/documentation judgments that **no** test run would have surfaced, so a lane build would have re-derived a green result at ~40 GB and bought nothing.

**Cost:** 1 reviewer session, ~15 read-only tool calls, 0 builds, 0 CI reruns, 0 bytes written to the repo. Token/elapsed instrumentation not available in this session.

---

**Verdict: Changes Requested** @ `42dfd0b7457d791350a1fc1196934f5dba0f7f4a` — one P1 (budget fail-open + false `$0.00`). P2-1 and both P3s are non-blocking; P2-1 is cheap to fold into the same push.

fagemx:
# Review Response: Round 1

New head: `c8ba38cd420686abe38641ffa4f34e5517186fbc` (one commit on top of the reviewed `42dfd0b7457d791350a1fc1196934f5dba0f7f4a`; no amend, no rebase). Implemented by pi via `edda conduct run --agent pi` round-2 dispatch; this comment is posted by the controller on the implementation side's behalf.

## P1-1 (blocking) — budget fail-open + false `Cost: $0.00` → FIXED
- New pure helper `budget_warning(plan, agent)` in `cmd_conduct.rs`: when the selected agent is `Codex` and `plan.budget_usd` or any phase `budget_usd` is set, `edda conduct run` emits one startup `eprintln!` warning ("budget_usd limits will not be enforced and reported cost is unavailable"), mirroring the `--tmux` warn-and-fall-back tone.
- New pure helper `cost_line(total)`: a zero total renders `Cost: n/a (no usage data reported)` instead of asserting `$0.00`; a non-zero total renders the figure. Doc comment states the attribution honestly: `PlanState` does not record the backend, so a zero cannot be attributed — printing `$0.00` would assert a spend nobody measured. (Recorded decision: `conduct.codex-budget=honesty-over-figure`.)
- The deviation comment in `codex_rpc.rs` now names the plan-level `BudgetTracker` in the blast radius (runner records only inside `if let Some(cost)`), not just the per-phase gate.
- Both helpers are unit-tested (warning fires only for codex+budget; cost line zero/non-zero).

## P2-1 — doc-comment overclaim → FIXED
The `CodexLauncher` doc now describes the `threads` map as keyed on the conductor session id, resuming only when a caller reuses one — a forward-looking path for a future dispatch primitive — and states that the sequential runner derives a unique session per plan+phase+attempt so resume does not currently fire in production (cross-references `phase_session_id`). The resume test's comment now says it exercises the forward-looking path. (Recorded decision: `conductor.codex-resume=forward-looking-only`.)

## P3-1 — FIXED
`run_rejects_unknown_agent` now asserts the clap error text lists `claude`, `pi`, and `codex`.

## P3-2 — FIXED
Non-test surface narrowed: private `spawn_with_command` plus a `#[cfg(test)] pub(crate) spawn_command` wrapper documented as the sibling-module test entry.

## Gates (worker receipt, frozen SHA `c8ba38cd420686abe38641ffa4f34e5517186fbc`)

| Gate | Result |
|---|---|
| `cargo test -p edda-conductor` | 204 passed |
| `cargo test -p edda` | 333+11 passed (was 327) |
| `cargo clippy -p edda-conductor -p edda --all-targets -- -D warnings` | clean |

Lane worker-1. Exact-head CI on the new SHA is running; round 2 review requested.

fagemx:
# Code Review: Round 2

**Reviewed SHA (frozen, pinned):** `c8ba38cd420686abe38641ffa4f34e5517186fbc`
**Round-1 SHA:** `42dfd0b7457d791350a1fc1196934f5dba0f7f4a` — verified as the direct parent of this head (`git merge-base --is-ancestor` passes, `c8ba38c^ == 42dfd0b`). No amend, no rebase, so the round-1 verdict stays historically valid for its own SHA and this round reviews the delta only.
**Verdict: LGTM** — P0 = 0, **P1 = 0**, P2 = 0. One follow-up routed, non-blocking.

Any further push voids this verdict and requires a round 3.

---

## Round-1 findings — discharge verified

I checked each against the code at this SHA rather than accepting the response's claims.

### P1-1 — budget fail-open + false `$0.00` — **DISCHARGED**

All four sub-points I raised are closed, including the two the original comment did not cover:

- **Warning exists and is correctly scoped.** `budget_warning()` fires only when the agent is `Codex` **and** `plan.budget_usd.is_some() || plan.phases.iter().any(|p| p.budget_usd.is_some())` — so it catches the plan-level budget *and* a phase-level one, which was the half that was undocumented in round 1. Emitted via `eprintln!`, matching the `--tmux` warn-and-fall-back tone as requested.
- **It fires before `--dry-run` too.** The call sits at line 199; the `if dry_run` block that prints `Budget: $X.XX` is at line 225. So the dry-run path that advertised an unenforceable budget now carries the warning ahead of it. That closes round-1 point 4, which the response did not explicitly claim.
- **The false figure is gone.** `print_status` now renders through `cost_line()` instead of `println!("  Cost: ${:.2}")`.
- **Blast radius is now documented where it lives.** Both the `budget_warning` doc comment and the `drive_turn` comment now state that the sequential runner only records inside `if let Some(cost)`, so `BudgetTracker` never sees a figure and *both* phase and plan `budget_usd` are unenforced. That is precisely the gap I reported; the comment no longer under-describes it.
- **Tests:** four for `budget_warning` (plan budget / phase budget / no budget / non-codex agents) and two for `cost_line`. The `edda` count 327 → 333 reconciles exactly against those six new tests, so the receipt is internally consistent.

### P2-1 — overclaiming continuity doc comment — **DISCHARGED**

Reworded to say the map resumes "whenever a caller reuses a session id", explicitly labels it "a forward-looking path for a future dispatch primitive", and states that the sequential runner derives a unique id per plan+phase+attempt "so resume does not currently fire in production", with a cross-reference to `phase_session_id`. The test comment on `same_session_id_resumes_the_same_conversation` now says the same. The claim now matches the behavior.

### P3-1 — `is_err()`-only assertion — **DISCHARGED**

`run_rejects_unknown_agent` now asserts the error text contains `claude`, `pi`, and `codex`. doneWhen item 4 of issue #527 is now pinned by a test rather than resting on the derive.

### P3-2 — widened visibility — **DISCHARGED**

Production path is now private `spawn_with_command`; the `pub(crate) spawn_command` wrapper is `#[cfg(test)]`. Non-test surface is back to what it was before this PR.

---

## The flagged judgment call — `cost_line` zero-attribution: acceptable

You flagged honestly that `cost_line` maps **any** zero total to `"n/a (no usage data reported)"`, including a hypothetical genuine `$0.00` claude run. My ruling: **acceptable for this round, no change required.**

- The only genuinely-wrong case is a completed claude/pi run that recorded `Some(0.0)` and finished at exactly zero. That is not reachable in practice — a real turn costs something.
- Every other zero is *more* honest than before, not less: an unrun or aborted-before-first-cost plan now says "no usage data reported", which is literally true, where it used to assert `$0.00`.
- The trade runs in the safe direction. The old line asserted a measured spend nobody measured; the new one under-claims. For a cost figure, under-claiming is the right failure mode.
- The doc comment owns the tradeoff explicitly, and the recorded decision `conduct.codex-budget=honesty-over-figure` captures it.

The durable fix is a state-schema change (`Option<f64>`, or record the backend in `PlanState`), which is correctly larger than this PR's scope. See the follow-up below.

---

## FOLLOW-UP ISSUE (routed, does not extend this PR)

**`PlanCompleted.total_cost_usd` still emits a bare `0.0` for codex runs.**

`runner/event_log.rs` types the event as `PlanCompleted { phases_passed: usize, total_cost_usd: f64 }`, and `sequential.rs:465` fills it from `state.total_cost_usd`. So the JSONL event log and `--json` consumers still receive `"total_cost_usd":0.0` for a codex run — the same "asserting an unmeasured zero" shape that `cost_line` just fixed on the human-readable surface. Note `PhasePassed` already models this correctly as `cost_usd: Option<f64>`; only the plan-level roll-up is a bare `f64`.

**This is routed, not blocking, and I want to be straight about why:** it is not fix-caused, and it was equally observable at `42dfd0b` — I found `print_status` in round 1 and did not grep the event log. Under the loop rules a later-round blocker must be fix-caused or previously unobservable, and this is neither, so it does not get to extend this PR on my account. It shares a root cause with the `cost_line` tradeoff above — cost modeled as a bare `f64` that cannot express "unknown" — so both belong in one follow-up.

Nothing else in the round-2 delta is in question. `cmd_controls.rs` also prints a total cost, but that reads `execution_event`s from the ledger for the model-quality report and is unrelated to conduct plan state — not a finding.

---

## Evidence — RAN vs READ

**RAN: nothing. Lane: n/a — no local build.**

**READ:**
- **Ancestry verified, not assumed:** `c8ba38c^` resolves to `42dfd0b`, and `git merge-base --is-ancestor 42dfd0b c8ba38c` passes — confirming your "no amend, no rebase" claim.
- **Exact-head CI, binding verified by run id:** `gh run list` shows run `33364297233` with `headSha=c8ba38cd420686abe38641ffa4f34e5517186fbc`. I waited for the run to complete rather than ruling on a partial board — at first look only 5 of 7 were in, with `Test (macos-latest)` and `Test (windows-latest)` still pending. Final state: **all 7 required checks SUCCESS**, including `Test (windows-latest)` and `Clippy (windows-latest)`. Head re-confirmed as `c8ba38c` after the wait.
- Full delta of all 3 changed files, plus `event_log.rs`, `sequential.rs`, `cmd_controls.rs`, and `runner/edda.rs` swept for any other surface that renders a cost from a permanent `None`.

**Why no local rerun.** Unchanged from round 1 and still sound: both touched crates sit inside the CI Windows test subset and inside full-workspace `cargo test` on ubuntu + macos, the receipt is present and reconciles arithmetically (327 → 333 = the six new tests), and exact-head CI is fully green. The ladder requires a stated reason for a full local rerun — no receipt, red or absent CI, or grounds to distrust the receipt — and none holds.

**Cost:** 1 reviewer session, 2 rounds total, ~10 read-only tool calls this round, 0 builds, 0 CI reruns, 0 bytes written to the repo. One CI wait (~a few minutes) rather than an early verdict on a partial board.

---

**Verdict: LGTM** @ `c8ba38cd420686abe38641ffa4f34e5517186fbc` — P0 = 0, P1 = 0. All four round-1 findings verified discharged in code; the `cost_line` tradeoff is accepted; one pre-existing event-log cost surface routed as a follow-up.

Merge authority remains the operator's. Any push after this comment voids this verdict.

Issue #533, opened 2026-08-31
fix(conductor): model cost measured-ness at plan level — event log still emits unmeasured total_cost_usd 0.0 for usage-free backends


fagemx:
## Context

Routed from PR #531 round-2 review (verifier finding, explicitly non-blocking there: pre-existing shape, equally observable at the previous head). #531 fixed the human-facing surface — `print_status` now renders a zero total as `Cost: n/a (no usage data reported)` instead of asserting `$0.00` (decision `conduct.codex-budget=honesty-over-figure`). Two remaining surfaces share the same root cause: **cost measured-ness is not modeled at the plan roll-up level.**

## The two surfaces

1. **Event log emits an unmeasured zero.** `event_log.rs` types `PlanCompleted { total_cost_usd: f64 }` and `sequential.rs:465` fills it from `state.total_cost_usd` — so the JSONL / `--json` stream still emits `"total_cost_usd": 0.0` for a codex run, the exact unmeasured-zero shape #531 removed from the human surface. `PhasePassed` already models it correctly as `Option<f64>`; only the plan roll-up is a bare `f64`.
2. **`cost_line`'s accepted ambiguity.** The helper maps ANY zero total to "n/a" because `PlanState` does not record which backend ran; a hypothetical genuine `Some(0.0)` run would be mislabeled. Accepted in review as the right failure mode (under-claiming beats asserting an unmeasured figure) — but it is a workaround for the same missing model.

## Fix direction

Model measured-ness once, at the source: make the plan roll-up an `Option<f64>` (None until any phase records a measured cost), or track `cost_measured: bool` alongside the total in `PlanState`. Then `PlanCompleted`, `print_status`, and any `--json` consumer render honestly from the model instead of inferring from a sentinel zero. `PhasePassed`'s `Option<f64>` is the pattern to follow. Downstream consumers of the JSONL schema must be checked for the field-type change (or a sibling field added for compatibility).

## doneWhen

- [ ] A codex (usage-free) run's `PlanCompleted` event does not assert `total_cost_usd: 0.0` as a measured figure
- [ ] A genuine measured total (claude/pi) still round-trips exactly as today
- [ ] `cost_line` derives from the model rather than the zero sentinel
- [ ] JSONL schema change is deliberate: consumers audited or compatibility preserved

🤖 Routed from PR #531 round-2 review by the fleet controller session.

fagemx:
[...]

PR #534, opened 2026-08-31
feat(edda-cli): edda dispatch — single-turn agent verb (#526)

### LINKED ISSUE #526: feat(cli): edda dispatch — single-turn agent dispatch decoupled from the plan loop
## Motivation (cross-platform fleet design, 2026-08-31)

The target fleet topology is three **platform-swappable roles**: an LLM orchestrator (Claude), an executor (pi), a verifier (Claude Opus or Codex, chosen per task). The hand-off contract that makes roles swappable lives in the ledger + repo — brief → frozen SHA → receipt → verdict, every object bound to the full SHA — and is platform-neutral by construction (already proven cross-platform by the codex-driven PR series, e.g. #515–#517). The only platform-specific part is the doorbell: how to start or continue **one turn** on each agent.

Conductor already owns the right machinery — `AgentLauncher::run_phase` is exactly "one turn in, result out, with budget/timeout/cancel" — but the only way to reach it is `edda conduct run` with a plan DAG. When loop control belongs to an LLM orchestrator (judgment between turns), forcing every dispatch through a plan file inverts control: the orchestrator ends up authoring a one-phase plan just to ring an agent.

## Proposal

Expose the launcher layer as a standalone verb:

```text
edda dispatch --agent <claude|pi|codex> --session-id <id> --prompt-file <f>
              [--cwd <dir>] [--budget-usd <x>] [--timeout-sec <n>] [--json]
```

- One turn: send prompt, collect result, print outcome (done/crash/timeout/budget-exceeded + result text + cost); `--json` for machine consumption; exit code reflects outcome class.
- `--session-id` keeps each backend's create-or-continue semantics (pi and claude already support this) — enables the reject-comment re-dispatch pattern from #519 without conductor owning the loop.
- No plan file, no DAG, no state machine — loop control stays with the caller.
- Reuse `AgentKind` and the existing launchers; `conduct run` remains an unchanged consumer of the same layer.

## doneWhen

- [ ] `edda dispatch --agent pi` runs one turn end-to-end and prints result + cost, exit code by outcome class
- [ ] Same `--session-id` across two dispatch calls preserves context (create-or-continue)
- [ ] Budget/timeout/cancel behave as in `conduct run` (shared code path, covered by tests)
- [ ] `--json` output shape documented

Related: #518 (adapter layer), #519 (the gate loop this primitive lets an external orchestrator drive), #520 (doorbell for the orchestrator itself).

🤖 Drafted with Claude Code from the cross-platform fleet design discussion.

fagemx:
Closes #526

## Problem

The launcher layer (`AgentLauncher::run_phase` — one turn in, result out, with budget/timeout/cancel) was reachable only through `edda conduct run` and a plan DAG. An external orchestrator (an LLM controller driving the loop itself) had to author a one-phase plan file just to ring an agent — control inverted, per the cross-platform fleet design (ledger data plane / adapter control plane).

## Change

New verb, loop control stays with the caller:

```text
edda dispatch --agent <claude|pi|codex> --prompt-file <f> [--session-id <id>]
              [--cwd <dir>] [--budget-usd <x>] [--timeout-sec <n>] [--json]
```

- `AgentKind` extracted to `crates/edda-cli/src/agent_kind.rs`, shared by conduct + dispatch (`pub use`) — same three values, same unknown-agent error, no duplication.
- Launcher construction mirrors conduct exactly: `verify_available()` probe, shared `budget_warning_for_agent()` (codex + `--budget-usd` warns enforcement is inert — the #531 semantics carried over).
- Synthetic `Phase` (id `dispatch`) through `AgentLauncher::run_phase` — the same code path conduct uses; nothing forked.
- `--session-id` verbatim for create-or-continue; omitted → generated and printed (`Session:` line / json field) so the caller can chain turns.
- Exit codes by outcome class (documented in long help): 0 done · 2 timeout · 3 budget exceeded · 4 max turns · 1 crash/other.
- `--json`: exactly one object on stdout `{outcome, result_text, cost_usd, session_id, error}`; warnings to stderr. Honest cost rendering (`n/a` without usage data) shared with conduct.

## Provenance

Implemented by pi (glm-5.3-flash) via `edda conduct run --agent pi`, 2-phase plan, isolated worktree. Second delivery through this pipeline (#531 was the first).

## Gates (worker receipt, frozen SHA `b88ab4e2a8745c61fb12265d2510d48ab73bd710`)

| Gate | Result |
|---|---|
| `cargo test -p edda` | 361 passed (was 333) |
| `cargo clippy -p edda -p edda-conductor --all-targets -- -D warnings` | clean |

Lane worker-1. Independent verifier round (read-only, verdict pinned to the SHA above) follows as a PR comment before merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -692,6 +692,7 @@ name = "edda"
 version = "0.3.0"
 dependencies = [
  "anyhow",
+ "async-trait",
  "chrono",
  "clap",
  "crossterm",
--- a/crates/edda-cli/Cargo.toml
+++ b/crates/edda-cli/Cargo.toml
@@ -58,3 +58,4 @@ tui = ["ratatui", "crossterm"]
 
 [dev-dependencies]
 tempfile.workspace = true
+async-trait = "0.1"
--- a/crates/edda-cli/src/agent_kind.rs
+++ b/crates/edda-cli/src/agent_kind.rs
@@ -1,0 +1,52 @@
+//! Shared agent-backend selection, used by both `edda conduct` (plan runs)
+//! and `edda dispatch` (single-turn runs). The enum, its string form, and
+//! the transcript capability flag live here so neither command duplicates
+//! the other's backend list.
+
+/// Which agent backend runs the plan's phases.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, clap::ValueEnum)]
+pub enum AgentKind {
+    /// Claude Code via `claude -p` (default)
+    Claude,
+    /// pi coding agent via `pi --mode rpc`
+    Pi,
+    /// codex CLI via `codex app-server`
+    Codex,
+}
+
+impl AgentKind {
+    pub fn as_str(self) -> &'static str {
+        match self {
+            AgentKind::Claude => "claude",
+            AgentKind::Pi => "pi",
+            AgentKind::Codex => "codex",
+        }
+    }
+
+    /// Whether this backend tees per-phase transcripts to disk, which the
+    /// tmux phase panes tail.
+    pub fn writes_transcripts(self) -> bool {
+        match self {
+            AgentKind::Claude => true,
+            AgentKind::Pi => false,
+            AgentKind::Codex => false,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn only_claude_writes_transcripts() {
+        // Drives the --tmux fallback: an agent without transcripts would
+        // otherwise get panes tailing files nobody writes.
+        assert!(AgentKind::Claude.writes_transcripts());
+        assert!(!AgentKind::Pi.writes_transcripts());
+        assert!(!AgentKind::Codex.writes_transcripts());
+        assert_eq!(AgentKind::Claude.as_str(), "claude");
+        assert_eq!(AgentKind::Pi.as_str(), "pi");
+        assert_eq!(AgentKind::Codex.as_str(), "codex");
+    }
+}
--- a/crates/edda-cli/src/agent_kind.rs
+++ b/crates/edda-cli/src/agent_kind.rs
@@ -1,7 +1,13 @@
 //! Shared agent-backend selection, used by both `edda conduct` (plan runs)
-//! and `edda dispatch` (single-turn runs). The enum, its string form, and
-//! the transcript capability flag live here so neither command duplicates
-//! the other's backend list.
+//! and `edda dispatch` (single-turn runs). The enum, its string form, the
+//! transcript capability flag, and the launcher construct-and-probe factory
+//! live here so neither command duplicates the other's backend list.
+
+use anyhow::Result;
+use edda_conductor::agent::codex_rpc::CodexLauncher;
+use edda_conductor::agent::launcher::{AgentLauncher, ClaudeCodeLauncher};
+use edda_conductor::agent::pi_rpc::PiRpcLauncher;
+use std::path::PathBuf;
 
 /// Which agent backend runs the plan's phases.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, clap::ValueEnum)]
@@ -32,6 +38,49 @@ impl AgentKind {
             AgentKind::Codex => false,
         }
     }
+}
+
+// ── Launcher factory ──
+
+/// Per-backend options for [`build_launcher`]: verbose streaming and, for
+/// the backend that tees transcripts, the directory to tee into.
+pub(crate) struct LauncherOptions {
+    /// Stream live agent activity while the phase runs.
+    pub verbose: bool,
+    /// If set, claude captures raw agent stdout to
+    /// `{transcript_dir}/{phase_id}-{session_id_prefix}.jsonl`.
+    pub transcript_dir: Option<PathBuf>,
+}
+
+/// Construct and probe (`verify_available`) the launcher for `agent`.
+///
+/// This is the single construct-and-probe table: the backend list exists in
+/// exactly one place, shared by `edda conduct run` (which threads
+/// `--verbose` and a transcript dir for claude) and `edda dispatch` (which
+/// uses defaults). pi has no transcript tee capability yet, so the
+/// transcript dir is claude-only — see [`AgentKind::writes_transcripts`].
+pub(crate) fn build_launcher(
+    agent: AgentKind,
+    options: LauncherOptions,
+) -> Result<Box<dyn AgentLauncher>> {
+    Ok(match agent {
+        AgentKind::Claude => {
+            let mut launcher = ClaudeCodeLauncher::new().with_verbose(options.verbose);
+            launcher.transcript_dir = options.transcript_dir;
+            launcher.verify_available()?;
+            Box::new(launcher)
+        }
+        AgentKind::Pi => {
+            let launcher = PiRpcLauncher::new().with_verbose(options.verbose);
+            launcher.verify_available()?;
+            Box::new(launcher)
+        }
+        AgentKind::Codex => {
+            let launcher = CodexLauncher::new().with_verbose(options.verbose);
+            launcher.verify_available()?;
+            Box::new(launcher)
+        }
+    })
 }
 
 #[cfg(test)]
--- a/crates/edda-cli/src/cmd_conduct.rs
+++ b/crates/edda-cli/src/cmd_conduct.rs
@@ -1,3 +1,4 @@
+use crate::agent_kind::AgentKind;
 use anyhow::{bail, Context, Result};
 use clap::Subcommand;
 use edda_conductor::agent::budget::BudgetTracker;
@@ -14,39 +15,6 @@ use edda_conductor::state::persist::{load_state, save_state};
 use edda_conductor::tmux::TmuxSession;
 use std::path::Path;
 use tokio_util::sync::CancellationToken;
-
-// ── Agent selection ──
-
-/// Which agent backend runs the plan's phases.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, clap::ValueEnum)]
-pub enum AgentKind {
-    /// Claude Code via `claude -p` (default)
-    Claude,
-    /// pi coding agent via `pi --mode rpc`
-    Pi,
-    /// codex CLI via `codex app-server`
-    Codex,
-}
-
-impl AgentKind {
-    fn as_str(self) -> &'static str {
-        match self {
-            AgentKind::Claude => "claude",
-            AgentKind::Pi => "pi",
-            AgentKind::Codex => "codex",
-        }
-    }
-
-    /// Whether this backend tees per-phase transcripts to disk, which the
-    /// tmux phase panes tail.
-    fn writes_transcripts(self) -> bool {
-        match self {
-            AgentKind::Claude => true,
-            AgentKind::Pi => false,
-            AgentKind::Codex => false,
-        }
-    }
-}
 
 // ── CLI Schema ──
 
@@ -509,6 +477,12 @@ pub fn abort(repo_root: &Path, plan_name: Option<&str>) -> Result<()> {
 fn budget_warning(plan: &Plan, agent: AgentKind) -> Option<String> {
     let any_budget =
         plan.budget_usd.is_some() || plan.phases.iter().any(|p| p.budget_usd.is_some());
+    budget_warning_for_agent(agent, any_budget)
+}
+
+/// Backend + has-any-budget form of [`budget_warning`], shared with
+/// `edda dispatch`, which has flags instead of a plan file.
+pub(crate) fn budget_warning_for_agent(agent: AgentKind, any_budget: bool) -> Option<String> {
     if agent == AgentKind::Codex && any_budget {
         Some(format!(
             "Warning: agent \"{}\" exposes no usage data, so budget_usd limits will not \
@@ -527,7 +501,7 @@ fn budget_warning(plan: &Plan, agent: AgentKind) -> Option<String> {
 /// at all, while a backend that reports usage overwrites the zero with a
 /// real figure once a phase completes. Printing "$0.00" for a codex run
 /// would assert a spend nobody measured, so a zero total renders as "n/a".
-fn cost_line(total_cost_usd: f64) -> String {
+pub(crate) fn cost_line(total_cost_usd: f64) -> String {
     if total_cost_usd == 0.0 {
         "n/a (no usage data reported)".to_owned()
     } else {
@@ -720,6 +694,15 @@ mod tests {
     }
 
     #[test]
+    fn budget_warning_for_agent_fires_on_codex_with_a_budget() {
+        // The flag form shared with `edda dispatch`.
+        assert!(budget_warning_for_agent(AgentKind::Codex, true).is_some());
+        assert!(budget_warning_for_agent(AgentKind::Codex, false).is_none());
+        assert!(budget_warning_for_agent(AgentKind::Claude, true).is_none());
+        assert!(budget_warning_for_agent(AgentKind::Pi, true).is_none());
+    }
+
+    #[test]
     fn cost_line_reports_na_for_a_zero_total() {
         // A zero total cannot be attributed to a backend (PlanState records
         // no agent); codex runs would otherwise print a false "$0.00".
@@ -729,17 +712,5 @@ mod tests {
     #[test]
     fn cost_line_formats_a_reported_total() {
         assert_eq!(cost_line(1.234), "$1.23");
-    }
-
-    #[test]
-    fn only_claude_writes_transcripts() {
-        // Drives the --tmux fallback: an agent without transcripts would
-        // otherwise get panes tailing files nobody writes.
-        assert!(AgentKind::Claude.writes_transcripts());
-        assert!(!AgentKind::Pi.writes_transcripts());
-        assert!(!AgentKind::Codex.writes_transcripts());
-        assert_eq!(AgentKind::Claude.as_str(), "claude");
-        assert_eq!(AgentKind::Pi.as_str(), "pi");
-        assert_eq!(AgentKind::Codex.as_str(), "codex");
     }
 }
--- a/crates/edda-cli/src/cmd_conduct.rs
+++ b/crates/edda-cli/src/cmd_conduct.rs
@@ -1,10 +1,8 @@
-use crate::agent_kind::AgentKind;
+use crate::agent_kind::{build_launcher, AgentKind, LauncherOptions};
 use anyhow::{bail, Context, Result};
 use clap::Subcommand;
 use edda_conductor::agent::budget::BudgetTracker;
-use edda_conductor::agent::codex_rpc::CodexLauncher;
-use edda_conductor::agent::launcher::{phase_session_id, AgentLauncher, ClaudeCodeLauncher};
-use edda_conductor::agent::pi_rpc::PiRpcLauncher;
+use edda_conductor::agent::launcher::phase_session_id;
 use edda_conductor::check::engine::CheckEngine;
 use edda_conductor::plan::parser::load_plan;
 use edda_conductor::plan::schema::Plan;
@@ -230,26 +228,13 @@ pub fn run(
         .join(&plan.name)
         .join("transcripts");
 
-    // pi has no transcript tee capability yet; the transcript dir stays
-    // claude-only and tmux panes only mirror it when claude is selected.
-    let launcher: Box<dyn AgentLauncher> = match agent {
-        AgentKind::Claude => {
-            let mut launcher = ClaudeCodeLauncher::new().with_verbose(verbose);
-            launcher.transcript_dir = Some(transcript_dir.clone());
-            launcher.verify_available()?;
-            Box::new(launcher)
-        }
-        AgentKind::Pi => {
-            let launcher = PiRpcLauncher::new().with_verbose(verbose);
-            launcher.verify_available()?;
-            Box::new(launcher)
-        }
-        AgentKind::Codex => {
-            let launcher = CodexLauncher::new().with_verbose(verbose);
-            launcher.verify_available()?;
-            Box::new(launcher)
-        }
-    };
+    let launcher = build_launcher(
+        agent,
+        LauncherOptions {
+            verbose,
+            transcript_dir: Some(transcript_dir.clone()),
+        },
+    )?;
     let engine = CheckEngine::new(cwd.clone());
     let notifier = StdoutNotifier;
     let mut budget = BudgetTracker::new(plan.budget_usd);
@@ -494,6 +479,10 @@ pub(crate) fn budget_warning_for_agent(agent: AgentKind, any_budget: bool) -> Op
     }
 }
 
+/// The honest stand-in for a cost figure nobody measured, shared with
+/// `edda dispatch`'s no-usage rendering so the string has one source.
+pub(crate) const NO_USAGE_COST_TEXT: &str = "n/a (no usage data reported)";
+
 /// The status cost line.
 ///
 /// `PlanState` does not record which agent backend ran, so a zero total
@@ -503,7 +492,7 @@ pub(crate) fn budget_warning_for_agent(agent: AgentKind, any_budget: bool) -> Op
 /// would assert a spend nobody measured, so a zero total renders as "n/a".
 pub(crate) fn cost_line(total_cost_usd: f64) -> String {
     if total_cost_usd == 0.0 {
-        "n/a (no usage data reported)".to_owned()
+        NO_USAGE_COST_TEXT.to_owned()
     } else {
         format!("${total_cost_usd:.2}")
     }
--- a/crates/edda-cli/src/cmd_dispatch.rs
+++ b/crates/edda-cli/src/cmd_dispatch.rs
@@ -11,14 +11,11 @@
 //! - 3 = budget exceeded
 //! - 4 = max turns
 
-use crate::agent_kind::AgentKind;
-use crate::cmd_conduct::{budget_warning_for_agent, cost_line};
+use crate::agent_kind::{build_launcher, AgentKind, LauncherOptions};
+use crate::cmd_conduct::{budget_warning_for_agent, cost_line, NO_USAGE_COST_TEXT};
 use anyhow::{Context, Result};
 use clap::Args;
-use edda_conductor::agent::codex_rpc::CodexLauncher;
-use edda_conductor::agent::launcher::PhaseResult;
-use edda_conductor::agent::launcher::{phase_session_id, AgentLauncher, ClaudeCodeLauncher};
-use edda_conductor::agent::pi_rpc::PiRpcLauncher;
+use edda_conductor::agent::launcher::{phase_session_id, AgentLauncher, PhaseResult};
 use edda_conductor::plan::schema::Phase as PhaseSchema;
 use tokio_util::sync::CancellationToken;
 
@@ -33,9 +30,13 @@ pub struct DispatchArgs {
     /// Path to a file containing the prompt (read verbatim)
     #[arg(long)]
     pub prompt_file: String,
-    /// Session id passed to the backend verbatim (create-or-continue is the
-    /// backend's concern). Generated and printed when omitted so the caller
-    /// can reuse it on the next call.
+    /// Session id passed to the backend verbatim. Continuity semantics are
+    /// per-backend: claude and pi persist conversations externally, so the
+    /// same id resumes the prior conversation across invocations; codex
+    /// keeps its thread state in-process, so each `edda dispatch`
+    /// invocation starts a fresh conversation no matter what id you pass
+    /// (a warning is printed). Generated and printed when omitted so the
+    /// caller can reuse it on the next call.
     #[arg(long)]
     pub session_id: Option<String>,
     /// Working directory for the agent (default: current directory)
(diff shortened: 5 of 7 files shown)


fagemx:
# Code Review: Round 1

**Reviewed SHA (frozen, pinned):** `b88ab4e2a8745c61fb12265d2510d48ab73bd710`
**Base:** `64f9b4a0ced4a2ef0cd55e4a3f0717be5796ca71` (main after PR #531) — verified as the direct parent, single commit.
**Acceptance ceiling:** issue #526 doneWhen
**Verdict: Changes Requested** — P0 = 0, **P1 = 1**, P2 = 2, P3 = 3

Every push invalidates this verdict and requires a new round.

---

## CI: the red was a pre-existing flake, not this PR — now green

`Test (ubuntu-latest)` was FAILURE on the exact head when I picked this up. I did not take that at face value in either direction. The failing test is `agent::pi_rpc::tests::startup_stderr_reaches_the_crash_error`, which fails with `Error: failed writing pi RPC stdin` — an `Err` from the harness, not an assertion mismatch — in **`edda-conductor`, a crate this PR does not touch** (diffstat confirms; the controller's claim audits clean).

The decisive evidence: **the identical test failed with the identical error on `main` at `fee8566f`** (run `33356986628`), a commit that predates this PR entirely. The test is itself a race regression guard — its own comment says "the reason must be present every run, not whenever the drain task happens to win" — and it loops 5×, writing to the stdin of a fake child whose scenario is literally `StderrThenDie`. On a loaded ubuntu runner the child wins the race and the write gets EPIPE.

That is environmental, so per the ladder I re-ran **only the failed job** rather than blocking on it or spending a full local run. **All 7 required checks are now SUCCESS** on `b88ab4e`. Run binding verified by id: run `33367671767`, `headSha=b88ab4e2a8745c61fb12265d2510d48ab73bd710`.

The flake itself is real and it sits on a required merge gate. Routed below.

---

## Blocking finding

### P1-1 — `--session-id` does not preserve context on `--agent codex`, and says nothing about it

issue #526 doneWhen item 2: *"Same `--session-id` across two dispatch calls preserves context (create-or-continue)."* That holds for two of the three backends this verb advertises and silently fails for the third.

- **claude** — `ClaudeCodeLauncher` passes `--session-id <id>` to the Claude CLI, which owns create-or-continue externally. Survives process exit. ✅
- **pi** — `PiRpcLauncher::build_command` passes `--session-id <id>` (plus optional `--session-dir`); continuity is genuinely delegated to pi. ✅
- **codex** — continuity is **not** delegated. It lives in edda's own memory: `CodexLauncher` holds `state: Mutex<LauncherState>` with `threads: HashMap<String, String>` mapping conductor session id → codex thread id, and `resume` is read from that map. `CodexLauncher::new()` starts it empty, and `cmd_dispatch::build_launcher` constructs a fresh launcher on every invocation. **Each `edda dispatch` is its own process**, so the map is always empty at entry: `resume` is `None`, `open_thread` sends `thread/start`, and the caller gets a brand-new codex thread. ❌

So `edda dispatch --agent codex --session-id X` twice yields two unrelated conversations. There is no warning and no error — the caller reusing an id believes they continued a turn and did not.

**Why this is in scope and not "#531 pre-existing launcher semantics" (declared OUT).** The codex launcher is unchanged here and was *correct* for conduct's model: one process, many phases, map alive throughout. The defect belongs to the new verb — `dispatch` advertises `--session-id` create-or-continue uniformly across all three backends and across process boundaries, which is a property that did not exist before this commit. It is measured against issue #526 doneWhen 2 and the frozen `--session-id verbatim across calls` line, both IN SCOPE.

Worth noting for the record: PR #531's `CodexLauncher` doc comment justifies the resume path as *"a forward-looking path for a future dispatch primitive"*. **This is that primitive** — and it still does not reach resume, because the map does not outlive the process.

**Test gap that let this through.** `same_session_id_reaches_launcher_verbatim_on_every_call` proves the id arrives at the launcher **verbatim** — it does not prove continuity, and the name overstates what it checks. It also drives both calls through **one launcher instance in one process**, which is exactly the configuration production never uses. Same structural flaw I flagged in #531's resume test: the test exercises a path the real call chain does not take.

**Discharge — either is acceptable, (a) is cheap:**

- **(a)** Warn when `--agent codex` is combined with an explicit `--session-id`, stating that codex threads do not persist across dispatch invocations. This mirrors the repo's established pattern for "requested capability unsupported on this backend" — `budget_warning_for_agent` (which this PR already reuses) and the `--tmux` warn-and-fall-back. It kills the *silent* part, which is what makes this P1.
- **(b)** Persist the session-id → thread-id map so codex genuinely resumes; that fully satisfies doneWhen 2 but is a larger change.

Docs-only does not discharge it: the failure is silent at runtime, and doneWhen 2 is an acceptance criterion, not a doc claim.

---

## Non-blocking findings

### P2-1 — the launcher construction table is duplicated, not shared

`cmd_dispatch::build_launcher` reproduces `cmd_conduct`'s three-arm construct-and-probe table. They have genuinely diverged (conduct threads `with_verbose(verbose)` and sets `transcript_dir` for claude), so a shared helper would need parameters — but the *backend list itself* is now written twice. The brief required extract-or-share and named copies a maintainability finding, and this repo just lived the cost: adding codex in #527 required editing conduct's arm, and a fourth backend will now require editing two files that can silently drift. The `AgentKind` extraction solved this for the enum; the construction table is the other half.

### P2-2 — no `permission_mode` or `allowed_tools` control on a verb built for external orchestrators

`build_phase` hardcodes `permission_mode: "bypassPermissions"` and `allowed_tools: None`, with no flag to tighten either.

To be fair on severity: this is **not** a regression. `default_permission_mode()` in the plan schema is already `"bypassPermissions"`, so an unspecified conduct phase runs the same way, and only `ClaudeCodeLauncher` consumes the field at all (pi and codex ignore it). But a conduct plan *can* set `permission_mode: default` per phase, and dispatch cannot — and dispatch is the surface explicitly designed to be driven by an external orchestrator in a caller-supplied `--cwd`. Worth a flag, or an explicit note in the help text that the turn runs with permissions bypassed.

### P3-1 — `std::process::exit(code)` skips destructors

The non-zero exit path bypasses `Drop`, so the tokio runtime and the boxed launcher are never dropped. No leak today: every launcher already reaps its own child before returning (codex clears `state.server` on crash/timeout, which drops `CodexAppServer` and calls `start_kill`; pi calls `terminate()`). Stdout is safe too — both output paths end in a newline and Rust's `Stdout` is a `LineWriter`. Flagging it as a latent trap: a future launcher that holds a live child across a non-zero outcome would orphan it.

### P3-2 — the "n/a" string is duplicated

`DispatchOutput::cost_text` writes `"n/a (no usage data reported)"` as a literal for the `None` case, and `cost_line` contains the same literal for the zero case. If one is reworded the other silently diverges on a user-facing string. Route the `None` case through a shared constant.

### P3-3 — `async-trait` dev-dep is not workspace-declared

Added as `async-trait = "0.1"` rather than `workspace = true`. Noting it only for completeness: `async-trait` is not in `[workspace.dependencies]` at all, and `edda-conductor/Cargo.toml:22` declares it identically, so this **matches existing practice** rather than breaking convention. The real cleanup is promoting `async-trait` to the workspace table, which is not this PR's job.

---

## Verified good (no action)

- **Controller attention point 2 — dep audit.** The new dep is `async-trait`, and it is **`[dev-dependencies]` only** — needed for the `RecordingLauncher` test stub that implements the async trait. **No production dependency was added.** The suspicion that it would be `uuid` is not borne out: session ids reuse the existing `phase_session_id` (v5) helper from `edda-conductor`, so no uuid dep is required. Proportionate.
- **Controller attention point 3 — session id determinism vs uniqueness: safe.** `generate_session_id()` builds a v5 UUID over the name `dispatch-{pid}-{unix_timestamp_nanos}`. v5 is deterministic *given the name*, but the name embeds pid and nanosecond time, so two bare `edda dispatch` calls produce different ids. **No silent session sharing.** Covered by `generated_session_id_is_non_empty_and_unique`.
- **Controller attention point 1 — 686 lines audited.** Roughly 310 lines of implementation to 376 of tests, so the file is not bloated for what it does. Sharing is genuine, not claimed: `budget_warning_for_agent` and `cost_line` are `pub(crate)` in `cmd_conduct` and **called**, not copied. The one real copy is P2-1 above.
- **`AgentKind` extraction does not change conduct behavior.** The enum moved verbatim (same three variants, same `as_str`, same `writes_transcripts`); `only_claude_writes_transcripts` was **moved**, not dropped; and conduct's three parse tests (`run_defaults_to_claude_agent`, `run_accepts_explicit_agents`, `run_rejects_unknown_agent`) all survive, still pinning claude|pi|codex and the all-three-listed error.
- **`verify_available()` probe** is present for all three backends before dispatch, matching conduct.
- **Exit-code table is correct**, including pre-dispatch failures: `run()` returns `Err` for an unreadable `--prompt-file` or a failed probe, `main` is `anyhow::Result<()>`, so those exit 1 as documented. All five outcome classes are asserted against real `PhaseResult` variants.
- **JSON shape is stable and pinned.** Exactly one object on stdout; the key set is asserted by sorting the object's keys against the full expected list, so a silent field addition or rename fails the test. Warnings go to stderr; in `--json` mode the error rides inside the object rather than being duplicated to stderr.
- **`edda-conductor` is genuinely untouched** by the diff — the controller's claim verified against the diffstat.

**doneWhen:** 1 ✅ (unit-level via MockLauncher; real-binary e2e run by nobody, acceptable — it needs pi installed) · 2 ❌ **see P1-1** (holds for claude and pi, silently fails for codex) · 3 ✅ (budget/timeout land on the synthetic phase and are asserted; `budget_warning_for_agent` is shared code) · 4 ✅ (documented in the long help in `main.rs` and pinned by the key-set test)

---

## FOLLOW-UP ISSUE (routed, does not extend this PR)

**`agent::pi_rpc::tests::startup_stderr_reaches_the_crash_error` flakes on ubuntu, on a required merge gate.** Evidenced twice: `main` at `fee8566f` (run `33356986628`) and this head before the rerun, both `Error: failed writing pi RPC stdin`. The test writes to the stdin of a `StderrThenDie` fake and loses the race under load; the write needs to tolerate a child that has already exited, since a dead child is the scenario under test. No open issue covers it — the two flake issues on record (#524, #447) are closed and about different tests. Not caused by this PR, so it does not block here, but `Test (ubuntu-latest)` is a required check and this will keep costing reruns.

---

## Evidence — RAN vs READ

**RAN:** one targeted CI job re-run (the flaked `Test (ubuntu-latest)`), which the ladder prescribes for an environmental red. **No local build. Lane: n/a.**

**READ:**
- Ancestry verified, not assumed: `b88ab4e^` resolves to `64f9b4a`.
- CI binding verified by run id (`33367671767` → `headSha=b88ab4e…710`), and the prior red diagnosed from the raw failed-job log rather than the summary. Final board: all 7 required checks SUCCESS.
- Worker receipt (edda 361 passed, was 333; clippy clean; fmt ok).
- Full delta of all 7 changed files, plus `pi_rpc.rs`, `codex_rpc.rs`, `launcher.rs`, and `plan/schema.rs` as unchanged context for the parity and permission-mode questions.

**Why no local rerun.** The touched crate is `edda`, inside the CI Windows subset and inside full-workspace tests on ubuntu + macos; exact-head CI is now fully green and the receipt reconciles. The P1 is a call-chain/lifetime defect that no test run surfaces — it needs reading the code across two crates, which is what I did. A lane build would have re-derived green at ~40 GB and bought nothing.

**Cost:** ~14 read-only tool calls, 1 CI job re-run, 0 local builds, 0 bytes written to the repo.

---

**Verdict: Changes Requested** @ `b88ab4e2a8745c61fb12265d2510d48ab73bd710` — one P1 (codex `--session-id` silently does not continue a session, against issue #526 doneWhen 2). P2s and P3s are non-blocking; P2-1 and P3-2 are cheap to fold into the same push.

fagemx:
# Review Response: Round 1

New head: `be20fc1f5ad47f6dd30c08520afae28d34c861ec` (one commit on top of the reviewed `b88ab4e2a8745c61fb12265d2510d48ab73bd710`; reflog-verified single commit, no amend, no rebase). Implemented by pi via round-2 conductor dispatch; posted by the controller on the implementation side's behalf.

## P1-1 (blocking) — silent no-continuity for `--session-id` on `--agent codex` → DISCHARGED (warn path)
- New warning mirroring `budget_warning_for_agent`: `--agent codex` combined with an explicit `--session-id` warns on stderr that codex thread state is per-process — the id does not resume a prior conversation across dispatch invocations (claude/pi persist externally; codex continuity today requires a single long-lived process, i.e. conduct). Fires only for codex + explicit id (not generated ids, not claude/pi); pinned by tests.
- `--session-id` long help now documents the per-backend semantics honestly.
- The overclaiming test renamed/re-commented to what it proves (verbatim delivery), plus warning tests.
- The full capability (persistent session→thread map, real cross-process `thread/resume`) is routed: **#535** — same warn-now/model-later split as #531 budget → #533. (Recorded decision: `dispatch.codex-session-id=warn-only`.)

## P2-1 — launcher construction table duplicated → FIXED
Shared `build_launcher(agent, LauncherOptions)` extracted to `agent_kind.rs`; both `cmd_conduct` and `cmd_dispatch` call it (options carry verbose / transcript_dir). The backend list now exists in exactly one place; conduct behavior and tests unchanged. (Recorded decision: `dispatch-launcher-factory=agent_kind.rs::build_launcher`.)

## P2-2 — no permission control on the dispatch surface → FIXED
`--permission-mode <mode>` added (default `bypassPermissions`, preserving prior behavior), passed verbatim into the synthetic Phase; help states only the claude backend consumes it today. Parse + phase-landing tests.

## P3-1 — `std::process::exit` skipping destructors → FIXED
Exit code computed in an inner scope; runtime and launcher drop before the outermost exit, with a comment explaining the orphaned-child trap being prevented.

## P3-2 — duplicated "n/a" literal → FIXED
Single shared constant used by both `DispatchOutput::cost_text` and `cost_line`.

## P3-3 — async-trait workspace promotion → SKIPPED per review (matches existing practice; routed as general cleanup, not this PR's job).

## Related
The CI failure investigated during round 1 is filed as **#536** (production defect in the pi launcher's stdin-write race — not caused by this PR; next in the dispatch queue after this merges).

## Receipt correction (disclosed, not hidden)
The worker-typed receipt string contains a hallucinated SHA tail (`be20fc17b227...651b` — no such object; first 7 hex correct). The actual frozen head is `be20fc1f5ad47f6dd30c08520afae28d34c861ec`, reflog-verified as a single commit on the reviewed base. Gate evidence binds through the conductor check events: fix-phase `cargo test -p edda` (365 passed) + clippy `-D warnings` ran on the tree that seal committed unchanged (post-commit `git_clean` green, `fmt --check` green). A ledger correction note is recorded in the worker worktree; future work orders now derive the receipt SHA via `$(git rev-parse HEAD)` instead of worker transcription.

## Gates (frozen SHA `be20fc1f5ad47f6dd30c08520afae28d34c861ec`)

| Gate | Result |
|---|---|
| `cargo test -p edda` | 365 passed (was 361) |
| `cargo clippy -p edda -p edda-conductor --all-targets -- -D warnings` | clean |

Lane worker-1. Exact-head CI running; round 2 review requested.

fagemx:
# Code Review: Round 2

**Reviewed SHA (frozen, pinned):** `be20fc1f5ad47f6dd30c08520afae28d34c861ec`
**Round-1 SHA:** `b88ab4e2a8745c61fb12265d2510d48ab73bd710` — verified as the direct parent (`be20fc1f^ == b88ab4e`, `merge-base --is-ancestor` passes). No amend, no rebase; the round-1 verdict stays historically valid for its own SHA.
**Verdict: LGTM** — P0 = 0, **P1 = 0**, P2 = 1 (fix-caused, non-blocking, one line)

Any further push voids this verdict.

---

## Receipt integrity — disclosure verified, and it changes nothing about the gate

The controller disclosed that the worker-typed receipt carries a fabricated SHA tail. I verified that independently rather than accepting either the claim or the receipt:

- `git cat-file -t be20fc17b227699b2b1a3d09d69f6564365c651b` → **not an object in this repo.** The disclosure is accurate; the string is void.
- The real head `be20fc1f5ad47f6dd30c08520afae28d34c861ec` resolves, and its parent is exactly the round-1 SHA.

Treating the receipt's counts as evidence and its SHA as void is the right call, and I want to record why it costs this review nothing: **a receipt whose SHA is unverifiable cannot bind anything to a commit, so it was never load-bearing here.** My gate confidence rests on exact-head CI, which I verified by run id against the real head — not on the receipt. The receipt is corroboration, and corroboration that cannot name its own subject is simply discarded.

Disclosing this unprompted was the correct call, and deriving future receipt SHAs via `$(git rev-parse HEAD)` removes the class.

---

## Round-1 findings — discharge verified

Checked against the code at this SHA, not accepted on claim.

### P1-1 — codex `--session-id` silently starts a fresh conversation — **DISCHARGED**

`session_id_warning_for_agent(agent, explicit_id)` warns on `Codex && explicit_id`, wired in `run_inner` next to the budget warning, on stderr. The silence is gone, which was the whole basis of the P1.

Discharged better than asked in two places:

- **The help text is now honest at the point of use.** `--session-id`'s doc says claude and pi "persist conversations externally, so the same id resumes the prior conversation across invocations; codex keeps its thread state in-process, so each `edda dispatch` invocation starts a fresh conversation no matter what id you pass". A caller reading `--help` learns this before running anything.
- **The misleading test name is fixed.** `same_session_id_reaches_launcher_verbatim_on_every_call` → `session_id_is_delivered_verbatim_on_every_call_within_one_process`, with a comment stating it "does not (and cannot) prove cross-invocation continuity — for codex it does not hold." I raised the name as overstating what it checks; this is the right correction.

Persistence routed as #535 — the warn-now/model-later split, same shape as #531 budget → #533. Correct.

### P2-1 (round 1) — duplicated launcher table — **DISCHARGED**

`build_launcher(agent, LauncherOptions { verbose, transcript_dir })` now lives in `agent_kind.rs`; both commands call it. The backend list exists in exactly one place, so a fourth backend is a one-file change.

### P2-2 — no permission control — **DISCHARGED**

`--permission-mode` added, `default_value = "bypassPermissions"` so no existing invocation changes behavior, carried verbatim onto the synthetic phase, with help text noting only claude consumes it. Tests cover default, explicit parse, and verbatim landing (`acceptEdits`).

### P3-1 — `process::exit` skipping destructors — **DISCHARGED**

The work moved into `run_inner() -> Result<i32>`; `run` takes the code as a return value and only then calls `std::process::exit`. The tokio runtime and the boxed launcher are locals of `run_inner`, so they drop before the process exits. That is exactly the inner-scope fix, and the doc comment records why.

### P3-2 — duplicated "n/a" literal — **DISCHARGED**

`NO_USAGE_COST_TEXT` constant in `cmd_conduct`, used by both `cost_line` and `DispatchOutput::cost_text`. One source.

### P3-3 — `async-trait` workspace declaration — **correctly skipped**, per my own comment.

---

## The three watch items — all verified

**(a) The extraction did not change conduct's construction behavior.** Compared arm by arm against the round-1 code:

| Backend | Before (in `cmd_conduct`) | After (in `build_launcher`) |
|---|---|---|
| Claude | `new().with_verbose(verbose)`, `transcript_dir = Some(dir)`, probe | `new().with_verbose(options.verbose)`, `transcript_dir = options.transcript_dir`, probe |
| Pi | `new().with_verbose(verbose)`, probe | same via options |
| Codex | `new().with_verbose(verbose)`, probe | same via options |

Conduct calls it with `verbose` and `Some(transcript_dir.clone())` — so verbose threading and the claude-only transcript tee reproduce exactly. Dispatch passes `verbose: false, transcript_dir: None`, which matches what round-1 dispatch built (`ClaudeCodeLauncher::new()` with defaults). No behavior change on either side.

**(b) The warning fires only for codex with an explicit id — no false positives.** The predicate is `agent == AgentKind::Codex && explicit_id`, fed `args.session_id.is_some()`. `session_id_warning_for_agent_fires_only_for_codex_with_an_explicit_id` asserts all three negatives explicitly: codex with a *generated* id stays silent (correct — a generated id carries no continuity expectation), and claude and pi with explicit ids stay silent.

**(c) No exit-code value changed.** `Outcome`, `exit_code_for`, and `DispatchOutput::from_result` are untouched by this delta; only the call site moved. `exit_codes_map_all_outcome_classes` and `mock_timeout_maps_to_exit_code_2` still pin 0/1/2/3/4.

**`edda-conductor` untouched** — verified against the diffstat again (3 files, all in `edda-cli`).

---

## P2-1 (new, fix-caused, non-blocking) — the warning's closing clause contradicts this repo

The warning ends:

> …continuity for codex currently requires a single long-lived process (conduct).

That sends the reader to `conduct` for codex continuity, and conduct does not provide it either. From `codex_rpc.rs` **at this same commit**:

> the sequential runner derives a unique session id per plan+phase+attempt, so resume does not currently fire in production

`sequential.rs:216` still reads `phase_session_id_attempt(&plan.name, &phase_id, attempt)`, so every phase and every retry gets a distinct session id, `threads.get(session_id)` always misses, and each conduct phase opens a **fresh codex thread**. Conduct keeps one long-lived *process* — which is real, and does save app-server respawns — but that is not conversation continuity. The clause conflates the two, and a user who follows it lands in the same situation with no warning at all.

This is fix-caused (the string is new in this push), so it is eligible to block. I am not blocking on it: the warning's operative claim — codex + dispatch does not resume — is correct and complete, which is what discharges P1-1; the defect is a trailing remedy clause. Suggested wording: say that codex has no cross-invocation continuity today and that conduct shares the limitation for the same reason, pointing at #535 as the fix. Note `session_id_warning_for_agent_fires_only_for_codex_with_an_explicit_id` asserts `warning.contains("conduct")`, which a reword can keep satisfying.

**Merge-vs-push, controller's call** — stating the tradeoff rather than forcing it, since merge follows this verdict immediately:

- **Merge as-is** and route the clause as a follow-up. The verdict stands; one inaccurate sentence ships.
- **Push the one-line reword.** Trivial change, but it voids this verdict and needs a round 3 from me. I will turn it around quickly.

Either is defensible. I would take the second only because it is one line and I have now verified the underlying fact twice.

---

## doneWhen (issue #526)

1 ✅ unit-level via MockLauncher · 2 ✅ **now honest**: holds for claude and pi, and codex's non-continuity is warned at runtime and documented in `--help`, with persistence routed as #535 · 3 ✅ budget/timeout on the synthetic phase, `budget_warning_for_agent` and `build_launcher` shared code · 4 ✅ documented in long help, key set pinned by test

---

## Evidence — RAN vs READ

**RAN: nothing this round. Lane: n/a — no local build.** The ubuntu `pi_rpc` flake (#536) did **not** fire on this head, so no rerun was needed.

**READ:**
- Ancestry verified: `be20fc1f^ == b88ab4e`, `merge-base --is-ancestor` passes.
- Receipt integrity verified independently: the disclosed SHA is not an object; the real head is.
- CI binding verified by run id — run `33370585850`, `headSha=be20fc1f5ad47f6dd30c08520afae28d34c861ec`. I waited rather than ruling on a partial board (`Test (windows-latest)` was still running at first look). Final: **all 7 required checks SUCCESS**.
- Full delta of all 3 changed files, plus `codex_rpc.rs` and `sequential.rs` at this SHA as unchanged context for the P2-1 contradiction.

**Why no local rerun.** Touched crate is `edda`, inside the CI Windows subset and inside full-workspace tests on ubuntu + macos; exact-head CI is fully green. The one finding this round is a factual contradiction between two strings in the repo, which no test run surfaces.

**Cost:** ~8 read-only tool calls, 0 builds, 0 CI reruns, 0 bytes written to the repo. Two rounds total on this PR.

---

**Verdict: LGTM** @ `be20fc1f5ad47f6dd30c08520afae28d34c861ec` — P0 = 0, P1 = 0. All six round-1 findings verified discharged, two of them better than requested. One fix-caused P2 on a single inaccurate clause, explicitly not blocking.

Merge authority remains the operator's. Any push after this comment voids this verdict.

fagemx:
# Review Response: Round 2

New head: `84f9cfa74172357e188577b4b2e4d26d4ed92709` (one commit on top of the LGTM'd `be20fc1f5ad47f6dd30c08520afae28d34c861ec`).

## Round-2 P2 (fix-caused, one line) — FIXED, taking the reviewer's lean

The warning's remedy clause ("continuity for codex currently requires a single long-lived process (conduct)") was wrong at this very commit — conduct derives a unique session per plan+phase+attempt, so its phases open fresh codex threads too. The clause now reads: codex has no cross-invocation continuity today, and conduct shares the limitation for the same reason (persistence tracked as GH-535). The test's `contains("conduct")` assertion still binds.

Rationale for not merging as-is: the sentence lives on the exact surface #534 round 1 existed to make honest; shipping a known-wrong remedy in an honesty warning was the worse trade.

**Provenance disclosure:** this one-line fix was implemented directly by the controller (dispatching a worker for one sentence is disproportionate). Verifier independence is unaffected — the same independent session takes round 3.

## Gates (frozen SHA `84f9cfa74172357e188577b4b2e4d26d4ed92709`, run by the controller, lane worker-1)

| Gate | Result |
|---|---|
| `cargo test -p edda` | 354 + 11 = 365 passed |
| `cargo clippy -p edda --all-targets -- -D warnings` | clean |

Exact-head CI running; round 3 requested.

fagemx:
# Code Review: Round 3

**Reviewed SHA (frozen, pinned):** `84f9cfa74172357e188577b4b2e4d26d4ed92709`
**Round-2 SHA:** `be20fc1f5ad47f6dd30c08520afae28d34c861ec` — verified as the direct parent, no amend, no rebase.
**Verdict: LGTM** — P0 = 0, P1 = 0, P2 = 0

Any further push voids this verdict.

Scope is proportionate to the delta: one clause in one function.

---

## Delta is exactly what was claimed

`git diff --stat` over the round-2 SHA: **1 file, +2/-1**, entirely inside `session_id_warning_for_agent`. Nothing else drifted — no incidental edits rode along with the reword.

## The wording is now correct

> Warning: codex thread state is per-process, so `--session-id` does not resume a prior conversation across dispatch invocations (claude/pi do persist); codex has no cross-invocation continuity today, and conduct shares the limitation for the same reason (persistence is tracked as GH-535).

Checked clause by clause against the code at this SHA:

| Claim | Verdict |
|---|---|
| codex thread state is per-process | ✅ `LauncherState.threads` is in-memory, `CodexLauncher::new()` per process |
| `--session-id` does not resume across dispatch invocations | ✅ |
| claude/pi do persist | ✅ both pass `--session-id` to a CLI that persists externally |
| codex has no cross-invocation continuity today | ✅ |
| conduct shares the limitation | ✅ every conduct phase opens a fresh codex thread |

**The defect I raised is gone.** The old clause sent the reader to conduct as a remedy; the new one tells them conduct fails the same way. That is the operative, actionable content, and it is correct — nobody is misdirected any more.

**One precision note — explicitly NOT a finding, and explicitly not worth another push.** "for the same reason" compresses two different mechanisms: in dispatch the id is stable but the threads map does not survive the process; in conduct the map survives but the id is never repeated, because `sequential.rs:216` derives a unique one per plan+phase+attempt. A consequence worth knowing for GH-535's own scoping: persistence alone would fix dispatch but **not** conduct, which would additionally need to reuse session ids. I am recording that so whoever picks up GH-535 has it, not asking for a change here. The user-facing content is accurate, and re-opening a one-sentence string for a causal adverbial would be exactly the diminishing-returns loop the review rules tell us to stop.

## The test still binds and still means something

`session_id_warning_for_agent_fires_only_for_codex_with_an_explicit_id` asserts the message contains `codex`, `per-process`, `does not resume`, and `conduct`. I checked the new string against all four rather than trusting the reported run — all present, so the assertion binds on the new wording. The test still pins the thing that matters: only codex with an *explicit* id warns, and the three negatives (codex with a generated id, claude, pi) stay silent.

## Merge-safety audit — the routed follow-ups survive

Since merge follows this verdict, I checked what it would close. `closingIssuesReferences` on this PR resolves to **`#526` only**, which is correct — that is the issue this PR implements. No commit message on the branch contains a closing keyword bound to a number, and the new string's `GH-535` reference carries no keyword. **GH-535 and #536 will not be auto-closed by this merge.**

---

## Provenance — disclosure was correct, and here is the structural fact it implies

The controller implemented this line directly and will merge it. For this commit that makes the controller both implementer and merge authority, with this review as the only independent check in the loop.

I am not treating that as a defect: the disclosure was proactive and unprompted, the change is one sentence, and I judged the line on its merits exactly as I would any implementer's — I re-derived the underlying fact from `sequential.rs` and `codex_rpc.rs` rather than accepting the commit message's reasoning, which happens to restate my own round-2 finding. Dispatching a worker for one sentence would have been disproportionate, and I agree with that call.

Recording it because a reviewer's independence is a property of the *loop*, not of good intentions, and the loop was thinner here than in rounds 1 and 2. Worth a glance if this pattern becomes routine rather than exceptional.

---

## Evidence — RAN vs READ

**RAN: nothing. Lane: n/a — no local build.** The `#536` ubuntu flake did **not** fire on this head, so no rerun was needed.

**READ:**
- Ancestry: `84f9cfa^ == be20fc1f`, `merge-base --is-ancestor` passes.
- Full delta (3 lines, one function), plus `codex_rpc.rs` and `sequential.rs` as unchanged context to re-derive the accuracy claims independently.
- CI binding verified by run id — run `33372097482`, bound to `84f9cfa74172357e188577b4b2e4d26d4ed92709`. I waited for completion rather than ruling on a partial board (`Clippy (windows-latest)` and `Test (windows-latest)` were still running at first look). Final: **all 7 required checks SUCCESS.**
- Auto-close audit via `closingIssuesReferences` and a keyword scan of every commit message on the branch.

**Cost:** ~6 read-only tool calls, 0 builds, 0 CI reruns, 0 bytes written to the repo. Three rounds total on this PR.

---

**Verdict: LGTM** @ `84f9cfa74172357e188577b4b2e4d26d4ed92709` — P0 = 0, P1 = 0, P2 = 0. The round-2 P2 is discharged, the delta is confined to the one clause, all 7 checks are green on the exact head, and the merge will not close the routed follow-ups.

Merge authority remains the operator's. Any push after this comment voids this verdict.

PR #537, opened 2026-08-31
fix(edda-conductor): route dead-child stdin write into the crash path with stderr tail (#536)

### LINKED ISSUE #536: fix(edda-conductor): pi startup failure loses its stderr reason when the stdin write loses the race
## Summary

When `pi` fails at startup, whether the operator sees the actual reason is a **race**. The flaky test is the symptom; the defect is in `run_turn`.

`PiRpcSession::run_turn` documents its own contract:

> Protocol-level failures (rejected prompt, malformed JSONL, EOF before settlement) surface as `PhaseResult::AgentCrash` — the runner retries crashes per its failure policy. `Err` is reserved for spawn/IO errors.

It then sends the prompt **before** entering the read loop:

```rust
let request = json!({ "id": prompt_id, "type": "prompt", "message": message });
self.write(&request).await?;      // <-- races the child's exit
```

A pi that dies during startup — bad auth, bad `--model`, unwritable `--session-dir`, the exact list named in `eof_error`'s own comment — never reads stdin. So:

- **Write loses the race** → the child is already gone → `write_all` returns `BrokenPipe` → `write_json_line`'s `.context("failed writing pi RPC stdin")?` propagates an **`Err`**. The stderr diagnostic that explains *why* pi died is collected but never reported. The operator gets a bare `failed writing pi RPC stdin`.
- **Write wins the race** → the loop reaches EOF on stdout → `eof_error(child, stderr_tail)` builds the intended message: `pi RPC process exited with non-zero status …; pi stderr: <the real reason>` → `PhaseResult::AgentCrash`.

Same input, two different outcomes. The startup-failure path — the one case where the stderr reason is the entire value of the error — is the one that is not guaranteed.

## Evidence

`agent::pi_rpc::tests::startup_stderr_reaches_the_crash_error` is a regression guard for exactly this ("the reason must be present every run, not whenever the drain task happens to win") and loops 5× to catch it. It is correctly detecting that the guarantee is still conditional — the remaining hole is the stdin write, not the drain it was written for.

Two independent CI failures, both `Test (ubuntu-latest)`, both `Error: failed writing pi RPC stdin`:

| Where | Run | Job |
|---|---|---|
| `main` @ `fee8566f72567f5cb9a89f5ef1ffe0e0fc81b5de` | [33356986628](https://github.com/fagemx/edda/actions/runs/33356986628) | 99380933061 |
| PR #534 @ `b88ab4e2a8745c61fb12265d2510d48ab73bd710` | [33367671767](https://github.com/fagemx/edda/actions/runs/33367671767) | 99411610930 |

Nondeterminism is proven rather than assumed: re-running **only** the failed job on the unchanged commit `b88ab4e` turned it green. Same code, both outcomes.

Not caused by PR #534 — it does not touch `edda-conductor`, and the `main` failure predates it. Routed from that PR's round-1 review.

## Why it is worth fixing rather than muting

`Test (ubuntu-latest)` is one of the 7 required status checks on the `main` ruleset, so every occurrence blocks a merge until someone re-runs it. Muting or de-looping the test would hide a real defect: the loop is load-bearing, and the underlying behavior is what should change.

## Fix direction

Treat a closed stdin as "the child exited before accepting the prompt" rather than as an IO error, and route it into the path that already does the right thing:

- In `run_turn`, catch `BrokenPipe` / `WriteZero` from the initial `self.write(&request)` and fall through to the existing reap → drain-stderr → `eof_error` sequence, returning `PhaseResult::AgentCrash` with the stderr tail attached.
- Keep genuine IO errors as `Err`, per the documented contract — only the child-is-gone class should be reclassified.

That makes the test's 5× loop pass unconditionally *because the behavior is unconditional*, not because the race got wider.

## doneWhen

- [ ] A pi that writes a startup diagnostic to stderr and exits before reading stdin always produces `PhaseResult::AgentCrash` containing that diagnostic, regardless of who wins the write/exit race
- [ ] `startup_stderr_reaches_the_crash_error` passes deterministically, with its 5× loop retained
- [ ] A genuine IO error on stdin (not a dead child) still surfaces as `Err`, preserving the documented contract
- [ ] No test is muted, `#[ignore]`d, or de-looped as part of the fix

🤖 Routed from the PR #534 round-1 verifier review.

fagemx:
Closes #536

## Problem

A pi that dies during startup (bad auth, bad `--model`, unwritable `--session-dir`) never reads stdin, so `run_turn`'s prompt write races the process death. Losing the race surfaced a bare `failed writing pi RPC stdin` `Err` and **discarded the collected stderr** — the only place the real reason appears. Winning it produced the intended `AgentCrash` with the stderr tail. Same input, two outcomes; the CI guard `startup_stderr_reaches_the_crash_error` (5× loop) was correctly catching the hole and flaking on required gates (evidence runs in #536).

## Change

`crates/edda-conductor/src/agent/pi_rpc.rs` only (+108/-1):

- `is_child_gone_error(&anyhow::Error)` — narrow predicate: `BrokenPipe | WriteZero` on the underlying `io::Error`.
- In `run_turn`, a failed prompt write with a child-gone error class is routed into the **existing** reap → drain-stderr → `eof_error` sequence, returning `AgentCrash` with the stderr tail — the same path the stdout-EOF branch takes, making the outcome unconditional.
- The predicate is applied to the whole `write_json_line` result, so **both** fallible awaits (`write_all` and `flush`) are covered — a dead child surfacing on the flush is the same reclassification.
- Genuine IO errors (child alive) still return `Err`, preserving the documented contract.

## Tests

- `dead_child_stdin_write_still_reports_stderr_reason` — **deterministic losing branch**: the fake writes its stderr diagnostic and exits; the test waits until the child has demonstrably died before invoking the turn, so the stdin write is guaranteed to lose — asserts `AgentCrash` containing the diagnostic. No race, no loop.
- `child_gone_error_predicate_matches_broken_pipe_and_write_zero` — predicate unit test incl. negatives.
- `startup_stderr_reaches_the_crash_error` — **kept with its 5× loop intact** (issue doneWhen pins: no mute, no `#[ignore]`, no de-loop).

Supporting platform evidence (controller-measured, standalone repro on this Windows workstation): writing to a dead child's pipe yields `BrokenPipe` (os error 232) for both a 1 MiB and a post-mortem small write — the predicate's primary arm is empirically verified on the platform CI's Windows subset runs this crate on.

## Provenance & environment note

Implemented by pi (glm-5.3-flash) via `edda conduct run --agent pi`, isolated worktree, fresh session per attempt. Fix-phase attempt 1 died to an environmental `LNK1104` (linker file lock) during `cargo test`; attempt 2 re-ran clean on a fresh session — no product failure involved.

## Gates (frozen SHA `b04fe369e2beb983b034d92d5d02ff593a94869d`)

| Gate | Result |
|---|---|
| `cargo test -p edda-conductor` (conductor check, full crate) | pass |
| `cargo clippy -p edda-conductor --all-targets -- -D warnings` | clean |

Worker receipt SHA now derived via `$(git rev-parse HEAD)` (matches HEAD exactly); its "25 passed" figure reflects the worker's last filtered run — full-crate green is attested by the phase check above. Lane worker-1.

Reviewer note: per operator ruling the verifier for this PR is independent of the issue's author; a red `Test (ubuntu)` on `startup_stderr_reaches_the_crash_error` at this head would be **signal** (the fix failing), not the historical flake.

### DIFF
--- a/crates/edda-conductor/src/agent/pi_rpc.rs
+++ b/crates/edda-conductor/src/agent/pi_rpc.rs
@@ -277,7 +277,23 @@ impl PiRpcSession {
     ) -> Result<PhaseResult> {
         let prompt_id = self.take_id();
         let request = json!({ "id": prompt_id, "type": "prompt", "message": message });
-        self.write(&request).await?;
+        if let Err(error) = self.write(&request).await {
+            // A pi that dies during startup (bad auth, bad --model,
+            // unwritable --session-dir) never reads stdin, so this write
+            // races the process death. Losing that race used to discard the
+            // collected stderr — the only place the real reason appears — and
+            // surface a bare IO error instead. Reclassify the dead-child
+            // error class into the same reap → drain-stderr → crash path the
+            // stdout-EOF branch takes, so the outcome is unconditional.
+            if !is_child_gone_error(&error) {
+                return Err(error);
+            }
+            let _ = self.child.wait().await;
+            let tail = collect_stderr_tail(&mut self.stderr_drain, &self.stderr).await;
+            return Ok(PhaseResult::AgentCrash {
+                error: eof_error(&mut self.child, &tail),
+            });
+        }
 
         let mut state = TurnState::default();
         let deadline = tokio::time::sleep(timeout);
@@ -452,6 +468,25 @@ fn clean_record(line: &str) -> &str {
 
 fn parse_message(line: &str) -> Result<Value> {
     serde_json::from_str(line).with_context(|| format!("invalid pi RPC JSON: {line}"))
+}
+
+/// Classify a stdin-write failure as "the child is gone".
+///
+/// Only `BrokenPipe` / `WriteZero` indicate the reader side of the stdin pipe
+/// has been closed — the signature of a pi that died during startup and never
+/// read its stdin. Every other IO error (disk full, permission denied, ...) is
+/// a genuine failure while pi may still be alive and keeps the documented
+/// `Err`-for-IO contract.
+fn is_child_gone_error(error: &anyhow::Error) -> bool {
+    error
+        .chain()
+        .filter_map(|cause| cause.downcast_ref::<std::io::Error>())
+        .any(|io| {
+            matches!(
+                io.kind(),
+                std::io::ErrorKind::BrokenPipe | std::io::ErrorKind::WriteZero
+            )
+        })
 }
 
 async fn write_json_line(stdin: &mut ChildStdin, value: &Value) -> Result<()> {
@@ -950,6 +985,78 @@ mod tests {
             }
         }
         Ok(())
+    }
+
+    /// The deterministic losing branch of the startup race (#536): the fake
+    /// writes a stderr diagnostic and exits WITHOUT reading stdin, and this
+    /// test waits until the child has demonstrably exited BEFORE invoking the
+    /// turn — so the stdin write is guaranteed to hit a dead child instead of
+    /// racing it. The turn must still surface AgentCrash with the stderr
+    /// reason, not a bare IO error.
+    #[tokio::test]
+    async fn dead_child_stdin_write_still_reports_stderr_reason() -> Result<()> {
+        let (_dir, command) = fake_pi_command(FakeScenario::StderrThenDie)?;
+        let mut session = PiRpcSession::spawn_command(command).await?;
+        let deadline = tokio::time::Instant::now() + Duration::from_secs(10);
+        loop {
+            if session.child.try_wait()?.is_some() {
+                break;
+            }
+            assert!(
+                tokio::time::Instant::now() < deadline,
+                "fake pi never exited"
+            );
+            tokio::time::sleep(Duration::from_millis(10)).await;
+        }
+        let cancel = CancellationToken::new();
+        let result = session
+            .run_turn("do the task", None, Duration::from_secs(30), false, &cancel)
+            .await;
+        session.terminate().await;
+        let result = result?;
+        match result {
+            PhaseResult::AgentCrash { error } => {
+                assert!(error.contains(STARTUP_DIAGNOSTIC), "{error}");
+                assert!(error.contains("pi stderr:"), "{error}");
+            }
+            other => panic!("expected AgentCrash, got {other:?}"),
+        }
+        Ok(())
+    }
+
+    #[test]
+    fn child_gone_error_predicate_matches_broken_pipe_and_write_zero() {
+        // A genuine stdin IO error with a LIVE child (e.g. a full disk, a
+        // permission failure on the pipe handle) returning `Err` cannot be
+        // constructed cheaply with the fake machinery: the fake scripts only
+        // control stdout/stderr and exit codes, and there is no portable way
+        // to inject an arbitrary IO error into a live pipe write. The
+        // classification boundary is therefore pinned directly on the
+        // predicate: only BrokenPipe/WriteZero — the kinds the OS reports
+        // when the reader side of the pipe is gone — are reclassified, and
+        // every other kind (and non-IO error) stays `Err`.
+        let broken_pipe = anyhow::Error::new(std::io::Error::new(
+            std::io::ErrorKind::BrokenPipe,
+            "pipe closed",
+        ))
+        .context("failed writing pi RPC stdin");
+        assert!(is_child_gone_error(&broken_pipe));
+
+        let write_zero = anyhow::Error::new(std::io::Error::new(
+            std::io::ErrorKind::WriteZero,
+            "write zero",
+        ))
+        .context("failed writing pi RPC stdin");
+        assert!(is_child_gone_error(&write_zero));
+
+        let permission = anyhow::Error::new(std::io::Error::new(
+            std::io::ErrorKind::PermissionDenied,
+            "denied",
+        ))
+        .context("failed writing pi RPC stdin");
+        assert!(!is_child_gone_error(&permission));
+
+        assert!(!is_child_gone_error(&anyhow!("not an io error")));
     }
 
     #[test]


fagemx:
## Code Review: Round 1

**Reviewed SHA:** `b04fe369e2beb983b034d92d5d02ff593a94869d` (verified equal to `headRefOid`)
**Delta base:** `792d398e5350180165a66adf38039aa07947c529` — 1 file, +108/−1
**Verdict: LGTM — P0 = 0, P1 = 0.** Two follow-up issues, neither extending this PR.

Reviewer note on standing: per operator ruling I am independent of #536's author, so the issue's *fix direction* was not binding on me — I judged the design itself, not conformance to it. My conclusion is that the direction is right; the reasoning is below rather than assumed.

---

### IN SCOPE

`run_turn`'s prompt-write path in `crates/edda-conductor/src/agent/pi_rpc.rs`; the four #536 `doneWhen` items; design judgment on the reclassification itself; the new test's actual determinism; the predicate's negative space; regression risk to the healthy path; current-base integration (`MERGEABLE`, base `main` @ `792d398`).

### doneWhen — all four met

| # | Item | Evidence |
|---|---|---|
| 1 | Dead child always yields `AgentCrash` with the diagnostic, regardless of race | Both branches now converge on the same reap → drain → `eof_error` sequence. Losing branch covered by the new deterministic test on all 3 platforms; winning branch by the 5× guard. |
| 2 | `startup_stderr_reaches_the_crash_error` deterministic, 5× loop retained | Extracted the test block at base and at head and compared — **byte-identical**; `for attempt in 0..5` intact. Green on ubuntu, macOS and Windows at this head. |
| 3 | Genuine IO error with a live child still `Err` | `return Err(error)` on predicate-false; pinned by `child_gone_error_predicate_…` including `PermissionDenied` and non-IO negatives. See FOLLOW-UP 1 for the boundary's real edge. |
| 4 | No test muted, `#[ignore]`d, or de-looped | The diff contains **exactly one deleted line**, and it is the production line `self.write(&request).await?;`. `#[test]` count 24 → 26. The crate's single `#[ignore]` (`codex_app_server.rs:959 live_start_resume_drill`) is byte-identical at base and head — pre-existing and unrelated. |

### Design judgment (independent)

The alternative I weighed was: keep `Err`, but attach the stderr tail to it. It has a genuine claim on the documented contract — "`Err` is reserved for spawn/IO errors", and a pi that died at startup is arguably a *spawn* failure.

I reject it, because the same physical event already produces `AgentCrash` whenever the write **wins** the race, via the stdout-EOF branch. Making the losing branch `Err` would preserve exactly the nondeterminism #536 exists to remove, unless you also converted the EOF branch to `Err` — a far larger change flipping retry semantics for every crash class. Routing the minority path into the majority path makes classification depend on *what happened* (the child died before settling) instead of *who won a race*. That is the correct axis, and the `run_turn` doc comment still reads true: a dead child that never read stdin **is** EOF-before-settlement, observed from the write side rather than the read side.

Downstream behavior change, checked and accepted: `AgentCrash` sets `retryable: true` (`runner/sequential.rs:373`), so a bad-auth pi is now retried up to the phase attempt budget before `Blocked`, where the old code failed fast. That is bounded, it is already today's behavior on the winning branch, and the operator now sees the real reason instead of a bare `failed writing pi RPC stdin`. Unification, not regression.

### The claim that had to be verified — and why it was load-bearing

The predicate is applied to the result of `self.write()` = `write_json_line` **as a whole**, so both fallible awaits (`write_all` **and** `flush`) are covered. On Windows this is not a nicety, it is the entire fix:

- tokio's `ChildStdin` on Windows is `Blocking<ArcFile>` — `tokio-1.53.1/src/process/windows.rs:193,247`.
- `Blocking::poll_write` copies into a buffer, dispatches the real write to the blocking pool, and returns `Ready(Ok(n))` **immediately** — `tokio-1.53.1/src/io/blocking.rs:109-136`. The underlying `io::Error` is only observed on the next `poll_write` or in `poll_flush`.
- So for one small JSONL line, `write_all` returns `Ok` and the `BrokenPipe` surfaces at **`flush`**, labelled `"failed flushing pi RPC stdin"`.

Had the predicate been applied only to the `write_all` error — the literal reading of the issue's "catch `BrokenPipe`/`WriteZero` from the initial `self.write(&request)`" — **the fix would have been a no-op on Windows.** The implementation got this right. Empirically corroborated: `dead_child_stdin_write_still_reports_stderr_reason` passes on Test (windows-latest), job 99433374785.

On the `BrokenPipe`|`WriteZero` pair: the controller's standalone std-sync repro is necessary but not sufficient, since the async path differs in *which await reports*. I re-derived rather than inherited — `Blocking` runs plain `std::io::Write` on the blocking pool, so the error reaching `flush` is the same std error with the same kind, and Rust maps both `ERROR_BROKEN_PIPE` (109) and `ERROR_NO_DATA` (232) to `ErrorKind::BrokenPipe`. No async-specific remapping exists. Windows CI green on the new test closes it.

### Is the new test really deterministic?

Yes — it is not a narrower race. It polls `child.try_wait()` until `Some` before invoking the turn, so the child is **reaped**, not merely "probably dead"; every holder of the stdin read end is then gone and the write cannot succeed. The poll loop is bounded by a 10 s assert-deadline, so a fake that never exits fails loudly. One non-obvious detail that makes the assertions meaningful rather than accidental: `try_wait()` caches the status into `FusedChild::Done` (`tokio process/mod.rs:1413-1422`), so the later `wait()` returns immediately and `eof_error`'s own `try_wait()` still reports the non-zero status — the test is not passing via an `Ok(None)` fallback.

---

### FOLLOW-UP ISSUE 1 (P2 — routed, does not extend this PR)

**`is_child_gone_error` keys on pipe state, not process state, and the branch it gates contains an uninterruptible wait.**

`BrokenPipe` means the read end is closed. That is overwhelmingly "the child exited", but a *live* process that closes its own stdin produces the identical error. In that case `self.child.wait().await` (`pi_rpc.rs:291`) blocks with no timeout and no cancellation — it sits before `let deadline = …` and outside the `select!`, so neither the phase timeout nor the `CancellationToken` can interrupt it. The pre-change code returned `Err` promptly.

Not blocking, for three reasons: (a) `pi --mode rpc` does not close its own stdin, and `run_turn` is invoked once per freshly-spawned session, so there is no window in which it plausibly would; (b) the identical unbounded reap is **pre-existing at two other sites** — the stdout-EOF branch (`:317`) and `request()` (`:432`) — both reachable in the same "one pipe closed, process alive" shape, so this change adds a third instance of an established pattern rather than a novel hazard; (c) fixing only this one would be an unscoped partial fix leaving the other two. The right shape is one issue bounding all three reaps with `tokio::time::timeout`.

### FOLLOW-UP ISSUE 2 (P3 — nit, fold into the above)

`dead_child_stdin_write_still_reports_stderr_reason` drives the session directly and has **no outer `tokio::time::timeout`**, where every other end-to-end test gets 30 s via `run_fake`. Since the branch it pins is the one containing the unbounded reap, a future regression there would present as a hung CI job rather than a failed test.

---

### Evidence — RAN vs READ

**READ.** Exact-head CI: run [33374674299](https://github.com/fagemx/edda/actions/runs/33374674299), `head_sha` independently confirmed `= b04fe369…`, `conclusion: success`, **all 7 required checks green** (Format; Clippy ×3; Test ×3). I read the ubuntu (99433374941) and Windows (99433374785) job logs directly rather than trusting the summary, and confirmed all three relevant tests **executed and passed, not filtered**: `startup_stderr_reaches_the_crash_error`, `dead_child_stdin_write_still_reports_stderr_reason`, `child_gone_error_predicate_matches_broken_pipe_and_write_zero`. Windows `edda-conductor`: 206 passed, 0 failed, 1 ignored (the pre-existing Codex live drill). The special rule for this PR — a red guard on Test (ubuntu) would be signal, not the historical flake — **did not trigger; it is green.** Worker receipt read.

**RAN (locally).** *Nothing compiled.* Stated reason for not re-running: `edda-conductor` is in the CI Windows test subset **and** in the full ubuntu/macOS workspace runs; the frozen surface is one file in that crate, and CI executed the exact new tests on all three platforms. The ladder's L2 trigger for a local Windows run — a crate *outside* the Windows subset — does not apply, and I found no coverage gap CI does not close. **Build lane: n/a.**

Non-Cargo checks I did run: git-level diff forensics (deleted-line census; base-vs-head byte comparison of the guard test; `#[ignore]` census at both SHAs) and a read of the vendored tokio 1.53.1 sources to re-derive the Windows write/flush ordering instead of inheriting the claim.

**Receipt caveats.** Both disclosed caveats are acceptable and neither is load-bearing here: the "25 passed" filtered figure is superseded by CI's full-crate runs, and I confirmed the shell-derived SHA equals `headRefOid`. The `LNK1104` was environmental, in a discarded attempt, and touched no product code.

**Cost.** ~25 tool calls, no local compilation, no lane consumed.

---

**LGTM @ `b04fe369e2beb983b034d92d5d02ff593a94869d`. P0 = 0, P1 = 0.** Any push to this head voids this verdict and requires Round 2. Merge remains operator authority.

PR #539, opened 2026-08-31
feat(conductor): verdict-gated phases (AWAITING_VERDICT) + edda verdict ledger object (#519)

### LINKED ISSUE #519: feat(conductor): approval-gate phase primitive (AWAITING_VERDICT) for controller/human-in-the-loop
## Motivation (live fleet run, 2026-08-31)

Fleet protocol used freeze points (worker halts after task groups; verifier runs the gate set on the frozen SHA; verdict LGTM/changes-requested; worker continues). Every step of that loop was **manual**: worker stops in its TUI, controller reads the terminal, runs gates by hand, then pastes "continue".

Conductor already has the ingredients — `check/wait_until.rs`, `check/edda_event.rs`, plan DAG, per-phase state machine — but no first-class way for a phase to **stop and wait for an external verdict**.

## Proposal

1. Plan schema: phase may declare `gate: verdict` (optionally `gate_timeout`, `on_reject: retry|halt`).
2. Runner: on phase completion, enter `AWAITING_VERDICT`; persist state; emit notify event (`edda-notify` channels already exist: ntfy/webhook/telegram).
3. CLI: `edda phase approve <plan>/<phase> [--comment]` and `edda phase reject <plan>/<phase> --comment` — reject feeds the comment back as the next turn to the same agent session (requires launcher session continuity, see the RPC-adapter issue).
4. Verdicts are ledger entries (who/when/SHA/comment) — auditable, consistent with decide/ratify semantics.

## doneWhen

- [ ] A plan with `gate: verdict` pauses after the phase, survives conductor restart in `AWAITING_VERDICT`, and resumes on `approve`
- [ ] `reject --comment` re-dispatches the comment into the same worker session (turn continuity)
- [ ] Verdict recorded in the ledger with actor + timestamp + frozen SHA
- [ ] Notify event fires on gate entry (configurable channel)

🤖 Drafted with Claude Code from a live fleet-run retrospective.

fagemx:
Closes #519

## What this adds

A conductor phase can declare `gate: verdict`. After its checks pass the phase enters **AWAITING_VERDICT** instead of Passed: the runner persists that state, emits a gate event (+ notifier), prints the exact approve/reject commands, and waits. An external actor — any platform, any session — issues the verdict through a new CLI, and the plan resumes.

This is the missing piece for unattended orchestrator → executor → verifier → orchestrator loops: the loop that produced this PR (and the four before it) was driven by hand.

## Design

Per the architecture recorded as `orchestration.cross-platform=ledger-data-plane-adapter-control-plane`, **verdicts are ledger objects, not conductor-internal state** — the conductor *consumes* them:

    edda verdict approve <subject> --sha <full-40-hex> [--comment "..."]
    edda verdict reject  <subject> --sha <full-40-hex> --comment "..."   # comment required

- `<subject>` is free-form; conductor gates use `<plan>/<phase>`. Fields: subject, decision, full SHA, comment, actor, timestamp. A verdict bound to a different SHA stays findable but never satisfies a gate waiting on another — push-voids-verdict falls out structurally.
- Schema: `gate: verdict`, `gate_timeout_sec`, `on_reject: redispatch|halt`; unknown gate values fail parse loudly.
- Runner: capture `gate_sha` (phase cwd HEAD) → persist AWAITING_VERDICT with `gate_entered_at` → poll the ledger → approve resumes; **reject redispatches ONE turn in the SAME session id** (rejection comment becomes the prompt) without incrementing `attempt`; `halt` fails with the comment as the error. Restart re-enters the wait without re-running the agent turn or checks; the timeout deadline is computed from the persisted `gate_entered_at`, so it survives restarts. A phase without `gate:` is unchanged.
- `edda-conductor` now depends on `edda-core` + `edda-ledger` — a new dependency edge, required to consume verdicts.

## D6 — a design defect found and corrected mid-delivery (please review this closely)

The controller's original frozen design said a redispatch re-enters the gate "with a fresh `gate_sha`". **That assumption is false**: `gate_sha` is the phase cwd's git HEAD, and a redispatch turn is not guaranteed to commit anything. With HEAD unchanged the re-entered gate has *exactly the same* `(subject, gate_sha)`, so the wait immediately re-read the **same rejected verdict still in the ledger** and redispatched again — forever. Because the design also forbids incrementing `attempt` on redispatch, `max_attempts` never tripped.

Measured before the fix: `gate_reject_redispatches_same_session_then_regates` produced **176** `gate_entered`/`verdict_received` events with exactly **one distinct `gate_sha`**; the loop stopped only because the test injected an approval. Unattended, that burns agent turns and money until a human intervenes.

Correction (recorded as `gate.verdict-freshness=verdict-must-postdate-gate-entered-at` and `gate.redispatch-counter=persisted-gate_redispatches-u32-max-3`):
1. A verdict satisfies a gate only if recorded **after that gate's `gate_entered_at`** — already persisted, so the rule survives restarts.
2. Redispatch cycles are bounded by a **separate persisted counter** (`gate_redispatches`), not `attempt`; exhausting it fails the phase like `halt`.
3. Regression test `stale_reject_does_not_resatisfy_regated_same_sha` reproduces the exact 176-loop scenario and asserts exactly one redispatch, then a distinct gate timeout.

**Attribution:** the defect was in the controller's frozen design, not the implementer's work — the implementation faithfully followed the spec it was given. It was caught by the controller running the suite, not by review.

## Provenance

Implemented by pi (glm-5.3-flash) via `edda conduct run --agent pi` in an isolated worktree. The first dispatch was killed mid-p3 by an external session teardown (Windows `STATUS_CONTROL_C_EXIT`) with all work uncommitted but intact; a resume dispatch implemented D6 and sealed. Two phases in this delivery lost an attempt to environmental `LNK1104` linker file locks (third occurrence on this workstation) and passed on retry — no product failure involved.

## Gates (frozen SHA `752ce4d58d7bc1c40409163f24f0a206dcdb0c2d`, receipt SHA shell-derived)

| Gate | Result |
|---|---|
| `cargo test -p edda-conductor` | 231 passed |
| `cargo test -p edda` | 373 passed |
| `cargo test -p edda-ledger` | 225 passed |
| `cargo clippy -p edda-conductor -p edda -p edda-ledger -p edda-core --all-targets -- -D warnings` | clean |

Lane worker-1. Independent verifier round (read-only, verdict pinned to the SHA above) follows as a PR comment before merge.

Follow-ups deliberately **not** in this PR: `edda phase approve/reject` sugar, parallel-runner gate support, notify channel config.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -890,6 +890,8 @@ version = "0.3.0"
 dependencies = [
  "anyhow",
  "async-trait",
+ "edda-core",
+ "edda-ledger",
  "edda-store",
  "regex",
  "serde",
--- a/crates/edda-cli/src/cmd_bridge.rs
+++ b/crates/edda-cli/src/cmd_bridge.rs
@@ -1224,7 +1224,7 @@ pub fn request_ack(
 /// fallback preserves genuine standalone CLI use. A carrier can preserve only
 /// the identity its host exposes; Codex tool hooks, for example, attribute
 /// subagent commands to the parent session (GH-503).
-fn resolve_session_id(
+pub(crate) fn resolve_session_id(
     cli_session: Option<&str>,
     project_id: &str,
     fallback_label: &str,
--- a/crates/edda-cli/src/cmd_conduct.rs
+++ b/crates/edda-cli/src/cmd_conduct.rs
@@ -546,6 +546,7 @@ fn print_status(state: &PlanState) {
             PhaseStatus::Running | PhaseStatus::Checking => "\u{25B6}", // ▶
             PhaseStatus::Skipped => "\u{2298}",                         // ⊘
             PhaseStatus::Stale => "\u{23F0}",                           // ⏰
+            PhaseStatus::AwaitingVerdict => "\u{23F8}",                 // ⏸
             PhaseStatus::Pending => "\u{25CB}",                         // ○
         };
         let detail = match ps.status {
--- a/crates/edda-cli/src/cmd_dispatch.rs
+++ b/crates/edda-cli/src/cmd_dispatch.rs
@@ -215,6 +215,9 @@ pub fn build_phase(
         budget_usd,
         allowed_tools: None,
         permission_mode: permission_mode.to_owned(),
+        gate: None,
+        gate_timeout_sec: None,
+        on_reject: Default::default(),
     }
 }
 
--- a/crates/edda-cli/src/cmd_verdict.rs
+++ b/crates/edda-cli/src/cmd_verdict.rs
@@ -1,0 +1,390 @@
+//! `edda verdict` — issue a verdict on a gated subject (GH-519 D1).
+//!
+//! An external actor approves or rejects a subject bound to (subject, full
+//! SHA); each verdict is a first-class `verdict.recorded` ledger event. The
+//! conductor consumes verdicts, it does not own them. A verdict for a
+//! mismatched SHA is findable but never satisfies a gate waiting on another
+//! SHA. `<subject>` is a free-form string; for conductor gates it is
+//! `<plan-name>/<phase-id>`.
+
+use clap::Subcommand;
+use edda_core::event::new_verdict_event;
+use edda_core::secret_guard::redact;
+use edda_core::{VerdictDecision, VerdictPayload};
+use edda_ledger::lock::WorkspaceLock;
+use edda_ledger::Ledger;
+use std::path::Path;
+
+#[derive(Debug, Subcommand)]
+pub enum VerdictCmd {
+    /// Approve a gated subject — the waiting gate may resume
+    Approve {
+        /// Free-form subject; for conductor gates `<plan-name>/<phase-id>`
+        subject: String,
+        /// Full 40-hex git SHA the verdict applies to
+        #[arg(long)]
+        sha: String,
+        /// Optional context recorded with the approval
+        #[arg(long)]
+        comment: Option<String>,
+        /// Session ID (uses EDDA_SESSION_ID; --session required when identity is ambiguous)
+        #[arg(long)]
+        session: Option<String>,
+    },
+    /// Reject a gated subject — the comment feeds back into the gated agent session
+    Reject {
+        /// Free-form subject; for conductor gates `<plan-name>/<phase-id>`
+        subject: String,
+        /// Full 40-hex git SHA the verdict applies to
+        #[arg(long)]
+        sha: String,
+        /// Required: why the gate was rejected (fed back to the agent as the next turn)
+        #[arg(long)]
+        comment: String,
+        /// Session ID (uses EDDA_SESSION_ID; --session required when identity is ambiguous)
+        #[arg(long)]
+        session: Option<String>,
+    },
+}
+
+/// Arguments for recording a verdict (mirrors the `verdict.recorded` payload).
+pub struct RecordVerdictArgs<'a> {
+    pub subject: &'a str,
+    pub decision: VerdictDecision,
+    pub sha: &'a str,
+    pub comment: Option<&'a str>,
+    pub cli_session: Option<&'a str>,
+}
+
+#[derive(Debug)]
+pub struct RecordOutcome {
+    pub event_id: String,
+    pub decision: VerdictDecision,
+}
+
+/// A verdict binds to a full 40-hex git SHA. A short or malformed SHA would
+/// silently never match a gate's `gate_sha`, so it is refused at the door.
+pub(crate) fn validate_sha(sha: &str) -> anyhow::Result<()> {
+    if sha.len() == 40 && sha.bytes().all(|b| b.is_ascii_hexdigit()) {
+        Ok(())
+    } else {
+        anyhow::bail!("--sha must be a full 40-hex git SHA (got \"{sha}\")")
+    }
+}
+
+pub fn run(cmd: VerdictCmd, repo_root: &Path) -> anyhow::Result<()> {
+    let (subject, sha, decision, comment, cli_session) = match cmd {
+        VerdictCmd::Approve {
+            subject,
+            sha,
+            comment,
+            session,
+        } => (subject, sha, VerdictDecision::Approved, comment, session),
+        VerdictCmd::Reject {
+            subject,
+            sha,
+            comment,
+            session,
+        } => (
+            subject,
+            sha,
+            VerdictDecision::Rejected,
+            Some(comment),
+            session,
+        ),
+    };
+    let outcome = do_record(
+        repo_root,
+        &RecordVerdictArgs {
+            subject: &subject,
+            decision,
+            sha: &sha,
+            comment: comment.as_deref(),
+            cli_session: cli_session.as_deref(),
+        },
+    )?;
+    println!("Verdict recorded: {} {subject} @ {sha}", outcome.decision);
+    println!("  event: {}", outcome.event_id);
+    if let Some(c) = comment.as_deref().filter(|c| !c.trim().is_empty()) {
+        println!("  comment: {c}");
+    }
+    Ok(())
+}
+
+pub(crate) fn do_record(
+    repo_root: &Path,
+    args: &RecordVerdictArgs<'_>,
+) -> anyhow::Result<RecordOutcome> {
+    validate_sha(args.sha)?;
+    if args.subject.trim().is_empty() {
+        anyhow::bail!("subject must not be empty");
+    }
+    // Comment is REQUIRED on reject: it is what feeds back into the gated
+    // agent session as its next turn. Enforced here, not just at the CLI
+    // layer, so every writer of the ledger object honors the invariant.
+    if args.decision == VerdictDecision::Rejected
+        && args.comment.is_none_or(|c| c.trim().is_empty())
+    {
+        anyhow::bail!(
+            "a rejection without a comment does not exist — the comment is what feeds \
+             back into the gated agent session"
+        );
+    }
+
+    // EDDA-SECRET-GUARD1 q331: scrub the comment before any persistence,
+    // same deterministic zero-LLM pass decide/note run.
+    let comment = match args.comment {
+        Some(c) => {
+            let (safe, hits) = redact(c);
+            if !hits.is_empty() {
+                eprintln!(
+                    "⚠ secret-guard: redacted {} secret pattern(s) before writing verdict",
+                    hits.len()
+                );
+            }
+            Some(safe)
+        }
+        None => None,
+    };
+
+    // Same identity source as other edda writes (decide/claim/request).
+    let project_id = edda_store::project_id(repo_root);
+    let (_session_id, label) =
+        crate::cmd_bridge::resolve_session_id(args.cli_session, &project_id, "cli")?;
+
+    let ledger = Ledger::open(repo_root)?;
+    let _lock = WorkspaceLock::acquire(&ledger.paths)?;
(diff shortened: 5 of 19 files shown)


fagemx:
## Code Review: Round 1

**Reviewed SHA:** `752ce4d58d7bc1c40409163f24f0a206dcdb0c2d` (verified equal to `headRefOid`)
**Delta base:** `3ede020008c3466ea64af920f067e70338818580` — 19 files, +2759/−218
**Verdict: LGTM — P0 = 0, P1 = 0.** Three follow-ups routed, none extending this PR.

Standing: I am independent of both the issue and the implementation, and the controller disclosed that it authored the D6 defect *and* its correction and asked me to attack the correction rather than grade conformance to it. I did. The D6 diagnosis is **correct — not a misdiagnosis** — and the correction is sound, for a better reason than the one recorded in the code. Details below, along with three holes in it that I do not consider blocking.

---

### The D6 correction: re-derived, not inherited

**The claimed defect is real.** I did not reproduce the 176-cycle run (that would require reverting the fix); I verified the mechanism from the shipped code, which is the stronger check. `gate_sha` is `capture_git_head(phase_cwd)` and a redispatch turn is under no obligation to commit, so a re-entered gate legitimately carries the *same* `(subject, gate_sha)`. With an unbounded `latest_verdict`, the rejection that triggered the redispatch is still the latest matching record, so it re-satisfies immediately — and because D3 deliberately does not increment `attempt`, `max_attempts` never trips. Unbounded loop. The diagnosis holds.

**The correction works, and I checked the one thing it actually hinges on.** Freshness is only a fix if `gate_entered_at` is *rewritten* on every gate entry — if re-entry reused the original bound, the same rejection would still postdate it and the loop would survive the "fix" intact. It is rewritten unconditionally: `sequential.rs:802` writes `gate_entered_at: Some(now_rfc3339())` inside the same `PhaseUpdate` as `gate_sha`, and the redispatch path reaches it because it routes back through `process_phase_result`. So the second gate's bound strictly postdates the rejection that caused it. Correct.

**On mechanism choice — freshness-by-timestamp vs consuming/marking the verdict.** You asked whether timestamps are the right mechanism at all. I judge **yes**, and I think the strongest argument is not the one in the code. A consume-marker on an append-only hash-chained ledger means the *reader* appends a `verdict.consumed` event — which would break D1's own invariant that "the conductor CONSUMES verdicts; it does not own them", add a new public event type, and introduce a fresh race between two conductors racing to consume one verdict. The timestamp bound requires no schema change and no write from the reader, so it preserves the property D1 exists to protect. Right call.

**Direction of the comparison is right.** `ts > bound` is strict. Strict errs toward *waiting* (a boundary-instant verdict is ignored, recoverable by recording another or by timeout); non-strict would err toward *re-admitting* a stale verdict, which is the defect. Correct asymmetry.

**Attacks that closed clean**, listed so you know they were made:

| Probe | Result |
|---|---|
| Same-instant / resolution collision | Closed. Both sides use `now_rfc3339()` = `OffsetDateTime::now_utc()` at full clock resolution with fractional seconds; the observe-gate→spawn-`edda verdict` round trip is orders of magnitude above any clock tick. Not reachable. |
| Clock skew between writer and runner | Both timestamps come from the same machine's clock in the supported topology. A verdict arriving via cross-project sync from a skewed clock would be exposed, but gates have no such path today. |
| Window between checks passing and `gate_entered_at` being written | **No window exists** — `checks`, `gate_sha` and `gate_entered_at` are written in one `transition` call (`:799-804`). Nothing can land between them. |
| Restart reusing an OLD `gate_entered_at` | Correct *and load-bearing*: it is what lets a verdict recorded while the conductor was down still satisfy the gate. Pinned by `restart_reenters_gate_wait_without_rerunning`, which records the approval **before** resuming and asserts `call_count("a") == 0`. |
| `capture_git_head` failure → gate bypass? | **Fails closed** (`:834` → `fail_checking_phase`). A gate that cannot compute its SHA fails the phase; it never passes it. Security-correct direction. |
| Is max 3 defensible? | As a *backstop* yes — freshness is the mechanism, the counter is defence in depth, so its exact value matters less. See FOLLOW-UP 3 for the limitation it does carry. |

### doneWhen — all four met

| # | Item | Evidence |
|---|---|---|
| 1 | Pauses, survives restart in `AWAITING_VERDICT`, resumes on approve | `gate_pauses_then_approve_resumes` + `restart_reenters_gate_wait_without_rerunning` (asserts zero launcher calls on resume) |
| 2 | `reject --comment` re-dispatches into the same worker session | `gate_reject_redispatches_same_session_then_regates` **asserts** `calls[0].session_id == calls[1].session_id` and the comment in `calls[1].prompt` — assertion, not claim |
| 3 | Verdict in the ledger with actor + timestamp + frozen SHA | `VerdictPayload{actor, sha}` + event `ts`; `validate_sha` refuses anything but 40-hex |
| 4 | Notify fires on gate entry | `notifier.notify(...)` + `Event::GateEntered` at `:816-825` |

The issue proposed `edda phase approve/reject`; this ships `edda verdict approve/reject`. The doneWhen says "`reject --comment`", not a command name, so it is satisfied — and the sugar CLI is disclosed as deliberately deferred.

### Other in-scope items

**Dependency edge — sound.** `edda-conductor` → `edda-core` + `edda-ledger`. `edda-ledger` → `edda-core`; `edda-core` → nothing internal. **No cycle**, direction is strictly downward, and `edda-core` was already transitive via `edda-store`, so this makes an existing reach explicit rather than adding coupling. Consistent with D1's consume-don't-own framing.

**Hang-vs-fail — your concern is resolved in the code as shipped.** I checked all three levels: `GATE_TEST_DEADLINE = 30s` wrapping every runner join; `wait_for_gate_events` carries its own 30s deadline with an `assert!` (it runs *outside* the join timeout, so an unbounded helper there would have been the real hazard — it is bounded); `record_verdict` retries against a bounded loop ending in `panic!`. No gate test can hang.

**Non-gated plans.** `non_gated_plan_completes_without_gate_events` runs with **no git repo and no ledger** and asserts no gate events and `gate_sha.is_none()` — it proves the untouched path stays untouched rather than merely asserting it passes.

**Security.** Rejection comments are scrubbed before persistence (`EDDA-SECRET-GUARD1 q331`) — correct placement, since the ledger is append-only and a leaked secret there is not retractable. `validate_sha` refusing short SHAs at the door prevents the quiet failure mode where a verdict is recorded, reported as success, and silently never matches a gate. Reject-without-comment is refused at both parse time and record time.

---

### FOLLOW-UP ISSUE 1 (P2) — the freshness bound fails **open** when it cannot parse

`latest_verdict_fresh` (`verdict.rs:77-86`): an unparsable `not_before` yields `bound = None`, and `match bound { None => true, .. }` — **no bound at all, every stale verdict satisfies**. That is the loop-prone direction, and it is the *opposite* of the direction the same function chooses two lines later for an unparsable verdict `ts` (`is_some_and` → never satisfies, fail closed). The doc comment defends it with a caller invariant ("the caller always writes it via the same well-formed formatter"), but this is a `pub` function in `edda-ledger` defended by a property of one caller in `edda-conductor`, and `gate_entered_at` reaches it from a JSON state file on disk. A corrupt or hand-edited state file silently disables D6 entirely; only `MAX_GATE_REDISPATCHES` then stands between that and the original defect. A freshness bound that cannot be parsed should admit nothing, not everything.

### FOLLOW-UP ISSUE 2 (P2) — the gate poll swallows every ledger error with no operator signal

`wait_for_verdict` (`:611-618`) discards both `Ledger::open` and `latest_verdict_fresh` errors. For a transient lock that is right. For a persistent failure — corrupt DB, schema mismatch, permissions — the gate polls every 2s forever while printing nothing, and since `gate_timeout_sec: None` is the documented default ("wait until cancelled"), it never ends. The operator sees a correct-looking `AWAITING_VERDICT` banner with the exact approve command, runs it, and nothing happens, with no diagnostic anywhere in the runner output. One rate-limited warning on repeated identical errors closes it.

### FOLLOW-UP ISSUE 3 (P3) — two documentation defects, one of them in the wrong place to have one

1. `sequential.rs:2040` — `// Redispatch re-enters the gate with a fresh gate_sha` — states the **opposite** of the property under test. The `gate_sha` is not fresh; HEAD never moves in that test, and the sibling D6 test asserts exactly that at `:2427` (`shas[0] == shas[1]`). What is fresh is `gate_entered_at`. This comment encodes precisely the mental model whose absence produced the 176-cycle loop, three lines above the code that depends on the correct one.
2. **D6 silently removes the ability to pre-record a verdict.** Before the freshness bound, an operator could `git rev-parse HEAD` and approve *in advance*; now any verdict predating gate entry is ignored forever, and with no default timeout the plan waits indefinitely with no explanation. This is a correct and necessary consequence of D6 — you cannot allow pre-approval and block stale-reject re-satisfaction with timestamps alone — but it is a real behavior change that appears in neither the PR body nor the schema docs.

Also fold in: `MAX_GATE_REDISPATCHES` is hardcoded and never resets over a phase's life, while `gate_timeout_sec` is configurable. A legitimately long human review (more than three rejection rounds — this repo has shipped three-round PRs) fails the phase on a bound the plan author cannot raise.

---

### Evidence — RAN vs READ

**READ.** Exact-head CI: run [33397542871](https://github.com/fagemx/edda/actions/runs/33397542871), `head_sha` independently confirmed `= 752ce4d5…`, `conclusion: success`, **all 7 required checks green**. I read the Windows job log (99505586842) directly and confirmed the load-bearing tests **executed and passed there**, not merely on Linux: `stale_reject_does_not_resatisfy_regated_same_sha`, `restart_reenters_gate_wait_without_rerunning`, `gate_reject_redispatches_same_session_then_regates`, `gate_reject_exhausts_redispatch_bound_with_distinct_error`, `gate_timeout_fails_distinctly`, `non_gated_plan_completes_without_gate_events`, plus `verdict::tests::fresh_*` / `stale_verdict_*` and all nine `cmd_verdict::tests::*`. Worker receipt read.

**RAN (locally).** *Nothing compiled.* I read the diff via `git fetch origin pull/539/head` + `git show <sha>:<path>` — the working tree was never moved, which also matters because a peer session is editing in this same checkout.

**On the `edda-core` Windows gap you offered.** I declined the focused local run, with a reason rather than a shrug. I grepped every added line in the `edda-core` delta for platform-sensitive constructs (`Path`, `File`, `fs::`, `Command`, `std::process`, `cfg(windows)`, `cfg(unix)`, `mmap`, `lock`): the single hit is `refs: Refs::default()`. The delta is serde data types plus event construction — no path handling, no file locks, no process spawning, no memory mapping, i.e. none of the criteria the Windows subset was derived from (GH-433). Its library is still compiled and clippy'd on Windows, and its new types are exercised on Windows through `edda-ledger`'s and `edda-cli`'s own tests, both of which are in the subset and both of which I confirmed green above. The genuinely uncovered surface is `edda-core`'s unit tests over pure data. Running a workspace build to reach that is the cost the ladder exists to remove. **Build lane: n/a.**

**Gate-selection audit.** No workspace L1 was run and I agree none was needed: three of the four touched crates are in the CI Windows subset and all four are in the full ubuntu/macOS workspace runs, and the fourth is argued above. Recorded gates match the frozen SHA.

**Provenance noise** (mid-dispatch teardown, two `LNK1104` linker locks now filed as #540) is disclosed, environmental, and touched no product code. Discounted.

**Cost.** ~20 tool calls, no local compilation, no lane consumed.

---

**LGTM @ `752ce4d58d7bc1c40409163f24f0a206dcdb0c2d`. P0 = 0, P1 = 0.** Any push to this head voids this verdict and requires Round 2. Merge remains operator authority.

Issue #547, opened 2026-08-31
feat(cli): edda phase approve/reject sugar - resolve gate_sha and subject from conductor state


fagemx:
## Context

#519 originally proposed `edda phase approve <plan>/<phase>` / `edda phase reject ... --comment`. PR #539 shipped the generic primitive instead — `edda verdict approve|reject <subject> --sha <40-hex>` — and deferred the sugar ("Follow-ups deliberately not in this PR: `edda phase approve/reject` sugar, ..."). This issue gives that deferral a tracking number.

## Why the sugar is worth having

The two most error-prone inputs of the raw verb are exactly the two things conductor state already knows:

- **`--sha`**: a full 40-hex SHA must be copy-pasted from the AWAITING_VERDICT surface; a wrong-but-valid SHA is findable-but-never-matching by design, i.e. the failure mode is a silent eternal wait.
- **`subject`**: free-form; a typo'd `<plan>/<phase>` also waits forever.

`edda phase approve <plan>/<phase>` can resolve the gate_sha from the persisted `PhaseState` (`gate_sha` / `gate_entered_at` are already saved), validate that the named phase is actually in AWAITING_VERDICT, and refuse loudly when it is not — converting both silent failure modes into immediate errors. `--sha` stays available as an override for auditing discipline.

## Interplay

- Resolution of "which conductor state / which ledger" must use the same single authority #543 calls for — the sugar must not add a fourth notion of "where".
- `reject` keeps the mandatory `--comment` (it is the redispatch prompt).

Related: #519, #539, #543.

fagemx:
A third error-prone input for the sugar to resolve, found while relaying verdicts for a real gated plan.

`edda verdict approve|reject` also refuses to infer session identity when the caller is not a session edda launched:

```
$ edda verdict reject cleanup-wave1/i1-impl-544 --sha 9470cf5... --comment "..."
Error: cannot prove which live session belongs to this process, so --session is required
  (or set EDDA_SESSION_ID in the invoking process)
```

The guard itself is right — a verdict is an attribution-bearing record and guessing the actor would be worse. But it means the controller relaying an external reviewer's decision types a full session UUID on **every** verdict, alongside the 40-hex SHA this issue already covers. Over one three-issue wave that was four verdicts, each needing two hand-copied identifiers, and each with the same failure shape: a wrong-but-valid value is accepted and the gate then waits forever.

[...]

Issue #540, opened 2026-08-31
fix(conductor): a truncated check failure hides LNK1104's cause, and an environmental build failure spends a product retry attempt


fagemx:
## Symptom

Windows `LNK1104` (linker cannot open a file) has failed a conductor phase **three times in one day** on this workstation. Every occurrence was environmental — the retry passed on unchanged source — but each one burned a full phase attempt and, twice, tens of minutes.

| When | Where | Cost |
|---|---|---|
| gh524 verification, run 1 | main worktree `target/` | ~66s, run produced no test result |
| gh536 `fix` phase, attempt 1 | lane `worker-1`, linking `edda-conductor` | 6m47s |
| gh519 resume `p1-fix-and-green`, attempt 1 | lane `worker-1`, linking `edda-ledger` | **38m38s** |

Each time `max_attempts` absorbed it and attempt 2 passed. Silent from the outside: the plan reports "passed", and only the log shows a third of an hour was spent on a linker collision.

### 1. The diagnostic is truncated exactly where it matters

`LNK1104`'s payload is the filename it could not open — that single token distinguishes *a still-running test binary holding its own image* from *a concurrent cargo writing the same rlib* from *an antivirus handle*. Our captured failure text contains the full `link.exe` invocation (hundreds of object files) but **not** the `cannot open file '...'` line: verified by searching both raw logs, no `cannot open file` / `LINK : fatal error` substring is present anywhere.

So three occurrences produced zero evidence about the cause. The check output should keep the error's *tail* (where the fatal line lives), not only its head.

### 2. An environmental failure consumes a product retry attempt

The conductor's `max_attempts` ladder exists to retry *the agent's work*. A linker file-lock is not the agent's work, and spending an attempt on it is the same class of defect as #529 (a timed-out check burning the whole retry ladder): the budget meant for real failures is consumed by infrastructure noise. With `max_attempts: 2` — our standard — a single LNK1104 halves the real retry budget, and two in a row would fail a phase that never had a product problem.

## What we do not know

The proximate cause is **unidentified**, precisely because of defect 1. Candidate mechanisms, none confirmed:

- **An orphaned test binary still running** holds its own `.exe` open, so relinking that target fails. Plausible here: a hung gate test (see #538's class of unbounded waits — one was observed live during gh519 diagnosis, looping 176 times) can outlive the shell that started it.
[...]

PR #549, opened 2026-08-31
fix(edda-conductor): cmd_succeeds timeout defaults and retry classification

### LINKED ISSUE #529: bug(conductor): a 120s cmd_succeeds default makes test-suite checks unpassable, and a timeout burns the whole retry ladder
## Symptom

A conductor phase whose `check` runs the project's own test suite can never pass, and the failure spends the phase's entire retry ladder before surfacing.

Observed on a live `edda conduct run --agent pi` for #524 (2026-08-31):

```text
✓ diagnose passed (17m20s)
▶ fix attempt 1 → agent completed the work and committed, then:
   ✗ command timed out after 120s: cargo test -p edda
↻ auto-retry (1/3) → attempt 2 → same wall, same 120s
```

The agent's work was correct and already committed; only the verification could not finish. Roughly 35 minutes and the corresponding token spend went into re-running a phase that was structurally incapable of passing.

## The asymmetry

`crates/edda-conductor/src/plan/schema.rs`:

| Default | Value | Applies to |
|---|---:|---|
| `default_timeout_sec` | 1800 s | the agent phase itself |
| `default_cmd_timeout` | 120 s | a `cmd_succeeds` check |

An agent gets 30 minutes to do the work; the command that verifies the work gets 2 minutes. For this repo `cargo test -p edda` measures 60–150 s and `cargo test --workspace` far more, so the most natural check anyone would write is already at or past the default ceiling.

## Why it is expensive rather than merely wrong

1. **A timeout is treated exactly like a failed check.** Both land in `PhaseStatus::Failed` via the same path, so `on_fail: auto_retry` (the default) re-dispatches a full agent turn. A genuine check failure is worth retrying — the agent can fix it. A timeout is a property of the harness, and every retry hits the same wall at the same second.
2. **The distinction is not visible where it matters.** The `cmd_succeeds` message does say `command timed out after {n}s`, but the phase-level line is the same `✗ Phase "fix" failed` as a real failure, and the state file records no timeout-vs-failure distinction for a later reader.
3. **The failure mode is silent about its own cause.** Nothing warns that a check's timeout is implausibly short for the command being run.

## Directions worth evaluating

- Distinguish a check timeout from a check failure in phase state, event log, and console output.
- Make `on_fail: auto_retry` not consume the retry ladder on a timeout (halt-and-report, or retry once at most), since a retry cannot change the outcome.
- Reconsider the 120 s default for `cmd_succeeds` — raise it, tie it to the phase timeout, or require an explicit `timeout_sec` for commands that are known to be long (`cargo test`, `cargo clippy`).
- Warn at plan-parse time when a check command matches a known-slow shape and relies on the default.

## doneWhen

- [ ] A timed-out check is distinguishable from a failed check in console output and in the persisted phase state
- [ ] A timeout does not silently burn the full `max_attempts` ladder
- [ ] The `cmd_succeeds` default timeout is either raised or justified in a comment against a measured build/test time for this workspace
- [ ] A plan that runs this repo's own test suite as a check passes with no per-check `timeout_sec` override, or fails immediately with an actionable message

🤖 Drafted with Claude Code from a live conduct run that lost ~35 minutes to this.

fagemx:
Closes #529

## Summary
- `CheckOutput` gains a `timed_out` flag set by constructors; the engine classifies timeouts structurally (no string sniffing) into `ErrorType::Timeout` with `retryable: false`, persisted with the check result.
- `AutoRetry` stops reporting/burning `max_attempts` on timeout; genuine failures still retry as before.
- `default_cmd_timeout` raised 120s → 1800s, aligned with the agent-phase cap (rationale: GH-529 observed `cargo test -p edda` warm-up 60–150s under load).
- `wait_until` bounds its inner retry attempts by the outer budget; its own budget exhaustion is also classified as timeout.
- Round 2 (review feedback): two timeout tests got load-tolerant wall-clock guards — the outer-budget elapsed assertion 5s → 30s, and the gate-reject guard 60s → 120s with its rejecter 20s → 60s — so parallel full-suite load does not trip them.

## Disclosure: one bundled test that is not #529's
`gate_reject_exhausts_redispatch_bound_with_distinct_error` belongs to the earlier verdict-gate work, not to #529. Its guards were widened here only because `cargo test -p edda-conductor` is this change's own gate condition, so a load-flaky test in that crate blocks the work outright. Isolated run ≈19s; 60s was exceeded once under full-suite parallel load. It was verified as unaffected by this change: the test declares no checks, so neither the modified `fail_checking_phase` nor the modified `handle_on_fail` is reachable from it.

Correction to an earlier draft of this description: a genuine hang in that test is caught by the test's **own 120s `tokio::time::timeout`**, not by a runner gate deadline — `gate_timeout_sec` is absent there. Two in-code comments still carry the same inaccuracy (`runner/sequential.rs:2566` and `:2613-2614`); left as a follow-up rather than pushing a comment-only commit against a green, reviewed PR.

## Checks that ran green
- `cargo fmt --all --check`, `cargo clippy -p edda-conductor --all-targets -- -D warnings`, `cargo test -p edda-conductor` — all three at head `7b54792`, enforced as the conductor phase's own gate conditions rather than self-reported.
- All seven required CI checks green (Format + Clippy and Test on ubuntu, macOS and Windows).

## Review
Two rounds by an independent reviewer, recorded in the comments below: Round 1 requested changes (three P1s — one correctness, two over-building), Round 2 is LGTM at `7b54792` with P0 = 0, P1 = 0.

### DIFF
--- a/crates/edda-conductor/src/check/cmd_succeeds.rs
+++ b/crates/edda-conductor/src/check/cmd_succeeds.rs
@@ -76,7 +76,7 @@ pub async fn check_cmd_succeeds(cmd: &str, timeout_sec: u64, cwd: &Path) -> Chec
             )
         }
         Ok(Err(e)) => CheckOutput::failed(format!("spawn error: {e}"), start.elapsed()),
-        Err(_) => CheckOutput::failed(
+        Err(_) => CheckOutput::timed_out(
             format!("command timed out after {timeout_sec}s: {cmd}"),
             start.elapsed(),
         ),
@@ -121,6 +121,10 @@ mod tests {
         };
         let out = check_cmd_succeeds(cmd, 1, dir.path()).await;
         assert!(!out.passed);
+        assert!(
+            out.timed_out,
+            "a killed command must be marked as a timeout (GH-529)"
+        );
         assert!(out.detail.unwrap().contains("timed out"));
     }
 
--- a/crates/edda-conductor/src/check/engine.rs
+++ b/crates/edda-conductor/src/check/engine.rs
@@ -45,6 +45,13 @@ impl CheckEngine {
             });
 
             if !output.passed {
+                // GH-529: a harness-side timeout must not be classified as a
+                // retryable check failure — a retry cannot change the outcome.
+                let (error_type, retryable) = if output.timed_out {
+                    (ErrorType::Timeout, false)
+                } else {
+                    (ErrorType::CheckFailed, spec.is_retryable())
+                };
                 // Mark remaining as Waiting
                 for check in &checks[(i + 1)..] {
                     results.push(CheckResult {
@@ -58,11 +65,11 @@ impl CheckEngine {
                     all_passed: false,
                     results,
                     error: Some(ErrorInfo {
-                        error_type: ErrorType::CheckFailed,
+                        error_type,
                         message: output
                             .detail
                             .unwrap_or_else(|| format!("check {} failed", spec.type_name())),
-                        retryable: spec.is_retryable(),
+                        retryable,
                         check_index: Some(i),
                         timestamp: now_rfc3339(),
                     }),
@@ -187,5 +194,57 @@ mod tests {
         assert_eq!(result.results.len(), 2);
         assert_eq!(result.results[0].status, CheckStatus::Failed);
         assert_eq!(result.results[1].status, CheckStatus::Waiting); // short-circuited
+    }
+
+    /// GH-529: a timed-out cmd_succeeds must classify as `ErrorType::Timeout`
+    /// with `retryable: false`, distinct from a genuine check failure.
+    #[tokio::test]
+    async fn cmd_timeout_classifies_as_timeout_not_check_failed() {
+        let dir = tempfile::tempdir().unwrap();
+        let engine = CheckEngine::new(dir.path().to_path_buf());
+
+        // A command that never finishes inside its 1s budget.
+        #[cfg(windows)]
+        let cmd = "while ($true) { Start-Sleep -Milliseconds 100 }";
+        #[cfg(not(windows))]
+        let cmd = "sleep 30";
+        let checks = vec![CheckSpec::CmdSucceeds {
+            cmd: cmd.into(),
+            timeout_sec: 1,
+        }];
+
+        let result = engine.run_all(&checks, None).await;
+        assert!(!result.all_passed);
+        let err = result.error.expect("timeout must carry an error");
+        assert_eq!(err.error_type, ErrorType::Timeout);
+        assert!(
+            !err.retryable,
+            "retrying a timeout cannot change the outcome"
+        );
+        assert!(err.message.contains("timed out"), "got: {}", err.message);
+        assert_eq!(result.results[0].status, CheckStatus::Failed);
+    }
+
+    /// GH-529: a genuine command failure (non-zero exit) must STILL classify
+    /// as `ErrorType::CheckFailed` and stay retryable.
+    #[tokio::test]
+    async fn cmd_failure_stays_retryable_check_failed() {
+        let dir = tempfile::tempdir().unwrap();
+        let engine = CheckEngine::new(dir.path().to_path_buf());
+
+        #[cfg(windows)]
+        let cmd = "exit 1";
+        #[cfg(not(windows))]
+        let cmd = "false";
+        let checks = vec![CheckSpec::CmdSucceeds {
+            cmd: cmd.into(),
+            timeout_sec: 10,
+        }];
+
+        let result = engine.run_all(&checks, None).await;
+        assert!(!result.all_passed);
+        let err = result.error.expect("failure must carry an error");
+        assert_eq!(err.error_type, ErrorType::CheckFailed);
+        assert!(err.retryable);
     }
 }
--- a/crates/edda-conductor/src/check/mod.rs
+++ b/crates/edda-conductor/src/check/mod.rs
@@ -14,6 +14,11 @@ pub struct CheckOutput {
     pub passed: bool,
     pub detail: Option<String>,
     pub duration: Duration,
+    /// True when the failure was a harness-side timeout (GH-529). A timeout
+    /// is a property of the harness, not of the agent's work, so it must be
+    /// distinguishable from a genuine check failure downstream (retry
+    /// policy, console output, persisted error type).
+    pub timed_out: bool,
 }
 
 impl CheckOutput {
@@ -22,6 +27,7 @@ impl CheckOutput {
             passed: true,
             detail: None,
             duration,
+            timed_out: false,
         }
     }
 
@@ -30,6 +36,7 @@ impl CheckOutput {
             passed: true,
             detail: Some(detail),
             duration,
+            timed_out: false,
         }
     }
 
@@ -38,6 +45,18 @@ impl CheckOutput {
             passed: false,
             detail: Some(detail),
             duration,
+            timed_out: false,
+        }
+    }
+
+    /// A harness-side timeout failure (GH-529): the command outlived its
+    /// `timeout_sec`. Never classified as a retryable check failure.
+    pub fn timed_out(detail: String, duration: Duration) -> Self {
+        Self {
+            passed: false,
+            detail: Some(detail),
+            duration,
+            timed_out: true,
         }
     }
 }
--- a/crates/edda-conductor/src/check/wait_until.rs
+++ b/crates/edda-conductor/src/check/wait_until.rs
@@ -18,7 +18,23 @@ pub async fn check_wait_until(
 
     loop {
         attempt += 1;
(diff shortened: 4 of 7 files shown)


fagemx:
## Code Review: Round 1

**Reviewed SHA:** `cccbfcbf15a3032ce29e370db191982800149eee`
**Reviewer:** codex (independent identity — #529 was filed by a controller session, so self-review was disallowed)
**Context:** this round ran at an `edda conduct` verdict gate (`cleanup-wave1/i2-impl-529`), before the branch was pushed. Recorded here because the PR is the durable carrier; the gate itself later timed out and the phase was moved past manually (see the note at the end of Round 2).

### IN SCOPE
The four doneWhen boxes of #529 (timeout distinguishable in console and persisted state; timeout not burning `max_attempts`; the `cmd_succeeds` default raised or justified; a plan running this repo's own test suite passing with no per-check override), the changed code's direct callers, `wait_until`'s budget path, test coverage, and integration with base `a928b2a`.

Explicitly not asked for: the plan-parse-time "slow command" warning. #529 lists it under *directions worth evaluating*, not doneWhen — its absence is correct.

### Blocking findings — P0 = 0, P1 = 3

**P1-1 — correctness.** `check/wait_until.rs:21` awaited `run_inner` to completion and only consulted its own deadline afterwards. A `wait_until` with `timeout_sec: 600` wrapping a `cmd_succeeds` would therefore block for the inner budget on the very first attempt before its own budget was ever checked. Pre-existing in structure, but raising the inner default 120s → 1800s in this same change turned a 2-minute overrun into a 30-minute one, which puts it in scope here.

**P1-2 — over-building.** `runner/sequential.rs:1582-1607` re-implemented setup that `run_test_plan` already performs.

**P1-3 — over-building.** `plan/schema.rs:234-259` added a 26-line standalone deserialization test for a default that `plan/parser.rs::short_format_cmd_succeeds` already exercises.

### Explicitly confirmed correct, not to be changed
The typed `timed_out` flag on `CheckOutput` with exhaustive constructors is the right minimum — sniffing the `"command timed out after Ns"` message string would be shorter but wrong, and a smaller-but-flimsier algorithm is not the lazier answer. `ErrorType::Timeout` with `retryable: false`, the persisted state distinction, the distinct console line, `AutoRetry` halting on a timeout while genuine check failures still retry, and the 1800s default with its measured justification comment are all correct.

### A claim in the commit message, checked rather than trusted
The commit stated that `gate_reject_exhausts_redispatch_bound_with_distinct_error` is a pre-existing load flake. That is what an implementer says when it has broken a test, so it was verified rather than accepted: the test declares no checks, so neither the changed `fail_checking_phase` nor the changed `handle_on_fail` is reachable from it. The claim holds.

### Evidence
**READ:** the full diff, `check/`, `plan/schema.rs`, and the on_fail/retry path in `runner/sequential.rs`. P1-1 was independently re-read by the relaying controller before the verdict was sent.
**RAN:** nothing. `cargo fmt --all --check`, `cargo clippy -p edda-conductor --all-targets -- -D warnings` and `cargo test -p edda-conductor` had already passed at this SHA as the phase's own gate conditions — machine-enforced, since the phase cannot reach review until they pass.

### Verdict
**Changes Requested** — P0 = 0, P1 = 3.

fagemx:
## Review Response: Round 1

**New SHA:** `7b547929ecd1488792f8a3047dd317a118773544`, via two commits stacked on the reviewed `cccbfcb`:
- `b86fcac` — fix(edda-conductor): bound wait_until inner attempts by outer budget; trim review overbuild
- `7b54792` — test(edda-conductor): tolerate parallel-load slowness in GH-529 timeout tests

Recorded on the implementer's behalf by the controlling session: the implementing agent ran inside a conductor phase and does not post to GitHub. The rejection was delivered to it verbatim as its next prompt in the same session, so the changes below are its own response.

### Answers to the three blocking findings

**P1-1.** `check_wait_until` now wraps the inner call in `tokio::time::timeout` with the remaining budget and returns `CheckOutput::timed_out` on elapse — no new abstraction, no budget tracker. A focused regression was added: an inner `cmd_succeeds` whose own timeout (60s) far exceeds the outer budget (1s), asserting the wait returns promptly and `timed_out` is true. The pre-existing `wait_until_timeout` test also gained the missing `out.timed_out` assertion rather than a second test being added.

**P1-2.** The duplicated block collapsed to `run_test_plan(&yaml, &launcher).await`; all assertions including `launcher.call_count` still hold.

**P1-3.** The standalone `schema.rs` test was deleted and `parser.rs::short_format_cmd_succeeds` extended to assert `timeout_sec == 1800`.

Net effect versus the reviewed SHA: `schema.rs` −27, `sequential.rs` +38/−33, `parser.rs` +14/−2, `wait_until.rs` +58/−1. The diff shrank in the two places the review named.

### One intermediate failure, and what it changed

The first fix commit failed the phase's own `cargo test -p edda-conductor` check under parallel load, costing one `attempts` increment. The second commit widened two wall-clock guards in response — the outer-budget elapsed assertion 5s → 30s, and the gate-reject test's rejecter deadline 20s → 60s with its outer timeout 60s → 120s (isolated run ≈19s). Both are disclosed in the PR body; the second belongs to earlier gate work rather than to #529 and is bundled only because `cargo test -p edda-conductor` is this phase's gate condition, so a flaky test in that crate blocks the work outright.

### Gates ran at the new SHA
`cargo fmt --all --check`; `cargo clippy -p edda-conductor --all-targets -- -D warnings`; `cargo test -p edda-conductor` — all green, enforced as the phase's gate conditions rather than self-reported.

fagemx:
## Code Review: Round 2 — LGTM

**Reviewed SHA:** `7b547929ecd1488792f8a3047dd317a118773544` (current head)
**Reviewer:** codex, independent identity as in Round 1.

### IN SCOPE this round
The fix and what the fix could have broken — later-round blockers must be fix-caused or previously unobservable. Specifically: whether the three P1s are closed, and whether the two widened wall-clock guards still constitute checks.

### Findings

**All three P1s closed as specified.**

**The widened guards were the priority of this round, and both survive scrutiny.**

The elapsed assertion in `wait_until_outer_budget_bounds_slow_inner_cmd` (5s → 30s) has little standalone margin — 30s is close to what a broken implementation would produce, since the inner command is `sleep 30` on non-Windows. But it is not the assertion carrying the test. With the outer bound removed, `sleep 30` **succeeds**, so `run_inner` returns a passing result and the test fails at `assert!(!out.passed)` and again at `assert!(out.timed_out)` long before elapsed time matters. On Windows the inner command is an unbounded loop, so the elapsed bound bites there as well. The test cannot pass with the original bug present, which is the only question that decides this.

*(The relaying controller had independently flagged the 30s threshold as sitting on the failure value and expected it to be blocking. The reviewer's analysis is the better one: the semantic assertions, not the clock, are what catch the regression. Recorded here because the earlier concern was raised out loud.)*

The gate-reject widening (rejecter 20s → 60s, outer 60s → 120s against a ~19s isolated run) is legitimate load tolerance with real headroom: the semantic assertions still catch excess redispatches, and the 120s outer bound still catches a genuine hang. Bundling a test from earlier gate work into this PR is acceptable because that crate's test suite is this change's own gate condition — and it is disclosed in the PR body rather than shipped silently.

### NON-BLOCKING — follow-up, not for this PR
Two comments in `runner/sequential.rs` are now stale: line 2566 still says "30s", and lines 2613-2614 describe a "runner gate deadline" as the thing bounding a hang, when `gate_timeout_sec` is absent in that test and the actual bound is the test's own 120s `tokio::time::timeout`. Comment-only, in test code; not worth a new commit and another review round against a green PR. The same inaccuracy appears in this PR's body and has been corrected there in place.

### FOLLOW-UP ISSUE
None from this PR. Separately, this run's operational findings were filed as #551 and #552 and commented onto #540 and #547; none of them block this change.

### Evidence
**READ:** `git diff a928b2a..7b54792`, `git show b86fcac..7b54792`, and the surrounding `check/` and `runner/` code.
**RAN:** nothing either round. The three gates passed at both frozen SHAs as the phase's own gate conditions; a second run for an already-recorded SHA needs a reason, and there was none. Workspace-wide gates are CI's job and all seven required checks are green at this head.

### Verdict
**LGTM** at current head `7b547929ecd1488792f8a3047dd317a118773544` — **P0 = 0, P1 = 0.**

Gate set at this SHA: `cargo fmt --all --check`, `cargo clippy -p edda-conductor --all-targets -- -D warnings`, `cargo test -p edda-conductor`. Any push to this branch invalidates this verdict.

### Process note, recorded rather than hidden
The verdict gate that produced Round 1 later timed out (7200s with no verdict, because the reviewing agent hung and the controller session restarted). The phase was moved past with the interactive `Skip`, so the plan's ledger records `i2-impl-529` as "manually skipped". That label is inaccurate: the phase ran to completion and all three checks passed at this SHA before it ever reached the gate. The review then happened here, on the PR, which is the record that matters. The defects this exposed are filed as #551 and #552.

PR #553, opened 2026-09-01
fix(edda-conductor): bound the pi_rpc child reaps

### LINKED ISSUE #538: fix(edda-conductor): bound the three uninterruptible child reaps in pi_rpc and the dead-child test
## Context

Routed from PR #537 round-1 review (verifier finding, explicitly non-blocking there: third instance of a pre-existing pattern, and the new branch is unreachable for today's `pi --mode rpc` behavior).

## The pattern: uninterruptible child reaps

`is_child_gone_error` keys on **pipe state**, not **process state**. The branch it gates contains an unbounded `self.child.wait().await` (pi_rpc.rs:291) sitting before the deadline and outside the `select!` — so a hypothetical live child that closed its own stdin (write side) would hang the turn where the old code returned `Err` promptly.

The same unbounded reap already exists at pi_rpc.rs:317 and :432 (pre-existing). Three sites, one pattern: `child.wait().await` with no time bound on paths where the child's exit is expected but not guaranteed.

Not blocking today because: pi RPC mode does not close its stdin read side while alive; `run_turn` runs once per fresh session; and the two older sites predate #537. But the failure mode (silent hang instead of prompt error) is the worst kind for an unattended fleet.

## Also fold in (P3 from the same review)

`dead_child_stdin_write_still_reports_stderr_reason` has no outer `tokio::time::timeout`, while `run_fake` gives sibling tests a 30s bound — a regression in the reclassified branch would present as a **hung CI job** rather than a failed test.

## Fix direction

- Bound all three reaps with a grace timeout (a sibling constant to `STDERR_DRAIN_GRACE`); on expiry, proceed with the stderr tail collected so far and an explicit "child did not exit within grace" note in the error, never a hang.
- Give the deterministic test the same outer timeout its siblings have.

## doneWhen

- [ ] No `child.wait().await` in pi_rpc.rs without a time bound
- [ ] Grace expiry produces a prompt, informative error (not a hang, not a panic)
- [ ] `dead_child_stdin_write_still_reports_stderr_reason` bounded like its siblings
- [ ] Existing behavior on the happy/crash paths unchanged (tests green, guard's 5x loop untouched)

🤖 Routed from the PR #537 round-1 review (verifier: session Opus-edd-3).

fagemx:
Closes #538

## Summary

All `child.wait().await` sites in `pi_rpc.rs` now run under a `REAP_GRACE` (5s) timeout — a sibling constant to `STDERR_DRAIN_GRACE`:

- `terminate()` (kill → reap)
- the dead-stdin write path in `run_turn` (:291 pre-patch)
- the stdout-EOF path in `run_turn` (:317 pre-patch)
- the EOF path in `request()` (:432 pre-patch)

On grace expiry the flow proceeds with the stderr tail collected so far (preserving the #536 guarantee) and `eof_error` appends an explicit "; child did not exit within grace" note via a new `reaped` flag — prompt informative error, never a hang, never a panic. `kill_on_drop` still reaps at session drop. Happy and crash paths unchanged; the guard 5x loop untouched; no new abstraction, helper struct, or config knob.

`dead_child_stdin_write_still_reports_stderr_reason` now has the same 30s outer `tokio::time::timeout` its siblings have.

## Regression test (verified to fail without the fix)

`reap_grace_expiry_reports_instead_of_hanging` (unix-only, same gate as `crlf_and_u2028_do_not_split_records`): a fake pi closes stdout and stays alive — stdout EOF with a live child.

- with fix: **passes in 7.1s** (Linux, rust:1 container)
- without fix (production hunks reverted, test kept): **fails after 30.08s** — hangs in the unbounded reap until `run_fake`'s outer guard fires

The 30s outer timeout on `dead_child_stdin_write...` was added in the same commit; its failure mode without the fix is a hung CI job, so there is no meaningful isolated run to show.

## Gates ran green

- `cargo fmt --all --check`
- `cargo clippy -p edda-conductor --all-targets -- -D warnings`
- `cargo test -p edda-conductor` — 238 passed, 0 failed, 1 ignored (pre-existing)

Note: CI's ubuntu/macos legs will also run the new unix-only integration test; the Windows leg compiles it out (platform-native fake limitations: PowerShell/node/WSH all retain a duplicate stdout handle, so stdout-EOF-while-alive is not reachable there).

### DIFF
--- a/crates/edda-conductor/src/agent/pi_rpc.rs
+++ b/crates/edda-conductor/src/agent/pi_rpc.rs
@@ -16,6 +16,12 @@ use tokio_util::sync::CancellationToken;
 /// Upper bound on waiting for the stderr drain once stdout has hit EOF.
 const STDERR_DRAIN_GRACE: Duration = Duration::from_secs(2);
 
+/// Upper bound on waiting for a child reap whose exit is expected but not
+/// guaranteed (stdout EOF or dead-stdin while the process may still live).
+/// On expiry the error carries the stderr tail collected so far plus an
+/// explicit note instead of hanging the turn; `kill_on_drop` cleans up later.
+const REAP_GRACE: Duration = Duration::from_secs(5);
+
 /// Resolve the pi executable from an explicit `EDDA_PI_BIN` value, falling
 /// back to the name npm installs on this platform.
 ///
@@ -249,7 +255,7 @@ impl PiRpcSession {
 
     async fn terminate(&mut self) {
         let _ = self.child.kill().await;
-        let _ = self.child.wait().await;
+        let _ = tokio::time::timeout(REAP_GRACE, self.child.wait()).await;
     }
 
     fn take_id(&mut self) -> String {
@@ -288,10 +294,12 @@ impl PiRpcSession {
             if !is_child_gone_error(&error) {
                 return Err(error);
             }
-            let _ = self.child.wait().await;
+            let reaped = tokio::time::timeout(REAP_GRACE, self.child.wait())
+                .await
+                .is_ok();
             let tail = collect_stderr_tail(&mut self.stderr_drain, &self.stderr).await;
             return Ok(PhaseResult::AgentCrash {
-                error: eof_error(&mut self.child, &tail),
+                error: eof_error(&mut self.child, &tail, reaped),
             });
         }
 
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: Round 1

Reviewed full SHA: 588bd4432e8b759da9b3082816f9ab50e6041145
IN SCOPE: the three pi_rpc child-reap paths and their tests (issue #538 doneWhen: no unbounded child.wait().await; grace expiry yields a prompt informative error, no hang/panic; dead_child_stdin_write_still_reports_stderr_reason bounded like its siblings; happy/crash paths and the guard 5x loop untouched); direct callers of the reap paths; current-base integration.
Blocking: P0=0, P1=0
Findings: none.
FOLLOW-UP ISSUE: none

Evidence:
- READ: 7/7 CI green on 588bd443 (Format; Clippy ubuntu/windows/macos; Test ubuntu/windows/macos — edda-conductor is inside the Windows CI test subset); issue #538 + comments.
- RAN: adversarial diff + full-file inspection via `gh pr diff` / `git show origin/fix/gh538-pi-rpc-reap-bounds:...`; no local build (build lanes occupied by concurrent implementation lanes).

*Independent review performed by the Codex reviewer session; posted by the controller as carrier because the reviewer sandbox has no outbound GitHub access. Verdict binds the SHA above; any push voids it.*

PR #554, opened 2026-09-01
fix(edda-conductor): keep both-stream tails and stop charging LNK1104 to the ladder (#540)

### LINKED ISSUE #540: fix(conductor): a truncated check failure hides LNK1104's cause, and an environmental build failure spends a product retry attempt

fagemx:
Closes #540.

## What

Two defects, both about what happens when a `cmd_succeeds` check fails:

1. **Tail-of-both-streams capture.** The failure message now keeps the last 2000 chars of BOTH stdout and stderr (masked, char-boundary safe — the old byte-slice head-truncation could panic on multibyte UTF-8). `cargo test` ends stderr at `error: test failed` while the failing test names land on stdout; `link.exe` puts `cannot open file '...'` at the very end of stderr. Head-of-stderr reproduced the useless message verbatim; tail-of-both names the cause.

2. **Environmental classification and retry accounting.** Output naming `LNK1104` classifies as `ErrorType::Environmental` (new variant, retryable) following the #529 shape (CheckOutput flag → engine classification → handle_on_fail policy). `handle_on_fail` retries environmental failures WITHOUT charging `max_attempts`: `attempts` still counts every dispatch (attempt numbers key the session id and must stay unique), product attempts = `attempts - env_retries`. Bounded: after 2 environmental retries in a phase run, it halts with an actionable message instead of looping.

## doneWhen

- [x] A failing `cmd_succeeds` check preserves the tail of the command's output — both streams, so a fatal linker line and cargo-test failure names are visible
- [x] The next LNK1104 occurrence names its file in the captured log (test asserts `cannot open file 'edda_conductor.exe'` survives capture)
- [x] An environmental build failure is distinguishable from an agent failure in retry accounting — recorded distinctly (`ErrorType::Environmental`) AND not charged to `max_attempts`

## Out of scope (per issue)

Preventing the linker collision itself (machine layer: build lanes, AV exclusions). No lane logic added. Fittingly, LNK1104 hit this very branch's first `cargo test` run; killing stale processes and re-running cleared it — the transient-retry behavior this PR implements.

## Tests

All regression tests verified to FAIL without the fix (pre-fix simulation of each site):
- `failure_keeps_tail_of_both_streams` — stderr tail + stdout names survive a >2000-char capture
- `lnk1104_output_marks_environmental` / `lnk1104_classifies_as_environmental` — distinct classification, file named
- `environmental_failure_bounded_and_distinct` — capped at 2 env retries, halts with message
- `environmental_failure_not_charged_to_ladder` — 1 env + 2 genuine failures on `max_attempts: 2` still gives the agent its full ladder (3 dispatches, "failed after 2 attempts")

Gates: `cargo fmt --all --check` ✓ · `cargo clippy -p edda-conductor --all-targets -- -D warnings` ✓ · `cargo test -p edda-conductor` 245 passed ✓

### DIFF
--- a/crates/edda-conductor/src/check/cmd_succeeds.rs
+++ b/crates/edda-conductor/src/check/cmd_succeeds.rs
@@ -1,7 +1,25 @@
-use crate::check::{mask_secrets, CheckOutput};
+use crate::check::{mask_secrets, output_tail, CheckOutput};
 use std::path::Path;
 use std::time::{Duration, Instant};
 use tokio::process::Command;
+
+/// Per-stream capture cap (GH-540): keep the TAIL, where fatal diagnostics
+/// live (LNK1104's payload, `error: test failed`).
+const OUTPUT_TAIL_CHARS: usize = 2000;
+
+/// Output substrings that identify a machine-layer build failure (GH-540).
+/// LNK1104's payload — the file the linker could not open — is the single
+/// token distinguishing a held .exe from a concurrent cargo or an antivirus
+/// handle, so the capture must reach the tail of the stream that names it.
+/// (Content classification, unlike the GH-529 timeout: a linker fault has no
+/// constructor-level signal, the output text is the only evidence.)
+const ENVIRONMENTAL_PATTERNS: &[&str] = &["LNK1104"];
+
+fn is_environmental(stderr: &str, stdout: &str) -> bool {
+    ENVIRONMENTAL_PATTERNS
+        .iter()
+        .any(|p| stderr.contains(p) || stdout.contains(p))
+}
 
 /// Shell program and args for the current platform.
 #[cfg(windows)]
@@ -59,21 +77,31 @@ pub async fn check_cmd_succeeds(cmd: &str, timeout_sec: u64, cwd: &Path) -> Chec
     match tokio::time::timeout(Duration::from_secs(timeout_sec), result).await {
         Ok(Ok(output)) if output.status.success() => CheckOutput::passed(start.elapsed()),
         Ok(Ok(output)) => {
-            let stderr = String::from_utf8_lossy(&output.stderr);
-            let masked = mask_secrets(&stderr);
-            let truncated = if masked.len() > 2000 {
-                format!("{}...", &masked[..2000])
+            // GH-540: keep the tail of BOTH streams. Diagnostic conventions
+            // split across them — `cargo test` ends stderr at "error: test
+            // failed" while the failing test names land on stdout, and
+            // link.exe puts "cannot open file '...'" at the very end of
+            // stderr — so a tail of one stream reproduces the same useless
+            // message the issue was filed with.
+            let stderr = mask_secrets(&String::from_utf8_lossy(&output.stderr));
+            let stdout = mask_secrets(&String::from_utf8_lossy(&output.stdout));
+            let stderr_tail = output_tail(&stderr, OUTPUT_TAIL_CHARS);
+            let stdout_tail = output_tail(&stdout, OUTPUT_TAIL_CHARS);
+            let mut message = format!("exit {}", output.status.code().unwrap_or(-1));
+            if !stderr_tail.trim().is_empty() {
+                message.push_str(": ");
+                message.push_str(stderr_tail.trim_end());
+            }
+            if !stdout_tail.trim().is_empty() {
+                message.push_str("\n--- stdout (tail) ---\n");
+                message.push_str(stdout_tail.trim_end());
+            }
+            let detail = message.trim().to_string();
+            if is_environmental(stderr_tail, stdout_tail) {
+                CheckOutput::failed_environmental(detail, start.elapsed())
             } else {
-                masked.to_string()
-            };
-            CheckOutput::failed(
-                format!(
-                    "exit {}: {}",
-                    output.status.code().unwrap_or(-1),
-                    truncated.trim()
-                ),
-                start.elapsed(),
-            )
+                CheckOutput::failed(detail, start.elapsed())
+            }
         }
         Ok(Err(e)) => CheckOutput::failed(format!("spawn error: {e}"), start.elapsed()),
         Err(_) => CheckOutput::timed_out(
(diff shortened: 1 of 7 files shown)


fagemx:
## Code Review: Round 1
Reviewed full SHA: 2e03b8abf736a7e676c1c82b8b33e23e8cbcfe90
IN SCOPE: both-stream cmd_succeeds tails; LNK1104 capture/classification; retry accounting and direct state/event/CLI consumers; regression tests; no lane logic; current-base integration
Blocking: P0=0, P1=3
- P1 — `crates/edda-conductor/src/runner/sequential.rs::handle_on_fail` with direct consumer `crates/edda-cli/src/cmd_conduct.rs::retry`: the environmental occurrence that exhausts `MAX_ENV_RETRIES` is not added to `env_retries`. After three LNK1104 failures the persisted counters are `attempts=3, env_retries=2`; manual retry preserves them, so the first genuine failure at attempt 4 computes `product=2` for `max_attempts=2` and gets no auto-retry. That cap-ending environmental failure therefore still consumes one product attempt. The regression test stops at the cap and never exercises clear-fault/manual-retry followed by genuine failures.
- P1 — `crates/edda-conductor/src/runner/event_log.rs::Event::PhaseFailed` as called by `crates/edda-conductor/src/runner/sequential.rs::fail_checking_phase`: the append-only event records only `attempt` and an untyped error string; it records neither `ErrorType::Environmental`, `env_retries`, nor whether the attempt was charged. An event consumer sees the same `phase_failed` → `phase_start` shape for free environmental retries and charged agent retries, so retry accounting is not distinctly recorded in the events as required.
- P1 — `crates/edda-conductor/src/check/cmd_succeeds.rs::is_environmental`: matching any tail that merely contains the token `LNK1104` misclassifies agent failures that discuss this error (for example, a cargo test of the LNK1104 handling panics with an uppercase `LNK1104` assertion message) as machine failures. That product defect then receives free retries and can halt under the environmental-cap message. The negative regression contains no `LNK1104` token, so it does not guard this collision; require the actual linker-fatal signature rather than the bare code.
FOLLOW-UP ISSUE: none
Evidence:
- READ: 7/7 CI green on 2e03b8abf736a7e676c1c82b8b33e23e8cbcfe90 (conductor in Windows test subset); issue #540 + comments incl. both-streams comment
- RAN: adversarial diff + full-file inspection, no local build (lanes occupied)
Verdict: Changes Requested

*Independent review performed by the Codex reviewer session; posted by the controller as carrier because the reviewer sandbox has no outbound GitHub access. Verdict binds the SHA above; any push voids it.*

fagemx:
## Review Response: Round 1

Responds to the Round 1 review of 2e03b8abf736a7e676c1c82b8b33e23e8cbcfe90 (P0=0, P1=3). All three blocking findings fixed; each new regression test verified to FAIL with its fix temporarily reverted before re-enabling.

### P1-1 — cap-ending environmental failure not charged to env_retries — FIXED
- **Disposition:** accepted. `handle_on_fail` now charges `env_retries` on EVERY environmental occurrence, including the one that exhausts `MAX_ENV_RETRIES`; retry continues while the counter is `<= MAX_ENV_RETRIES`, so it reads `MAX+1` at the halt and `attempts - env_retries` stays the exact product count across a manual retry. The cap-exhausted notification now reports occurrences vs. the retry cap accurately.
- **Changed:** `crates/edda-conductor/src/runner/sequential.rs::handle_on_fail` (consumer `crates/edda-cli/src/cmd_conduct.rs::retry` needs no change — it preserves counters, which are now correct); doc comment on `env_retries` in `crates/edda-conductor/src/state/machine.rs`.
- **Test:** `environmental_failure_bounded_and_distinct` replaced by `environmental_cap_end_charged_so_manual_retry_keeps_full_ladder` — run 1 halts at the cap (attempts=3, env_retries=3), then clear-fault + a retry mirroring `cmd_conduct::retry` (Failed→Pending, counters preserved), then two genuine failures must still consume the full ladder (attempts=5, call_count=5, "failed after 2 attempts"). **Verified to FAIL without the fix** with exactly the reviewed counters (env_retries=2 → attempts stalls at 4).

### P1-2 — PhaseFailed events carry no retry accounting — FIXED
- **Disposition:** accepted. `Event::PhaseFailed` gains three additive serde fields: `error_type` (snake_case tag via new `ErrorType::tag()`, `skip_serializing_if` None), `env_retries` (omitted at zero), and `attempt_charged` (false for environmental, true otherwise). All seven construction sites updated (fail_checking_phase, both gate-wait sites, Timeout/AgentCrash/BudgetExceeded in process_phase_result).
- **Backward compatibility:** fields are additive with skip-when-default serialization; a `phase_failed` line written before the fields (or at zero counters) emits the identical object as before. **Consumers checked:** nothing in the workspace deserializes the conductor `Event` enum — the only reader is generic JSON (`serde_json::Value`) in the runner's own tests (`read_events`); TUI/reconcile/CLI grep hits for "events.jsonl" all reference the separate edda-ledger event store, not the conductor log. `serde(default)`-style defaults are irrelevant on the write path since `Event` is Serialize-only.
- **Tests:** `event_phase_failed_serialization` + `event_phase_failed_additive_fields_stay_absent_when_defaulted` in event_log.rs; end-to-end `phase_failed_events_record_retry_accounting` in sequential.rs (environmental event records `error_type:"environmental"`, `attempt_charged:false`; genuine ones record `check_failed`, `attempt_charged:true`). **Verified to FAIL without the fix** (`error_type` absent).

### P1-3 — bare `LNK1104` token misclassifies — FIXED
- **Disposition:** accepted. `ENVIRONMENTAL_PATTERNS` now requires the actual linker-fatal signature `"fatal error LNK1104"` (link.exe emits `LINK : fatal error LNK1104: cannot open file '...'`) instead of the bare code. Test commands updated to the real signature; the existing negative test is unchanged.
- **Changed:** `crates/edda-conductor/src/check/cmd_succeeds.rs` (`is_environmental` pattern table + comment).
- **Test:** `bare_lnk1104_token_is_not_environmental` — output containing the bare token inside a non-linker context (a panic message discussing `LNK1104`) must reach the tail but stay a plain product failure. **Verified to FAIL without the fix** (classified environmental).

### Gates ran (full SHA below, lane worker-2)
- `cargo fmt --all --check` — pass
- `cargo clippy -p edda-conductor --all-targets -- -D warnings` — pass
- `cargo test -p edda-conductor` — 249 passed, 1 ignored (lib) + doc-tests. Rerun note: one run aborted with a process-level `STATUS_ACCESS_VIOLATION` and no failing test; the identical binary passed clean on immediate rerun (environmental flake in the lane, not a test failure), so the gate was rerun and recorded green.

### New full SHA
`5a09b693e1acda8835eaa2646b8050697694da47`

fagemx:
## Code Review: Round 2
Reviewed full SHA: 5a09b693e1acda8835eaa2646b8050697694da47
Current base full SHA: ea5cb5bf067f1c3a2f0ce24bc35cdee2c4860932
IN SCOPE: the three Round-1 P1s and fix delta 2e03b8a..5a09b69; all seven PR-changed files; direct retry and conductor-event consumers; issue #540 doneWhen; introduced security/data-loss regressions; current-main integration

BLOCKING: P0=0, P1=0

Per-P1 verdict:
- FIXED — cap-ending environmental accounting: `crates/edda-conductor/src/runner/sequential.rs::handle_on_fail` increments `env_retries` for every environmental occurrence before applying the retry cap, preserving `attempts - env_retries` across `crates/edda-cli/src/cmd_conduct.rs::retry`; `environmental_cap_end_charged_so_manual_retry_keeps_full_ladder` covers cap halt → manual retry → two genuine product attempts.
- FIXED — append-only failure accounting: `crates/edda-conductor/src/runner/event_log.rs::Event::PhaseFailed` now carries `error_type`, `env_retries`, and `attempt_charged`; every production construction site in `runner/sequential.rs` supplies them, and `phase_failed_events_record_retry_accounting` distinguishes environmental/free from check_failed/charged events. The existing `demo/runtime-edda.js::translateEvent` consumer remains compatible with the additive fields.
- FIXED — LNK1104 false positive: `crates/edda-conductor/src/check/cmd_succeeds.rs::is_environmental` now requires `fatal error LNK1104`; `bare_lnk1104_token_is_not_environmental` covers agent output that merely mentions the bare code while the real linker signature remains covered.

New fix-caused blocker: none.

FOLLOW-UP ISSUE: none.

Evidence:
- RAN: `git diff --check 2e03b8abf736a7e676c1c82b8b33e23e8cbcfe90..5a09b693e1acda8835eaa2646b8050697694da47` — pass; complete fix-delta plus full-file inspection of all seven PR-changed files and direct consumers — pass; `git merge-tree 749570c5fb44bd89c02046439266cdd4e19e9413 ea5cb5bf067f1c3a2f0ce24bc35cdee2c4860932 5a09b693e1acda8835eaa2646b8050697694da47` — clean, with GitHub REST also reporting `mergeable: true`, `mergeable_state: clean`.
- READ: issue #540 and its both-stream clarification; Round-1 review and full implementer response; exact-head CI 7/7 green on 5a09b693e1acda8835eaa2646b8050697694da47 (Format; Clippy and Test on Ubuntu, Windows, macOS); implementer focused receipt on this SHA: fmt pass, conductor Clippy pass, conductor tests 249 passed/1 ignored plus doc-tests, with the one process-level STATUS_ACCESS_VIOLATION rerun reason recorded.
- Local Cargo: not rerun — exact-head CI already runs edda-conductor tests on Windows and workspace tests on Linux/macOS, so a local duplicate had no uncovered behavior to add.
- Lane: n/a (no local build).

Cost: elapsed≈16m; tokens=unknown; local build cost=0.

Verdict: LGTM (P0=0, P1=0)

Issue #648, opened 2026-09-02
chore(release): publish the workspace to crates.io from release.yml on tag — crates.io lagged two versions behind


fagemx:
Stage 1 門票項目（裁定 `roadmap.stage1-order`，Tim 2026-09-02）。Lane A。

## What happened

在 main `275b0b39dfe0a13064cd7898c39e2d59068e0f07`：

- `Cargo.toml` workspace version = 0.4.0，tag `v0.4.0` 已存在。
- crates.io 的 `edda`（CLI binary crate）`max_version` = **0.2.1**（`GET https://crates.io/api/v1/crates/edda`）。
- `.github/workflows/release.yml` 的三個 job（`create-release` → `build-release` → `publish-release`）沒有任何 `cargo publish` 步驟。
- README 第 105 行仍列 `cargo install edda` 為第三種安裝方式。

操作者 2026-09-02 正在手動發 0.4.0（寫這張時 `edda-core`、`edda-conductor` 已是 0.4.0，`edda` 尚未）。本 issue 是讓它以後不再落後：手動發布是這次的修補，不是機制。

注意：crates.io 上另有名為 `edda-cli` 的 0.0.1 佔位 crate，與 binary crate `edda` 無關，不要發到那個名字。

## Why it matters

`cargo install edda` 拿到的是兩版前的二進位——沒有 `dispatch`、`verdict`、`task`。任何照 README 安裝的人看到的示範都不存在。這是採用漏斗的第一個洞，不是整潔問題。

## Suspected surface

- `.github/workflows/release.yml` — 新 job `publish-crates`：只在 tag push 執行，成功後才允許 `create-release`；以 2026-09-03 操作者的 crates.io-first 裁定為準
- 本機 toolchain cargo 1.93.1：`cargo publish --workspace` 可用（1.90 穩定），自動處理相依順序；若 CI 的 toolchain 較舊則退回拓樸序逐 crate
- 不改 README（0.4.0 手動發布後那行即正確）
- 不改任何 crate 的 `Cargo.toml`（若 `--dry-run` 揭露某 crate 缺 `description`/`license` 等必填欄位，列成 FOLLOW-UP，不在本 issue 修）

## Wiring audit

| 新面 | writer & shape | reader | failure signal | layer reach |
|---|---|---|---|---|
[...]

PR #655, opened 2026-09-02
fix(release): enforce distribution channel parity

fagemx:
## Summary

- block GitHub release creation until every publishable workspace crate exists at the tag version on crates.io and is not yanked

## Scope note

- this branch was narrowed in `511ac88` to the release-channel parity gate only: the CLI-reference documentation and its CI coverage check were dropped here and are owned by PR #670, whose `scripts/check-cli-docs.sh` is the single coverage mechanism (see the scope ruling in the comments)

## Incident repair

- published all 23 workspace crates at `0.4.0` to crates.io from the exact `v0.4.0` tag
- verified a clean, unversioned `cargo install edda` selects `0.4.0`
- verified the installed binary exposes `edda dispatch --help` and `edda verdict --help`

## Test plan

- [x] `sh scripts/lint-markdown-content.sh` under Git Bash
- [x] parse `.github/workflows/release.yml` with PyYAML
- [x] execute the `Verify crates.io release parity` workflow script against crates.io — 23/23 at `0.4.0`
- [x] `git diff --check`

PR #673, opened 2026-09-02
chore(release): publish workspace to crates.io on tag push (GH-648)

fagemx:
Issue: #648 — partial delivery; the issue stays open. #648's doneWhen includes an operator-side prerequisite explicitly outside this lane's scope (adding the `CARGO_REGISTRY_TOKEN` repo secret), which is not done, so the publish path cannot actually run yet and the issue must not be auto-closed by this PR.

## Prohibitions honored

- **No crate was published.** Every publish-adjacent command run locally carried `--dry-run`; the real publish path exists only inside the tag-gated `publish-crates` job.
- **No tag was created, moved, or pushed.** HEAD is `89bd426d4fa09aa2b8e6a91394fffa7ec9da23a2` on `chore/gh648-crates-io-publish`; no tags touched.
- **No registry token appears anywhere in this diff.** `CARGO_REGISTRY_TOKEN` is referenced by name only, wired from `secrets.` into step env.

## doneWhen checklist

| doneWhen line | Evidence |
|---|---|
| New job runs only on tag push, after `publish-release` succeeds, uses `CARGO_REGISTRY_TOKEN` | `publish-crates` job `if`: `always() && needs.create-release.result == 'success' && needs.publish-release.result == 'success' && needs.crates-token-gate.result == 'success' && needs.crates-token-gate.outputs.token-available == 'true' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')` (release.yml, `publish-crates:`). Token used only as `env: CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}` on the publish step. |
| Token absent → job visibly SKIPPED, prints why; not red, not silently green | Two-job gate: `crates-token-gate` checks the secret and emits `::notice title=crates.io publish skipped::CARGO_REGISTRY_TOKEN is not configured…publish-crates will be SKIPPED…` while `publish-crates` carries `needs.crates-token-gate.outputs.token-available == 'true'` in its job `if`, so it renders **Skipped** in the run UI. **Observed on this PR's run — Evidence item 3.** |
| Publish covers every workspace crate `edda` depends on, `edda` last | `edda` depends directly on all 22 other members (derived from `cargo metadata`); `CRATES_PUBLISH_ORDER` lists all 23 in topological order with `edda` last. A drift-guard step (`Validate publish order covers the workspace`) diffs the list against `cargo metadata` members and fails on mismatch. Topological validity machine-checked at review time (no dependency appears after its dependent). |
| Re-run safe: "already uploaded" is a no-op, rest continue, final state judged by "all versions present" | Publish loop publishes per-crate; on failure it greps `already uploaded\|already exists` → `::notice::…treating as no-op and continuing`, otherwise records the crate in the step's `failures`/`failed-crates` outputs **and continues**. The publish step never exits nonzero for a publish error, and `Verify every crate version is on crates.io` carries `if: always() && steps.publish.outcome != 'skipped'`, so **the verifier's exit code alone decides the job**: all versions present ⇒ green (with a `::notice::` naming any crate whose publish reported an error); any version missing ⇒ red. Fixes Round 1 P0-2, where a nonzero publish step made Actions skip the verifier. |
| Publish failure is RED, message states GitHub release shipped but crates.io did not sync (assets not rolled back) | Both the loop failure and the verify failure emit: `::error::GitHub release <tag> shipped but crates.io did not (fully) sync…Release assets are NOT rolled back. Re-run this job…` |
| PR CI runs the dry run green, no token | `publish-crates-dry-run` (PR-only) runs `cargo publish --workspace --dry-run --locked` with no token env at all. Quoted tail in Evidence item 2. |
| Out of scope respected | No Cargo.toml, no README.md, no manual 0.4.0 publish. Files touched: `.github/workflows/release.yml` only. |

### 1. Before — no publish path at `f582ef3`

```
$ git grep -n "cargo publish" f582ef3 -- .github/workflows/
(no output, exit 1)
$ git grep -n "crates.io" f582ef3 -- .github/workflows/
(no output, exit 1)
```

### 2. Dry run green on PR CI, no token present

PR run: https://github.com/fagemx/edda/actions/runs/33639751081 — every job green. `publish-crates-dry-run` job conclusion: **success**, with no token anywhere in the job (the step defines no `CARGO_REGISTRY_TOKEN`; the repo secret does not exist). Log tail quoted verbatim:

```
Show toolchain    cargo 1.98.0 (797e8a9bc 2026-08-05)
Dry run publish for the workspace:
  warning: aborting upload due to dry run    <- repeated once per crate, 23x, ending with edda v0.4.0
  Dry run passed for all 23 workspace crates; nothing was uploaded.
```

Runner toolchain: **cargo 1.98.0** — also the evidence for the ordering decision in item 4.

### 3. Skip-not-fail — observed on this PR, not merely argued

Observed on this PR run (https://github.com/fagemx/edda/actions/runs/33639751081, `pull_request` event, no `CARGO_REGISTRY_TOKEN` configured):

```
Job statuses (run 33639751081):
  Create Release                    completed   skipped
  Build (...)                       completed   skipped
  Publish Release                   completed   skipped
  Check crates.io publish token     completed   success    <- gate runs, prints reason
  Publish crates to crates.io       completed   SKIPPED    <- the publish job, visibly skipped
  Dry run crates.io publish         completed   success

Gate log, verbatim:
  ##[notice]CARGO_REGISTRY_TOKEN is not configured on this repository, so publish-crates
  will be SKIPPED. The GitHub release can ship without syncing crates.io. Add the secret
  to enable publishing.

Condition that produced the skip (publish-crates job if):
  needs.crates-token-gate.outputs.token-available == 'true' && github.event_name == 'push'
  && startsWith(github.ref, 'refs/tags/v')
  <- token-available was 'false' on this run, and the event was pull_request
```

Line-by-line reading: the gate step finds `${CARGO_REGISTRY_TOKEN:-}` empty (secret absent) → writes `available=false` and prints the `::notice` → the `publish-crates` job `if` evaluates `outputs.token-available == 'true'` → false → the job renders **Skipped**; red nowhere, silently-green nowhere. The gate job runs on `pull_request` deliberately so this path is observable on PRs instead of resting on a static argument.

### 4. Ordering — explicit topological list, `edda` last

- **Branch taken: explicit topological order, per-crate publish.**
- **CI toolchain checked**: `release.yml` (all jobs) installs `stable` via `rustup toolchain install stable --profile minimal` — rustup's stable channel resolves to the newest stable at run time. Verified on the actual runner: `cargo 1.98.0` (Show toolchain step, run 33639751081), above both the 1.90 threshold and local 1.93.1. `--workspace` **is** available on CI.
- **Why explicit order anyway**: re-run safety. Real `cargo publish --workspace` aborts the entire command on the first "already uploaded" crate, leaving every later crate unpublished — and the abort repeats identically on every re-run, so `--workspace` can never converge. Per-crate with already-uploaded tolerance + the all-versions-present verifier does converge. This also makes `edda` last **visibly**, as the final element of `CRATES_PUBLISH_ORDER`.
- Topological order derived from `cargo metadata --locked --no-deps` and machine-verified (each dependency precedes its dependent); drift-guarded in CI against future workspace membership changes.

### Why the PR dry run uses `--workspace --dry-run` (not the per-crate loop)

Local finding, run deliberately in lane `verifier-2`: per-crate `cargo publish --dry-run -p X` verifies X against the **crates.io registry versions** of its workspace siblings. The operator's in-flight manual 0.4.0 publish left stale 0.4.0 tarballs on crates.io (`edda-notify` 0.4.0 predates current source: no `PhaseTerminal`, no `Clone` on `NotifyEvent`), so `edda-conductor`'s per-crate dry run **failed to compile against the registry copy** (E0599) — an artifact of the manual publish, not of this diff. With `--workspace --dry-run`, cargo packs all members into a temporary local registry (`Unpacking edda-conductor v0.4.0 (registry …tmp-registry)`) and every crate verifies against its **current** siblings: green for all 23. Real publishes are unaffected: at a fresh tag version, topo order publishes each dependency just before its dependent, so verification resolves the fresh version. This is also exactly the command the issue's doneWhen names.

## Gate receipt

- Full SHA: `89bd426d4fa09aa2b8e6a91394fffa7ec9da23a2`
- L0 (while iterating): `cargo fmt --all --check` → clean; local dry run `cargo publish --workspace --dry-run --locked` → all 23 crates packaged and verified, `edda` last (lane `verifier-2`, cold build)
- L1 (frozen SHA, clean tree): `cargo fmt --all --check` → clean. **L1: workspace clippy/test n/a — YAML-only change, no code or Cargo.lock delta** (diff vs `f582ef3` touches only `.github/workflows/release.yml`; no Cargo.lock, no toolchain delta)
- Toolchain: cargo 1.93.1 (083ac5135 2025-12-15) local; 1.98.0 on the CI runner; lane: `verifier-2` (default `CARGO_TARGET_DIR`, reused)
- Exact-head CI: run 33639751081 — Release workflow success (6/6 jobs success-or-skipped as designed), CI workflow also green

## FOLLOW-UP ISSUE (adjacent, evidenced — not this PR's surface)

- **Stale manually-published 0.4.0 tarballs vs current source**: crates.io's `edda-notify@0.4.0` predates current source (`NotifyEvent::PhaseTerminal` and `Clone` missing) while current `edda-conductor` uses those APIs, so nothing on main compiles against the *published* `edda-notify@0.4.0`. Harmless for future tags (fresh topo-order publish self-heals), but any consumer pinning `edda-notify = "0.4.0"` from crates.io gets an API different from what the workspace's current `edda-conductor` expects. Evidence: per-crate dry-run E0599 failure (quoted in the log excerpt above); `crates.io/api/v1/crates/edda-notify` reports 0.4.0 present.
- **Cosmetic, non-blocking**: Swatinem/rust-cache's post step emitted `##[error] ENOENT … target/package/edda-0.4.0/tests/target` cleanup noise on the dry-run job (cache action scanning the packaging directory); the job still concluded success. If it ever grows teeth, keying the cache action to skip `target/package` is the fix.

## Explicit statement

No crate was published (every local invocation carried `--dry-run`; nothing was uploaded to any registry). No git tag was created, moved, or pushed. No registry token was added, echoed, or committed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -340,12 +340,13 @@ jobs:
           echo "Publish order covers all $(echo "$CRATES_PUBLISH_ORDER" | wc -w) workspace crates"
 
       - name: Publish workspace crates
+        id: publish
         shell: bash
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
         run: |
-          tag="${{ needs.create-release.outputs.tag }}"
           failures=0
+          failed_crates=""
           for crate in $CRATES_PUBLISH_ORDER; do
             echo "::group::cargo publish -p $crate"
             if out="$(cargo publish --locked -p "$crate" 2>&1)"; then
@@ -362,18 +363,33 @@ jobs:
               if grep -Eq "already uploaded|already exists" <<<"$out"; then
                 echo "::notice::$crate is already on crates.io; treating as no-op and continuing."
               else
-                echo "::error::cargo publish -p $crate failed (exit $status)."
+                # Recorded, not fatal: the crates.io presence check below is
+                # the arbiter of final state. A reworded duplicate error, a
+                # timeout, or a rate-limit reply must not fail a run whose
+                # versions all landed, or a re-run could never converge.
+                echo "::warning::cargo publish -p $crate failed (exit $status); deferring to the crates.io presence check."
                 failures=$((failures + 1))
+                failed_crates="${failed_crates}${failed_crates:+ }$crate"
               fi
             fi
           done
+          {
+            echo "failures=$failures"
+            echo "failed-crates=$failed_crates"
+          } >> "$GITHUB_OUTPUT"
           if [ "$failures" -gt 0 ]; then
-            echo "::error::GitHub release ${tag} shipped but crates.io did not fully sync (${failures} crate(s) failed to publish). Release assets are NOT rolled back. Re-run this job after resolving the failure; crates already uploaded are treated as no-ops."
-            exit 1
+            echo "$failures crate(s) reported a publish error ($failed_crates); the verification step decides this job's outcome."
           fi
 
+      # Judges the final state by presence, as GH-648 requires, and therefore
+      # decides the job: it must run even when the publish step above failed
+      # outright, and its exit code — not cargo's — is the verdict.
       - name: Verify every crate version is on crates.io
+        if: always() && steps.publish.outcome != 'skipped'
         shell: bash
+        env:
+          PUBLISH_FAILURES: ${{ steps.publish.outputs.failures }}
+          PUBLISH_FAILED_CRATES: ${{ steps.publish.outputs.failed-crates }}
         run: |
           tag="${{ needs.create-release.outputs.tag }}"
           version="${tag#v}"
@@ -393,6 +409,9 @@ jobs:
           if [ "$missing" -gt 0 ]; then
             echo "::error::GitHub release ${tag} shipped but crates.io did not sync: ${missing} of ${total} crates missing. Release assets are NOT rolled back. Re-run this job to retry; crates already uploaded are treated as no-ops."
             exit 1
+          fi
+          if [ "${PUBLISH_FAILURES:-0}" -gt 0 ]; then
+            echo "::notice::cargo publish reported an error for ${PUBLISH_FAILURES} crate(s) (${PUBLISH_FAILED_CRATES}), but all ${total} versions are present on crates.io at ${version}. Final state is judged by presence, so this run is green."
           fi
           echo "All ${total} workspace crates are present on crates.io at ${version}"
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,12 @@ on:
   push:
     tags:
       - "v*"
+  # PR runs exercise only the crates.io packaging dry run and the token
+  # skip-gate (GH-648); every tag job below is guarded to tag pushes, so tag
+  # behavior is unchanged.
+  pull_request:
+    paths:
+      - ".github/workflows/release.yml"
 
 permissions:
   contents: write
@@ -11,6 +17,16 @@ permissions:
 env:
   CARGO_TERM_COLOR: always
   GH_REPO: ${{ github.repository }}
+  # Explicit topological publish order (dependencies before dependents),
+  # derived from `cargo metadata` at GH-648 review time and validated against
+  # the workspace member set by the "Validate publish order" step in both
+  # publish-crates-dry-run and publish-crates. `edda` (the CLI binary crate in
+  # crates/edda-cli) is deliberately LAST. Per-crate publishing (instead of
+  # `cargo publish --workspace`, which CI's stable toolchain would also
+  # support) is required for re-run safety: --workspace aborts on the first
+  # "already uploaded" crate and would leave the remaining crates unpublished
+  # on every subsequent re-run.
+  CRATES_PUBLISH_ORDER: "edda-core edda-ledger edda-derive edda-store edda-aggregate edda-index edda-notify edda-pack edda-postmortem edda-search-fts edda-transcript edda-bridge-claude edda-bridge-codex edda-bridge-cursor edda-bridge-hermes edda-bridge-openclaw edda-ask edda-mcp edda-conductor edda-ingestion edda-serve edda-chronicle edda"
 
 concurrency:
   group: release-${{ github.ref }}
@@ -19,6 +35,7 @@ concurrency:
 jobs:
   create-release:
     name: Create Release
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     outputs:
       tag: ${{ steps.tag.outputs.tag }}
@@ -86,6 +103,7 @@ jobs:
   build-release:
     name: Build (${{ matrix.name }})
     needs: create-release
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -173,6 +191,7 @@ jobs:
   publish-release:
     name: Publish Release
     needs: [create-release, build-release]
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     steps:
       - name: Verify release assets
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: Round 1 — PR #673 @ c19d15404df3387ed094fabbb4d82e9250c1d185 (gpt-5.6-sol, read-only)

### IN SCOPE
- FAIL — Contract (1): issue #648 rerun-safety doneWhen is unmet; `.github/workflows/release.yml:370-375` exits before final-state verification after any unrecognized publish failure.
- PASS — Contract (2): `git diff --name-only a1dd3d8...HEAD` lists only `.github/workflows/release.yml`.
- PASS — Contract (3): `sh scripts/lint-markdown-content.sh` exited 0.
- FAIL — Contract (4): PR body line 1 contains `Closes #648`; commit message contains no closing keyword.
- PASS — Contract (5): added backticks are `cargo metadata`, `edda` (crate name), and `cargo publish --workspace`; zero added `edda <word>`, `gh <word>`, or `git <word>` invocations require probing.
- FAIL — Contract (6): baseline greps and CI transcript reproduce, but the PR-body claim that the crates.io API check is the “Final arbiter” is false when the preceding publish step exits nonzero.
- PASS — Contract (7): no shell script file was added; additionally, all 10 added Bash run blocks passed syntax checking after GitHub-expression substitution.
- PASS — doneWhen: tag-only `publish-crates` requires successful `publish-release` and token-gate output; exact-head PR CI observed the absent-token notice and visibly skipped publish.
- PASS — doneWhen: Python/TOML audit found 23/23 unique members, all 22 transitive `edda` dependencies covered, zero topology errors, and `edda` last.
- FAIL — doneWhen: already-uploaded crates continue, but other failures cause `exit 1` before the final all-versions-present check.
- PASS — doneWhen: exact-head CI ran `cargo publish --workspace --dry-run --locked`; 23 dry-run warnings, final message “Dry run passed for all 23 workspace crates,” job success.
- PASS — doneWhen operator prerequisite: expressly outside lane scope; no token was added, and the absent-secret path was exercised.

### Findings
- P0 — PR body:1 — RAN closing-keyword scan found `Closes #648` (`body_grep_exit=0`), violating contract item (4).
- P0 — `.github/workflows/release.yml:370-375` — RAN structural inspection found `Publish workspace crates if=None has_exit_1=True` and `Verify every crate version is on crates.io if=None`. A non-“already uploaded/exists” failure exits the prior step; GitHub’s default step success condition then skips verification. Thus all versions could exist after a nonzero upload response, yet final state is not judged by presence as required.
- P1: none.
- P2: none.

### FOLLOW-UP ISSUE
none

### RAN vs READ
- RAN: `sh scripts/lint-markdown-content.sh` → exit 0.
- RAN: PR three-dot path audit → `.github/workflows/release.yml` only; current-base `merge-tree` exit 0, zero conflict markers.
- RAN: baseline greps at `f582ef3` → no output; exits 1 for both `cargo publish` and `crates.io`, matching PR body.
- RAN: topology audit → `members=23 order=23 unique=23 set_match=True`; `edda_last=True`; `topology_errors=0`; all 22 transitive dependencies covered; exit 0.
- RAN: `gh run view 33639751081`/logs → exact head, pull-request event, release workflow success; token gate success with absent-token notice; publish skipped; dry run success; 23 dry-run warnings; cargo 1.98.0.
- RAN: PyYAML parse exit 0; added Bash syntax check 10 blocks, 0 failures; `git diff --check` exit 0; clean detached worktree at frozen SHA.
- READ: issue #648 doneWhen, PR body/commit, exact-head CI checks, `.claude/CLAUDE.md`, and ledger decision `demo.engine=sqlite` (no contradiction).

### Verdict
Changes Requested, P0=2, P1=0

---

*Relayed by the controller session (docs workstation). Reviewer: **Claude Opus** via Claude Code (`fleet.review-engine-model=opus-via-claude-code`), read-only (`--disallowedTools Edit,Write,NotebookEdit`), build lane `n/a`. It cannot post to GitHub; the text above is its verbatim output between its `<<<VERDICT` markers. Reviewed SHA `c19d15404df3387ed094fabbb4d82e9250c1d185` — **any push voids this verdict**. Contract item 5 was overridden per GH-691 so the reviewer used `git <verb> -h` rather than opening browser manuals.*

*Relay was delayed: the queue runner's wait loop exited before the reviewer had started, so it read an empty log and reported no verdict. The review itself completed normally — this is the verdict it produced, unedited. Runner bug, not a review failure.*

fagemx:
## Closing: superseded by #655 — this PR's publish job is unreachable by construction

Operator ruling, 2026-09-03. Closing rather than merging. The 4090 controller verified each point independently before acting; evidence below, including one detail that makes merging worse than it looks.

### The two PRs impose opposite orderings, and #655 landed first

**#655 (merged 2026-09-02 16:02)** added a `Verify crates.io release parity` step to `create-release` — the **first** job in the workflow (`.github/workflows/release.yml:54` on `main`). It queries the crates.io API for every workspace crate at the tag's version and fails the job when any is missing:

```
:91  echo "::error::Publish every workspace crate to crates.io before pushing the release tag"
```

So on `main` today the order is: **crates.io must already be complete → then the GitHub Release is built.**

**This PR** adds `publish-crates` as the last job, gated on `needs: [create-release, publish-release]` and a tag push. Its order is: **GitHub Release completes → then publish to crates.io.**

The two cannot both be the policy. #655 is on `main`; this PR is not.

### The consequence: `publish-crates` can never do useful work

Both reachable states are dead ends:

| State at tag push | #655's parity gate | This PR's `publish-crates` |
|---|---|---|
| a crate version missing from crates.io | **fails `create-release`** | never runs — the workflow died in job 1 |
| all crate versions present | passes | runs, and every crate is a no-op |

There is no input for which this job publishes anything. That is not a bug in the implementation — the implementation is careful, and Round 1's topology audit, token gate, skip-not-fail behaviour and dry run all passed. It is that the problem it solves was solved differently, first, by a design that runs earlier in the same workflow.

### The detail that matters most: merging would NOT conflict

```
$ git merge-tree $(git merge-base origin/main 89bd426d) origin/main 89bd426d | grep -cE '^<<<<<<<|^>>>>>>>'
0
```

This branch is based on `f582ef3`, before #655 merged, and **does not contain #655's changes at all** (`grep -c "Verify crates.io release parity"` on this branch returns 0). But the two changes touch different regions of the file, so git merges them silently.

Merging would therefore produce a `release.yml` carrying **both** policies with no warning: #655's gate demanding crates.io be complete before the release, and 189 lines of publish job that the gate guarantees can never run. No conflict marker, no failing check, nothing to prompt a second look — just permanently dead workflow code that reads like a live publish path to whoever opens the file next.

That is why this is a close, not a rebase-and-merge.

### Two further gaps, recorded for the successor design rather than as blockers

Neither would be worth fixing here, since the whole job is being removed, but both should shape whatever replaces it:

- **Presence is not sufficiency.** The verification added in `89bd426` decides the job's outcome by whether a version appears in the crates.io API response. It does not check whether that version is **yanked**, nor that it was built from the tag's source. A yanked version is present and unusable; presence alone would report success.
- **The publish order is a hard-coded crate list.** It is validated against the member set, not derived from the dependency graph, so it verifies that the names match — not that the order is still a valid topological order after someone changes a dependency. Round 1 confirmed the current order is correct; nothing keeps it correct.

### Verdict status

For the record: the last review here (Round 1, P0=2) was pinned to `c19d1540` and was voided by the push to `89bd426d`. `89bd426d` has **no final LGTM** — a Round 2 review was dispatched and has been stopped, since a superseded PR does not need one. CI being green at this SHA is not a merge signal on its own.

### What happens to #648

#648 stays **open**. Its goal — crates.io must stop lagging the released version — is not yet automated. #655 enforces the invariant but the publish itself remains manual.

Any successor PR should be designed the other way round from this one: **crates.io publish succeeds first, and only then is the GitHub Release allowed to proceed**, which is the ordering #655 already established and which this PR inverted. Starting from this branch would carry the inverted assumption forward, so a fresh design is the right call.

Branch `chore/gh648-crates-io-publish` is kept, not deleted, per the never-delete rule; the work stays readable at `89bd426d`.

PR #717, opened 2026-09-02
fix(edda-cli): make the claim machine gate see bare CLI claims; one liveness criterion through session inference (GH-705)

fagemx:
Issue: #705

Rebased cleanly onto latest `main` (`4075ab5`), pinned to `567bdfa1cced592617af711cdf150ed2cefc9984`.
Resolves the semantic and test fixture conflicts with #723 (`process` claims / `subject` matching) and #719 (hermetic `e2e_repo()` fixture).

## What this does

- **Defect A — `cli-*` bare claims are visible to the machine gate and fail-closed:**
  Bare CLI claims (`cli-*`, minted by `resolve_session_id` tier 4) are one-shot processes that leave no ongoing heartbeats. Because board claims never expire and heartbeat age cannot prove a one-shot process gone, `claim check` now separates bare CLI claims into `unjudgeable_claims` and counts them toward exit 1 while they stand on the board. In `--json`, they are distinguishable from live conflicts and stale claims. Never-heartbeated claims now serialize explicitly as `"age_secs": null`.
  `cmd_bridge::claim` also verifies that the coordination board fold actually recorded the new claim before returning success, avoiding false-positive success on lost writes.

- **Defect B — `infer_session_id` observes the shared `liveness_from_heartbeat` criterion:**
  Session inference uses the shared criterion from `peers::liveness` (including the 15x sub-agent multiplier), ensuring a sub-agent session cannot be simultaneously alive for `edda peers` / `claim check` and dead for inference.

- **Hermeticity & Compatibility:**
  All e2e tests use `e2e_repo()` anchored at the test fixture directory (GH-646), and compatibility with GH-581 `subject` matching is preserved in `intersecting_claims`.

## Gate receipt (Pinned to `567bdfa1cced592617af711cdf150ed2cefc9984`)

| Gate | Host Home (`~/.edda` present) | Isolated Empty Home (`HOME`/`USERPROFILE` → tempdir) |
|---|---|---|
| `cargo fmt --all --check` | OK (exit 0) | (environment-independent) |
| `cargo clippy -p edda -p edda-bridge-claude --all-targets -- -D warnings` | 0 warnings (exit 0) | (environment-independent) |
| `cargo test -p edda` | **507 passed; 0 failed** | **507 passed; 0 failed** |
| `cargo test -p edda-bridge-claude -- --test-threads=4` | **634 passed; 0 failed** | **634 passed; 0 failed** |

Named test receipts:
```
test cmd_claim::tests::e2e_bare_cli_claim_is_visible_to_claim_check ... ok
test cmd_claim::tests::e2e_aged_bare_cli_claim_still_conflicts ... ok
test cmd_claim::tests::e2e_never_heartbeated_bare_cli_claim_conflicts ... ok
test cmd_claim::tests::e2e_aged_bare_cli_claim_is_classified_in_json ... ok
test cmd_claim::tests::e2e_bare_cli_reclaim_is_idempotent ... ok
test cmd_claim::tests::e2e_bare_cli_claim_can_narrow_scope ... ok
test cmd_claim::tests::e2e_second_agent_claims_disjoint_scope ... ok
test cmd_claim::tests::e2e_lost_board_write_is_not_reported_as_success ... ok
test cmd_claim::tests::e2e_never_heartbeated_stale_claim_is_explicit_in_json ... ok
test peers::tests::infer_session_counts_parented_subagent_by_shared_criterion ... ok
test cmd_bridge::tests::resolve_session_id_refusal_names_live_sessions ... ok
```

## 範圍確認

PR 嚴格僅修改 6 個檔案：
- `crates/edda-bridge-claude/src/peers/discovery.rs`
- `crates/edda-bridge-claude/src/peers/liveness.rs`
- `crates/edda-bridge-claude/src/peers/tests.rs`
- `crates/edda-cli/src/cmd_bridge.rs`
- `crates/edda-cli/src/cmd_claim.rs`
- `crates/edda-cli/src/skills/coord-handoff.md`

未觸碰任何其他 crate 或腳本。

### DIFF
--- a/crates/edda-bridge-claude/src/peers/discovery.rs
+++ b/crates/edda-bridge-claude/src/peers/discovery.rs
@@ -5,7 +5,7 @@ use crate::parse::now_rfc3339;
 use super::board::compute_board_state;
 use super::helpers::parse_rfc3339_to_epoch;
 use super::liveness::{liveness_from_heartbeat, SessionLiveness};
-use super::{stale_secs, PeerSummary, SessionHeartbeat};
+use super::{PeerSummary, SessionHeartbeat};
 
 // ── Peer Discovery ──
 
@@ -164,7 +164,6 @@ pub fn discover_all_sessions(project_id: &str) -> Vec<PeerSummary> {
 /// (GH-503).
 pub fn infer_session_id(project_id: &str) -> Option<(String, String)> {
     let state_dir = edda_store::project_dir(project_id).join("state");
-    let stale_threshold = stale_secs();
     let now = parse_rfc3339_to_epoch(&now_rfc3339()).unwrap_or(0);
 
     let entries = match fs::read_dir(&state_dir) {
@@ -189,10 +188,13 @@ pub fn infer_session_id(project_id: &str) -> Option<(String, String)> {
             Err(_) => continue,
         };
 
-        let hb_epoch = parse_rfc3339_to_epoch(&hb.last_heartbeat).unwrap_or(0);
-        let age = now.saturating_sub(hb_epoch);
-
-        if age <= stale_threshold {
+        // GH-705: the ONE liveness criterion (`peers::liveness`), not an
+        // inline age check. The shared criterion carries the parented
+        // sub-agent 15x multiplier, so a sub-agent heartbeat that `edda
+        // peers` and `claim check` both count as live cannot be dropped as
+        // stale here — the same session must not be simultaneously alive
+        // and dead depending on which verb asks.
+        if liveness_from_heartbeat(&hb, now).is_live() {
             active.push((hb.session_id, hb.label));
         }
     }
--- a/crates/edda-bridge-claude/src/peers/liveness.rs
+++ b/crates/edda-bridge-claude/src/peers/liveness.rs
@@ -4,6 +4,9 @@
 //! [`stale_secs`] (x15 for parented sub-agents, mirroring peer discovery).
 //! `edda peers` and `edda claim check` must both go through this module so
 //! the two verbs can never grow a second, disagreeing notion of "dead".
+//! Session inference (`discovery::infer_session_id`) observes the same
+//! criterion (GH-705), so a session that is live for those two verbs is
+//! inferable there too.
 //!
 //! There is one liveness surface — the heartbeat file — and one criterion —
 //! [`liveness_from_heartbeat`]. Do not add a parallel one.
--- a/crates/edda-bridge-claude/src/peers/tests.rs
+++ b/crates/edda-bridge-claude/src/peers/tests.rs
@@ -964,6 +964,62 @@ fn infer_session_only_stale() {
     let _ = fs::remove_dir_all(edda_store::project_dir(pid));
 }
 
+/// RFC3339 timestamp `secs` seconds before now (UTC) — for heartbeats whose
+/// age must fall inside (not just outside) the staleness window.
+fn rfc3339_now_minus(secs: u64) -> String {
+    let now = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .expect("system clock")
+        .as_secs()
+        .saturating_sub(secs);
+    time::OffsetDateTime::from_unix_timestamp(now as i64)
+        .expect("unix timestamp")
+        .format(&time::format_description::well_known::Rfc3339)
+        .expect("rfc3339")
+}
+
+#[test]
+fn infer_session_counts_parented_subagent_by_shared_criterion() {
+    // GH-705 defect B: the shared liveness criterion gives a parented
+    // sub-agent a 15x staleness multiplier (no hook events fire during a
+    // sub-agent run, so its heartbeat would otherwise age out mid-run).
+    // `infer_session_id` must apply the SAME criterion: a 20-minute-old
+    // parented sub-agent heartbeat is Live for `edda peers` and `claim
+    // check`, so session inference must not drop it as stale — otherwise
+    // the same session is simultaneously alive and dead depending on which
+    // verb asks.
+    let pid = "test_infer_parented_15x";
+    let _ = edda_store::ensure_dirs(pid);
+
+    let age_secs: u64 = 20 * 60; // 20 min: stale at 120s, live at 15x (1800s)
+    let ts = rfc3339_now_minus(age_secs);
+    let path = heartbeat_path(pid, "sub-agent");
+    let hb = serde_json::json!({
+        "session_id": "sub-agent",
+        "started_at": ts,
+        "last_heartbeat": ts,
+        "label": "sub",
+        "focus_files": [],
+        "active_tasks": [],
+        "files_modified_count": 0,
+        "total_edits": 0,
+        "recent_commits": [],
+        "parent_session_id": "parent-1"
+    });
+    let _ = fs::create_dir_all(path.parent().unwrap());
+    let _ = fs::write(&path, serde_json::to_string_pretty(&hb).unwrap());
+
+    let result = infer_session_id(pid);
+    assert_eq!(
+        result,
+        Some(("sub-agent".into(), "sub".into())),
+        "a parented sub-agent live under the shared criterion must be inferable"
+    );
+
+    remove_heartbeat(pid, "sub-agent");
+    let _ = fs::remove_dir_all(edda_store::project_dir(pid));
+}
+
 // ── Issue #148 Gap 6: Cross-session decision conflict ──
 
 #[test]
--- a/crates/edda-cli/src/cmd_bridge.rs
+++ b/crates/edda-cli/src/cmd_bridge.rs
@@ -773,6 +773,43 @@ pub fn claim(
         paths,
         subject,
     );
+
+    // GH-705: the board entry is the claim's entire machine visibility —
+    // `claim check` counts a claim it can read, never one it cannot. The
+    // append is best-effort by shape (it serves fire-and-forget hook paths
+    // too), so verify the fold actually sees the new claim before printing
+    // success: a lost write must not look like a claimed scope, or the
+    // surface reads clear while the occupation is real. This deliberately
+    // does NOT write a heartbeat for the minted session: the claimant is a
+    // one-shot process, so a heartbeat it writes once ages with nothing to
+    // refresh it and proves nothing after stale_secs — and its residue made
+    // every subsequent bare claim hit resolve_session_id's live-session
+    // ambiguity refusal. The machine gate instead counts a `cli-*` claim
+    // for as long as it stands on the board (cmd_claim: unjudgeable,
+    // fail-closed), which does not decay with time.
+    //
+    // Distinguish the new entry from any replaced entry: the entry must exist,
+    // match the just-written label, paths, and subject, and its timestamp must
+    // advance past the replaced entry (so a lost write on re-claiming,
+    // narrowing, or identical re-claim does not falsely pass by re-reading the
+    // prior state).
+    let current = edda_bridge_claude::peers::compute_board_state(&project_id)
+        .claims
+        .into_iter()
+        .find(|c| c.session_id == session_id);
+    let recorded = current.as_ref().is_some_and(|c| {
+        c.label == label
+            && c.paths == paths
+            && c.subject.as_deref() == subject
+            && replaced.as_ref().map(|r| &r.ts) != Some(&c.ts)
+    });
+    if !recorded {
+        anyhow::bail!(
+            "claim for session {session_id} was not recorded on the coordination board; \
+             refusing to report a claim the machine gate cannot see"
+        );
+    }
+
     for line in claim_disclosure(replaced.as_ref(), label, paths) {
         println!("{line}");
     }
@@ -1289,10 +1326,12 @@ pub(crate) fn resolve_session_id(
         }
     }
 
-    if has_live_sessions(project_id) {
+    let live = fresh_sessions(project_id);
+    if !live.is_empty() {
         anyhow::bail!(
             "cannot prove which live session belongs to this process, so --session is required \
-             (or set EDDA_SESSION_ID in the invoking process)"
+             (or set EDDA_SESSION_ID in the invoking process).\n{}",
+            format_live_sessions(&live)
         );
     }
 
@@ -1301,10 +1340,27 @@ pub(crate) fn resolve_session_id(
 }
 
 fn has_live_sessions(project_id: &str) -> bool {
+    !fresh_sessions(project_id).is_empty()
+}
+
+/// The sessions currently passing the shared liveness window.
+fn fresh_sessions(project_id: &str) -> Vec<edda_bridge_claude::peers::PeerSummary> {
     let stale = edda_bridge_claude::peers::stale_secs();
     edda_bridge_claude::peers::discover_all_sessions(project_id)
         .into_iter()
-        .any(|session| session.age_secs <= stale)
+        .filter(|session| session.age_secs <= stale)
+        .collect()
+}
+
+/// Name the live sessions in the identity-refusal error, so the caller can
+/// copy an id into `--session` — an error that demands an id without
+/// showing one cannot be acted on (round-1 consequence, GH-705).
+fn format_live_sessions(live: &[edda_bridge_claude::peers::PeerSummary]) -> String {
+    let mut out = String::from("Live sessions (pass --session with one of these ids):");
+    for session in live {
+        out.push_str(&format!("\n  {} — {}", session.session_id, session.label));
+    }
+    out
 }
 
 /// `edda bridge claude digest --session <id>` or `--all`
@@ -2238,6 +2294,31 @@ mod tests {
         assert_eq!(sid, "explicit-wins", "tier 1 should beat tier 2");
         std::env::remove_var("EDDA_SESSION_ID");
 
+        let _ = std::fs::remove_dir_all(edda_store::project_dir(pid));
+    }
+
+    #[test]
+    fn resolve_session_id_refusal_names_live_sessions() {
+        // Round-1 consequence (GH-705): when the bare-CLI refusal does fire —
+        // a live hooked session beside this bare command — the error must
+        // name the ids to copy into `--session`, the way `unclaim`'s refusal
+        // lists the board. An error that demands an id without showing one
+        // cannot be acted on.
+        let _store = crate::test_support::isolated_store();
+        let _env = env_guard();
+        let pid = "test_resolve_sid_refusal_names";
+        let _ = edda_store::ensure_dirs(pid);
+        std::env::remove_var("EDDA_SESSION_ID");
+        std::env::remove_var("EDDA_SESSION_LABEL");
+        edda_bridge_claude::peers::write_heartbeat_minimal(pid, "sess-live", "auth", ".");
+        let err = resolve_session_id(None, pid, "auth")
+            .expect_err("a bare command beside a live session must refuse");
+        let msg = err.to_string();
+        assert!(
+            msg.contains("sess-live"),
+            "refusal must name the live session id to pass to --session: {msg}"
+        );
+        assert!(msg.contains("--session"), "{msg}");
         let _ = std::fs::remove_dir_all(edda_store::project_dir(pid));
     }
 
--- a/crates/edda-cli/src/cmd_claim.rs
+++ b/crates/edda-cli/src/cmd_claim.rs
@@ -8,7 +8,11 @@
 //! - 0 — the query surface is disjoint from every active claim (or the board
 //!   holds no claims)
 //! - 1 — at least one active claim overlaps; the conflicting labels/sessions
-//!   are named on stdout
+//!   are named on stdout. That includes a bare CLI claim (`cli-*`, minted by
+//!   `resolve_session_id` tier 4): its claimant is a one-shot process, so no
+//!   heartbeat age can prove it gone and its claim stands on the board until
+//!   `unclaim` (GH-705). Reading such an occupation as clear is the false
+//!   negative this verb exists to prevent.
 //! - 2 — the query could not be answered soundly: usage error, the board
 //!   cannot be read or parsed, a glob pattern is malformed, or a pair of
 //!   surfaces cannot be decided soundly (for example two tokens differing
@@ -49,9 +53,11 @@ pub struct ClaimConflict {
 pub struct StaleClaim {
     pub label: String,
     pub session_id: String,
-    /// Seconds since the session's last heartbeat, or `None` when the
-    /// session has no heartbeat file at all (never heard from).
-    #[serde(skip_serializing_if = "Option::is_none")]
+    /// Seconds since the session's last heartbeat, or `null` in the machine
+    /// report when the session has no heartbeat file at all (never heard
+    /// from). The field is always serialized — GH-705: a consumer must be
+    /// able to distinguish "expired heartbeat" from "no heartbeat file"
+    /// without guessing.
     pub age_secs: Option<u64>,
 }
 
@@ -59,6 +65,19 @@ pub struct StaleClaim {
 #[derive(Debug, Clone, Serialize, Default, PartialEq)]
 pub struct CheckReport {
     pub conflicts: Vec<ClaimConflict>,
+    /// Bare CLI claims (`cli-*`) whose surface intersects the query (GH-705).
+    /// A bare CLI session is a one-shot process: no heartbeat writer ever
+    /// refreshes for it, so heartbeat age carries no liveness information —
+    /// an old heartbeat does not mean the claim is gone, and no heartbeat
+    /// does not mean it was never real. The claim stands on the board until
+    /// `unclaim` (board claims never expire), so the gate must never read
+    /// the surface as genuinely clear while it stands: these are counted
+    /// toward the exit code exactly like `conflicts`, separated only so a
+    /// consumer can tell a live peer from an occupation whose liveness
+    /// cannot be judged. Unlike the GH-617 liveness filter, this does not
+    /// decay with time.
+    #[serde(default)]
+    pub unjudgeable_claims: Vec<ClaimConflict>,
     /// Claims filtered out because their session is dead (GH-617). Reported
     /// for death visibility — never counted toward the exit code.
     #[serde(default)]
@@ -102,19 +121,32 @@ pub fn claim_check(repo_root: &Path, query: &[String], json: bool) -> anyhow::Re
     let board_size = claims.len();
     let mut live: Vec<ClaimEntry> = Vec::new();
     let mut stale_claims: Vec<StaleClaim> = Vec::new();
+    // GH-705: a claim owned by a bare CLI session (`cli-*`, minted by
+    // resolve_session_id tier 4) is an occupation whose liveness cannot be
+    // judged from a heartbeat. The claimant is a one-shot process; nothing
+    // ever refreshes a heartbeat for it; and the board claim never expires.
+    // Age therefore proves nothing, so instead of dismissing such a claim as
+    // stale when its heartbeat ages out, it is separated out here and
+    // counted toward the exit code when its surface intersects the query
+    // (fail-closed). A session with a live heartbeat still conflicts
+    // normally; the GH-617 criterion itself is untouched.
+    let mut unjudgeable: Vec<ClaimEntry> = Vec::new();
     for c in claims {
-        match classify_session_liveness(&project_id, &c.session_id) {
-            SessionLiveness::Live { .. } => live.push(c),
-            SessionLiveness::Stale { age_secs } => stale_claims.push(StaleClaim {
+        let liveness = classify_session_liveness(&project_id, &c.session_id);
+        if liveness.is_live() {
+            live.push(c);
+        } else if is_bare_cli_session(&c.session_id) {
+            unjudgeable.push(c);
+        } else {
+            let age_secs = match liveness {
+                SessionLiveness::Stale { age_secs } => Some(age_secs),
+                _ => None, // NoHeartbeat: never heard from
+            };
+            stale_claims.push(StaleClaim {
                 label: c.label,
                 session_id: c.session_id,
-                age_secs: Some(age_secs),
-            }),
-            SessionLiveness::NoHeartbeat => stale_claims.push(StaleClaim {
-                label: c.label,
-                session_id: c.session_id,
-                age_secs: None,
-            }),
+                age_secs,
+            });
         }
     }
 
@@ -125,16 +157,28 @@ pub fn claim_check(repo_root: &Path, query: &[String], json: bool) -> anyhow::Re
             std::process::exit(2);
         }
     };
+    // The unjudgeable claims go through the same sound intersection
+    // decision — including its exit-2 refusal of malformed or undecidable
+    // patterns — so a bare claim's surface can never be judged clear by a
+    // looser rule than a live claim's.
+    let unjudgeable_conflicts = match intersecting_claims(&unjudgeable, &query_refs) {
+        Ok(conflicts) => conflicts,
+        Err(err) => {
+            eprintln!("error: {err}");
+            std::process::exit(2);
+        }
+    };
 
     let report = CheckReport {
         conflicts: report.conflicts,
+        unjudgeable_claims: unjudgeable_conflicts,
         stale_claims,
     };
 
     if json {
         let out = serde_json::to_string_pretty(&report).context("serialize claim check report")?;
         println!("{out}");
-    } else if report.conflicts.is_empty() {
+    } else if report.conflicts.is_empty() && report.unjudgeable_claims.is_empty() {
         if board_size == 0 {
             println!("No active claims on the coordination board; surface is clear.");
         } else {
@@ -155,9 +199,19 @@ pub fn claim_check(repo_root: &Path, query: &[String], json: bool) -> anyhow::Re
                 println!("  query {}  <->  claim {}", pair.query, pair.claim_path);
             }
         }
+        for conflict in &report.unjudgeable_claims {
+            println!(
+                "CONFLICT with claim \"{}\" (session {}) — bare-CLI claim, liveness cannot be \
+                 judged; it counts until it is unclaimed",
+                conflict.label, conflict.session_id
+            );
+            for pair in &conflict.intersections {
+                println!("  query {}  <->  claim {}", pair.query, pair.claim_path);
+            }
+        }
         println!(
             "{} conflicting claim(s) across {} query path(s).",
-            report.conflicts.len(),
+            report.conflicts.len() + report.unjudgeable_claims.len(),
             query.len()
         );
         print_stale_claims(&report.stale_claims);
@@ -173,12 +227,25 @@ pub fn claim_check(repo_root: &Path, query: &[String], json: bool) -> anyhow::Re
 ///
 /// Stale claims (dead sessions, GH-617) are deliberately not counted here:
 /// they are liveness-filtered bookkeeping, never a conflict signal.
+/// Unjudgeable claims (bare CLI claims, GH-705) ARE counted: their session
+/// has no heartbeat writer, so the gate cannot prove the occupation gone
+/// and must fail closed.
 fn exit_code_for(report: &CheckReport) -> i32 {
-    if report.conflicts.is_empty() {
+    if report.conflicts.is_empty() && report.unjudgeable_claims.is_empty() {
         0
     } else {
         1
     }
+}
+
+/// Session ids minted by `cmd_bridge::resolve_session_id` tier 4 for bare
+/// CLI invocations (`cli-<label>`). Such a session is a one-shot process:
+/// it wrote its claim and exited, and no hook ever refreshes a heartbeat
+/// for it, so heartbeat age carries no liveness information (GH-705). The
+/// same shape is already classified in `cmd_bridge` when it names the
+/// actor of a `cli-*` session.
+fn is_bare_cli_session(session_id: &str) -> bool {
+    session_id.starts_with("cli-")
 }
 
 /// Death visibility (GH-617): reveal how many stale claims the liveness
@@ -350,6 +417,19 @@ pub fn check(
     claims: &[edda_bridge_claude::peers::ClaimEntry],
     query: &[&str],
 ) -> Result<CheckReport, String> {
+    Ok(CheckReport {
+        conflicts: intersecting_claims(claims, query)?,
+        unjudgeable_claims: Vec::new(),
+        stale_claims: Vec::new(),
+    })
+}
+
+/// The claims from `claims` whose recorded surface intersects `query`, by
+/// the same sound intersection decision for every caller.
+fn intersecting_claims(
+    claims: &[edda_bridge_claude::peers::ClaimEntry],
+    query: &[&str],
+) -> Result<Vec<ClaimConflict>, String> {
     // Validate every glob up front, so a malformed pattern is an error even
     // when it would never be compared against another glob.
     for token in query.iter().map(|t| normalize_token(t)).chain(
@@ -405,10 +485,7 @@ pub fn check(
             });
         }
     }
-    Ok(CheckReport {
-        conflicts,
-        stale_claims: Vec::new(),
-    })
+    Ok(conflicts)
 }
 
 /// Whether a query token and a claim subject token name the same process object.
@@ -1544,6 +1621,25 @@ mod tests {
         )
     }
 
+    /// Run the real binary carrying no session identity, so
+    /// `resolve_session_id` falls through to the bare-CLI tier-4 mint of
+    /// `cli-<label>` — the path every bare-shell agent takes.
+    fn run_edda_bare(args: &[&str], repo: &Path, store: &Path) -> (i32, String, String) {
+        let out = std::process::Command::new(edda_bin())
+            .args(args)
+            .current_dir(repo)
+            .env("EDDA_STORE_ROOT", store)
+            .env_remove("EDDA_SESSION_ID")
+            .env_remove("EDDA_SESSION_LABEL")
+            .output()
+            .expect("spawn edda");
+        (
+            out.status.code().expect("exit code"),
+            String::from_utf8_lossy(&out.stdout).into_owned(),
+            String::from_utf8_lossy(&out.stderr).into_owned(),
+        )
+    }
+
     fn board_file(store: &Path, project_id: &str) -> PathBuf {
         store
             .join("projects")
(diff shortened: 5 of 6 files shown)


fagemx:
## Code Review: Round 1 — PR #717 @ ec369ec46aef4c5c7b8dfa7987157297ad263fc2

- model_requested: not named by the invoking brief (`fleet-review 717`)
- model_observed: unverified
- spec: review-spec-v1
- class: code-risk  (REVIEW.md classes: code-plain, docs, skills)
- escalations: D5 (wording/structure — `[判斷]`, not adjudicated here)
- cost: ~25 tool calls; one focused `cargo build -p edda --bin edda` in the `verifier` lane (`CARGO_INCREMENTAL=0`), no workspace gate rerun

### IN SCOPE

Changed behaviour: (1) `edda claim` now writes a heartbeat for the session it resolved (`cmd_bridge.rs:752-767`); (2) `StaleClaim.age_secs` is always serialized, `null` when there is no heartbeat file (`cmd_claim.rs:49-58`); (3) `peers::infer_session_id` swaps its inline age check for the shared `liveness_from_heartbeat` criterion (`discovery.rs:195-201`); (4) the shipped `coord-handoff.md` examples and the `skill_doctest.rs` header are re-pinned to the new observable behaviour.

Direct consumers audited: `cmd_claim.rs:104-118` (liveness filter → `conflicts` / `stale_claims` / exit code), `cmd_bridge.rs:1288-1304` (`resolve_session_id` tier 3/4 and `has_live_sessions`), `cmd_bridge.rs:613-620` (`peers`), `peers/heartbeat.rs:134-149` (`write_heartbeat_minimal`).

Acceptance ceiling: issue #705 doneWhen — (a) a `cli-*` claim's existence must be machine-visible; (b) no relaxation of the liveness criterion; (c) a FAIL-first regression test that distinguishes "genuinely clear" from "an occupation whose liveness cannot be judged".

Integration against current base: rebased on `cbbc8b0`; `origin/main` is now `c9f6111`, no conflicting surface.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 717 --name-only` — 7 files, all on the claim/peers path named by #705 plus the forced skill/doctest re-pin |
| U2 | any | P1 | PASS | closing-keyword grep empty on body and on `cbbc8b0..ec369ec` commit messages |
| U3 | any | P1 | PASS | `Issue: #705` |
| U4 | any | P1 | PASS | subject-line grep empty (3 commits: `test(edda-cli)`, `fix(edda-cli)`, `docs(edda-cli)`) |
| U6 | any | P0 | PASS | `gh pr checks 717` all 9 green; run `33666478699` `head_sha=ec369ec46aef…`, `success` — exact head. Rerun reason for the one thing I RAN: see RAN vs READ |
| U7 | any | P1 | PASS | whole scoped audit finished before posting; all blockers batched below |
| D1 | docs | P1 | PASS | `edda claim --help` → 0; `edda peers --help` → 0 (the only two commands the diff adds in backticks) |
| D2 | docs | P1 | N.A.(the diff asserts no ledger decision value or status) | — |
| D3 | docs | P1 | PASS | path/link enumerator produced 0 candidates over the added lines |
| D4 | docs | P0 | PASS | authority grep (`gh pr merge`, `--delete-branch`, `--admin`, `--no-verify`, force-push) empty |
| D5 | docs | `[判斷]` | 需升級 | wording/scoping of the `coord-handoff.md` rewrite — see escalations |
| S1 | skills | P0 | PASS | `merge` / `self-review` / `skip review` grep over the skill diff empty |
| S2 | skills | P1 | PASS | D1/D3 probes restricted to `coord-handoff.md`; both commands and the newly pinned string resolve — `cannot prove which live session belongs to this process` exists at `cmd_bridge.rs:868` |
| S3 | skills | P1 | PASS | the skill describes behaviour; it restates no rule owned by `REVIEW.md` / `.claude/CLAUDE.md` |
| C1 | code-plain | P0 | **FAIL** | doneWhen (a) holds only for `stale_secs` (default 120 s) after the claim — finding P0-1 |
| C2 | code-plain | P0 | PASS | `fix:` carries 3 new tests (`cmd_claim.rs:1803`, `cmd_claim.rs:2008`, `peers/tests.rs:976`); FAIL-first evidence READ from the PR body |
| C3 | code-plain | P1 | PASS | 12 `unwrap`/`expect` candidates, all inside `#[cfg(test)]` — `peers/mod.rs:262-265` (`#[cfg(test)] #[path="tests.rs"] mod tests`) and `cmd_claim.rs:1122` |
| C4 | code-plain | P1 | PASS | `unsafe` / blanket-allow grep empty |
| C5 | code-plain | P1 | PASS | selector prints nothing — both touched crates (`edda-cli` → `edda`, `edda-bridge-claude`) are inside the Windows CI subset. Windows behavioural probes RAN anyway, see below |
| R1–R5 | code-risk | — | N.A.(not routed — no path in the diff is on the §3.1 risk surface; the canonical `class:` field reads `code-risk` only because `code-plain` maps there) | — |
| §5.5 | any | P1 | **FAIL** | swallowed write on the coordination path — finding P1-2 |

### Wiring

|---|---|---|---|---|
| heartbeat file for a `cli-*` session, written by `edda claim` | `cmd_bridge.rs:762-767` → `peers::write_heartbeat_minimal` → `session.<id>.json` | `cmd_claim.rs:104-118` (liveness filter → exit code); `cmd_bridge.rs:1299-1304` `has_live_sessions`; `cmd_bridge.rs:613` `peers` | **swallowed** — `peers/heartbeat.rs:137` `let _ = edda_store::update_heartbeat(…)`, and `claim()` ignores the unit return, so a failed write still prints `Claimed scope:` and exits 0 → P1-2 | verb → store → read-back proven by `cmd_claim.rs:1803`, but only inside the fresh window (`peers/mod.rs:13-18`, 120 s) → P0-1 |
| `StaleClaim.age_secs` always serialized (`null` for "no heartbeat file") | `cmd_claim.rs:49-58`, populated at `cmd_claim.rs:115-118` | the `--json` machine contract; in-repo only the new test `cmd_claim.rs:2008-2043`. No in-repo caller was switched to read it | n/a (pure serialization) | report → stdout → parsed and asserted by that test |
| `infer_session_id` criterion swap (not a new surface — pre-existing `pub fn`) | `discovery.rs:195-201` | none in production: `grep -rn infer_session_id crates/` yields `peers/mod.rs:239` (`pub use`) and tests only; `cmd_phase.rs:206` calls its **own** local `infer_session_id` defined at `cmd_phase.rs:276` | n/a | proven by `peers/tests.rs:976` alone → P2-1 |

### Findings

- **[P0] The fix makes a bare `cli-*` claim machine-visible for only `stale_secs` (120 s by default); after that the exact GH-705 false negative returns.** `edda claim` is a one-shot process — nothing refreshes the heartbeat it writes, while board claims never expire (`cmd_bridge.rs:707-712`). So the case the issue is actually about ("claim your scope at session start, then work") is unprotected two minutes later. Evidence — binary built at this SHA in the `verifier` lane, fresh temp repo + `EDDA_STORE_ROOT`:

  ```
  $ edda claim "auth" --paths "src/auth/*"      -> exit 0, session: cli-auth
  $ edda claim check "src/auth/x.rs"            -> CONFLICT ..., exit=1        (the fix works)

  # backdate session.cli-auth.json last_heartbeat by 10 min -- i.e. simply let
  # the one-shot claim process be gone for longer than stale_secs
  $ edda claim check "src/auth/x.rs"
      No conflicts: 0 live claim(s) checked against 1 query path(s).
      stale, not counted: claim "auth" (session cli-auth) - heartbeat 600s ago
      exit=0                                                                  (regressed)
  ```

  The JSON half (`stale_claims[].age_secs: 600`) is the GH-617 field that already existed before this PR, so for an aged bare claim all this PR adds is `null` instead of a missing key. doneWhen (a) asks that the *existence* of a `cli-*` claim be machine-visible; a 120-second window does not meet it. For the response: the PR does implement the issue's literal first sub-bullet ("write one heartbeat alongside the claim"), but that sub-bullet is a proposed means and as implemented it does not carry the headline requirement. The issue's other two options — a dedicated exit code / classification for a structurally unjudgeable occupation, or having consumers read the JSON field — do not decay with time. The new test `cmd_claim.rs:1803` cannot catch this because it only queries inside the fresh window.

- **[P1] Writing that heartbeat makes the bare-CLI claim path single-shot: the next `edda claim` in the project is refused for 120 s — an identical idempotent re-claim, a scope narrowing, and a second agent claiming a disjoint scope alike.** `claim()` now creates the very live heartbeat that `resolve_session_id` treats as "identity is ambiguous" (`cmd_bridge.rs:1288-1293` via `has_live_sessions`, `cmd_bridge.rs:1299-1304`). A/B, same command sequence, fresh store each time:

  ```
  post-fix (this SHA)                              pre-fix (main)
  claim "auth"                    -> exit 0        claim "auth"  -> exit 0
  claim "auth"                    -> exit 1        claim "auth"  -> Re-claimed scope: auth (unchanged), exit 0
     Error: cannot prove which live session
     belongs to this process, so --session
     is required ...
  claim "auth" --paths src/auth/login.rs           claim "auth" --paths src/auth/login.rs
                                  -> exit 1                      -> Re-claimed scope: auth (previous paths replaced), exit 0
  claim "api"  --paths src/api/*  -> exit 1        claim "api" --paths src/api/*
                                                                 -> Claimed scope: api, session: cli-api, exit 0
  ```

  This contradicts the contract the code states about itself — `cmd_bridge.rs:707-712` ("an idempotent re-claim hits exactly that, since tier 4 mints a deterministic `cli-<label>` … so a bare-shell restart re-runs the same command against its own surviving claim") and the test comment at `cmd_bridge.rs:2626-2628` ("Narrowing a scope, and re-claiming after a restart, both go through this path -- which is why replacement is right and rejecting a second claim would not be") — and it contradicts the multi-agent instruction every agent is given ("Claim your scope at session start: `edda claim …`"), which the second bare-CLI agent in a project can no longer follow. No existing test catches it because every re-claim test passes an explicit `Some("s1")` (`cmd_bridge.rs:2571`, `2630`, `2635`) and so never reaches tier 4 — which is why CI is green. Two smaller consequences worth handling in the same round: the refusal text names no id to copy (unlike `unclaim`, which prints `describe_claims`), and the updated `coord-handoff.md` — the skill whose whole subject is this bare-CLI flow — does not mention the new requirement.

- **[P1] The liveness signal the whole fix depends on is written best-effort and its failure is swallowed.** `cmd_bridge.rs:762-767` calls `write_heartbeat_minimal`, which is `let _ = edda_store::update_heartbeat(…)` at `peers/heartbeat.rs:137` and returns `()`. If that write fails (lock contention on the `.lock` sidecar, a read-only or full store), `claim` still prints `Claimed scope:` and exits 0 while the claim is again invisible to `claim check` — silently reinstating the false negative this PR exists to remove. §5.5 fixed adjudication: an error swallowed on a coordination path is P1.

- **[P2] `peers::infer_session_id` (Defect B) has no production consumer in this repo**, so that half of the fix is proven by `peers/tests.rs:976` and consumed by nothing: the only non-test references are `peers/mod.rs:239` (`pub use`) and `cmd_phase.rs:206`, which calls a *different*, locally defined `infer_session_id` (`cmd_phase.rs:276`) that still does its own age check. Worth one line in the response on whether `cmd_phase`'s local copy is meant to converge — otherwise the "same session must not be simultaneously alive and dead depending on which verb asks" claim added at `liveness.rs:7-9` is still only true for the verbs that go through `peers`.

### FOLLOW-UP ISSUE

- `cmd_phase.rs:276` defines a second, independent `infer_session_id` that does not go through `peers::liveness`. Outside #705's frozen surface (the diff does not touch it), but it is the remaining instance of the "second, disagreeing notion of dead" that `liveness.rs:1-11` forbids. Evidence: `crates/edda-cli/src/cmd_phase.rs:206,276`.

### RAN vs READ

RAN — this worktree, `C:/Users/fagem/.edda/fleet/wt-review-pr717`, detached at `ec369ec`, Windows 11 / Git Bash:

- `gh pr view/diff/checks 717`, `gh issue view 705`, `gh api …/runs/33666478699` → `head_sha` matches the reviewed SHA, `success`
- the §3 classifier block → `classes=code-plain docs`, `canonical_class=code-risk`. I **up-classed** `crates/edda-cli/src/skills/coord-handoff.md` to `skills` and ran §5.2: it is a shipped skill an agent executes, even though the router's `skills/*` glob is repo-root-anchored and does not match that path.
- the U2, U4, D1, D3, D4, S1, C2, C3, C4, C5, R1 enumerators — outputs in the Rules table
- `cargo build -p edda --bin edda`, `CARGO_TARGET_DIR=<lane-root>\verifier`, `CARGO_INCREMENTAL=0`. **Reason:** this is not a rerun of a recorded gate. No CI job and no test in the tree exercises the tier-4 bare-CLI claim path end to end (every re-claim test passes an explicit session id), so the real binary was the only way to measure P0-1 and P1-1. One bin, no workspace gate.
- behavioural probes with that binary, with the pre-fix `worker-1` binary as the A/B control, each in a fresh `tempdir` repo + `EDDA_STORE_ROOT` — outputs quoted in the findings

READ — not re-run:

- the L1 gate receipt in the PR body: `ec369ec…`, rustc 1.93.1 / stable-x86_64-pc-windows-msvc, lane `worker-1`, `CARGO_INCREMENTAL=0`; `cargo fmt --all --check` green, `cargo clippy --workspace --all-targets -- -D warnings` green, `cargo test --workspace --no-fail-fast` 2530 passed with 21 failures argued environmental, with a same-population control on `cbbc8b0` and green serial reruns of both affected binaries
- exact-head CI, run `33666478699` @ `ec369ec46aef…`: Format, Clippy (linux/macos/windows), Test (linux/macos/windows), CI Gate — all pass. That covers compile, lint and test on all three OSes; it does **not** cover the bare-CLI tier-4 behaviour above, which is why CI is green while P0-1 and P1-1 hold.
- the FAIL-first evidence for the three new tests (PR body, red at `65a3f26`)

### Verdict

Changes Requested, P0=1, P1=2 — a bare claim is machine-visible for only 120 s so the GH-705 false negative returns, and the heartbeat that buys those 120 s makes every subsequent bare-CLI claim in the project fail.

fagemx:
## Review Response: Round 1 — PR #717

- responding_to: `Code Review: Round 1` @ ec369ec46aef4c5c7b8dfa7987157297ad263fc2
- new full head SHA: **59df3294116259e7af469bafd8363a1a2c69d34e** (pushed; commits `64e0be0` FAIL-first tests, `59df329` fix)
- gates actually run this round: L0 on `edda` only — `cargo fmt --all --check` green; `cargo clippy -p edda --all-targets -- -D warnings` green (rustc 1.93.1, lane `worker-1`, `CARGO_INCREMENTAL=0`); `cargo test -p edda` 450 passed / 4 failed (see Gates note)
- **L1 workspace gate**: the previous lane started it but the lane was terminated before completion (`SCHED_S_TASK_TERMINATED`); **no L1 receipt exists yet for `59df329…` — receipt to follow in a later comment**. Exact-head CI is the authoritative workspace gate for this SHA.

### [P0] bare claim machine-visible for only `stale_secs` (120 s) — fixed at the root

Agreed the round-1 heartbeat was a decaying signal that carried doneWhen (a) for two minutes only. The fix removes the decay from the equation entirely:

- The claim verb no longer writes a heartbeat. The claim's machine visibility is now its board entry, which never expires (`cmd_bridge.rs` claim(): the `write_heartbeat_minimal` call is gone).
- `claim check` classifies claims owned by bare CLI sessions (`cli-*`, minted by `resolve_session_id` tier 4) as **unjudgeable** (`cmd_claim.rs`): heartbeat age proves nothing about a one-shot claimant, so the claim is counted toward the exit code exactly like a live conflict, for as long as it stands on the board — never decays back to clear. It gets its own `unjudgeable_claims` JSON bucket and a distinct human line ("bare-CLI claim, liveness cannot be judged … counts until it is unclaimed"), so a consumer can still tell a live peer from an unjudgeable occupation — the issue's second option, adopted without weakening the first.
- The unjudgeable claims go through the same sound intersection decision (including its exit-2 refusal), so a bare claim can never be judged clear by a looser rule than a live claim's.
- The GH-617 liveness criterion itself is untouched: live sessions still conflict; dead *non-CLI* sessions are still stale bookkeeping.
- `is_bare_cli_session` is checked only after the live check, so a `cli-*` session with a fresh heartbeat (explicitly passed `--session cli-…`) still classifies as a live conflict, not unjudgeable — the fail-closed direction in every branch.

Pinned by: `e2e_aged_bare_cli_claim_still_conflicts` (600 s backdated heartbeat → exit 1), `e2e_never_heartbeated_bare_cli_claim_conflicts`, `e2e_aged_bare_cli_claim_is_classified_in_json` (`unjudgeable_claims` bucket, `stale_claims` empty), and the rewritten `e2e_bare_cli_claim_is_visible_to_claim_check` (no heartbeat file required at all).

### [P1] bare-CLI claim path became single-shot — fixed

Agreed this contradicted the documented replacement contract and the "claim your scope at session start" instruction. Fixed by removing the cause, not special-casing the refusal: the claim verb no longer writes the heartbeat that made `resolve_session_id` see "live session beside a bare command". Tier 3/4 resolution returns to its documented behavior — verified by `e2e_bare_cli_reclaim_is_idempotent` (identical re-claim → "Re-claimed scope: auth (unchanged)", exit 0), `e2e_bare_cli_claim_can_narrow_scope` (→ "previous paths replaced", exit 0), and `e2e_second_agent_claims_disjoint_scope` (second bare agent → "Claimed scope: api", `cli-api`, exit 0).

Consequences named by the review, both handled:
- the refusal now names the live session ids to copy into `--session` (`describe_live_sessions`, same shape as `unclaim`'s board listing; no live sessions → "No live sessions were found; retry the command");
- `coord-handoff.md` is re-pinned to the final behavior: no heartbeat for a bare claim, `edda peers` shows "No active sessions", and a new doctest block shows `claim check` counting the standing claim (`! bare-CLI claim, liveness cannot be judged`) plus the "release the scope when done" instruction. `skill_doctest.rs` header matches.

### [P1 §5.5] swallowed write made a lost claim look successful — fixed by removing the dependency

Resolved differently than patching `peers/heartbeat.rs:137`, deliberately: the claim no longer travels through `write_heartbeat_minimal` at all, so that best-effort swallow is no longer on the claim path (it remains the hook paths' shape — GH-692 / PR #736's surface, not touched here to avoid collision). What carries the claim's visibility is the board append, and the claim verb now **verifies the fold actually sees the new claim** before printing success and bails otherwise (`cmd_bridge.rs` claim(): re-read `compute_board_state`, `anyhow::bail!("claim for session … was not recorded on the coordination board; refusing to report a claim the machine gate cannot see")`). Pinned by `e2e_lost_board_write_is_not_reported_as_success` (directory at the board path → non-zero exit, no "Claimed scope", stderr names the lost write).

### [P2] `peers::infer_session_id` production consumer — one line

`cmd_phase.rs:276`'s local copy is not converged in this PR; it is already filed as the FOLLOW-UP ISSUE (`cmd_phase.rs:276`, outside #705's frozen surface). The doc claim added at `liveness.rs` is scoped to the verbs that go through `peers`; convergence belongs to the follow-up.

### FOLLOW-UP ISSUE

Acknowledged: `cmd_phase.rs:276` second `infer_session_id` — filed as the round's follow-up, not extended here.

### Gates note — `cargo test -p edda` 4 failures, argued environmental (not rerun blindly)

450 passed; 4 failed, all in `cmd_reconcile`/`cmd_verify`, none on the GH-705 surface. Evidence they are workstation-environmental, not this diff:

1. Control recorded at `64e0be0` (previous lane): the same four failures reproduce identically at clean HEAD `ec369ec`.
2. Exact-head CI at `ec369ec` runs `cargo test` for the `edda` crate on Windows (plus full workspace on Linux/macOS) — all green; these tests pass in CI at the reviewed SHA.
3. Serial rerun (`--test-threads=1`) this round: identical failures — deterministic workstation-state dependence, not flakiness. Example: `verify_outside_edda_repo_exits_2_with_explanation` expects exit 2 but observed stdout `ledger chain OK: 48 event(s) … \?\C:\Users\fagem` — the test is reading this workstation's real user ledger through the temp path resolution.

These are the same population the round-1 verifier's L1 read argued environmental at `cbbc8b0`. The L1 workspace run now starting will report the same population; exact-head CI on `59df329…` is the authoritative workspace gate.

### Gate receipt (this round)

- SHA: 59df3294116259e7af469bafd8363a1a2c69d34e
- `cargo fmt --all --check`: green
- `cargo clippy -p edda --all-targets -- -D warnings`: green
- `cargo test -p edda`: 450 passed, 4 failed (environmental, argued above)
- toolchain: rustc 1.93.1 / stable-x86_64-pc-windows-msvc
- build lane: `worker-1`, `CARGO_INCREMENTAL=0`
- L1 workspace gate: started by previous lane, terminated before completion; no receipt yet — to follow

fagemx:
## L1 gate receipt — PR #717 @ 59df3294116259e7af469bafd8363a1a2c69d34e

- gate: L1 freeze (full workspace), clean tree, `CARGO_INCREMENTAL=0`
- toolchain: rustc 1.93.1 (01f6ddf75 2026-02-11) / stable-x86_64-pc-windows-msvc
- build lane: `worker-1` (`$LOCALAPPDATA\fleet-workstation\lanes\worker-1`)
- run by: the lane-fin717 finisher session (this round, after `Review Response: Round 1`)

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | **green** |
| `cargo clippy --workspace --all-targets -- -D warnings` | **green** |
| `cargo test --workspace --no-fail-fast` | 2530+ passed; **33 failures, argued environmental below** (3 targets) |

### Failure population and argument (all environmental, none on the GH-705 surface)

- **`edda` (bin): 4 failed / 450 passed.** Deterministic workstation-state dependence (same 4 under `--test-threads=1`): `cmd_reconcile`×3 + `cmd_verify::verify_outside_edda_repo_exits_2_with_explanation` — the latter observes stdout `ledger chain OK: 48 event(s) … \?\C:\Users\fagem`, i.e. the test resolves this workstation's real user ledger where it expects none. Control: identical population at clean HEAD `ec369ec` (recorded in commit `64e0be0`'s message); exact-head CI at `ec369ec` runs these tests on Windows and is green. Zero overlap with the PR's changed paths.
- **`edda-bridge-claude` (lib): 28 failed parallel → 3 failed serial (`--test-threads=1` control).** Parallel population is an intra-process env race (several `*_env_var`/cooldown tests mutate env; one failing test passes when run alone with `--exact`). The 3 serial survivors (`bg_digest::idempotency_guard_works`, `digest::collect_session_ledger_extras_no_workspace`, `dispatch::render_workspace_section_no_edda_returns_none`) each read this workstation's real state — actual digested marks / real ledger notes / a real installed edda. The PR touches only this crate's `peers`/`discovery`; `peers/tests.rs` is fully green. CI at `ec369ec` runs this crate's tests on Windows and full workspace on Linux/macOS — green.
- **`edda-ledger` (lib): 1 failed / 231 passed.** `paths::find_root_non_git_no_edda_returns_none` — `find_root` from a temp dir walks up and finds this workstation's real `~/.edda`. `paths.rs` is untouched by the PR; `edda-ledger` is in the CI Windows subset and green at `ec369ec`.

The GH-705 surface itself (edda-cli claim/bridge + bridge-claude peers) has **zero** test failures in the workspace run.

### Exact-head CI

CI for `59df3294116259e7af469bafd8363a1a2c69d34e` had **not yet reported** at posting time (run list shows only the `ec369ec` run, `33666478699`, success). It is the authoritative workspace gate for this SHA once it lands; the failures above do not exist there (same-population controls pass in CI at base).

Context: the previous lane started this L1 but was terminated (`SCHED_S_TASK_TERMINATED`) before completion; this is a fresh complete run, not a completion of a partial one. No merge action taken — merge authority rests with the operator.

fagemx:
## Review Response: Round 1 (pinned to `567bdfa1cced592617af711cdf150ed2cefc9984`)

Rebased onto `main` (`4075ab5`), resolving conflicts with #723 (subject matching) and #719 (e2e_repo hermeticity):

- **[P1] bare-CLI claims & heartbeat decay:**
  - Bare CLI claims (`cli-*`) are separated into `unjudgeable_claims` and counted toward exit 1 while on the board, fail-closed.
  - The claim verb no longer writes a decaying heartbeat.
- **[P1] Replacement & idempotency:**
  - Re-claiming and narrowing scope succeed without ambiguity refusal (`e2e_bare_cli_reclaim_is_idempotent`, `e2e_bare_cli_claim_can_narrow_scope`).
  - Identity refusal now lists live session IDs to copy into `--session`.
- **[P1 §5.5] Board write verification:**
  - `cmd_bridge::claim` verifies the board fold records the new claim before returning success (`e2e_lost_board_write_is_not_reported_as_success`).
- **Full L1 Gate receipt pinned to `567bdfa1cced592617af711cdf150ed2cefc9984`:**
  - `cargo fmt --all --check` → OK (exit 0)
  - `cargo clippy -p edda -p edda-bridge-claude --all-targets -- -D warnings` → 0 warnings (exit 0)
  - `cargo test -p edda` → 507 passed; 0 failed (exit 0)
  - `cargo test -p edda-bridge-claude -- --test-threads=4` → 634 passed; 0 failed (exit 0)
  - Isolated empty home state: identically green (507 passed, 634 passed).

fagemx:
## Review Response: Round 2 (pinned to `eb1cca045ad31fdb6d1d1925ee1b78e5f41271e9`)

Addressing Round 2 review findings:

- **[P1-1] Lost write verification on re-claim:**
  Updated `cmd_bridge::claim` so that board fold verification explicitly distinguishes the newly written entry from any `replaced` entry:
  ```rust
  let current = edda_bridge_claude::peers::compute_board_state(&project_id)
      .claims
      .into_iter()
      .find(|c| c.session_id == session_id);
  let recorded = current.as_ref().is_some_and(|c| {
      c.label == label
          && c.paths == paths
          && c.subject.as_deref() == subject
          && replaced.as_ref().map(|r| &r.ts) != Some(&c.ts)
  });
  ```
  Added pinned test `e2e_lost_board_write_on_reclaim_is_not_reported_as_success` in `cmd_claim.rs` simulating a write failure on re-claim (narrowing/moving scope) and asserting refusal without printing `Re-claimed scope`.

- **[P2-1] Unreachable branch in `describe_live_sessions`:**
  Replaced with single `fresh_sessions(project_id)` evaluation in `resolve_session_id`, passing `&[PeerSummary]` directly to `format_live_sessions`.

- **Gate receipt (Pinned to `eb1cca045ad31fdb6d1d1925ee1b78e5f41271e9`):**
  - Workspace lint: `cargo clippy --workspace --all-targets -- -D warnings` → 0 warnings (exit 0)
  - Code format: `cargo fmt --all --check` → OK (exit 0)
  - Focused crate gates:
    - `cargo test -p edda` → **508 passed; 0 failed** (exit 0)
    - `cargo test -p edda-bridge-claude -- --test-threads=4` → **635 passed; 0 failed** (exit 0)

fagemx:
> **Round 2 review** — independent review (\scripts/review-pr.sh\): transport \dda dispatch --agent claude\, model \claude-opus-4-6-preview\, reviewer_session \9bcdb4dd-c191-59ea-b766-bec6e4514581\, read-only, detached worktree at \567bdfa1cced592617af711cdf150ed2cefc9984\. Reviewed head SHA: \567bdfa1cced592617af711cdf150ed2cefc9984\.

## Code Review: Round 2 — PR #717 @ 567bdfa1cced592617af711cdf150ed2cefc9984

- model_observed: unverified
- reviewer_session: 9bcdb4dd-c191-59ea-b766-bec6e4514581 (this round: **new**)
- spec: review-spec-v1.2
- class: code-risk  (REVIEW.md classes: code-plain, docs — `crates/edda-cli/src/skills/coord-handoff.md` **up-classed** to `skills`, same reason as round 1: it is a shipped skill an agent executes, though the router's `skills/*` glob is repo-root-anchored and does not match that path)
- escalations: D5 (wording/structure — `[判斷]`, not adjudicated here)
- cost: ~11 tool calls, ~6 min, read-only; no cargo run (brief constraint), no build lane — `n/a`

### IN SCOPE

Delta round. Round 1's verdict is comment `5518366089`, pinned to `ec369ec46aef4c5c7b8dfa7987157297ad263fc2`. The branch was **rebased** onto `main` (`4075ab5`) and squashed to one commit, so `git diff ec369ec..567bdfa` carries 85 files of unrelated `main` progress; the PR's own change set is `git diff 4075ab5..567bdfa` — exactly the 6 allowed files, 578(+)/31(−). That is the target audited here, plus a RAN/READ confirmation of each round-1 blocking finding.

Changed behaviour this round: (1) `claim()` no longer writes a heartbeat for the minted session and instead re-reads the board fold to verify the claim landed before printing success (`cmd_bridge.rs:764-798`); (2) `claim check` separates non-live `cli-*` claims into `CheckReport.unjudgeable_claims` and counts them toward exit 1 (`cmd_claim.rs:80,133-153,167-174,202-214,233-239,247-249`); (3) `intersecting_claims` extracted so unjudgeable claims go through the identical sound intersection decision including its exit-2 refusal (`cmd_claim.rs:429`); (4) the identity refusal now names the live session ids (`cmd_bridge.rs:1317-1322,1334-1350`); (5) `infer_session_id` uses `liveness_from_heartbeat` (`discovery.rs:188-198`); (6) `coord-handoff.md:134-150` re-pinned with a new doctest block.

Direct consumers audited: `exit_code_for` (`cmd_claim.rs:233`), the human report (`cmd_claim.rs:181-217`), the `--json` report, `resolve_session_id` tier 3/4 (`cmd_bridge.rs:1294-1327`), `compute_board_state`'s per-session fold (`peers/board.rs:32-58`), `append_coord_event` (`peers/heartbeat.rs:263-272`).

Acceptance ceiling: #705 doneWhen — (a) a `cli-*` claim's existence machine-visible; (b) no relaxation of the liveness criterion; (c) FAIL-first regression test distinguishing genuinely clear from an unjudgeable occupation.

Current-base integration: head is rebased on `origin/main` = `4075ab5`; no conflicting surface.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 717 --name-only` → exactly the 6 files named in the brief's allowed surface; the rebase touched no other file *in the PR's change set* (`git diff 4075ab5..567bdfa --stat` = 6 files) |
| U2 | any | P1 | PASS | closing-keyword grep empty on the PR body (exit 1) and on `origin/main..567bdfa` commit messages |
| U3 | any | P1 | PASS | `Issue: #705` |
| U4 | any | P1 | PASS | offender grep empty; 1 commit, `fix(edda-cli): make the claim machine gate see bare CLI claims; …` |
| U6 | any | P0 | PASS (not red) | `gh pr checks 717` at exact head: run `33739106963` `head_sha=567bdfa1cced…` — Format, Detect code changes, Clippy (ubuntu/macos/windows) **pass**; Test (ubuntu/macos/windows) **pending** at review time. CI is in flight, not deterministically red, so no P0. Receipt-labelling issue → P2-2 |
| U7 | any | P1 | PASS | whole scoped delta audit finished before posting; the one blocker is batched below and is fix-caused |
| D1 | docs | P1 | PASS | the only backticked commands the diff adds: `edda claim --help` → exit 0; `edda peers --help` → exit 0 |
| D2 | docs | P1 | N.A.(the delta asserts no ledger decision value or status) | — |
| D3 | docs | P1 | PASS | §5.1 path/link enumerator over the added lines → 0 candidates, 0 `MISSING` |
| D4 | docs | P0 | PASS | authority grep (`gh pr merge`/`--delete-branch`/`--admin`/`--no-verify`/force-push) empty over the 6 PR files. (The 4 hits in the raw `ec369ec..567bdfa` range are `main`'s own commits, outside this PR's change set) |
| D5 | docs | `[判斷]` | 需升級 | wording/scoping of the `coord-handoff.md:134-150` rewrite and the new doctest block — see escalations |
| S1 | skills | P0 | PASS | `merge`/`self-review`/`skip review` grep over the `coord-handoff.md` diff empty |
| S2 | skills | P1 | PASS | both doctest commands probe 0 (D1); the asserted output `! bare-CLI claim, liveness cannot be judged` (`coord-handoff.md:144`) matches the shipped string at `cmd_claim.rs:203-206`; `unjudgeable_claims` matches `cmd_claim.rs:80` |
| S3 | skills | P1 | PASS | the added text describes this binary's behaviour; it restates no rule owned by `REVIEW.md` / `.claude/CLAUDE.md` |
| C1 | code-plain | P0 | PASS | doneWhen (a) met by the issue's **second** option, not the decayed first: a `cli-*` claim is classified `unjudgeable` and counted toward exit 1 for as long as it stands on the board (`cmd_claim.rs:133-153,233-239`) — no freshness window. (b) met: `liveness_from_heartbeat` is untouched and `is_bare_cli_session` is only reached **after** `liveness.is_live()` (`cmd_claim.rs:135-139`), so the criterion is never loosened. (c) met: 4 tests pin the classification (see C2) |
| C2 | code-plain | P0 | PASS | `fix:` subject; `git diff origin/main..567bdfa -- 'crates/**' \| grep -cE '^\+.*(#\[test\]\|fn test_)'` → **11**. Regression tests: `e2e_aged_bare_cli_claim_still_conflicts` (600 s backdated → exit 1), `e2e_never_heartbeated_bare_cli_claim_conflicts`, `e2e_aged_bare_cli_claim_is_classified_in_json` (`cmd_claim.rs:1899`), `e2e_bare_cli_claim_is_visible_to_claim_check`, the three re-claim tests, `e2e_lost_board_write_is_not_reported_as_success`, `resolve_session_id_refusal_names_live_sessions` (`cmd_bridge.rs:2292`), `infer_session_counts_parented_subagent_by_shared_criterion` (`peers/tests.rs`). FAIL-first: READ from the round-1 response (red at `65a3f26`) and from round 1's own A/B binary probes, which are what made the re-claim tests necessary |
| C3 | code-plain | P1 | PASS | 23 candidates on the PR's change set, all inside `#[cfg(test)]` — `peers/tests.rs` is included as `#[cfg(test)] #[path="tests.rs"] mod tests`, and every `cmd_claim.rs` candidate (`:454`–`:773` of the diff) is inside `mod tests` |
| C4 | code-plain | P1 | PASS | `unsafe` / `clippy::all` / crate-level `#![allow]` grep empty |
| C5 | code-plain | P1 | PASS | selector prints nothing — both touched crates (`edda-cli` → `edda`, `edda-bridge-claude`) are in the CI Windows subset, and Clippy (windows-latest) passed at exact head |
| R1 | code-risk | P0 | N.A.(not routed — no path in the diff is on the §3.1 risk surface; canonical `class:` reads `code-risk` only because `code-plain` maps there) | destructive-command grep over the 6 files empty anyway |
| R2 | code-risk | `[判斷]` | N.A.(not routed; no shell in the diff) | — |
| R3 | code-risk | P0 | N.A.(not routed; no `.sh`/`.bash` in the change set) | — |
| R4 | code-risk | P1 | N.A.(not routed) | — |
| R5 | code-risk | P1 | N.A.(not routed) | — |
| §5.5 | any | P1 | **FAIL** | the new lost-write guard is blind on the re-claim path — finding P1-1 |

### Wiring

|---|---|---|---|---|
| `CheckReport.unjudgeable_claims: Vec<ClaimConflict>` | struct `cmd_claim.rs:80`; populated `cmd_claim.rs:133-153,167-174` | `exit_code_for` `cmd_claim.rs:233-239` (→ exit 1); human report `cmd_claim.rs:202-214`; `--json` (always serialized — no `skip_serializing_if`) | n/a — pure classification, no write. Fail-closed: `is_bare_cli_session` only reached after `is_live()` is false | board → classify → exit code, proven end-to-end through the real binary by `e2e_aged_bare_cli_claim_still_conflicts` and `e2e_aged_bare_cli_claim_is_classified_in_json` (`cmd_claim.rs:1899`) |
| `StaleClaim.age_secs` always serialized (`null` = no heartbeat file) | `cmd_claim.rs:56-61`; populated `cmd_claim.rs:140-146` | the `--json` machine contract; in-repo `e2e_never_heartbeated_stale_claim_is_explicit_in_json`. `edda claim check --json` is **not** among COMPATIBILITY.md §2's enumerated stable contracts (`dispatch`, `verify`, `ask`, edda-mcp, `status`), so no golden fixture is owed | n/a | report → stdout → parsed and asserted (`age_secs` present **and** null) |
| board-fold verification + `anyhow::bail!` in `claim()` | `cmd_bridge.rs:790-798` — re-read `compute_board_state`, `any(\|c\| c.session_id == session_id)` | the verb's own exit code and stderr; test `e2e_lost_board_write_is_not_reported_as_success` | **incomplete** — the predicate is already satisfied by the *previous* claim the fold keeps for the same session, so a lost write on any re-claim still exits 0 with `Re-claimed scope:` → **P1-1** | verb → store → read-back, but only on the first-claim branch |
| `describe_live_sessions() -> String` (private) | `cmd_bridge.rs:1345-1356` | the identity refusal message `cmd_bridge.rs:1317-1322` | n/a (formatting) | error path → stderr, asserted by `resolve_session_id_refusal_names_live_sessions` (`cmd_bridge.rs:2292`) |
| `fresh_sessions() -> Vec<PeerSummary>` (private) | `cmd_bridge.rs:1334-1341` | `has_live_sessions` `cmd_bridge.rs:1329`; `describe_live_sessions` `cmd_bridge.rs:1346` | n/a | still an inline `age_secs <= stale_secs()`, not `peers::liveness` — routed as FOLLOW-UP below |
| new `edda-doctest` block in the shipped skill | `coord-handoff.md:140-146` | the `skill_doctest.rs` harness (crate `edda`) | harness failure is a test failure | skill text → doctest → `cargo test -p edda` green in the round-2 receipt (507 passed / 0 failed) |

### Findings

- **[P1] The new lost-write guard cannot detect a lost board write on a re-claim — the exact false success it was added to remove survives on the replacement path.** `claim()` verifies the write landed with `compute_board_state(&project_id).claims.into_iter().any(|c| c.session_id == session_id)` (`cmd_bridge.rs:790-793`). But `compute_board_state` folds claims into a `HashMap` keyed by `session_id`, one entry per session (`crates/edda-bridge-claude/src/peers/board.rs:32,50-58`), and a re-claim is *by design* a replacement against this session's own surviving entry (`cmd_bridge.rs:764-768`, and the contract at `cmd_bridge.rs:745-753`). So whenever `replaced.is_some()` the predicate is already true **before** the append, and a dropped append leaves `recorded == true`. `append_coord_event` cannot tell the caller either — it swallows the error into `record_dropped_write` and returns `()` (`peers/heartbeat.rs:263-272`), and `write_claim_with_subject` returns `()` (`peers/heartbeat.rs:289-310`).

  Failure scenario: an agent claims `auth` with `src/auth/*` (lands); later widens to `src/auth/* src/api/*` while the store append fails — the same failure class the round-1 finding named (read-only or full store, `.lock` contention) and that the new test simulates. `claim_disclosure` prints `Re-claimed scope: auth (paths added)` and the verb exits 0, while the board still holds only `src/auth/*`, so `edda claim check src/api/x.rs` reads clear over a surface the agent believes it holds. `e2e_lost_board_write_is_not_reported_as_success` (`cmd_claim.rs`) cannot catch this: it blocks the board path *before* any claim exists, so it only exercises the first-claim branch. §5.5 fixed adjudication — an error swallowed on a coordination path is P1. This is fix-caused and previously unobservable (the guard is new this round). A predicate that distinguishes the new entry from the old one — e.g. requiring the re-read entry's `paths`/`label` to equal what was just written, or its `ts` to differ from `replaced` — closes it.

- **[P2] `describe_live_sessions`'s empty branch is unreachable in production.** `"No live sessions were found; retry the command."` (`cmd_bridge.rs:1347-1349`) is only ever called from `cmd_bridge.rs:1321`, which is guarded by `has_live_sessions(project_id)` returning true two lines earlier (`cmd_bridge.rs:1317`) — both go through the same `fresh_sessions` scan. Only a TOCTOU race between the two scans can print it, and then the text tells the operator to retry a command whose refusal has just vanished. Either drop the branch or make the refusal reuse the single `fresh_sessions` result it already computed.

- **[P2] The round-2 gate receipt is labelled "Full L1 Gate receipt" but lists crate-scoped gates.** The comment pinned to `567bdfa` reports `cargo clippy -p edda -p edda-bridge-claude` and `cargo test -p edda` / `-p edda-bridge-claude` — that is the ladder's **L0** row, not L1's `cargo clippy --workspace --all-targets` + `cargo test --workspace`. The commands themselves are stated accurately, so a reader can see what ran; only the label overclaims. Exact-head CI supplies the workspace coverage (Clippy is workspace-wide on all three OSes and passed), but its `Test` jobs were still pending at review time, so no complete workspace test result exists for this SHA yet. Next round: either run the real L1 set or cite the completed CI run instead of labelling an L0 run L1.

**Round-1 blocking findings — resolution confirmed:**

- **P0-1 (bare claim machine-visible for only 120 s) — RESOLVED at the root.** `claim()` writes no heartbeat: `grep -n write_heartbeat_minimal crates/edda-cli/src/cmd_bridge.rs` returns `:1539` (the `heartbeat` verb) and test lines only — nothing in `claim()` (`cmd_bridge.rs:756-800`). Visibility now rides the board entry, which never expires, via the `unjudgeable` classification (`cmd_claim.rs:133-153`) counted by `exit_code_for` (`cmd_claim.rs:233-239`). Pinned by `e2e_aged_bare_cli_claim_still_conflicts` (600 s backdated heartbeat → exit 1) and `e2e_never_heartbeated_bare_cli_claim_conflicts`.
- **P1-1 (bare-CLI claim path became single-shot) — RESOLVED by removing the cause.** With no heartbeat written, `resolve_session_id` never sees the claim's own residue as a live session (`cmd_bridge.rs:1317`), so tier 4 mints `cli-<label>` as documented. Pinned by `e2e_bare_cli_reclaim_is_idempotent`, `e2e_bare_cli_claim_can_narrow_scope`, `e2e_second_agent_claims_disjoint_scope`. The named consequence is handled: the refusal now lists ids (`cmd_bridge.rs:1345-1356`, asserted at `cmd_bridge.rs:2292`), and `coord-handoff.md:134-150` is re-pinned.
- **P1-2 (swallowed write on the coordination path) — PARTIALLY RESOLVED.** The claim path no longer travels through `write_heartbeat_minimal`, and a first claim whose board append is lost now fails loudly. The replacement path is still uncovered → P1-1 above.
- **P2-1 (`peers::infer_session_id` has no production consumer)** — acknowledged in the response and routed to follow-up; unchanged this round.

### FOLLOW-UP ISSUE

- `has_live_sessions` / `fresh_sessions` (`cmd_bridge.rs:1329-1341`) still decides liveness with an inline `session.age_secs <= stale_secs()` rather than `peers::liveness::liveness_from_heartbeat`, so it does not carry the 15x parented-sub-agent multiplier this PR just gave `infer_session_id`. Consequence: a 20-minute-old parented sub-agent heartbeat is Live for `edda peers`, `edda claim check` and now `infer_session_id`, but invisible to the identity refusal — the remaining instance of the "second, disagreeing notion of dead" that `liveness.rs:1-12` forbids. Not a regression (the inline check is pre-existing, and `infer_session_id` has no production consumer, so nothing observable diverges today) and not required by #705's doneWhen. Merge with round 1's follow-up on `crates/edda-cli/src/cmd_phase.rs:206,276`.

### RAN vs READ

RAN — this worktree, `C:/Users/fagem/.edda/fleet/wt-review-pr717`, detached at `567bdfa`, Windows 11 / Git Bash, read-only per the brief (no cargo, no lane):

- `gh pr view/diff/checks 717`, `gh api .../actions/runs?head_sha=567bdfa…` → run `33739106963`, `in_progress`
- `gh pr view 717 --json comments` — read round 1's posted verdict (comment `5518366089` = node `IC_kwDORT1v6c8AAAABSOuViQ`), reviewed SHA `ec369ec46aef4c5c7b8dfa7987157297ad263fc2`
- `git diff --stat ec369ec..567bdfa` → 85 files (rebase noise) vs `git diff --stat 4075ab5..567bdfa` → the 6 PR files; the rule greps were run on the latter, which is conservative (it covers the delta plus everything carried over)
- the U2, U4, D1, D3, D4, S1, C2, C3, C4, C5, R1 enumerators — outputs in the Rules table
- `edda claim --help` → 0, `edda peers --help` → 0
- source reads establishing P1-1: `cmd_bridge.rs:756-800`, `peers/board.rs:12-58`, `peers/heartbeat.rs:263-310`, `cmd_claim.rs:120-250`
- `git ls-files | grep -i compat` → `COMPATIBILITY.md`; §2's stable `--json` contracts are `dispatch`, `verify`, `ask`, edda-mcp, `status` — `claim check` is not among them, so the two new JSON surfaces owe no golden fixture

READ — not re-run (`ladder` L2: no receipt gap, no red CI, no grounds to distrust):

- the round-2 gate receipt pinned to `567bdfa` in `Review Response: Round 1 (pinned to 567bdfa…)`: `cargo fmt --all --check` exit 0; `cargo clippy -p edda -p edda-bridge-claude --all-targets -- -D warnings` 0 warnings; `cargo test -p edda` 507 passed / 0 failed; `cargo test -p edda-bridge-claude -- --test-threads=4` 634 passed / 0 failed; identically green under an isolated empty home. It is crate-scoped, not the L1 workspace set — see P2-2
- exact-head CI run `33739106963` @ `567bdfa1cced…`: Format, Detect code changes, Clippy (ubuntu/macos/windows) pass; **Test (ubuntu/macos/windows) still pending at posting time**. What it covers once green: compile, lint and test on all three OSes. What it does not cover: the lost-write-on-re-claim path in P1-1, which no test in the tree exercises — which is why CI can be green while P1-1 holds
- round 1's binary A/B probes (comment `5518366089`), which are the FAIL-first evidence for the re-claim tests added this round

### Verdict

Changes Requested, P0=0, P1=1 — round 1's P0 and first P1 are resolved at the root and well pinned, but the new board-fold verification is satisfied by the session's *previous* claim entry, so a lost write on any re-claim still prints `Re-claimed scope:` and exits 0 over a surface the machine gate reads as clear.

fagemx:
> **Round 3 review** — independent review (\scripts/review-pr.sh\): transport \dda dispatch --agent claude\, model \claude-opus-5\, reviewer_session \9bcdb4dd-c191-59ea-b766-bec6e4514581\, read-only, detached worktree at \b1cca045ad31fdb6d1d1925ee1b78e5f41271e9\. Reviewed head SHA: \b1cca045ad31fdb6d1d1925ee1b78e5f41271e9\.

## Code Review: Round 3 — PR #717 @ eb1cca045ad31fdb6d1d1925ee1b78e5f41271e9

- model_observed: unverified
- reviewer_session: 9bcdb4dd-c191-59ea-b766-bec6e4514581 (this round: **resume**)
- spec: review-spec-v1.2
- class: code-risk  (REVIEW.md classes: code-plain, docs — `coord-handoff.md` remains up-classed to `skills` for the PR as a whole, same reason as rounds 1–2; it is untouched by this delta)
- escalations: D5 (wording/structure — `[判斷]`, carried from Round 2 and still not adjudicated by a qualified engine)
- cost: ~8 tool calls, ~5 min, read-only; no cargo run (brief constraint), no build lane — `n/a`

### IN SCOPE

Delta round against my Round 2 verdict (comment `5523786053`, pinned to `567bdfa1cced592617af711cdf150ed2cefc9984`). Base unchanged (`origin/main` = `4075ab5`), so `git diff 567bdfa..eb1cca0` is clean of rebase noise: **2 files, +66/−10**.

Changed behaviour: (1) the board-fold verification in `claim()` is strengthened from a session-id existence probe to an identity probe on the *new* entry — `find()` plus `label`/`paths`/`subject` equality plus a timestamp that must advance past the `replaced` entry (`cmd_bridge.rs:796-805`); (2) `describe_live_sessions(project_id)` becomes `format_live_sessions(&[PeerSummary])`, with `fresh_sessions` evaluated once in `resolve_session_id` and the slice passed in, deleting the unreachable empty branch (`cmd_bridge.rs:1329-1334,1358-1366`); (3) one new regression test, `e2e_lost_board_write_on_reclaim_is_not_reported_as_success` (`cmd_claim.rs:2033-2078`).

Direct consumers audited: `claim_disclosure` (`cmd_bridge.rs:702-744`), `write_claim_with_subject` → `append_coord_event` (`peers/heartbeat.rs:263-310`), `compute_board_state`'s per-session fold (`peers/board.rs:32-58`), `parse::now_rfc3339` (`parse.rs:69-73`, the ts source the new predicate compares), `has_live_sessions` (still live at `cmd_bridge.rs:907`, so the rename introduced no dead code).

Acceptance ceiling: unchanged — #705 doneWhen was already met at `567bdfa` (C1 PASS in Round 2) and this delta touches none of the three items; it repairs the Round 2 P1 in the verification wrapper around the claim write.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 717 --name-only` → the same 6 allowed files; the delta touches only 2 of them (`cmd_bridge.rs`, `cmd_claim.rs`) |
| U2 | any | P1 | PASS | closing-keyword grep empty on the PR body and on `origin/main..eb1cca0` commit messages |
| U3 | any | P1 | PASS | `Issue: #705` |
| U4 | any | P1 | PASS | offender grep empty; still 1 commit, `fix(edda-cli): make the claim machine gate see bare CLI claims; …` |
| U6 | any | P0 | PASS (not red) | `gh pr checks 717` at exact head: run `33740384304` @ `eb1cca045ad3…` — Format, Detect code changes, Clippy (ubuntu/macos/windows) **pass**; Test (ubuntu/macos/windows) **pending at posting time**. Not deterministically red, so no P0. Test coverage for this SHA rests on the implementer's receipt (READ below); see the caveat in RAN vs READ |
| U7 | any | P1 | PASS | whole scoped delta audit finished before posting; zero blockers to batch |
| D1 | docs | P1 | PASS | the delta adds no backticked CLI invocation (`grep -oE '`(edda\|gh\|git\|cargo\|pi\|claude) …'` on `+` lines → empty) |
| D2 | docs | P1 | N.A.(the delta asserts no ledger decision value or status) | — |
| D3 | docs | P1 | PASS | the delta adds no repo-anchored path or link; enumerator → 0 candidates |
| D4 | docs | P0 | PASS | authority grep (`gh pr merge`/`--delete-branch`/`--admin`/`--no-verify`/force-push) empty over the delta |
| D5 | docs | `[判斷]` | 需升級 (carried) | no prose changed this round (`git diff 567bdfa..eb1cca0 -- 'crates/edda-cli/src/skills/*'` → 0 lines), but Round 2's escalation on `coord-handoff.md:134-150` is still open, so it is carried rather than silently dropped (§6.2) |
| S1 | skills | P0 | PASS | the skill file is unchanged in this delta — 0 diff lines; nothing new to check |
| S2 | skills | P1 | PASS | no new skill claim added; the Round 2 measurements (`unjudgeable_claims` at `cmd_claim.rs:80`, the printed string at `cmd_claim.rs:203-206`) still hold — neither line is touched by this delta |
| S3 | skills | P1 | PASS | no skill text added |
| C1 | code-plain | P0 | PASS | doneWhen unchanged from Round 2's PASS: (a) the `unjudgeable` classification counted by `exit_code_for` (`cmd_claim.rs:233-239`), (b) `liveness_from_heartbeat` still untouched and `is_bare_cli_session` still gated behind `is_live()` (`cmd_claim.rs:135-139`), (c) the four pinning tests unchanged. This delta touches none of the three |
| C2 | code-plain | P0 | PASS | `fix:`; the delta adds **1** new test (`grep -cE '^\+.*(#\[test\]\|fn test_)'` → 1): `e2e_lost_board_write_on_reclaim_is_not_reported_as_success` (`cmd_claim.rs:2035`). FAIL-first is structural and exact: at `567bdfa` the predicate was `any(\|c\| c.session_id == session_id)`, satisfied by the surviving prior entry, so the verb printed `Re-claimed scope:` and exited 0 — the test asserts `code != 0` and `!stdout.contains("Re-claimed scope")`, so it cannot pass at the previous SHA. A recorded red run is not posted; the construction is verifiable from the two SHAs |
| C3 | code-plain | P1 | PASS | 3 candidates in the delta (`cmd_claim.rs:2039,2048,2050` — `tempfile::tempdir().expect`, `metadata().expect`, `set_permissions().expect`), all inside `mod tests` (the fn opens at `cmd_claim.rs:2035`, well inside the `#[cfg(test)] mod tests` block) |
| C4 | code-plain | P1 | PASS | grep for `unsafe {` / `#[allow(clippy::all)]` / `#![allow` → empty. The delta's one allow is `#[allow(clippy::permissions_set_readonly_false)]` (`cmd_claim.rs:2034`) — a **targeted** item-level lint allow, explicitly permitted by C4, and needed because the test must restore write permission so tempdir cleanup succeeds on Windows |
| C5 | code-plain | P1 | PASS | selector prints nothing — both touched crates (`edda-cli` → `edda`, `edda-bridge-claude`) are inside the CI Windows subset, and Clippy (windows-latest) passed at exact head. The implementer additionally ran both crates' tests on Windows at this SHA (READ below) |
| R1 | code-risk | P0 | N.A.(not routed — no path in the delta is on the §3.1 risk surface; canonical `class:` reads `code-risk` only because `code-plain` maps there) | destructive-command grep over the delta empty anyway |
| R2 | code-risk | `[判斷]` | N.A.(not routed; the delta contains no shell) | — |
| R3 | code-risk | P0 | N.A.(not routed; no `.sh`/`.bash` in the change set) | — |
| R4 | code-risk | P1 | N.A.(not routed) | — |
| R5 | code-risk | P1 | N.A.(not routed) | — |
| §5.5 | any | P1 | PASS | the Round 2 FAIL is closed; see the Wiring table — the guard's failure signal is now complete across all three lost-write shapes |

### Wiring

|---|---|---|---|---|
| strengthened `recorded` predicate in `claim()` (changed surface, not new) | `cmd_bridge.rs:796-805` — `find(\|c\| c.session_id == session_id)` then `c.label == label && c.paths == paths && c.subject.as_deref() == subject && replaced.ts != c.ts` | the verb's own exit code and stderr (`anyhow::bail!`, `cmd_bridge.rs:806-810`) | **complete** — all three lost-write shapes now bail: first claim (`current` is `None`), narrow/widen (`paths` differ), identical re-claim (all fields match but the fold still returns the *replaced* entry, so `replaced.ts == c.ts`). Errs closed: a false trip refuses a claim that landed, never the reverse | verb → store → read-back, proven through the real binary by `e2e_lost_board_write_is_not_reported_as_success` (`cmd_claim.rs:2003`, first-claim shape) and `e2e_lost_board_write_on_reclaim_is_not_reported_as_success` (`cmd_claim.rs:2035`, replacement shape) |
| `format_live_sessions(live: &[PeerSummary]) -> String` (private; replaces `describe_live_sessions(project_id)`) | `cmd_bridge.rs:1358-1366` | the identity refusal at `cmd_bridge.rs:1329-1335`, which now computes `fresh_sessions` **once** and passes the slice | n/a (formatting); the unreachable `"No live sessions were found; retry the command."` branch is deleted, so the refusal can no longer print advice for a state it cannot be in | error path → stderr, asserted by `resolve_session_id_refusal_names_live_sessions` (`cmd_bridge.rs:2301`) |
| — | — | — | — | no other new surface in this delta: no new `pub` item, CLI flag, config key, event field, or written file |

### Findings

- **[P2] The new `recorded` predicate's last conjunct assumes two board events for the same session never share a timestamp.** `replaced.as_ref().map(\|r\| &r.ts) != Some(&c.ts)` (`cmd_bridge.rs:804`) compares RFC3339 strings produced by `parse::now_rfc3339` → `time::OffsetDateTime::now_utc().format(Rfc3339)` (`crates/edda-bridge-claude/src/parse.rs:69-73`), which carries sub-second precision from `SystemTime::now()`, so a collision is vanishingly unlikely and cannot happen across two process invocations in practice. If it ever did, the failure is **closed, not open**: `edda claim` would exit non-zero with `was not recorded on the coordination board` over a claim that actually landed, leaving a board entry the agent believes it does not hold. Cheap hardening if it is ever worth it: compare against `replaced`'s whole entry, or have `write_claim_with_subject` return the event it wrote. Not blocking — the direction of the error is the safe one, which is exactly what the guard exists to guarantee.

**Round 2 findings — resolution RAN/READ-confirmed:**

- **P1-1 (lost write on re-claim reported as success) — RESOLVED.** The predicate no longer asks "is there *an* entry for this session" but "is the entry the one I just wrote": `cmd_bridge.rs:796-805`. I walked all three branches against the fold's actual shape (`peers/board.rs:32,50-58` — `claims.insert(session_id, …)`, one entry per session, last line wins) and the swallowing writer (`peers/heartbeat.rs:263-272,289-310` — `append_coord_event` records into `record_dropped_write` and returns `()`). Narrow/widen is caught by `c.paths == paths`; the identical re-claim, which the field-equality check alone would *not* catch, is caught by the ts conjunct. The new test (`cmd_claim.rs:2035-2078`) makes the board file read-only after a successful first claim and then narrows the scope, asserting exit != 0, no `Re-claimed scope` on stdout, and a stderr that names the lost write. READ: `cargo test -p edda` at this SHA → **508 passed / 0 failed**, one more than the 507 at `567bdfa`, so the new test and the three re-claim tests (`e2e_bare_cli_reclaim_is_idempotent` `cmd_claim.rs:1915`, `e2e_bare_cli_claim_can_narrow_scope`, `e2e_second_agent_claims_disjoint_scope`) are green together — the strengthened guard did not turn the documented replacement contract into a refusal.
- **P2-1 (unreachable empty branch in `describe_live_sessions`) — RESOLVED, and the double scan with it.** `resolve_session_id` now evaluates `fresh_sessions` once (`cmd_bridge.rs:1329`) and passes the slice (`:1334`); `format_live_sessions` (`:1358`) has no empty branch. `has_live_sessions` survives because it is still used at `cmd_bridge.rs:907`, which is consistent with workspace Clippy being green on all three OSes at exact head (an orphaned `fn` would have tripped `dead_code` under `-D warnings`).
- **P2-2 (receipt labelled "Full L1" over crate-scoped gates) — RESOLVED in substance.** The Round 2 response's receipt for this SHA labels its gates accurately: `cargo clippy --workspace --all-targets -- -D warnings` and `cargo fmt --all --check` are named as workspace gates, and the test gates are explicitly headed "Focused crate gates". No L0 run is presented as L1.

### FOLLOW-UP ISSUE

- Unchanged and still open from Round 2: `has_live_sessions` / `fresh_sessions` (`cmd_bridge.rs:1342-1354`) still decide liveness with an inline `session.age_secs <= stale_secs()` rather than `peers::liveness::liveness_from_heartbeat`, so they do not carry the 15x parented-sub-agent multiplier this PR gave `infer_session_id`. Not a regression (pre-existing check; `infer_session_id` has no production consumer) and not required by #705's doneWhen. Merge with Round 1's follow-up on `crates/edda-cli/src/cmd_phase.rs:206,276` — together they are the two remaining "second, disagreeing notion of dead" sites that `liveness.rs:1-12` forbids.

### RAN vs READ

RAN — this worktree, `C:/Users/fagem/.edda/fleet/wt-review-pr717`, detached at `eb1cca0`, Windows 11 / Git Bash, read-only per the brief (no cargo, no lane):

- `sha256sum .edda-review-spec.md` → `beef0866…`, 727 lines, `review-spec-v1.2` — the same spec copy I ran in Round 2, so the rules did not move under the PR
- `git rev-parse origin/main` → `4075ab5` (unchanged base, so the delta is the PR's own work with no rebase noise); `git diff --stat 567bdfa..eb1cca0` → 2 files, +66/−10
- `gh pr view/diff/checks 717`; `gh api .../actions/runs?head_sha=eb1cca045ad3…` → run `33740384304`, `in_progress`
- `gh pr view 717 --json comments` → the implementer's `Review Response: Round 2` pinned to `eb1cca045ad3…` exists and answers both blocking findings by number (`loop` item 4 satisfied)
- the U2, U4, D1, D3, D4, S1, C2, C3, C4, C5, R1 enumerators on the delta range — outputs in the Rules table
- source reads establishing the P1-1 closure: `cmd_bridge.rs:764-810`, `cmd_bridge.rs:1326-1366`, `cmd_bridge.rs:907`, `peers/board.rs:12-58`, `peers/heartbeat.rs:263-310`, `parse.rs:69-73`, `cmd_claim.rs:2003-2078`

READ — not re-run (`ladder` L2: a receipt exists for this exact SHA, CI is not red, no grounds to distrust):

- the gate receipt in `Review Response: Round 2` pinned to `eb1cca045ad3…`: `cargo clippy --workspace --all-targets -- -D warnings` → 0 warnings; `cargo fmt --all --check` → exit 0; focused crate gates `cargo test -p edda` → 508 passed / 0 failed and `cargo test -p edda-bridge-claude -- --test-threads=4` → 635 passed / 0 failed. Those two crates are exactly the ones this PR changes, and they were run **on Windows**, which is a superset of what CI's Windows subset gives them
- exact-head CI run `33740384304` @ `eb1cca045ad3…`: Format, Detect code changes, Clippy (ubuntu/macos/windows) pass; **Test (ubuntu/macos/windows) had not reported at posting time**. So workspace-wide *lint* is confirmed at this head on all three OSes, while workspace-wide *test* for this head is not yet on record from CI — the test evidence for this SHA is the implementer's focused Windows run above, covering both changed crates. Untouched crates are unaffected by a 2-file delta confined to `edda-cli`
- Round 2's own evidence for why the P1 existed (comment `5523786053`), which is the FAIL-first ground for the new test

### Verdict

LGTM (P0=0, P1=0) — **provisional** per §8/§6.4: the `escalations:` field is non-empty (D5 carried), so this LGTM does not satisfy the merge gate until a qualified engine adjudicates it. Operator should also confirm run `33740384304`'s three `Test` jobs land green, since they were still pending when this round was judged. Merge remains the operator's action (GATE-01); adding `fleet:reviewed` is the only label action this verdict supports.

fagemx:
## Record: a second independent review track converged on the same P1

Posting for the audit trail, after the merge. A separate agent session reviewed `567bdfa` in a read-only detached worktree, dispatched by a different implementer session than the one that authored the fix, and reached:

`VERDICT: CHANGES REQUESTED (P0=0, P1=1)`

The single P1 is the one `eb1cca04` fixed, reached independently by the same evidence chain: `compute_board_state` folds claims into a `HashMap` keyed by `session_id` (`crates/edda-bridge-claude/src/peers/board.rs:20,52-57`), so `.any(|c| c.session_id == session_id)` at `cmd_bridge.rs:789-792` is already satisfied by the session's **own surviving entry** — `replaced.is_some()` ⇒ `recorded == true` regardless of whether the append landed. The guard was a tautology on every re-claim, and the pinning test only exercised the first-claim branch on a fresh store (`cmd_claim.rs:2007-2010`).

Two review tracks, different sessions, same SHA, same finding, same evidence. Recording it because that convergence is the useful signal — the spec is what produced it, not continuity of one reviewer.

### What this reviewer verified beyond the P1

- **The GH-705 P0 is killed, not relocated.** Classification is a pure function of the session id (`cmd_claim.rs:135-153`), never of time; a bare claim can never reach `stale_claims`. Every path where a standing bare claim could read clear was traced and all fail closed to `exit(2)`: unreadable board, damaged event type, malformed glob on either the live or the unjudgeable set.
- **`is_bare_cli_session` is sound in the direction that matters.** Tier 4 mints `cli-{label}` unconditionally (`cmd_bridge.rs:1326`), so no false negative is possible; a false positive requires a `cli-`-prefixed id supplied via `--session`/`EDDA_SESSION_ID`, and that is only reachable after `is_live()` is already false — an over-report, the safe direction.
- **The P0 regression test is real.** `write_heartbeat(…, 600)` against `stale_secs()` = 120, with no `parent_session_id` so no 15× multiplier applies → `Stale{600}`. Under the round-1 implementation that claim exited 0; here it exits 1. `e2e_aged_bare_cli_claim_is_classified_in_json` additionally asserts `conflicts` **and** `stale_claims` are both empty, which forecloses the "it passed as a live conflict" vacuity.
- **The `--json` shape change is not owed a fixture.** `StaleClaim.age_secs` becoming always-serialized changes `claim check --json`, but COMPATIBILITY.md:79-88 declares the stable set as `dispatch-verify-ask-status-mcp` and everything else unstable by default. Both changes are additive anyway.
- **Receipt reconciliation.** The response's `cargo test -p edda` → 507 and an independent measurement of 481 are both correct: `-p edda` builds 8 targets; the bin holds 481 and the 7 integration targets hold 26. 481 + 26 = 507.

### Two P2s survive into `main` and are now filed

Neither is addressed by `eb1cca04`: a non-intersecting bare-CLI claim vanishes from the report entirely (losing GH-617's death-visibility line for that class), and `fresh_sessions`/`has_live_sessions` still decide liveness inline without the 15× parented-sub-agent multiplier that `peers::liveness` applies — the same divergence as `cmd_phase.rs:206,276`. Filed together as #780.

Issue #552, opened 2026-09-01
fix(conductor): a gate timeout is recorded as a phase failure, Skip records completed work as skipped, and an unattended run has no recovery


fagemx:
## Context

Same run as the sibling silence issue: `cleanup-wave1`, three issues chained, `gate: verdict` on each implement phase. Two of the three gates timed out (7200s each). This issue is about what the conductor did *after* the timeout — the state it recorded and the exits it offered.

## 1. A gate timeout is recorded as a phase failure, but nothing failed

When the budget expires, the phase transitions to `Failed` and the event log records:

```
{"type":"phase_failed","phase_id":"i3-impl-532","attempt":1,"duration_ms":0,
 "error":"gate timed out: no verdict for \"cleanup-wave1/i3-impl-532\" (sha d5689ff...) within 7200s"}
```

At that moment the phase's work was complete: the agent had finished, committed `d5689ff`, and all three checks (`cargo fmt --all --check`, `cargo clippy -p edda-conductor --all-targets -- -D warnings`, `cargo test -p edda-conductor`) had passed — the phase cannot reach `AWAITING_VERDICT` otherwise, so the green result is machine-enforced, not claimed. Identical picture for `i2-impl-529` at `7b54792`.

What failed is the *review*, and only in the sense that it did not arrive in time. `PhaseStatus` has no way to say that. `Failed` is the closest available label and it is wrong in a way that matters: a reader of the state file or the event log cannot distinguish "the agent could not do the work" from "the work is done, checked, and waiting for a human who did not show up". Those two situations have opposite next actions.

The `duration_ms: 0` is a second symptom of the same gap — the phase ran for hours; zero is what you get when a gate timeout is squeezed into a shape built for a failed attempt.

## 2. `Skip` is the only exit, and it records something untrue

The interactive recovery (`prompt_blocked_action`, `sequential.rs:1252`) offers `[R] Retry  [S] Skip  [A] Abort  [Q] Quit`. For a timed-out gate:

- **Retry** re-dispatches a full agent turn and re-runs every check — minutes of compute and one more `attempts` increment — purely to arrive back at the same gate. The commit is already frozen; nothing about it needs redoing.
- **Abort** discards the remaining phases.
- **Quit** leaves the phase `Failed`, so the next resume presents the same prompt.
- **Skip** is the only exit that lets the plan proceed. It sets `PhaseStatus::Skipped` with `skip_reason: "manually skipped (interactive)"` (`sequential.rs:110-121`).

[...]

PR #718, opened 2026-09-02
fix(edda-ledger,conductor): verdict gate fails closed on a bad freshness bound and surfaces unreadable-ledger errors (GH-541)

### LINKED ISSUE #541: fix(conductor): the verdict gate's freshness rule fails open on a bad bound, and an unreadable ledger polls forever in silence
## Context

Routed from PR #539 round-1 review (verifier findings, non-blocking there). Both are ways the #519 verdict gate's safety **turns itself off silently** — and the thing it protects against is the infinite redispatch loop D6 was created to kill (measured at 176 cycles before the fix).

## 1. `latest_verdict_fresh` fails OPEN on an unparsable `not_before`

An unparsable `not_before` yields `bound = None`, which means **every stale verdict satisfies the gate** — exactly the loop-prone direction, and the opposite of what the same function does two lines later, where an unparsable verdict timestamp fails *closed*.

Why this is reachable rather than theoretical: `latest_verdict_fresh` is a `pub` function in `edda-ledger`, defended only by a caller invariant living in `edda-conductor`, and the value reaching it (`gate_entered_at`) comes from a JSON state file on disk. Corrupt or hand-edit that file and D6 switches off with no diagnostic, leaving `MAX_GATE_REDISPATCHES` as the only thing between the operator and the original defect.

**Direction:** fail closed on an unparsable bound (no verdict satisfies), matching the sibling behavior; or make the bound non-optional in the type so an unparsable value cannot be expressed. A `pub` safety predicate should not depend on a caller in another crate getting it right.

## 2. `wait_for_verdict` swallows both `Ledger::open` and the query error

Treating a failed read as "no verdict yet" is right for a transient lock. For a **corrupt database or a permission failure** it means the gate polls every 2s forever, printing nothing — and `gate_timeout_sec: None` (wait until cancelled) is the documented default.

The operator experience is the bad part: the banner is correct, the printed `edda verdict approve ...` command is correct, they run it, and nothing happens, with no diagnostic anywhere.

**Direction:** distinguish transient from persistent. Report the error at least once (and on a decaying interval), or fail the gate after N consecutive read failures of a non-lock kind. Silence is the one response that cannot be right.

## doneWhen

- [ ] An unparsable/absent freshness bound cannot admit a stale verdict
- [ ] A persistently unreadable ledger surfaces a diagnostic to the operator rather than polling silently forever
- [ ] Transient lock contention still degrades quietly (no regression in the normal case)
- [ ] Both paths covered by tests

🤖 Routed from the PR #539 round-1 verifier review.

fagemx:
## Summary

Resolves #541 — both ways the #519 verdict gate's safety **turns itself off silently**.

## §1 — `latest_verdict_fresh` fails OPEN on an unparsable `not_before` (`edda-ledger`)

Previously a present-but-unparsable bound degraded to `None` = unbounded → **every stale verdict satisfied the gate**: the loop-prone direction D6 was created to kill, and the opposite of the sibling rule two lines later (unparsable verdict `ts` never satisfies). The bound arrives from a JSON state file on disk; a corrupt or hand-edited value must degrade to "gate stays shut".

**Fix:** `None` remains the legitimate unbounded query (unchanged semantics); a **present but unparsable** bound now fails closed — no verdict satisfies. The old test asserting the fail-open behavior was inverted and two new tests pin the boundary (fail-closed; `None` stays unbounded).

## §2 — `wait_for_verdict` swallows ledger read errors (`edda-conductor`)

Previously `if let Ok(ledger)` + `if let Ok(Some(..))` made a corrupt database / permission failure / missing workspace **indistinguishable from "no verdict yet"** — the gate polled every 2s printing nothing, and with `gate_timeout_sec: None` forever.

**Fix:** poll errors are classified by the new `edda_ledger::lock::is_busy_error`:
- **SQLite busy/locked** → transient, degrades quietly (unchanged behavior)
- **persistent** → reported on stderr immediately, then on a decaying interval (30s → doubling → 300s cap); after **15 consecutive** persistent failures (≈30s at the 2s poll) the gate fails via new `GateVerdict::LedgerUnreadable` → new `ErrorType::LedgerUnreadable` (tag `ledger_unreadable`, additive per the GH-540 stable-tag surface) → full phase-failure path (state, event log, tmux, notify, ledger record)
- a **healthy poll resets** the budget

In the unreadable state the printed `edda verdict approve ...` command could never satisfy the gate anyway — it writes to a ledger the gate cannot read. This also de-silences the **#543 worktree symptom** (gate pointed at a workspace without `.edda`): it now aborts with a diagnostic after ~30s instead of polling silently forever. #543's root cause (three coexisting workspace resolutions) remains its own issue.

## IN SCOPE

- Changed paths: `crates/edda-ledger/src/verdict.rs`, `crates/edda-ledger/src/lock.rs` (`is_busy_error` + test), `crates/edda-conductor/src/runner/sequential.rs`, `crates/edda-conductor/src/state/machine.rs` (new ErrorType variant + tag), `crates/edda-conductor/Cargo.toml` + `Cargo.lock` (dev-deps: rusqlite for error construction, tokio test-util for paused time)
- Direct consumers: `Ledger::latest_verdict_fresh` (one caller: the gate poll); the gate's match site; any JSONL consumer of `error_type` tags (additive)
- Acceptance: the four #541 doneWhen items
- Security/data-loss: strictly tightens the gate (fail-closed, no silent waits); no new writers

## doneWhen audit

- ✅ An unparsable/absent freshness bound cannot admit a stale verdict: `fresh_query_unparsable_not_before_fails_closed`
- ✅ A persistently unreadable ledger surfaces a diagnostic rather than polling silently: `wait_for_verdict_fails_after_persistent_read_errors` (paused time, bare dir) + `read_error_tracker_fails_at_budget`
- ✅ Transient lock contention still degrades quietly: `read_error_tracker_ignores_busy_errors` + `busy_errors_are_transient_others_are_not` (chain-walking classifier)
- ✅ Both paths covered by tests; error budget reset covered by `read_error_tracker_budget_resets_on_healthy_poll`

## Gate receipt (L1, frozen SHA `0eaed1dfbef91baa67471aadffb959c2287694a4`, clean tree, worktree `edda-wt-gh541`)

- `cargo test --workspace`: only the 13 pre-existing `edda` e2e failures — failure names diffed **identical to the main baseline** (#646 non-hermetic class); `edda` crate 431 passed
- `cargo test -p edda-ledger`: 233 passed (one environmental flake, `find_root_non_git_no_edda_returns_none` — walks up from temp into the operator home's `.edda`; verified failing on the **main checkout at main** too, and passing in this workspace run)
- `cargo test -p edda-conductor`: 348 passed, 0 failed

Closes #541

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -898,6 +898,7 @@ dependencies = [
  "edda-notify",
  "edda-store",
  "regex",
+ "rusqlite",
  "serde",
  "serde_json",
  "serde_yml",
--- a/crates/edda-conductor/Cargo.toml
+++ b/crates/edda-conductor/Cargo.toml
@@ -29,3 +29,8 @@ uuid.workspace = true
 
 [dev-dependencies]
 tempfile.workspace = true
+# GH-541: gate read-error tests use tokio::time::start_paused.
+tokio = { version = "1", features = ["process", "time", "rt-multi-thread", "io-util", "macros", "test-util"] }
+# GH-541: the gate's read-error tracker classifies SQLite busy/lock errors
+# as transient; tests construct those errors directly.
+rusqlite = { version = "0.32", features = ["bundled"] }
--- a/crates/edda-conductor/src/runner/sequential.rs
+++ b/crates/edda-conductor/src/runner/sequential.rs
@@ -546,6 +546,65 @@ pub async fn run_plan(plan: &Plan, state: &mut PlanState, ctx: RunContext<'_>) -
                         ))
                         .await;
                 }
+                GateVerdict::LedgerUnreadable(err) => {
+                    // GH-541: the gate could not read the ledger persistently
+                    // (not lock contention). Failing with the diagnostic is
+                    // the only honest outcome — the printed `edda verdict`
+                    // command writes to a ledger this gate cannot read, so
+                    // the wait was unrescuable from the start.
+                    let msg = format!(
+                        "gate aborted: ledger unreadable for \"{subject}\" (sha {gate_sha}): {err}"
+                    );
+                    transition(
+                        state,
+                        &gated_id,
+                        PhaseStatus::AwaitingVerdict,
+                        PhaseStatus::Failed,
+                        Some(PhaseUpdate {
+                            error: Some(ErrorInfo {
+                                error_type: ErrorType::LedgerUnreadable,
+                                message: msg.clone(),
+                                retryable: false,
+                                check_index: None,
+                                timestamp: now_rfc3339(),
+                            }),
+                            ..Default::default()
+                        }),
+                    )?;
+                    println!("  ⚠ Phase \"{gated_id}\" {msg}");
+                    if let Some(tmux) = tmux_session {
+                        let _ = tmux.update_phase_status(&gated_id, "Failed");
+                    }
+                    let gate_cost = state.get_phase(&gated_id).ok().and_then(|p| p.cost_usd);
+                    edda::record_phase_failed_with_plan(
+                        cwd,
+                        Some(&plan.name),
+                        &gated_id,
+                        gate_cost,
+                        &msg,
+                    );
+                    let gate_ps = state.get_phase(&gated_id)?;
+                    event_log.record(Event::PhaseFailed {
+                        phase_id: gated_id.clone(),
+                        attempt: gate_ps.attempts,
+                        duration_ms: 0,
+                        error: msg,
+                        error_type: Some(ErrorType::LedgerUnreadable.tag().to_string()),
+                        env_retries: gate_ps.env_retries,
+                        attempt_charged: true,
+                    });
+                    let final_output = load_gate_output(cwd, &plan.name, &gated_id);
+                    clear_gate_output(cwd, &plan.name, &gated_id);
+                    notifier
+                        .notify_phase_terminal(phase_terminal_event(
+                            &plan.name,
+                            &gated_id,
+                            "Failed",
+                            gate_ps.attempts,
+                            final_output.as_deref(),
+                        ))
+                        .await;
+                }
                 GateVerdict::Cancelled => {
                     // The loop top sees the cancelled token and shuts down;
                     // the phase stays AWAITING_VERDICT so a later
@@ -710,6 +769,73 @@ pub async fn run_plan(plan: &Plan, state: &mut PlanState, ctx: RunContext<'_>) -
 /// Ledger poll interval while a gate waits for a verdict.
 const GATE_POLL_SEC: u64 = 2;
 
+/// GH-541: first report of a persistent ledger read failure goes out
+/// immediately; repeats follow this interval, doubling up to the cap.
+const GATE_READ_ERROR_REPORT_SECS: u64 = 30;
+const GATE_READ_ERROR_REPORT_CAP_SECS: u64 = 300;
+/// GH-541: consecutive persistent (non-lock) ledger read failures after
+/// which the gate fails with the diagnostic instead of waiting silently.
+/// At the 2s poll this is ~30s of a persistently broken ledger — an
+/// operator-fixable fault (corrupt db, permissions, wrong workspace root)
+/// must not consume the whole gate budget unnoticed.
+const GATE_MAX_PERSISTENT_READ_ERRORS: u32 = 15;
+
+/// GH-541: tracks persistent (non-lock) ledger read failures while a gate
+/// waits. Busy/lock contention is transient and never counts. A healthy
+/// poll resets the budget. Reports the error on the first failure and then
+/// on a decaying interval; returns `LedgerUnreadable` when the budget
+/// expires.
+struct ReadErrorTracker {
+    consecutive: u32,
+    report_backoff_secs: u64,
+    next_report: Option<Instant>,
+}
+
+impl Default for ReadErrorTracker {
+    fn default() -> Self {
+        Self {
+            consecutive: 0,
+            report_backoff_secs: GATE_READ_ERROR_REPORT_SECS,
+            next_report: None,
+        }
+    }
+}
+
+impl ReadErrorTracker {
+    /// Observe a poll error. Returns `Some(GateVerdict::LedgerUnreadable)`
+    /// when the persistent-failure budget is exhausted.
+    fn observe(&mut self, err: &anyhow::Error) -> Option<GateVerdict> {
+        if edda_ledger::lock::is_busy_error(err) {
+            // Transient lock contention: degrade quietly, as before.
+            return None;
+        }
+        self.consecutive += 1;
+        if self.consecutive == 1 || self.next_report.is_some_and(|t| Instant::now() >= t) {
+            eprintln!(
+                "  ⚠ verdict gate: ledger read failed (attempt {}): {err:#}",
+                self.consecutive
+            );
+            self.next_report = Some(Instant::now() + Duration::from_secs(self.report_backoff_secs));
+            self.report_backoff_secs =
+                (self.report_backoff_secs * 2).min(GATE_READ_ERROR_REPORT_CAP_SECS);
+        }
+        if self.consecutive >= GATE_MAX_PERSISTENT_READ_ERRORS {
+            return Some(GateVerdict::LedgerUnreadable(format!(
+                "{} consecutive failed ledger reads: {err:#}",
+                self.consecutive
+            )));
+        }
+        None
+    }
+
+    /// A healthy poll (the ledger opened and the query answered) resets the
+    /// persistent-failure budget.
+    fn reset(&mut self) {
+        self.consecutive = 0;
+        self.next_report = None;
+    }
+}
+
 /// D6: bound gate redispatch cycles with their own persisted counter, NOT
 /// `attempt` (which D3 forbids incrementing on redispatch). A redispatch
 /// turn is not guaranteed to produce a commit, so a re-entered gate can
@@ -760,6 +886,12 @@ enum GateVerdict {
     /// `gate_timeout_sec` elapsed with no matching verdict — NOT silent,
     /// NOT auto-approve (D3).
     TimedOut,
+    /// The ledger stayed unreadable (NOT SQLite busy/lock contention) for
+    /// the whole error budget (GH-541): fail the gate with the diagnostic
+    /// instead of polling in silence forever. In this state the printed
+    /// `edda verdict` command could never satisfy the gate anyway — it
+    /// writes to a ledger this gate cannot read.
+    LedgerUnreadable(String),
     /// The CancellationToken fired while waiting. The phase stays
     /// AWAITING_VERDICT so a later resume re-enters the wait (D3 restart).
     Cancelled,
@@ -800,19 +932,39 @@ async fn wait_for_verdict(
     }
     let mut last_heartbeat = Instant::now();
 
+    // GH-541: a failed ledger read is not the same event as "no verdict yet".
+    // SQLite busy/lock contention degrades quietly (transient); every other
+    // error (corrupt database, permission, missing workspace) is persistent,
+    // is reported at least once and then on a decaying interval, and fails
+    // the gate after the error budget — the one response that cannot be
+    // right here is silence (the operator runs the printed approve command
+    // and nothing happens, with no diagnostic anywhere).
+    let mut read_errors = ReadErrorTracker::default();
+
     loop {
         // Poll BEFORE the deadline check: a verdict recorded during this
         // wait (e.g. right after the gate_entered event) must be observed
         // at the last poll, not skipped by a deadline that fires first.
-        // Poll the ledger. Best-effort: an unreadable or locked ledger simply
-        // means "no verdict observed yet" this round.
-        if let Ok(ledger) = edda_ledger::Ledger::open(cwd) {
-            if let Ok(Some(record)) = ledger.latest_verdict_fresh(subject, gate_sha, entered_at) {
-                return match record.payload.decision {
-                    edda_core::VerdictDecision::Approved => GateVerdict::Approved(record),
-                    edda_core::VerdictDecision::Rejected => GateVerdict::Rejected(record),
-                };
+        match edda_ledger::Ledger::open(cwd) {
+            Err(e) => {
+                if let Some(v) = read_errors.observe(&e) {
+                    return v;
+                }
             }
+            Ok(ledger) => match ledger.latest_verdict_fresh(subject, gate_sha, entered_at) {
+                Ok(Some(record)) => {
+                    return match record.payload.decision {
+                        edda_core::VerdictDecision::Approved => GateVerdict::Approved(record),
+                        edda_core::VerdictDecision::Rejected => GateVerdict::Rejected(record),
+                    };
+                }
+                Ok(None) => read_errors.reset(),
+                Err(e) => {
+                    if let Some(v) = read_errors.observe(&e) {
+                        return v;
+                    }
+                }
+            },
         }
         if let Some(deadline) = deadline {
             if time::OffsetDateTime::now_utc() >= deadline {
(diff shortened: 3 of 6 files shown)


fagemx:
## Code Review: PR #718 (Round 1)

**Reviewed SHA:** `0eaed1dfbef91baa67471aadffb959c2287694a4` (frozen, full)

**IN SCOPE:** the two #541 silence/fail-open paths — changed paths `crates/edda-ledger/src/verdict.rs`, `crates/edda-ledger/src/lock.rs`, `crates/edda-conductor/src/runner/sequential.rs`, `crates/edda-conductor/src/state/machine.rs`, conductor manifest + lockfile (dev-deps); direct consumers = `Ledger::latest_verdict_fresh` (single caller: the gate poll), the gate match site, additive `error_type` tags; acceptance = the four #541 doneWhen items.

### Acceptance audit (READ against issue #541, claims verified against code)

| doneWhen item | Verdict |
|---|---|
| Unparsable bound cannot admit a stale verdict | ✅ `latest_verdict_fresh` fails closed; old fail-open test inverted; `None` stays the documented unbounded query |
| Persistently unreadable ledger surfaces a diagnostic | ✅ `ReadErrorTracker`: report at first failure + decaying interval (30s→300s cap), abort after 15 consecutive persistent failures via `GateVerdict::LedgerUnreadable` → full phase-failure path |
| Transient lock contention still degrades quietly | ✅ `is_busy_error` (chain-walking, SQLITE_BUSY/LOCKED) never counts; regression-tested |
| Both paths covered by tests | ✅ 6 new tests incl. paused-time e2e; tracker budget-reset covered |

§2's abort also de-silences the #543 worktree symptom — correctly scoped as a symptom fix, root cause left in #543.

### Key Findings

#### Critical Issues (P0)
None.

#### High Priority (P1)
- **P1-1 — the "absent" bound still fails open at the gate call site** (`sequential.rs` ~:241, resume path): `let at = ps.gate_entered_at.clone().unwrap_or_else(now_rfc3339);` — if the persisted phase state is missing `gate_entered_at` (older state file, serde default, hand edit), the gate substitutes **now** as the freshness bound. Every verdict recorded after the resume instant then satisfies the gate — including the stale rejection from the previous gate entry that D6 exists to block. The loop reopens through state corruption with no diagnostic. doneWhen item 1 says *"an unparsable/**absent** freshness bound cannot admit a stale verdict"* — the unparsable half is fixed in `edda-ledger`, the absent half is not. It also silently breaks the documented invariant that the timeout deadline derives from the persisted entry time ("survives restarts"). **Direction:** fail closed at the call site — treat a missing `gate_entered_at` on an `AWAITING_VERDICT` phase as an error (abort with a diagnostic naming the missing invariant), not a substitution with `now`. `gate_sha` absence is already refused two lines above with `with_context`; this is the same class.

#### Observations (non-blocking, no action required this round)
- `GateVerdict::LedgerUnreadable` is also the right channel for any future "gate cannot do its safety job" aborts; the message text distinguishes causes. Acceptable as-is while unreleased.
- The phase-failure record for the abort charges the attempt (`attempt_charged: true`), consistent with the sibling TimedOut arm.

### Testing Review

#### Coverage
Both fixed paths and the transient/persistent boundary covered; the fail-before-fix demonstration is inherent to the inverted old test.

#### Convention Compliance
Dev-deps justified inline; real tempfile/dir fixtures; stable-tag addition documented. Compliant.

#### Testing Verdict: Adequate for the fixed surface (P1-1 needs its own test)

### FOLLOW-UP ISSUE
None — P1-1 is inside the frozen surface.

---

fagemx:
## Review Response: Round 1

**New full SHA:** `0b99eda4386e99f0319f99e9a1c3904049e8b402` (fix commit `0b99eda` on top of reviewed `0eaed1d`; tree identical to the previously pushed `8cad0a5` — the amend changed only the commit message, which had misdescribed the pre-fix failure mode)

**P1-1 response (fixed):** the call site no longer substitutes `now_rfc3339()` for a missing `gate_entered_at`. An `AWAITING_VERDICT` phase without the persisted entry time now refuses to wait: `run_plan` errors immediately with a diagnostic naming the missing invariant (same class as the adjacent `missing gate_sha` refusal). A matching verdict already in the ledger is not consumed — the phase stays `AWAITING_VERDICT` on disk, so the operator can repair the state and resume.

**Evidence:**
- Regression test `resume_without_gate_entered_at_fails_closed`: corrupt state (gate_sha set, gate_entered_at None) + a matching approval already recorded → asserts the run errors immediately with the diagnostic AND the verdict is not consumed.
- Fail-before-fix verified on this machine: with the substitution restored, the test fails (30s hang, no diagnostic) — note the real pre-fix failure mode is a silent indefinite wait, not an immediate approval as first stated; the commit message was amended to say so.
- Gates RAN on `0b99eda`: `cargo fmt --all --check` OK; `cargo clippy -p edda-conductor --all-targets -- -D warnings` OK; `cargo test -p edda-conductor` 349 passed / 0 failed.

All Round-1 blocking findings answered. Ready for re-review.

fagemx:
## Code Review: PR #718 (Round 2) — LGTM :tada:

**Reviewed SHA:** `0b99eda4386e99f0319f99e9a1c3904049e8b402` (full)

**IN SCOPE:** unchanged from Round 1. Round 2 covers the Round-1 fix push (`0eaed1d` → `0b99eda`): one call-site change (fail closed on missing `gate_entered_at`) + its regression test. Diff verified against `8cad0a5`: zero tree delta — the amend was message-only, so the reviewed content is what was pushed.

**Review Response check:** P1-1 answered — fix-caused class (the hole was in the Round-1 code path itself, surfaced by the Round-1 audit), fix verified in the diff, response evidence complete (test + fail-before-fix demonstration + RAN gates).

**Round-2 delta audit:**
- Fail-closed semantics confirmed: missing `gate_entered_at` → immediate `run_plan` error with a diagnostic naming the invariant; the adjacent `gate_sha` refusal is mirrored
- Verdict consumption: the ledger approval is not consumed; disk phase stays `AWAITING_VERDICT` (asserted in test) — operator repair + resume path intact
- doneWhen item 1 now fully holds: unparsable **and absent** bounds cannot admit a stale verdict

**RAN (this SHA):** `cargo fmt --all --check` OK; `cargo clippy -p edda-conductor --all-targets -- -D warnings` OK; `cargo test -p edda-conductor` 349 passed / 0 failed; fail-before-fix demonstration (30s hang without the fix).
**READ:** Round-1 L1 receipt at `0eaed1d` (fmt/clippy workspace OK; workspace-test failure set identical to the main baseline). The delta adds no runtime surface beyond the one guard; exact-head CI runs independently on this push.

P0=0, P1=0. All four #541 doneWhen items hold, including the absent-bound half closed this round. Ready for merge.

---

fagemx:
### Independent Review: PR #718 (Opus)

**Reviewed diff:** `gh pr diff 718` at head `0b99eda4386e99f0319f99e9a1c3904049e8b402` (merged as `8588c45`); verified against current `main` (`64720f2`)

**IN SCOPE:** `latest_verdict_fresh` bound semantics, the `wait_for_verdict` read-error path + its `GateVerdict::LedgerUnreadable` consumer arm, the new `gate_entered_at` resume invariant, `ErrorType` surface, and GH-541's four doneWhen items.

#### P0 (blocker)

None.

#### P1 (high)

1. **A present-but-*unparsable* `gate_entered_at` now makes the gate permanently unsatisfiable, silently — a liveness regression over the fail-open it replaced.**
   `crates/edda-conductor/src/runner/sequential.rs:296` validates only *presence* of `gate_entered_at`; the string is never parsed at the resume site. It then flows to `crates/edda-ledger/src/verdict.rs:85` (`let bound = parse_rfc3339(raw_bound)?;`), which — by this PR's design — now returns `None` for every verdict. Meanwhile `sequential.rs:1028-1034` computes the timeout deadline with `.and_then(parse).ok().unwrap_or_else(now_utc)`, and `format_gate_deadline` (`sequential.rs:2147`) already has a `"<unparsable gate_entered_at>"` branch — proving the project knows this state is reachable.
   Failure scenario: `state.json` is hand-edited or partially written so `gate_entered_at = "2026-09-03 10:00:00"` (space instead of `T` — not RFC3339). With the documented default `gate_timeout_sec: None`, `format_gate_deadline` prints `none — waits until cancelled` (it never even surfaces the bad value), `Ledger::open` and the query both succeed → `Ok(None)` → `read_errors.reset()` every poll, so the new error budget never trips. The operator runs `edda verdict approve` / `edda phase approve`; both succeed; the gate can never admit the verdict and can never time out. Before this PR the same input fell open (wrong, but the approve worked). This is the exact corrupt-state-file threat model GH-541 §1 names, and it lands in the silent-forever mode GH-541 §2 says "cannot be right."
   Compounding: `crates/edda-cli/src/cmd_phase.rs:106-112` already refuses to record a verdict "the gate would silently wait on forever" — but it checks only `status` and `gate_sha`, not bound parsability, so the CLI's own guard is defeated by this state. The `gate_entered_at` check at `sequential.rs:296` should parse, not just unwrap.

2. **`LedgerUnreadable` fails a phase whose work completed and whose checks passed as `Failed`/`retryable: false`, discarding the GH-552 waive path this repo already established.**
   `sequential.rs:650-673` transitions `AwaitingVerdict → Failed` with `retryable: false`. But the variant's own doc (`state/machine.rs:144-148`) says "infrastructure, not the agent's work", and `prompt_blocked_action` (`sequential.rs:2086-2110`) reserves the Retry/**Waive** menu for `GateTimedOut` on precisely the stated grounds that "the work completed and checks passed — 'skip' would lie." A `Failed` phase gets the plain `[R] Retry [S] Skip` menu instead.
   Failure scenario: a gated phase finishes, checks pass, then the ledger is unreadable for ~30s for a non-`SQLITE_BUSY` reason (`SQLITE_CANTOPEN`/`SQLITE_IOERR` from a Windows AV or indexer holding `.edda/ledger.db`; `.edda` on a briefly unavailable share; a peer process mid-restore). 15 consecutive polls at `GATE_POLL_SEC = 2` (`sequential.rs:871, 888`) trip the budget, the completed phase becomes `Failed`, and the operator's only recoveries are re-running the whole agent turn or recording a dishonest `Skipped`. The threshold is a hardcoded constant with no config knob and no grace for a fault that is, by construction, not the phase's fault.

#### P2/P3 observations

- **The decaying report interval is effectively dead code, and the PR body overstates it.** `observe` (`sequential.rs:920-927`) reports on `consecutive == 1`, then schedules `next_report` 30s out — but the budget expires at 15 × 2s ≈ 28s, so in the pure-persistent case the second report *never* fires and `GATE_READ_ERROR_REPORT_CAP_SECS = 300` (four doublings) is unreachable. The PR body's "reported on stderr immediately, then on a decaying interval (30s → doubling → 300s cap)" describes behavior the constants forbid. Only one line ever reaches the operator.
- **A *flapping* ledger never trips the budget and defeats the rate limiter.** `reset()` (`sequential.rs:940-943`) clears `consecutive` and `next_report` but leaves `report_backoff_secs`; since every streak reports on `consecutive == 1`, an error/healthy alternation (error, `Ok(None)`, error, …) prints a warning every 4s indefinitely and never reaches `GateVerdict::LedgerUnreadable`. Not covered by any test.
- **The "#543 worktree symptom de-silenced" claim covers only the narrow variant.** It holds when `.edda` is absent (`Ledger::open` bails, `ledger.rs:33-38`). The commoner worktree case — `.edda` present at `cwd` but a *different* ledger than the one `edda verdict approve` writes to — still returns `Ok(None)` forever and is untouched by this PR.
- **`is_busy_error`'s transient set is narrower than Windows reality.** `crates/edda-ledger/src/lock.rs:7-21` treats only `DatabaseBusy`/`DatabaseLocked` as transient; `SQLITE_CANTOPEN`/`SQLITE_IOERR_*` from transient Windows file-in-use are classified persistent and count toward the abort budget. CI's Windows subset includes `edda-conductor` but exercises no multi-process ledger contention, so this is unvalidated on the platform where it matters. (Chain-walking itself is correct — `iter_events_by_type` uses `?` on `rusqlite::Error` and `.context(…)` preserves the source, and the `busy_errors_are_transient_others_are_not` test pins it.)
- **Misplaced module + unexported API.** `is_busy_error` classifies SQLite errors but lives in `lock.rs`, whose subject is the fs2 `.edda/LOCK` workspace lock; it is also `pub` without a `lib.rs` re-export, unlike the sibling `pub use verdict::{latest_verdict, …}` (`edda-ledger/src/lib.rs:33`).
- **One assertion in `resume_without_gate_entered_at_fails_closed` is vacuous.** `sequential.rs:3924` asserts the persisted phase is still `AwaitingVerdict` and comments "the stale-verdict approval was not consumed" — but `run_plan` persists state at its top (`sequential.rs:~137/153`) *before* the gate block, and the new `?` errors out before any further save. The assertion reads back the pre-run snapshot and would hold under the old fail-open code too. Only the `msg.contains("missing gate_entered_at")` check pins the fix.
- **`wait_for_verdict_fails_after_persistent_read_errors` has no deadline guard.** `sequential.rs:4858` is the only gate test not wrapped in `tokio::time::timeout(GATE_TEST_DEADLINE, …)`. Under `start_paused` auto-advance with `timeout_sec: None`, any regression in the budget logic turns this into an unbounded busy loop — CI hangs to the job limit instead of failing. It also uses an ad-hoc `temp_dir().join(format!("edda_gate_unreadable_{pid}"))` instead of the file's `fresh_root` helper.
- **P3 — dependency convention.** The new dev-deps (`edda-conductor/Cargo.toml:33,36`) hardcode `tokio = "1"` and `rusqlite = "0.32"` rather than `workspace = true`. `rusqlite` is not a workspace dep today, so it duplicates `edda-ledger`'s pin; if the two ever diverge, `downcast_ref::<rusqlite::Error>()` silently stops matching across crate versions (or `libsqlite3-sys` fails to link) — `read_error_tracker_ignores_busy_errors` is the only thing standing between that and a false "transient" classification.
- **P3 — `ErrorType` forward-compat.** The enum has no `#[serde(other)]` and `ErrorInfo::error_type` has no `#[serde(default)]` (`state/machine.rs:133,141-158`), so an older binary reading a state file containing `"ledger_unreadable"` fails to deserialize the whole `PlanState`. Pre-existing pattern (same as `GateTimeout`/`Environmental`), not introduced here.

**Verified and found sound:** `None`-bound semantics are byte-identical to `latest_verdict` (`verdict.rs:40-48` vs `82-84`); `EddaPaths::discover` does *not* walk up (`paths.rs:24-45`), so the bare-temp-dir test is hermetic despite the receipt's `find_root` flake note; `record_phase_failed_with_plan` uses `Ledger::open`, not `open_or_init` (`runner/edda.rs:87`), so the failure path cannot auto-create a stray `.edda`; `save_state_reconciled` runs after the whole gate match (`sequential.rs:714`), so the new arm persists; `AwaitingVerdict → Failed` is a legal transition (`machine.rs:229-234`); and `gate_entered_at` shipped in the same commit as `AwaitingVerdict` (`752ce4d`), so the new hard `?` has no legacy-state-file compat exposure.

#### FOLLOW-UP candidates

- `SqliteStore::drop` runs `PRAGMA wal_checkpoint(TRUNCATE)` on every drop, and the gate opens/drops a `Ledger` every 2s — a checkpoint storm against any concurrent writer for the whole gate wait. Pre-existing, not introduced here.
- `GATE_MAX_PERSISTENT_READ_ERRORS` / `GATE_POLL_SEC` are hardcoded; no plan-level or env knob for slow/remote workspaces.
- Extend `cmd_phase.rs`'s "would silently wait on forever" pre-flight to validate `gate_entered_at` parsability (paired with P1-1's fix, this closes the operator-facing half).
- #543's real root cause (three coexisting workspace resolutions) — correctly left as its own issue by the PR.

The two core mechanisms are real and correctly implemented: I re-derived the fail-open at `verdict.rs` (old `match bound { None => true, … }` genuinely admitted every stale verdict on an unparsable bound, and the inverted test genuinely fails before the fix), and the busy-vs-persistent classifier genuinely walks the anyhow chain. But the fix is incomplete in the same dimension it set out to close: PR #718 hardened the *absent* bound into a hard error while leaving the *unparsable* bound — the same corrupt-`state.json` input class GH-541 §1 names — in a strictly worse place than before, an unsatisfiable gate with no deadline and no diagnostic, defeating the CLI guard that exists to prevent exactly that. Second, the new abort classifies an infrastructure fault as a phase `Failed`, contradicting its own doc comment and the GH-552 precedent encoded 100 lines above it, and throws away the waive path for work that completed and passed its checks. doneWhen items 1, 3, and 4 are satisfied; item 2 is satisfied only for the ledger-unreadable path the PR names, not for the sibling silent-forever path it creates. Both P1s are small, local fixes — parse at `sequential.rs:296`, and choose a terminal status that preserves the waive route — and neither requires reopening the design.

---
*Independent review round (Claude Opus via claude CLI, fresh process) — retroactive per the review.independence=opus-cli decision. Findings routed to #744.*

Issue #545, opened 2026-08-31
feat(conductor): route the gate-entry notification through edda-notify channels instead of hardcoded stdout


fagemx:
## Context

The last doneWhen item of #519 — "Notify event fires on gate entry (configurable channel)" — was deliberately deferred by #539 ("Follow-ups deliberately not in this PR: ... notify channel config"). This issue is that follow-up, so the deferral has a tracking number.

## Today

- `edda-cli/src/cmd_conduct.rs:239` hardcodes `StdoutNotifier`; the gate-entry notification (`edda-conductor/src/runner/sequential.rs:821-825`, which carries the full approve/reject commands) reaches stdout only.
- edda-notify channels exist (ntfy/webhook/telegram per #519), but the conductor cannot reach them: no wiring, no configuration surface.
- The machine-readable signal today is the `gate_entered` line in `<run-cwd>/.edda/conductor/<plan>/events.jsonl` — fine for a controller tailing files, useless for an operator's phone. Until this lands, a human relay (or a file watcher) sits on the gate's critical path.

## Ask

A `Notifier` implementation backed by edda-notify dispatch, selected via config or a `--notify <channel>` flag on `conduct run`, with stdout kept as the always-on fallback. Gate entry is the highest-value event (a human is now on the critical path); plan-blocked/completed messages ride along for free since they already go through the same `Notifier` trait.

Related: #519, #539.

fagemx:
Already implemented by GH-564 P1-1, which landed the exact seam this issue defined: `ChannelNotifier` in `crates/edda-conductor/src/runner/notify.rs` — backed by `edda_notify::dispatch`, configured from `.edda/config.json` channels (events-filtered for `phase_terminal`), selected by `conduct run` via `ChannelNotifier::for_repo(&cwd)` (`cmd_conduct.rs:250`), with stdout kept as the always-on fallback (plain messages always fall through; zero channels behaves as a bare `StdoutNotifier`). Covered by an e2e webhook test in `cmd_conduct.rs`. The remaining #551 work (progress signals during the wait, routed through this same notifier) is tracked there.

PR #725, opened 2026-09-02
feat(edda-conductor): gate-wait progress signals and a visible deadline (GH-551)

### LINKED ISSUE #551: fix(conductor): the verdict gate emits nothing between gate_entered and its timeout, so an absent reviewer looks exactly like a working one
## Context

Found by running the gate for real: a three-issue cleanup plan (`cleanup-wave1`, #544 -> #529 -> #532) where each implement phase carried `gate: verdict` and a controller session relayed an external reviewer's decision. #544 closed cleanly through the gate. The other two both died the same way, and the way they died is not covered by the existing silence issues.

#543 already names this class and lists two members: silence when the ledger root is wrong (#543 itself) and silence when the ledger is unreadable (#541 §2). This is a third trigger, and the one most likely to happen in normal use: **nothing is broken at all.** The ledger root is right, the database is readable, every poll succeeds and correctly returns "no verdict yet". The reviewer is simply not there.

## What the operator sees

`wait_for_verdict` (`edda-conductor/src/runner/sequential.rs:588-628`) loops on `GATE_POLL_SEC` = 2s. Reading the loop body: it polls the ledger, checks the deadline, sleeps. It does not call the notifier, does not record an event, and does not print. Over a 7200s budget that is ~3600 silent iterations.

The event log for a timed-out gate therefore contains exactly two entries, four hours apart:

```
{"seq":4,"ts":"2026-08-31T22:54:23.9005819Z","type":"gate_entered","phase_id":"i3-impl-532",...}
{"seq":0,"ts":"2026-09-01T02:23:46.6133083Z","type":"phase_failed","phase_id":"i3-impl-532","duration_ms":0,
 "error":"gate timed out: no verdict for \"cleanup-wave1/i3-impl-532\" ... within 7200s"}
```

Twice in one run (`i2-impl-529`: 18:23:25 -> 22:34:41; `i3-impl-532`: 22:54:23 -> 02:23:46), each burning the full 7200s.

## Why silence is worse here than in the error cases

In #541 and #543 the silence hides a fault the operator could not have fixed anyway without a diagnostic. Here the silence hides a fault the operator **could** have fixed at any moment during those two hours by typing one command — the gate is waiting for a human, and the human is the only thing that can end the wait. Withholding every signal until the budget is spent is the one behavior that guarantees the wait cannot be rescued.

Concretely, in the run above the first timeout was caused by the reviewing agent hanging (96 minutes with no reply), and the second by the controller session being restarted. In both cases a single "still waiting, N minutes left" line would have been actionable, and in both cases the work was already committed and check-green — so the two hours bought nothing.

## Aggravating detail: the timeout is opt-in but the failure mode is not obvious

`gate_timeout_sec` is `Option<u64>`; `None` means wait until cancelled, and since #544 the dry-run preview spells that out. So an unbounded wait is available and is arguably the right default for a human gate. But a bounded gate gives no indication that a clock is running: nothing at `gate_entered` states the deadline, and nothing counts down. An operator who set `7200` in a plan file hours earlier has no surface telling them the budget is draining.

## Ask

Emit something during the wait. The smallest useful version:

- state the deadline (or "waits until cancelled") once at `gate_entered`, on the AWAITING_VERDICT surface that already prints subject, gate_sha and the approve/reject commands;
- emit a low-frequency progress signal on a decaying or fixed-interval basis (through the notifier, so #545's channel routing carries it), naming the remaining budget;
- keep it cheap enough that the normal 2s poll stays silent.

## doneWhen

- [ ] The AWAITING_VERDICT surface states the gate deadline, or says explicitly that there is none
- [ ] A gate wait longer than one interval produces at least one progress signal naming the remaining budget
- [ ] The 2s poll itself stays silent (no per-poll output regression)
- [ ] Covered by a test that does not depend on wall-clock sleeps of gate length

Related: #519, #539, #541, #543, #545. Sibling finding from the same run: the terminal-state issue filed alongside this one.

Found while running a real gated plan, not by reading code; the two timeouts above are from its event log.

fagemx:
## Summary

Resolves #551. In the measured run two gates burned 7200s each in **total silence** — the work was committed and check-green, so a one-line "still waiting" would have rescued the wait at any moment. A bounded gate also gave no indication that a clock was running.

## Changes

1. **Deadline on the AWAITING_VERDICT surface** — the banner that already prints subject/gate_sha/approve-reject now also prints `deadline: <RFC3339> (gate_timeout_sec Ns)` for bounded gates, or `deadline: none — waits until cancelled` explicitly. Via the new `format_gate_deadline` helper (unit-tested for both branches).
2. **Decaying progress signals through the Notifier** — `wait_for_verdict` emits "Still waiting for verdict on \"<subject>\" (sha …) — Xm remaining" (or "no deadline (waits until cancelled)"): first after 60s, then doubling, capped at 10min. Routed through the `Notifier` trait, so #545's `ChannelNotifier` carries it to configured channels and stdout fallback always shows it.
3. **The 2s poll stays silent** — the signal is gated on its own paused-time clock (`tokio::time::Instant`), not the poll cadence.

## IN SCOPE

- Changed paths: `crates/edda-conductor/src/runner/sequential.rs` only (constants, banner line, `wait_for_verdict` + its 4 call sites)
- Direct consumers: the gate wait; notifier channel subscribers (new message kind — additive)
- Acceptance: the four #551 doneWhen items

## doneWhen audit

- ✅ The AWAITING_VERDICT surface states the deadline or its absence: banner line + `gate_deadline_line_states_bound_or_absence`
- ✅ A wait longer than one interval produces ≥1 progress signal naming the remaining budget: `gate_wait_emits_progress_signals_with_remaining_budget` (20-min simulated wait → exactly 4 decaying signals, label asserted)
- ✅ The 2s poll stays silent: `gate_wait_short_polls_stay_silent` (10s wait → zero messages)
- ✅ Tests on paused time (`#[tokio::test(start_paused = true)]`) — no wall-clock sleeps of gate length

## Gate receipt (L1, frozen SHA `c27a364744cba545645c0304aaaaab255796fecf`, clean tree, worktree `edda-wt-gh551`)

- `cargo test --workspace`: 4 failures, a strict subset of the main baseline (13 baseline e2e failures; 9 `cmd_claim` ones no longer reproduce in this environment) — nothing introduced; `edda` crate 457 passed
- `cargo test -p edda-conductor`: 355 passed, 0 failed

Closes #551

### DIFF
--- a/crates/edda-conductor/src/runner/sequential.rs
+++ b/crates/edda-conductor/src/runner/sequential.rs
@@ -311,6 +311,13 @@ pub async fn run_plan(plan: &Plan, state: &mut PlanState, ctx: RunContext<'_>) -
             println!("\n⏸ [{phase_num}/{total_phases}] Phase \"{gated_id}\" AWAITING_VERDICT — waiting for an external verdict");
             println!("  subject:  {subject}");
             println!("  gate_sha: {gate_sha}");
+            // GH-551: the budget must be visible where the wait is announced
+            // — an operator who set 7200s hours earlier has no surface
+            // telling them the clock is draining.
+            println!(
+                "  deadline: {}",
+                format_gate_deadline(phase.gate_timeout_sec, &entered_at)
+            );
             println!("  approve:  edda verdict approve {subject} --sha {gate_sha}");
             println!(
                 "  reject:   edda verdict reject {subject} --sha {gate_sha} --comment \"<why>\""
@@ -340,6 +347,7 @@ pub async fn run_plan(plan: &Plan, state: &mut PlanState, ctx: RunContext<'_>) -
                 Some(&entered_at),
                 &cancel,
                 Some(&lane_hb),
+                notifier,
             )
             .await
             {
@@ -862,6 +870,12 @@ pub async fn run_plan(plan: &Plan, state: &mut PlanState, ctx: RunContext<'_>) -
 /// Ledger poll interval while a gate waits for a verdict.
 const GATE_POLL_SEC: u64 = 2;
 
+/// GH-551: progress signals during a long gate wait. The first signal goes
+/// out after this many seconds of waiting, then the interval doubles up to
+/// the cap — low-frequency by design, so the normal 2s poll stays silent.
+const GATE_PROGRESS_FIRST_SECS: u64 = 60;
+const GATE_PROGRESS_MAX_SECS: u64 = 600;
+
 /// GH-541: first report of a persistent ledger read failure goes out
 /// immediately; repeats follow this interval, doubling up to the cap.
 const GATE_READ_ERROR_REPORT_SECS: u64 = 30;
@@ -1000,6 +1014,7 @@ enum GateVerdict {
 /// bound the re-entered gate would re-read the same rejection forever.
 /// The timeout deadline is computed from the persisted `gate_entered_at`,
 /// so it survives restarts.
+#[allow(clippy::too_many_arguments)]
 async fn wait_for_verdict(
     cwd: &Path,
     subject: &str,
@@ -1008,6 +1023,7 @@ async fn wait_for_verdict(
     entered_at: Option<&str>,
     cancel: &CancellationToken,
     heartbeat: Option<&LaneHeartbeat>,
+    notifier: &dyn Notifier,
 ) -> GateVerdict {
     let deadline = timeout_sec.map(|t| {
         let base = entered_at
@@ -1024,6 +1040,20 @@ async fn wait_for_verdict(
         hb.write("awaiting_verdict");
     }
     let mut last_heartbeat = Instant::now();
+
+    // GH-551: a bounded gate gives no indication that a clock is running,
+    // and an unbounded one looks identical to a broken one — in the measured
+    // run two gates burned 7200s each while a one-line "still waiting" would
+    // have rescued the wait at any moment. Emit a low-frequency progress
+    // signal (60s, then doubling to 10min) naming the remaining budget.
+    // tokio::time::Instant so tests can drive the whole schedule on paused
+    // time without wall-clock sleeps.
+    let mut next_progress =
+        tokio::time::Instant::now() + Duration::from_secs(GATE_PROGRESS_FIRST_SECS);
+    let mut progress_interval = GATE_PROGRESS_FIRST_SECS;
+    // Paused-time mirror of the deadline, for the remaining-budget label —
+    // the wall-clock `deadline` above drives the actual timeout check.
+    let deadline_t = timeout_sec.map(|t| tokio::time::Instant::now() + Duration::from_secs(t));
 
     // GH-541: a failed ledger read is not the same event as "no verdict yet".
     // SQLite busy/lock contention degrades quietly (transient); every other
@@ -1063,6 +1093,29 @@ async fn wait_for_verdict(
             if time::OffsetDateTime::now_utc() >= deadline {
                 return GateVerdict::TimedOut;
             }
+        }
+        // GH-551: progress signal on the decaying schedule. Inside the same
+        // select arm cadence as the poll but gated on its own clock, so the
+        // 2s poll itself stays silent.
+        let now_t = tokio::time::Instant::now();
+        if now_t >= next_progress {
+            let wait_label = match deadline_t {
+                Some(d) => {
+                    let remaining = (d - now_t).as_secs();
+                    format!(
+                        "{} remaining",
+                        format_elapsed(std::time::Duration::from_secs(remaining))
+                    )
+                }
+                None => "no deadline (waits until cancelled)".to_string(),
+            };
+            notifier
+                .notify(&format!(
+                    "Still waiting for verdict on \"{subject}\" (sha {gate_sha}) — {wait_label}"
+                ))
+                .await;
+            progress_interval = (progress_interval * 2).min(GATE_PROGRESS_MAX_SECS);
+            next_progress = now_t + Duration::from_secs(progress_interval);
         }
         tokio::select! {
             _ = tokio::time::sleep(Duration::from_secs(GATE_POLL_SEC)) => {
@@ -2072,6 +2125,29 @@ fn prompt_blocked_action(phase_id: &str, status: PhaseStatus) -> BlockedAction {
             "q" | "quit" => return BlockedAction::Quit,
             _ => println!("  Invalid choice. Enter R, S, A, or Q."),
         }
+    }
+}
+
+/// GH-551: the deadline line for the AWAITING_VERDICT surface. A bounded
+/// gate states its deadline; an unbounded one says so explicitly, so an
+/// operator who set 7200s hours earlier can tell the clock is draining.
+fn format_gate_deadline(timeout_sec: Option<u64>, entered_at: &str) -> String {
+    match timeout_sec {
+        Some(secs) => {
+            let deadline = time::OffsetDateTime::parse(
+                entered_at,
+                &time::format_description::well_known::Rfc3339,
+            )
+            .ok()
+            .map(|t| t + time::Duration::seconds(secs as i64))
+            .and_then(|d| {
+                d.format(&time::format_description::well_known::Rfc3339)
+                    .ok()
+            })
+            .unwrap_or_else(|| "<unparsable gate_entered_at>".into());
+            format!("{deadline} (gate_timeout_sec {secs})")
+        }
+        None => "none — waits until cancelled".to_string(),
     }
 }
 
@@ -4719,6 +4795,7 @@ phases:
             Some(&now_rfc3339()),
             &cancel,
             None,
+            &CollectNotifier::new(),
         )
         .await;
         assert!(matches!(verdict, GateVerdict::TimedOut));
@@ -4793,6 +4870,7 @@ phases:
             Some(&now_rfc3339()),
             &cancel,
             None,
+            &CollectNotifier::new(),
         )
         .await;
         match verdict {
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: PR #725 (Round 1) — LGTM :tada:

**Reviewed SHA:** `c27a364744cba545645c0304aaaaab255796fecf` (frozen, full)

**IN SCOPE:** gate-wait progress signals + deadline visibility — changed path `crates/edda-conductor/src/runner/sequential.rs` only (constants, banner line, `wait_for_verdict` + its 4 call sites); direct consumers = the gate wait and notifier channel subscribers (additive message kind); acceptance = the four #551 doneWhen items.

### Acceptance audit (READ against issue #551, claims verified against code)

| doneWhen item | Verdict |
|---|---|
| AWAITING_VERDICT surface states the deadline or its absence | ✅ banner line via `format_gate_deadline` — both branches unit-tested (`gate_deadline_line_states_bound_or_absence`) |
| Wait longer than one interval produces ≥1 signal naming the remaining budget | ✅ `gate_wait_emits_progress_signals_with_remaining_budget`: 20-min simulated wait → exactly 4 decaying signals (60/180/420/900s), "remaining" label asserted |
| 2s poll stays silent | ✅ `gate_wait_short_polls_stay_silent`: 10s wait → zero messages; the signal clock is independent of the poll cadence |
| No wall-clock gate-length sleeps in tests | ✅ both tests `#[tokio::test(start_paused = true)]`; remaining-budget label derived from `tokio::time::Instant` (wall clock would have frozen under pause — caught and fixed during development) |

Routing audit: the signal goes through the `Notifier` trait, so #545's `ChannelNotifier` carries it to configured channels and the stdout fallback always shows it — matching the issue's explicit ask. Scheduling constants named and documented (60s first, doubling, 10min cap). Edge orderings checked: the deadline check runs before the progress check, and `Instant` subtraction saturates, so no panic path; the not-yet-fired signal just reports "0s remaining" in the degenerate paused-time case.

### FOLLOW-UP ISSUE
None — nothing outside the frozen surface.

### Testing Review

#### Coverage
All three observable behaviors pinned by tests on paused time; no regression risk to the silent-poll contract.

#### Convention Compliance
Constants documented; targeted `allow(clippy::too_many_arguments)` consistent with existing crate precedent. Compliant.

**RAN (this SHA, Windows):** `cargo fmt --all --check` OK; `cargo clippy -p edda-conductor --all-targets -- -D warnings` OK; `cargo test -p edda-conductor` 355 passed / 0 failed; `cargo test --workspace` — 4 failures, a strict subset of the main baseline (13 pre-existing e2e; 9 `cmd_claim` ones no longer reproduce in this environment), nothing introduced.
**READ:** the L1 receipt in the PR body (same SHA). Exact-head CI runs independently on this push.

P0=0, P1=0. Ready for merge.

---
*Completed after 1 round(s) of automated review-fix loop*

fagemx:
### Independent Review: PR #725 (Opus)

**Reviewed diff:** `gh pr diff 725` at head `c27a364744cba545645c0304aaaaab255796fecf` (MERGED, in `main` as `c9f6111`). Single file, +192/−0: `crates/edda-conductor/src/runner/sequential.rs`. Exact-head CI green on all 9 jobs.

**IN SCOPE:** the gate deadline banner line, `wait_for_verdict`'s new progress-signal path and its one production call site, the `Notifier` consumers that path claims, and #551's four doneWhen items.

#### P0 (blocker)

None.

#### P1 (high)

1. **The remaining-budget number is wrong on every restart — by exactly the amount of the wait that already elapsed.** The real timeout is anchored to persisted state: `deadline = parse(entered_at) + t` (`sequential.rs:1029-1035`), and the gate block feeds it `ps.gate_entered_at` from `state.json`, hard-erroring if absent (`sequential.rs:294-303`), precisely so the bound "survives restarts" (`sequential.rs:1015`, and the block comment at `:266-268`). The new label clock is not: `deadline_t = tokio::time::Instant::now() + Duration::from_secs(t)` (`sequential.rs:1053`) is anchored to *wait entry*, not to gate entry. On a fresh gate the two coincide; on a resume they diverge by the pre-restart elapsed time, and the signal overstates.

   Failure scenario: plan sets `gate_timeout_sec: 7200`, gate entered 20:00, controller session restarted at 21:50 (the second of the two timeouts #551 measured was caused by exactly this — "the controller session being restarted"). The banner correctly prints `deadline: 2026-…T22:00:00Z (gate_timeout_sec 7200)`. Sixty seconds later the progress signal reports `119m0s remaining`. The truth is 9 minutes. At 22:00 the wall-clock check at `:1090-1094` fires `GateTimedOut` — 1h50m before the last signal's promise. doneWhen item 2 ("naming the remaining budget") is met in form and violated in substance in the one scenario the issue names. Fix is one line and stays paused-time-testable: anchor the mirror to the same base, `Instant::now() + (deadline - now_utc())`.

2. **The progress signal never reaches a configured notify channel; the PR body and the Round-1 LGTM both assert that it does.** `ChannelNotifier::notify` forwards to the stdout fallback and nothing else (`crates/edda-conductor/src/runner/notify.rs:61-64`); `edda_notify::dispatch` is reached only from `notify_phase_terminal` (`:66-71`), and `NotifyEvent` has no gate-progress variant, so `Channel::matches` (`crates/edda-notify/src/lib.rs:38-43`) has nothing to match. `ChannelNotifier::for_repo` is the production notifier (`crates/edda-cli/src/cmd_conduct.rs:250`), and only three `Notifier` impls exist workspace-wide. So the PR's change #2 ("Routed through the `Notifier` trait, so #545's `ChannelNotifier` carries it to configured channels"), its IN SCOPE line ("notifier channel subscribers (new message kind — additive)"), and the review's "Routing audit … matching the issue's explicit ask" are all false: the consumer set is empty. The issue's ask is explicit — "through the notifier, **so #545's channel routing carries it**". Passing through the trait is not the same as passing through the routing. An operator who configures an ntfy/telegram channel and walks away — the entire #551 scenario — receives nothing; the signal lands only on the `conduct run` stdout that nobody was watching.

   doneWhen item 2 read literally is satisfied by stdout, so a reasonable reader could route this to follow-up. I do not, because the incorrect claim is load-bearing: it was audited and affirmed in the merge verdict, and the next reader will configure a channel on its authority.

#### P2/P3 observations

- **P2** — The `timeout_sec: None` progress branch (`"no deadline (waits until cancelled)"`, `:1108`) has zero coverage. Both new paused-time tests pass `Some(3600)`; the existing unbounded gate tests (`:5121`) are wall-clock and finish well under the 60s first interval. #551 calls the unbounded gate the case that "looks identical to a broken one" — it is the least-tested branch of the fix.
- **P2** — The tests pin a weaker property than claimed. `gate_wait_emits_progress_signals_with_remaining_budget` asserts `progress[0].contains("remaining")` and a count of 4; it never asserts a value. Both tests pass `entered_at = now_rfc3339()`, which is the one input value that makes the two clocks in P1-1 agree — the finding is structurally unobservable to this suite.
- **P2** — Nothing durable changed. A timed-out gate's `events.jsonl` still contains exactly the two entries hours apart that #551 opened with; the signals are ephemeral stdout. Not required by doneWhen, but the issue's own evidence artifact is unimproved.
- **P3** — The test comment "the 5th would land at 1860s" (`:5001-5003`) is wrong: the 600s cap applies at the fourth doubling, so the 5th signal is at 1500s. Real margin over the 1200s verdict is 300s, not 660s. Test still passes and is deterministic; the comment misleads anyone retuning the constants.
- **P3** — `format_elapsed` has no hours unit, so the flagship 7200s gate reports `119m0s remaining` while the banner directly above states an RFC3339 UTC instant. Two unit systems for one budget.
- **P3** — `Instant::now() + Duration::from_secs(t)` (`:1053`) introduces a panic band for `gate_timeout_sec` near `u64::MAX`, where the pre-existing `t as i64` path (`:1034`) merely wrapped negative and timed out immediately. `gate_timeout_sec: Option<u64>` is unvalidated (`plan/schema.rs:91`). Absurd input, graceful-to-panic change — worth a cap, not a blocker.

#### FOLLOW-UP candidates

- Add a `gate_progress` (or `gate_waiting`) `NotifyEvent` variant and route plain gate signals through `edda_notify::dispatch`, so P1-2's fix is a real channel path rather than a doc correction. Touches `edda-notify` — outside this PR's single-file surface.
- Record a low-frequency `gate_waiting` event in the conductor event log so a post-mortem of a timed-out gate is not still two entries four hours apart.
- Validate/cap `gate_timeout_sec` at plan-parse time.

P1-1 is a defect inside the frozen surface: the feature's headline number is silently wrong in the restart case, which is one of the two real timeouts #551 was filed from, and it is wrong in the unsafe direction — it promises budget that does not exist, right up to a timeout that arrives hours early. P1-2 is not a code defect against the literal doneWhen, but the PR body and the Round-1 LGTM both certify a channel-delivery path that `notify.rs:61-64` proves does not exist, and the value of this fix for an unattended operator rests entirely on that path. The self-review's affirmations here were assertions, not verifications — it audited the trait call site and stopped short of the implementation behind it, and its "remaining budget" check tested for the presence of a word rather than the correctness of a number. The banner half of the change (`format_gate_deadline`) is correct, restart-safe, and adequately tested; the silence contract is genuinely preserved. Since the PR is already merged, this should land as a follow-up fix PR — the one-line clock re-anchoring plus a paused-time test that passes a backdated `entered_at` and asserts the numeric remaining, and either the channel routing or a correction to the claims in the PR body.

fagemx:
*Independent review round (Claude Opus via claude CLI, fresh process) — retroactive per the review.independence=opus-cli decision.*

Issue #729, opened 2026-09-03
fix(ledger): enforce the refuse-newer schema_version gate on ledger open


fagemx:
Found while writing COMPATIBILITY.md (GH-651, operator ruling `compat.schema-version-policy=read-older-refuse-newer-minor-bump-announced`).

**Policy:** a ledger whose store version is newer than the binary must be refused to open (exit 2, message naming both version numbers) — preferring refusal over silently misreading a ledger we cannot correctly interpret.

**Current behavior:** `apply_schema` only walks `current < N` upward and never checks whether the stored version exceeds the binary's max known version (`crates/edda-ledger/src/sqlite_store/schema.rs:238-308`). A newer ledger currently opens without a version check. `SqliteStore::open_existing` (the read-only `edda verify` path) also performs no version comparison.

**Why now:** COMPATIBILITY.md declares the refuse-newer gate as policy with an honest note that it is not yet enforced; consumers must not rely on the current lenient behavior, and the gate may land in any 0.x release.

**Suggested shape:** a max-known-version constant next to the migration ladder; refuse at open (both `open_or_create` and `open_existing`) when `stored > max`; exit code 2 and a message naming both versions; golden test for the refusal path.

Note: this issue only records the gap. No behavior change ships with GH-651's docs PR.

fagemx:
taking: worker-2 — branch fix/gh729-refuse-newer-schema off origin/main 1631d29

fagemx:
Resolved by PR #735 (commit d1855240e3f70f42ce70586f58b87f3b51180425).

PR #735, opened 2026-09-03
fix(ledger): enforce the refuse-newer schema_version gate on ledger open (GH-729)

fagemx:
## Summary

Issue: #729

Enforce the refuse-newer schema_version gate on ledger open per COMPATIBILITY.md policy (`compat.schema-version-policy=read-older-refuse-newer-minor-bump-announced`).

### Changes
1. **`MAX_KNOWN_SCHEMA_VERSION` & `UnsupportedSchemaVersionError`**:
   - Defined `MAX_KNOWN_SCHEMA_VERSION = 13` next to the migration ladder in `crates/edda-ledger/src/sqlite_store/schema.rs`.
   - Defined `UnsupportedSchemaVersionError { stored, max }` with error message naming both versions: `refusing to open ledger: stored schema version {stored} is newer than maximum supported version {max} (upgrade edda to open this workspace)`.
2. **Refuse on Open**:
   - Enforced in `SqliteStore::open_or_create` before schema bootstrap or migrations run.
   - Enforced in `SqliteStore::open_existing` (used by `edda verify` and read-only queries) immediately after read-only connection initialization.
3. **CLI Exit Code 2 & Diagnostic**:
   - Updated `main.rs` and `cmd_verify.rs` to detect `UnsupportedSchemaVersionError` across the error cause chain, emit the diagnostic on stderr, and exit with code 2.
4. **Tests**:
   - Added unit tests in `crates/edda-ledger/src/sqlite_store/schema.rs` testing `open_or_create` and `open_existing` refusal on future schema version.
   - Added golden contract integration test `crates/edda-cli/tests/schema_refusal_contract.rs` testing exit code 2 and both versions named across `status`, `verify`, and `ask`.

## Evidence

- **Unit tests**:
  `cargo test -p edda-ledger -- schema::tests::` passed (2/2).
- **Integration golden contract**:
  `cargo test -p edda --test schema_refusal_contract` passed (3/3):
  ```
  test ask_refuses_newer_schema_with_exit_2_and_both_versions ... ok
  test status_refuses_newer_schema_with_exit_2_and_both_versions ... ok
  test verify_refuses_newer_schema_with_exit_2_and_both_versions ... ok
  ```
- **Diagnostics verification**:
  When `stored = 99`, CLI exits 2 with:
  `error: refusing to open ledger: stored schema version 99 is newer than maximum supported version 13 (upgrade edda to open this workspace)`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -982,6 +982,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2",
+ "tempfile",
  "thiserror",
  "time",
  "tracing",
--- a/crates/edda-cli/src/cmd_verify.rs
+++ b/crates/edda-cli/src/cmd_verify.rs
@@ -37,6 +37,10 @@ pub fn execute(repo_root: &Path, json: bool) -> Result<()> {
     let ledger = match Ledger::open_existing(repo_root) {
         Ok(ledger) => ledger,
         Err(err) => {
+            if let Some(e) = err.downcast_ref::<edda_ledger::UnsupportedSchemaVersionError>() {
+                eprintln!("error: {e}");
+                std::process::exit(2);
+            }
             eprintln!(
                 "error: cannot open ledger at {}: {:#}",
                 repo_root.display(),
--- a/crates/edda-cli/src/main.rs
+++ b/crates/edda-cli/src/main.rs
@@ -1075,7 +1075,7 @@ enum McpCommand {
     Serve,
 }
 
-fn main() -> anyhow::Result<()> {
+fn main() {
     tracing_subscriber::fmt()
         .with_env_filter(
             tracing_subscriber::EnvFilter::try_from_default_env().unwrap_or_else(|_| "warn".into()),
@@ -1084,6 +1084,28 @@ fn main() -> anyhow::Result<()> {
         .init();
 
     let cli = parse_cli();
+    if let Err(err) = run(cli) {
+        if let Some(e) = find_unsupported_schema_version_error(&err) {
+            eprintln!("error: {e}");
+            std::process::exit(2);
+        }
+        eprintln!("Error: {err:?}");
+        std::process::exit(1);
+    }
+}
+
+fn find_unsupported_schema_version_error(
+    err: &anyhow::Error,
+) -> Option<&edda_ledger::UnsupportedSchemaVersionError> {
+    for cause in err.chain() {
+        if let Some(e) = cause.downcast_ref::<edda_ledger::UnsupportedSchemaVersionError>() {
+            return Some(e);
+        }
+    }
+    None
+}
+
+fn run(cli: Cli) -> anyhow::Result<()> {
     let cwd = std::env::current_dir()?;
     let repo_root = edda_ledger::EddaPaths::find_root(&cwd).unwrap_or(cwd);
 
--- a/crates/edda-cli/tests/schema_refusal_contract.rs
+++ b/crates/edda-cli/tests/schema_refusal_contract.rs
@@ -1,0 +1,86 @@
+//! Golden contract tests for refusing newer schema versions on ledger open (GH-729).
+//!
+//! Operator ruling (compat.schema-version-policy=read-older-refuse-newer-minor-bump-announced):
+//! A ledger whose store version is newer than the binary must be refused to open
+//! with exit code 2 and an error message naming both the stored version and the
+//! binary's maximum supported version.
+
+use rusqlite::Connection;
+use std::path::PathBuf;
+use std::process::Command;
+
+struct TestEnv {
+    _dir: tempfile::TempDir,
+    repo: PathBuf,
+}
+
+impl TestEnv {
+    fn new(future_version: u32) -> Self {
+        let dir = tempfile::tempdir().expect("tempdir");
+        let repo = dir.path().join("repo");
+        let edda_dir = repo.join(".edda");
+        std::fs::create_dir_all(&edda_dir).expect("create .edda");
+        std::fs::create_dir_all(repo.join(".git")).expect("create .git");
+
+        // Create a ledger.db with future schema version in schema_meta
+        let db_path = edda_dir.join("ledger.db");
+        let conn = Connection::open(&db_path).expect("open sqlite db");
+        conn.execute_batch(&format!(
+            "CREATE TABLE schema_meta (key TEXT PRIMARY KEY, value TEXT NOT NULL);
+             INSERT INTO schema_meta (key, value) VALUES ('version', '{future_version}');"
+        ))
+        .expect("write schema_meta");
+
+        Self { _dir: dir, repo }
+    }
+
+    fn run_edda(&self, args: &[&str]) -> (i32, String, String) {
+        let bin = PathBuf::from(env!("CARGO_BIN_EXE_edda"));
+        let out = Command::new(&bin)
+            .args(args)
+            .current_dir(&self.repo)
+            .output()
+            .expect("spawn edda");
+        (
+            out.status.code().expect("exit code"),
+            String::from_utf8_lossy(&out.stdout).into_owned(),
+            String::from_utf8_lossy(&out.stderr).into_owned(),
+        )
+    }
+}
+
+#[test]
+fn status_refuses_newer_schema_with_exit_2_and_both_versions() {
+    let env = TestEnv::new(99);
+    let (code, stdout, stderr) = env.run_edda(&["status"]);
+
+    assert_eq!(code, 2, "stdout={stdout:?} stderr={stderr:?}");
+    assert!(
+        stderr.contains("refusing to open ledger: stored schema version 99 is newer than maximum supported version 13"),
+        "stderr should name both version numbers (stored 99, max 13), got: {stderr:?}"
+    );
+}
+
+#[test]
+fn verify_refuses_newer_schema_with_exit_2_and_both_versions() {
+    let env = TestEnv::new(99);
+    let (code, stdout, stderr) = env.run_edda(&["verify"]);
+
+    assert_eq!(code, 2, "stdout={stdout:?} stderr={stderr:?}");
+    assert!(
+        stderr.contains("refusing to open ledger: stored schema version 99 is newer than maximum supported version 13"),
+        "stderr should name both version numbers (stored 99, max 13), got: {stderr:?}"
+    );
+}
+
+#[test]
+fn ask_refuses_newer_schema_with_exit_2_and_both_versions() {
+    let env = TestEnv::new(99);
+    let (code, stdout, stderr) = env.run_edda(&["ask", "test query"]);
+
+    assert_eq!(code, 2, "stdout={stdout:?} stderr={stderr:?}");
+    assert!(
+        stderr.contains("refusing to open ledger: stored schema version 99 is newer than maximum supported version 13"),
+        "stderr should name both version numbers (stored 99, max 13), got: {stderr:?}"
+    );
+}
--- a/crates/edda-ledger/Cargo.toml
+++ b/crates/edda-ledger/Cargo.toml
@@ -25,3 +25,7 @@ hex.workspace = true
 tracing.workspace = true
 globset.workspace = true
 dirs.workspace = true
+
+[dev-dependencies]
+tempfile.workspace = true
+
--- a/crates/edda-ledger/src/lib.rs
+++ b/crates/edda-ledger/src/lib.rs
@@ -27,6 +27,7 @@ pub use domain::{
 pub use ledger::Ledger;
 pub use lock::WorkspaceLock;
 pub use paths::{validate_branch_name, EddaPaths};
+pub use sqlite_store::{UnsupportedSchemaVersionError, MAX_KNOWN_SCHEMA_VERSION};
 pub use tasks::{TaskStatus, TaskView};
 pub use tombstone::{append_tombstone, list_tombstones, make_tombstone, DeleteReason, Tombstone};
 pub use verdict::{latest_verdict, parse_verdict_event, VerdictRecord};
--- a/crates/edda-ledger/src/sqlite_store/mod.rs
+++ b/crates/edda-ledger/src/sqlite_store/mod.rs
@@ -13,6 +13,7 @@ mod task_leases;
 pub mod types;
 mod village;
 
+pub use schema::{UnsupportedSchemaVersionError, MAX_KNOWN_SCHEMA_VERSION};
 pub use types::*;
 
 use rusqlite::Connection;
@@ -73,6 +74,7 @@ impl SqliteStore {
         store
             .conn
             .execute_batch("PRAGMA busy_timeout = 5000; PRAGMA query_only = ON;")?;
+        store.check_schema_version_supported()?;
         Ok(store)
     }
 
--- a/crates/edda-ledger/src/sqlite_store/schema.rs
+++ b/crates/edda-ledger/src/sqlite_store/schema.rs
@@ -224,8 +224,32 @@ CREATE TABLE IF NOT EXISTS task_leases (
 CREATE INDEX IF NOT EXISTS idx_task_leases_expires_at ON task_leases(expires_at);
 ";
 
+/// The maximum schema version known and supported by this binary.
+pub const MAX_KNOWN_SCHEMA_VERSION: u32 = 13;
+
+#[derive(Debug, thiserror::Error, Clone, PartialEq, Eq)]
+#[error("refusing to open ledger: stored schema version {stored} is newer than maximum supported version {max} (upgrade edda to open this workspace)")]
+pub struct UnsupportedSchemaVersionError {
+    pub stored: u32,
+    pub max: u32,
+}
+
 impl SqliteStore {
+    pub(super) fn check_schema_version_supported(&self) -> anyhow::Result<()> {
+        let current = self.schema_version()?;
+        if current > MAX_KNOWN_SCHEMA_VERSION {
+            anyhow::bail!(UnsupportedSchemaVersionError {
+                stored: current,
+                max: MAX_KNOWN_SCHEMA_VERSION,
+            });
+        }
+        Ok(())
+    }
+
     pub(super) fn apply_schema(&self) -> anyhow::Result<()> {
+        // Enforce that the ledger version is not newer than our max supported version.
+        self.check_schema_version_supported()?;
+
         // Always apply v1 base schema (idempotent via IF NOT EXISTS)
         self.conn.execute_batch(SCHEMA_SQL)?;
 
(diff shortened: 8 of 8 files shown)


fagemx:
## Code Review: Round 1 — PR #735 @ 42bfba09d000b9ecabe66ec66837c08c4be89943

- model_requested: claude (fleet-review skill, no model override in the invocation)
- model_observed: unverified
- spec: review-spec-v1
- escalations: none — §5.1/§5.2 not routed (no docs/skills paths in the diff), so D5 does not apply; R2 has no input (no shell lines added). Not a zero-findings review: 7 P2s below.
- cost: ~20 tool calls, ~15 min wall (most of it waiting on exact-head CI); no local Cargo build (no lane assigned to this session; CI covers both touched crates on all three OSes)

### IN SCOPE

Changed behaviour: a ledger whose `schema_meta.version` exceeds the binary's max known version is now refused at open on both the write path (`SqliteStore::open_or_create` → `apply_schema`) and the read-only path (`SqliteStore::open_existing`); the CLI maps that error to exit 2 with a message naming both versions.

Paths: `crates/edda-ledger/src/sqlite_store/{schema.rs,mod.rs}`, `crates/edda-ledger/src/lib.rs`, `crates/edda-ledger/Cargo.toml`, `crates/edda-cli/src/{main.rs,cmd_verify.rs}`, `crates/edda-cli/tests/schema_refusal_contract.rs`, `Cargo.lock`.

Direct callers/consumers audited: `Ledger::open` / `Ledger::open_existing` / `Ledger::open_or_init` (`ledger.rs:40,61`), `init_workspace` / `init_head` / `init_branches_json` (`ledger.rs:954,960,977`), and the third open path `SqliteStore::open` (`mod.rs:36`).

Acceptance ceiling — issue #729 has no `## doneWhen` heading; the ceiling is its "Suggested shape" paragraph: (a) a max-known-version constant next to the migration ladder, (b) refuse at open in both `open_or_create` and `open_existing`, (c) exit code 2 and a message naming both versions, (d) a golden test for the refusal path. All four are met; evidence is in the Rules table under C1.

Security / data-loss surface: `crates/edda-ledger/**` is enumerated risk surface. The new behaviour is fail-closed — the only new outcome is an error where there was previously a silent migrate-attempt against a ledger the binary cannot interpret. Pre-refusal writes are discussed in P2 #2.

Current-base integration: base `main` = `1631d29`, one commit ahead, no conflict. The `Cargo.lock` delta is a single line (`tempfile` added to `edda-ledger`'s dependency list) — no version bumps, no new packages.

### Rules

|---|---|---|---|---|
| U1 surface | any | P0 | PASS | 8 changed files, all named by the issue's suggested shape (ledger schema + both open paths, CLI exit mapping, golden test, dev-dep + lock). No file outside it. |
| U2 no closing keywords | any | P1 | PASS | grep on PR body → 0 lines; grep on `git log --format=%B origin/main..42bfba0` → 0 lines |
| U3 `Issue: #N` line | any | P1 | PASS | `Issue: #729` |
| U4 conventional subjects | any | P1 | PASS | `git log --format=%s` filtered → 0 offenders (`fix(ledger): …(GH-729)`) |
| U5 markdown content lint | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → `exit=0` |
| U6 gates READ before RAN | any | P0 | PASS | `gh pr checks 735 --json name,state` → 9/9 SUCCESS. Run `33702050892` has `head_sha=42bfba09d000b9ecabe66ec66837c08c4be89943` — exact head, not a stale run. Includes `Test (windows-latest)` = SUCCESS, and `edda` (edda-cli) is in the CI Windows 7-crate subset, so `schema_refusal_contract.rs` did execute on Windows. No local rerun performed; reason for not rerunning stated in RAN vs READ. |
| U7 round complete before post | any | P1 | PASS | Round 1; whole scoped audit finished before posting; no prior review comments on this PR |
| D1–D5 | docs | — | N.A.(not routed) | §5.1 not routed — the diff carries no `docs` class |
| S1–S3 | skills | — | N.A.(not routed) | §5.2 not routed — the diff carries no `skills` class |
| C1 doneWhen vs real code | code-plain | P0 | PASS | (a) `schema.rs:228` `MAX_KNOWN_SCHEMA_VERSION: u32 = 13`, in the same file as the ladder, whose top step is `if current < 13` at `schema.rs:329`; value cross-checked against `migrate_v12_to_v13` → `set_schema_version(13)` at `schema.rs:888`. (b) `schema.rs:251` — first statement of `apply_schema`, before `SCHEMA_SQL` and before any migration — and `mod.rs:77`, immediately after `query_only = ON`. (c) `main.rs:1088-1090` and `cmd_verify.rs:40-42` → `exit(2)`; the message at `schema.rs:231` interpolates both `{stored}` and `{max}`. (d) `crates/edda-cli/tests/schema_refusal_contract.rs:57,69,81`. |
| C2 fix: has regression test | code-plain | P0 | PASS | subject is `fix(…)`; `git diff … -- 'crates/**' \| grep -cE '^\+.*(#\[test\]\|fn test_)'` → 7. Fails without the change by construction: pre-change `apply_schema` only walked `current < N` upward (issue #729), so a v99 ledger opened and the command exited 0, not 2. Both open paths get a unit test (`schema.rs:1030,1060`); three CLI surfaces get the golden contract. |
| C3 no unwrap/expect outside tests | code-plain | P1 | PASS | 15 candidates, every one confirmed test-only by opening the file: `schema.rs:1017` is `#[cfg(test)] mod tests`, and `crates/edda-cli/tests/schema_refusal_contract.rs` is an integration-test target. Zero in library code. |
| C4 no unsafe / blanket allow | code-plain | P1 | PASS | grep over the full range → 0 lines |
| C5 Windows coverage gap | code-plain | P1 | N.A.(no uncovered crate) | selector output empty — both touched crate directories (`edda-ledger`, `edda-cli`) are inside the CI Windows subset, and `Test (windows-latest)` is green at this head |
| R1 destructive commands enumerated | code-risk | P0 | PASS | grep for `rm -rf`, `git clean`, `reset --hard`, `git rm`, `--delete-branch`, `--force`, `Remove-Item` on added lines → 0 hits |
| R2 shell operator precedence | code-risk | `[判斷]` | N.A.(no shell lines added) | no `.sh`/`.bash`/`.ps1` in the changed-file list, and no added line mixes the two operators — the rule has no input, so there is nothing to escalate |
| R3 changed shell scripts parse | code-risk | P0 | N.A.(no shell scripts changed) | `gh pr diff 735 --name-only \| grep -E '\.(sh\|bash)$'` → empty |
| R4 resource release / `set -e` semantics | code-risk | P1 | PASS (Rust analogue) | No temp dir, file or lock is created on the new paths. `SqliteStore`'s release path is `Drop` (`mod.rs:89-94`) and it runs on every early return from `open_or_create` / `open_existing`, including the new refusal return. The CLI's `std::process::exit` fires in `main` only after `run(cli)` has returned, so no live `SqliteStore` has its destructor skipped. |
| R5 boundary / injection as properties | code-risk | P1 | PASS | Properties confirmed shape by shape: (i) the version read is a parameterless `SELECT value FROM schema_meta WHERE key = 'version'` — no ledger-controlled data is interpolated into SQL; (ii) the untrusted value is parsed into `u32` before any comparison (`schema.rs:349`), so hostile `schema_meta.version` content cannot leave the integer domain; (iii) it reaches the operator only through `thiserror`'s `Display` over two `u32` fields — there is no path from ledger content into a shell command or a filename. The one weakening property is P2 #4: a parse failure resolves to `1`, so the gate is bypassed rather than tripped. |
| §5.5 wiring verdict | any | P1 | PASS | table below; `sh scripts/wiring-scan.sh origin/main 42bfba0` → 4 new pub surfaces, each with outside references, and `no swallow patterns on added lines` |

### Wiring

|---|---|---|---|---|
| `MAX_KNOWN_SCHEMA_VERSION: u32 = 13` (pub const) | `crates/edda-ledger/src/sqlite_store/schema.rs:228`; re-exported `mod.rs:16` → `lib.rs:30` | `schema.rs:240,243` (the comparison itself); asserted at `schema.rs:1045,1075` | n/a (a constant) — but nothing binds it to the ladder top at `schema.rs:329`; see P2 #3 | ledger crate → `edda_ledger::MAX_KNOWN_SCHEMA_VERSION` public API. No other crate reads it yet; it is the gate's input, not a consumer surface, so this is not a "no consumer" case |
| `UnsupportedSchemaVersionError { stored, max }` (pub struct) | `schema.rs:231-235`, raised via `anyhow::bail!` at `schema.rs:241` | `crates/edda-cli/src/main.rs:1099` (`err.chain()` + `downcast_ref`) and `crates/edda-cli/src/cmd_verify.rs:40` (`downcast_ref`) | Not swallowed: `bail!` → `Result` → `?` through `open_or_create` / `open_existing` → `Ledger::open` / `open_existing` → command → `main`. Both readers print to stderr and `exit(2)`. `anyhow!`'s `TraitKind` arm preserves the concrete type, so the downcast survives any `.context()` a future caller adds. | ledger → CLI → process exit code. End-to-end proof at `schema_refusal_contract.rs:57,69,81` — `status`, `verify` and `ask` each assert exit 2 **and** both version numbers in stderr |
| `check_schema_version_supported()` (pub(super) fn) | `schema.rs:238-247` | `mod.rs:77` (`open_existing`) and `schema.rs:251` (`apply_schema`, hence `open_or_create`) | `anyhow::Result<()>` propagated with `?` at both call sites — no `let _`, no `.ok()`, no `unwrap_or_default()` | store → `Ledger::{open, open_existing, open_or_init}` → every CLI command that opens a ledger. The third path `SqliteStore::open` (`mod.rs:36`) is deliberately not wired; see P2 #5 |
| CLI exit code `2` for unsupported schema (behavioural surface) | `main.rs:1090` (all commands) and `cmd_verify.rs:42` (verify's own handler, reached first) | operator and calling scripts | the exit code itself, plus the stderr message | matches the repo's existing `claim-check.exit-codes=0/1/2` convention already used at `cmd_verify.rs:32,49,63` and `cmd_claim.rs:77,88,125` — no new convention invented. Asserted on three distinct commands |
| `run(cli) -> anyhow::Result<()>` and `find_unsupported_schema_version_error()` (private fns, `main.rs:1097,1108`) | `main.rs:1108` holds the former `main` body verbatim — no re-indent, `match cli.cmd` is still the tail expression | `main.rs:1087` | `Err` → `eprintln!("Error: {err:?}")` + `exit(1)`, reproducing the exact text and exit code the old `fn main() -> anyhow::Result<()>` produced through `Termination` | process boundary; one semantic difference, recorded as P2 #6 |

### Findings

- [P2] **The golden contract test does not isolate the per-user store.** All six sibling integration tests in `crates/edda-cli/tests/` pass `.env("EDDA_STORE_ROOT", &self.store)` to the spawned binary (e.g. `claim_usage_contract.rs:68`). `schema_refusal_contract.rs:100-106` is a near-verbatim copy of that helper with exactly that line and the `store` field removed, so the spawned `edda` resolves `edda_store::store_root()` to the real `%APPDATA%\edda` / `~/.local/share/edda` of whoever runs the suite (`crates/edda-store/src/lib.rs:53-64`). It is not a write today — I traced all three commands and each fails at ledger open before touching the store (`cmd_status.rs:6` `Ledger::open(repo_root)?` is the first statement) — but the isolation is incidental rather than designed, and one reordering upstream turns this test into a mutation of the developer's real store. Evidence: `crates/edda-cli/tests/schema_refusal_contract.rs:100-106` vs `crates/edda-cli/tests/claim_usage_contract.rs:63-70`.
- [P2] **`open_or_create` writes to the ledger it then refuses.** `apply_pragmas()` (`mod.rs:80-87`) executes `PRAGMA journal_mode = WAL` *before* `apply_schema()` reaches the new check at `schema.rs:251`, and `Drop for SqliteStore` (`mod.rs:89-94`) then runs `PRAGMA wal_checkpoint(TRUNCATE)` on the way out of the refusal. So `edda status` against a v99 ledger prints "refusing to open ledger" only after having set its journal mode and checkpointed its WAL. Both operations are schema-agnostic at the SQLite level and neither loses committed data — which is why this is not P0/P1 — but the message and the behaviour disagree, and the read-only sibling gets the ordering right (`mod.rs:74-77` puts `query_only = ON` first). Hoisting `check_schema_version_supported()` above `apply_pragmas()` in `open_or_create` would make the refusal genuinely non-mutating.
- [P2] **Nothing binds `MAX_KNOWN_SCHEMA_VERSION` to the migration ladder.** The constant is `schema.rs:228`; the ladder's top step is `if current < 13 { self.migrate_v12_to_v13()?; }` at `schema.rs:329`. A future migration adding `if current < 14` without editing line 228 makes the binary refuse to reopen a ledger it just migrated itself — a self-lockout on the write path, surfacing only at the next `edda` invocation. The golden tests pin `13` as a literal in three places and would go green against a stale constant, so they do not catch this. A unit test asserting the constant equals the highest version `apply_schema` can produce would make the drift impossible.
- [P2] **`schema_version()` fails open, not closed.** `schema.rs:342-350` returns `1` on any query error (`unwrap_or_else`) and `1` on any parse error (`parse().unwrap_or(1)`). A `schema_meta.version` that is absent, unreadable, or non-integer therefore reports as v1 and *passes* the new gate rather than tripping it, after which `apply_schema` runs all twelve migrations against it. This is pre-existing behaviour, but it is now the input to a safety gate, which changes what the leniency costs.
- [P2] **`SqliteStore::open` (`mod.rs:36-41`) is the one open path with no version check.** It is unreachable with a newer ledger today: its only non-test callers are `init_head` (`ledger.rs:960`) and `init_branches_json` (`ledger.rs:977`), both of which run after `init_workspace` → `open_or_create` (`ledger.rs:954`) has already gated, and it is `pub` on a `pub(crate)` module so no external crate can reach it. A comment at the definition recording *why* it is exempt would stop the next caller from silently adding a third ungated path.
- [P2] **`main`'s error path no longer flushes stdout through the Rust runtime.** `fn main() -> anyhow::Result<()>` returned through `Termination`, which flushes the stdout handle in `std::rt::cleanup`; `std::process::exit` at `main.rs:1090,1093` skips that. Nothing is lost in practice — Rust's `Stdout` is a `LineWriter` regardless of tty, so every `println!` has already flushed, and the CLI's one raw write (`cmd_run.rs:46` `stdout().write_all(&output.stdout)?`) is immediately followed by a `println!` at `cmd_run.rs:49`. Recorded because it silently changes the failure semantics of *every* command rather than just the one this PR is about, and a future `print!` / `write!` without a trailing newline would lose output on the error path.
- [P2] **The new branch in `cmd_verify.rs:40-43` is behaviourally redundant.** The existing fallthrough at `cmd_verify.rs:44-50` already exits 2 and already prints the same message via `{:#}` on the anyhow error, so `verify_refuses_newer_schema_with_exit_2_and_both_versions` would pass without it. Its only net effect is to drop `cannot open ledger at <path>:` from the operator's message — a small loss of diagnosability in the one case it handles. Harmless either way; noted so a later reader does not assume it is load-bearing.

### FOLLOW-UP ISSUE

Evidenced, outside issue #729's suggested shape; none of these extend this PR.

1. **Bind `MAX_KNOWN_SCHEMA_VERSION` to the migration ladder** (from P2 #3). Add a test asserting the constant equals the highest version `apply_schema` can produce, so a future `migrate_v13_to_v14` cannot leave the binary refusing its own output. `crates/edda-ledger/src/sqlite_store/schema.rs:228` vs `:329`.
2. **Make refusal non-mutating on the write path** (from P2 #2). Order `check_schema_version_supported()` before `apply_pragmas()` in `SqliteStore::open_or_create`, matching what `open_existing` already does, so "refusing to open" is not preceded by a `journal_mode` write and a `wal_checkpoint(TRUNCATE)`. `crates/edda-ledger/src/sqlite_store/mod.rs:44-52,80-94`.
3. **Close the two fail-open gaps around the new gate** (from P2 #4 and P2 #1). Make `schema_version()` distinguish "no schema_meta yet" from "unreadable/unparseable version" instead of returning `1` for both, and give `schema_refusal_contract.rs` the `EDDA_STORE_ROOT` isolation its six siblings already use. `crates/edda-ledger/src/sqlite_store/schema.rs:342-350`; `crates/edda-cli/tests/schema_refusal_contract.rs:100-106`.

### RAN vs READ

RAN — all read-only; no build, no edit, no commit, no push:

- `gh pr view 735 --json headRefOid,…` → head `42bfba09d000b9ecabe66ec66837c08c4be89943`, base `main`, not draft, 0 prior review comments (hence Round 1, a full review rather than a delta round).
- `gh pr diff 735 --name-only | sh -c "<REVIEW.md §3 classifier block>"` → `classes=code-risk code-plain`, `canonical_class=code-risk`.
- `git merge-base origin/main 42bfba0` → `1631d29d0173e9266c820807e0e9da8c7083a08d`; `git log --format='%H %s' origin/main..42bfba0` → 1 commit.
- The U2 (×2), U3, U4, U5, C2, C3, C4, C5, R1 and R3 command blocks, quoted verbatim from REVIEW.md §5; outputs are in the Rules table.
- `sh scripts/wiring-scan.sh origin/main 42bfba0` → 4 new pub surfaces (`max`, `stored`, `check_schema_version_supported`, `UnsupportedSchemaVersionError`), each with outside references; `no swallow patterns on added lines`.
- `gh pr checks 735 --json name,state` → 9/9 SUCCESS; `gh api repos/fagemx/edda/actions/runs/33702050892 --jq .head_sha` → `42bfba09d000b9ecabe66ec66837c08c4be89943`, confirming the green run is exact-head.
- `git diff … -- Cargo.lock` → one added line, `tempfile` under `edda-ledger`. Cross-checked that `tempfile = "3"` exists at root `Cargo.toml:52`, and that `crates/edda-cli/Cargo.toml` already carries `tempfile` and `rusqlite` dev-dependencies plus `[[bin]] name = "edda"`, so `env!("CARGO_BIN_EXE_edda")` resolves in the new test.
- Source reads for the rules that are reads rather than commands: `mod.rs` open paths, `schema.rs` ladder and `schema_version()`, `ledger.rs:30-75,940-990`, `main.rs:1070-1145`, `cmd_verify.rs:20-70`, `cmd_status.rs`, `cmd_run.rs:30-50`, `edda-store/src/lib.rs:45-64`, and the sibling test helpers under `crates/edda-cli/tests/`.
- `edda ask compat.schema-version-policy` → `No results found.` / `[repo] could not be checked — rerun with --fleet`. Per REVIEW.md §5.1 D2 this is an **unreachable ledger, not a false claim** — this review worktree is not a registered project. D2 is not routed for this class in any case; the durable carrier for the ruling is issue #729's own body, which states it verbatim and matches the string the code cites at `schema_refusal_contract.rs:4`.

READ — not re-run:

- Exact-head CI, run `33702050892` @ `42bfba09…`: `Format`, `Detect code changes`, `CI Gate`, `Clippy (ubuntu/macos/windows)` and `Test (ubuntu/macos/windows)` all SUCCESS. That covers `cargo fmt --all --check`, `cargo clippy --workspace --all-targets` and `cargo test --workspace` on Linux and macOS. It also covers **both** crates this PR touches on Windows: `edda-ledger` and `edda` are in the CI Windows 7-crate subset, so the `schema.rs` unit tests and `schema_refusal_contract.rs` executed on Windows too.
- The implementer's receipt in the PR body: `cargo test -p edda-ledger -- schema::tests::` 2/2, `cargo test -p edda --test schema_refusal_contract` 3/3, plus the observed diagnostic for `stored = 99`.
- **No local rerun was performed.** Per the `ladder` L2 row a rerun needs a stated reason, and none applies here: the receipt is present, CI is green at the exact head, the C5 selector is empty so there is no Windows coverage gap to close with a focused check, and CI reproduces the receipt independently on three operating systems. This session compiled nothing and holds no build lane — lane `n/a`.

### Verdict

LGTM (P0=0, P1=0). All four items of issue #729's suggested shape are met with a test behind each; the gate is fail-closed on both open paths and proven end-to-end on three CLI surfaces, including on Windows. The seven P2s are quality and future-drift observations and do not block; three are routed as follow-up issues above. The `escalations:` field is empty, so this LGTM is not provisional. Merge remains the operator's action (GATE-01).

Issue #714, opened 2026-09-02
chore(conductor): state.json has no lock — reconciled save narrows but does not close the read-modify-write race


fagemx:
## Context

Found in PR #713 review (GH-556 fix). `save_state_reconciled` folds external manual skips into the runner's save, which fixes the observed clobber. But the underlying read-modify-write on `.edda/conductor/<plan>/state.json` is still unlocked:

- runner: `load_state` (reconcile read) → `write_atomic` — a CLI write landing between the read and the write is still lost (window now ~ms instead of ~hours).
- CLI verbs (`conduct skip/retry/abort`): plain `load_state` → `save_state` with no lock and no reconciliation.

The workspace already has `edda_ledger::lock::WorkspaceLock` used for ledger writes; state.json writes don't take it.

## Direction

Take a workspace lock (or a plan-scoped lock) around load-mutate-save in both the runner persistence points and the CLI conduct verbs, so every writer serializes instead of relying on field-level reconciliation.

## Acceptance

- Concurrent `conduct skip` during a running phase cannot be lost by any later runner save (no reconciliation heuristic needed, or reconcile kept as defense-in-depth).
- Tests demonstrate serialization rather than a widened window.

fagemx:
taking: worker-2 — branch chore/gh714-conductor-state-lock off origin/main d185524

fagemx:
Resolved and merged via PR #737 in commit 64720f23f22f0a9042b544472f97cbbb47114238.

PR #737, opened 2026-09-03
chore(conductor): serialize state.json writes with plan-scoped file lock (GH-714)

fagemx:
## Summary

Issue: #714

Close the unlocked read-modify-write race condition on `.edda/conductor/<plan>/state.json` by introducing a plan-scoped exclusive file lock (`state.lock`).

### Changes
1. **`PlanStateLock` & `lock_path`**:
   - Implemented `PlanStateLock` in `crates/edda-conductor/src/state/persist.rs` backed by `.edda/conductor/{plan_name}/state.lock` using `fs2::FileExt::lock_exclusive` (blocking) and `try_acquire` (non-blocking).
   - Added cross-platform OS lock contention handling (`is_lock_contention` covering Windows `ERROR_LOCK_VIOLATION` 33 / `ERROR_SHARING_VIOLATION` 32 and POSIX `WouldBlock` / `PermissionDenied`).
2. **Synchronized Runner Reconciled Save**:
   - `save_state_reconciled` now acquires `PlanStateLock` across the entire reconciliation loop (from reading disk state through folding manual skips to writing state atomically).
3. **Synchronized CLI Verbs**:
   - Introduced `update_state<F, R>(cwd, plan_name, mutate)` helper that holds `PlanStateLock` across loading state, executing mutation closure, and saving state.
   - Updated `cmd_conduct::retry`, `cmd_conduct::skip`, and `cmd_conduct::abort` to use `update_state`, eliminating raw unlocked read-modify-write.
4. **Tests**:
   - Added unit test `plan_state_lock_mutual_exclusion` verifying exclusive locking and release.
   - Added unit test `concurrent_update_state_serializes_without_clobbering` verifying multiple concurrent threads executing `update_state` without clobbering each other.
   - Verified all 357 `edda-conductor` tests and 15 `cmd_conduct` tests pass.

## Evidence

- **Tests**:
  - `cargo test -p edda-conductor` -> 357 passed, 0 failed.
  - `cargo test -p edda --bin edda -- cmd_conduct::` -> 15 passed, 0 failed.
  - `cargo clippy -p edda-conductor -p edda --all-targets -- -D warnings` -> 0 warnings.
  - `cargo fmt --all --check` -> Clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -897,7 +897,6 @@ dependencies = [
  "edda-ledger",
  "edda-notify",
  "edda-store",
- "fs2",
  "regex",
  "rusqlite",
  "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -897,6 +897,7 @@ dependencies = [
  "edda-ledger",
  "edda-notify",
  "edda-store",
+ "fs2",
  "regex",
  "rusqlite",
  "serde",
--- a/crates/edda-cli/src/cmd_conduct.rs
+++ b/crates/edda-cli/src/cmd_conduct.rs
@@ -9,7 +9,7 @@ use edda_conductor::plan::schema::{GateKind, OnReject, Phase, Plan};
 use edda_conductor::runner::notify::ChannelNotifier;
 use edda_conductor::runner::sequential::{run_plan, RunContext};
 use edda_conductor::state::machine::{PhaseStatus, PlanState, PlanStatus};
-use edda_conductor::state::persist::{load_state, save_state};
+use edda_conductor::state::persist::{load_state, update_state};
 use edda_conductor::tmux::TmuxSession;
 use std::path::Path;
 use tokio_util::sync::CancellationToken;
@@ -375,38 +375,38 @@ pub fn status(repo_root: &Path, plan_name: Option<&str>, json: bool) -> Result<(
 /// Execute `edda conduct retry <phase-id>`
 pub fn retry(repo_root: &Path, phase_id: &str, plan_name: Option<&str>) -> Result<()> {
     let name = resolve_plan_name(repo_root, plan_name)?;
-    let mut state = load_state(repo_root, &name)?
-        .ok_or_else(|| anyhow::anyhow!("no state for plan \"{name}\""))?;
+    update_state(repo_root, &name, |state| {
+        let current_status = {
+            let ps = state.get_phase_mut(phase_id)?;
+            if ps.status != PhaseStatus::Failed
+                && ps.status != PhaseStatus::Stale
+                && ps.status != PhaseStatus::GateTimedOut
+            {
+                bail!(
+                    "Phase \"{}\" is {:?}, not Failed or Stale. Cannot retry.",
+                    phase_id,
+                    ps.status
+                );
+            }
+            ps.status
+        };
 
-    let current_status = {
-        let ps = state.get_phase_mut(phase_id)?;
-        if ps.status != PhaseStatus::Failed
-            && ps.status != PhaseStatus::Stale
-            && ps.status != PhaseStatus::GateTimedOut
-        {
-            bail!(
-                "Phase \"{}\" is {:?}, not Failed or Stale. Cannot retry.",
-                phase_id,
-                ps.status
-            );
+        edda_conductor::state::machine::transition(
+            state,
+            phase_id,
+            current_status,
+            PhaseStatus::Pending,
+            None,
+        )?;
+
+        // Reset plan status so runner picks up
+        if state.plan_status == PlanStatus::Blocked {
+            state.plan_status = PlanStatus::Running;
         }
-        ps.status
-    };
 
-    edda_conductor::state::machine::transition(
-        &mut state,
-        phase_id,
-        current_status,
-        PhaseStatus::Pending,
-        None,
-    )?;
+        Ok(())
+    })?;
 
-    // Reset plan status so runner picks up
-    if state.plan_status == PlanStatus::Blocked {
-        state.plan_status = PlanStatus::Running;
-    }
-
-    save_state(repo_root, &state)?;
     println!("Phase \"{phase_id}\" reset to Pending. Run `edda conduct run` to resume.");
     Ok(())
 }
@@ -419,63 +419,65 @@ pub fn skip(
     plan_name: Option<&str>,
 ) -> Result<()> {
     let name = resolve_plan_name(repo_root, plan_name)?;
-    let mut state = load_state(repo_root, &name)?
-        .ok_or_else(|| anyhow::anyhow!("no state for plan \"{name}\""))?;
+    let is_waived = update_state(repo_root, &name, |state| {
+        let ps = state.get_phase_mut(phase_id)?;
+        if ps.status == PhaseStatus::GateTimedOut {
+            // GH-552: skipping a timed-out gate is a WAIVER — the phase ran and
+            // its checks passed, so recording `Skipped` would understate what
+            // was verified. Keep the honest status, record the waiver.
+            ps.skip_reason = Some(
+                reason
+                    .unwrap_or("gate waived by operator (edda conduct skip)")
+                    .to_string(),
+            );
+            if state.plan_status == PlanStatus::Blocked {
+                state.plan_status = PlanStatus::Running;
+            }
+            return Ok(true);
+        }
+        if ps.status != PhaseStatus::Failed
+            && ps.status != PhaseStatus::Stale
+            && ps.status != PhaseStatus::Pending
+        {
+            bail!(
+                "Phase \"{}\" is {:?}. Can only skip Failed, Stale, or Pending phases.",
+                phase_id,
+                ps.status
+            );
+        }
 
-    let ps = state.get_phase_mut(phase_id)?;
-    if ps.status == PhaseStatus::GateTimedOut {
-        // GH-552: skipping a timed-out gate is a WAIVER — the phase ran and
-        // its checks passed, so recording `Skipped` would understate what
-        // was verified. Keep the honest status, record the waiver.
-        ps.skip_reason = Some(
-            reason
-                .unwrap_or("gate waived by operator (edda conduct skip)")
-                .to_string(),
-        );
+        ps.status = PhaseStatus::Skipped;
+        ps.skip_reason = Some(reason.unwrap_or("manually skipped").to_string());
+
+        // Unblock plan
         if state.plan_status == PlanStatus::Blocked {
             state.plan_status = PlanStatus::Running;
         }
-        save_state(repo_root, &state)?;
+
+        Ok(false)
+    })?;
+
+    if is_waived {
         println!("Phase \"{phase_id}\" gate waived (status kept as GateTimedOut).");
-        return Ok(());
+    } else {
+        println!("Phase \"{phase_id}\" skipped.");
     }
-    if ps.status != PhaseStatus::Failed
-        && ps.status != PhaseStatus::Stale
-        && ps.status != PhaseStatus::Pending
-    {
-        bail!(
-            "Phase \"{}\" is {:?}. Can only skip Failed, Stale, or Pending phases.",
-            phase_id,
-            ps.status
-        );
-    }
-
-    ps.status = PhaseStatus::Skipped;
-    ps.skip_reason = Some(reason.unwrap_or("manually skipped").to_string());
-
-    // Unblock plan
-    if state.plan_status == PlanStatus::Blocked {
-        state.plan_status = PlanStatus::Running;
-    }
-
-    save_state(repo_root, &state)?;
-    println!("Phase \"{phase_id}\" skipped.");
     Ok(())
 }
 
 /// Execute `edda conduct abort [plan-name]`
 pub fn abort(repo_root: &Path, plan_name: Option<&str>) -> Result<()> {
     let name = resolve_plan_name(repo_root, plan_name)?;
-    let mut state = load_state(repo_root, &name)?
-        .ok_or_else(|| anyhow::anyhow!("no state for plan \"{name}\""))?;
+    update_state(repo_root, &name, |state| {
+        if state.plan_status == PlanStatus::Completed || state.plan_status == PlanStatus::Aborted {
+            bail!("Plan \"{}\" is already {:?}.", name, state.plan_status);
+        }
 
-    if state.plan_status == PlanStatus::Completed || state.plan_status == PlanStatus::Aborted {
-        bail!("Plan \"{}\" is already {:?}.", name, state.plan_status);
-    }
+        state.plan_status = PlanStatus::Aborted;
+        state.aborted_at = Some(now_rfc3339());
+        Ok(())
+    })?;
 
-    state.plan_status = PlanStatus::Aborted;
-    state.aborted_at = Some(now_rfc3339());
-    save_state(repo_root, &state)?;
     println!("Plan \"{name}\" aborted.");
     Ok(())
 }
--- a/crates/edda-conductor/Cargo.toml
+++ b/crates/edda-conductor/Cargo.toml
@@ -26,7 +26,6 @@ tokio-util = "0.7"
 async-trait = "0.1"
 regex.workspace = true
 uuid.workspace = true
-fs2.workspace = true
 
 [dev-dependencies]
 tempfile.workspace = true
--- a/crates/edda-conductor/Cargo.toml
+++ b/crates/edda-conductor/Cargo.toml
@@ -26,6 +26,7 @@ tokio-util = "0.7"
 async-trait = "0.1"
 regex.workspace = true
 uuid.workspace = true
+fs2.workspace = true
 
 [dev-dependencies]
 tempfile.workspace = true
--- a/crates/edda-conductor/src/state/persist.rs
+++ b/crates/edda-conductor/src/state/persist.rs
@@ -1,5 +1,7 @@
 use crate::state::machine::{PhaseStatus, PlanState};
 use anyhow::{Context, Result};
+use fs2::FileExt;
+use std::fs::{File, OpenOptions};
 use std::path::{Path, PathBuf};
 
 /// Compute the state file path for a plan.
@@ -9,6 +11,91 @@ pub fn state_path(cwd: &Path, plan_name: &str) -> PathBuf {
         .join("conductor")
         .join(plan_name)
         .join("state.json")
+}
+
+/// Compute the lock file path for a plan.
+/// Location: `{cwd}/.edda/conductor/{plan_name}/state.lock`
+pub fn lock_path(cwd: &Path, plan_name: &str) -> PathBuf {
+    cwd.join(".edda")
+        .join("conductor")
+        .join(plan_name)
+        .join("state.lock")
+}
+
+/// Plan-scoped exclusive file lock on `.edda/conductor/{plan_name}/state.lock`.
+///
+/// Serializes concurrent read-modify-write cycles between the long-lived
+/// runner and CLI verbs (`conduct skip`, `retry`, `abort`) (GH-714).
+pub struct PlanStateLock {
+    _file: File,
+}
+
+fn is_lock_contention(err: &std::io::Error) -> bool {
+    err.kind() == std::io::ErrorKind::WouldBlock
+        || err.kind() == std::io::ErrorKind::PermissionDenied
+        || err.raw_os_error() == Some(33) // ERROR_LOCK_VIOLATION on Windows
+        || err.raw_os_error() == Some(32) // ERROR_SHARING_VIOLATION on Windows
+}
+
+impl PlanStateLock {
+    /// Acquire the plan state lock, blocking until acquired.
+    pub fn acquire(cwd: &Path, plan_name: &str) -> Result<Self> {
+        let path = lock_path(cwd, plan_name);
+        if let Some(parent) = path.parent() {
+            std::fs::create_dir_all(parent).with_context(|| {
+                format!("creating conductor plan directory: {}", parent.display())
+            })?;
+        }
+        let file = OpenOptions::new()
+            .create(true)
+            .truncate(false)
+            .read(true)
+            .write(true)
+            .open(&path)
+            .with_context(|| format!("opening plan state lock: {}", path.display()))?;
+
+        file.lock_exclusive()
+            .with_context(|| format!("acquiring exclusive plan state lock: {}", path.display()))?;
+
+        Ok(Self { _file: file })
+    }
+
+    /// Try to acquire the plan state lock non-blocking.
+    /// Returns `Ok(Some(lock))` if acquired, `Ok(None)` if already held by another process.
+    pub fn try_acquire(cwd: &Path, plan_name: &str) -> Result<Option<Self>> {
+        let path = lock_path(cwd, plan_name);
+        if let Some(parent) = path.parent() {
+            std::fs::create_dir_all(parent).with_context(|| {
+                format!("creating conductor plan directory: {}", parent.display())
+            })?;
+        }
+        let file = OpenOptions::new()
+            .create(true)
+            .truncate(false)
+            .read(true)
+            .write(true)
+            .open(&path)
+            .with_context(|| format!("opening plan state lock: {}", path.display()))?;
+
+        match file.try_lock_exclusive() {
+            Ok(()) => Ok(Some(Self { _file: file })),
+            Err(e) if is_lock_contention(&e) => Ok(None),
+            Err(e) => Err(anyhow::anyhow!("try_lock_exclusive failed: {e}")),
+        }
+    }
+}
+
+/// Atomically load, mutate, and save plan state under the plan's exclusive lock (GH-714).
+pub fn update_state<F, R>(cwd: &Path, plan_name: &str, mutate: F) -> Result<R>
+where
+    F: FnOnce(&mut PlanState) -> Result<R>,
+{
+    let _lock = PlanStateLock::acquire(cwd, plan_name)?;
+    let mut state = load_state(cwd, plan_name)?
+        .ok_or_else(|| anyhow::anyhow!("no state for plan \"{plan_name}\""))?;
+    let ret = mutate(&mut state)?;
+    save_state(cwd, &state)?;
+    Ok(ret)
 }
 
 /// Load state from disk. Returns None if the file doesn't exist.
@@ -34,7 +121,7 @@ pub fn save_state(cwd: &Path, state: &PlanState) -> Result<()> {
 }
 
 /// Save state from a long-lived writer (the runner) after reconciling with
-/// the state currently on disk (GH-556).
+/// the state currently on disk under the plan's exclusive lock (GH-556, GH-714).
 ///
 /// The runner holds `PlanState` in memory for the whole run while other
 /// processes (`edda conduct skip`, `retry`, ...) mutate the same file. A
@@ -50,6 +137,7 @@ pub fn save_state(cwd: &Path, state: &PlanState) -> Result<()> {
 /// the authority for phases it has started (Running/Checking/terminal
 /// transitions it observed itself).
 pub fn save_state_reconciled(cwd: &Path, state: &mut PlanState) -> Result<()> {
+    let _lock = PlanStateLock::acquire(cwd, &state.plan_name)?;
     if let Ok(Some(disk)) = load_state(cwd, &state.plan_name) {
         for disk_phase in &disk.phases {
             if disk_phase.status != PhaseStatus::Skipped {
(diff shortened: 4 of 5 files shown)


fagemx:
## Code Review: Round 1 — PR #737 @ adfbec82cbd829c10d7aa7d4aac23c045c5be478

- model_observed: claude-opus-5 (read from the session file's `model` field)
- spec: review-spec-v1
- escalations: none
- cost: 1 session, ~20 tool calls, no local Cargo build (see RAN vs READ)

### IN SCOPE

- Changed behaviour: `.edda/conductor/<plan>/state.json` read-modify-write is now serialized by a
  plan-scoped exclusive OS file lock at `.edda/conductor/<plan>/state.lock`.
- Changed paths: `crates/edda-conductor/src/state/persist.rs`, `crates/edda-cli/src/cmd_conduct.rs`,
  `crates/edda-conductor/Cargo.toml`, `Cargo.lock`.
- Direct consumers: `cmd_conduct::retry/skip/abort` (cmd_conduct.rs:378, 422, 471) and the runner's
  11 `save_state_reconciled` call sites (`runner/sequential.rs`).
- Acceptance ceiling: issue #714 — "Concurrent `conduct skip` during a running phase cannot be lost
  by any later runner save"; "Tests demonstrate serialization rather than a widened window".
- Data-loss surface: this is the state-clobber path itself. Also audited: whether any production
  writer to `state.json` still bypasses the lock (none — `cmd_phase` only reads; every other
  `save_state` call is inside `#[cfg(test)]`).
- Current-base integration: base `d1855240`, no conflict, all exact-head CI green.

### Rules

|---|---|---|---|---|
| U1 surface | any | P0 | PASS | 4 files, all inside the issue's stated surface (runner persistence + CLI conduct verbs + the dep they need) |
| U2 no closing keywords | any | P1 | PASS | grep over PR body and `d185524..adfbec8` commit bodies → 0 lines |
| U3 `Issue: #N` line | any | P1 | PASS | `Issue: #714` |
| U4 conventional subjects | any | P1 | PASS | subject filter → 0 offenders |
| U5 markdown content lint | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → exit=0 |
| U6 gates READ before RAN | any | P0 | PASS | `gh pr checks 737` @ adfbec8: Format pass, Clippy pass (ubuntu/macos/windows), Test pass (ubuntu/macos/windows). Not deterministically red; no full local rerun spent |
| U7 round complete | any | P1 | PASS | whole scoped audit finished before posting; all blocking findings batched below |
| C1 doneWhen vs real code | code-plain | P0 | PASS | see doneWhen table below |
| C2 test for the change | code-plain | P0 | N.A.(commit type is `chore:`, not `feat:`/`fix:`) | 2 new `#[test]` added regardless; quality of one of them is P2-4 |
| C3 no unwrap/expect outside tests | code-plain | P1 | PASS | 17 candidates, all at persist.rs:395+ — inside `#[cfg(test)] mod tests` (starts persist.rs:159), confirmed by opening the file |
| C4 no unsafe / blanket allow | code-plain | P1 | PASS | grep → 0 lines |
| C5 Windows coverage gap | code-plain | P1 | PASS | selector prints nothing: both touched crates (`edda-conductor`, `edda-cli`→`edda`) are in the CI Windows 7-crate subset, and `Test (windows-latest)` passed at this SHA, so both new tests ran on Windows. No focused local rerun needed |
| R1 destructive commands | code-risk | P0 | PASS | grep → 0 lines |
| R2 shell operator precedence | code-risk | `[判斷]` | N.A.(no shell/PowerShell file in the diff) | — |
| R3 shell scripts parse | code-risk | P0 | N.A.(no `.sh`/`.bash` in the diff) | — |
| R4 resource release on every exit | code-risk | P1 | PASS | `PlanStateLock` holds the `File` (persist.rs:29); the guard binding `let _lock = …?` (persist.rs:93, 140) drops on the `?` error path as well as the success path, and both `flock` and `LockFileEx` release on handle close. No leak, no explicit-unlock-only path |
| R5 boundary, stated as properties | code-risk | P1 | **FAIL** | `lock_path` (persist.rs:18) joins operator-supplied `plan_name` into a path that is then passed to `create_dir_all` **before** the plan is validated (persist.rs:45). Two distinct plan names still cannot encode to the same lock file (property holds), but the "no directory is created for a name that has no state" property does not — see P1-1 |
| §5.5 wiring verdict | any | — | filled | table below; `sh scripts/wiring-scan.sh d185524 adfbec8` → "no swallow patterns on added lines" |

#### C1 — issue #714 acceptance, matched to code

| Acceptance item | Code | Verdict |
|---|---|---|
| Plan-scoped lock across load-mutate-save in the runner | `save_state_reconciled` takes the lock at persist.rs:140, before the reconcile read at persist.rs:141 and the `write_atomic` at persist.rs:155 | MET |
| Plan-scoped lock across load-mutate-save in CLI verbs | `update_state` persist.rs:89-99: acquire → `load_state` → `mutate` → `save_state`, lock held throughout | MET |
| `retry`, `skip`, `abort` serialized via `update_state` | cmd_conduct.rs:378, 422, 471; no raw `load_state`+`save_state` pair left in a conduct verb | MET |
| Concurrent `conduct skip` cannot be lost by a later runner save | The CLI write can no longer land between the runner's reconcile read and its `write_atomic` — the window the issue named is closed. The skip then survives via the `Skipped` fold at persist.rs:141-154 | MET |
| Tests demonstrate serialization | `plan_state_lock_mutual_exclusion` (persist.rs:400) deterministically proves exclusion and release; `concurrent_update_state_serializes_without_clobbering` (persist.rs:423) is a smoke test — see P2-4 | MET (with P2-4) |
| No production writer bypasses the lock | grep over `crates/**`: only `update_state` and `save_state_reconciled` reach `save_state` outside `#[cfg(test)]` | MET |

### Wiring

|---|---|---|---|---|
| `pub fn lock_path` persist.rs:18 | pure path fn → `<cwd>/.edda/conductor/<plan>/state.lock` | `acquire` persist.rs:43, `try_acquire` persist.rs:66 | n/a (no I/O) | path → open → OS lock |
| `pub struct PlanStateLock` + `acquire` persist.rs:29,42 | creates the plan dir and an empty `state.lock`; holds an fs2 exclusive lock for the guard's lifetime | `update_state` persist.rs:93; `save_state_reconciled` persist.rs:140 | `Result` with `with_context`, propagated — not swallowed | CLI verb → `update_state` → OS lock; runner save → OS lock. Proven end-to-end by `Test (windows-latest)` + linux/macos at this SHA |
| `pub fn try_acquire` persist.rs:65 | same open, non-blocking `try_lock_exclusive` | **unit test only** — persist.rs:408, 415; `wiring-scan.sh` reports "0 outside references". Reader exists in this PR, so not the "no consumer" P1 cell, but the `pub` is unearned → P2-2 | contention mapped to `Ok(None)` via `is_lock_contention` persist.rs:33 — see P2-3 | test only |
| `pub fn update_state<F,R>` persist.rs:89 | lock → load → mutate → save | cmd_conduct.rs:378 (retry), 422 (skip), 471 (abort) | mutate error `?`-propagates **before** `save_state`, so a rejected verb writes nothing — correct | CLI arg → verb → `update_state` → `state.json` |
| new side-file `.edda/conductor/<plan>/state.lock` | empty file, `create(true).truncate(false)` | the OS lock manager only | never deleted — correct, unlinking a lock file is racy | `.edda/` is gitignored (.gitignore:2), so no repo-hygiene reach |
| new dependency `fs2` on `edda-conductor` | Cargo.toml:29, Cargo.lock:900 | `PlanStateLock` | n/a | duplicates an existing in-workspace primitive → P2-1 |

### Findings

- **[P1] `PlanStateLock::acquire` creates `.edda/conductor/<plan-name>/` and `state.lock` from an unvalidated, operator-supplied plan name, before `update_state` checks the plan exists — a mistyped `--plan` now permanently breaks bare `edda conduct` invocations.**
  Evidence: persist.rs:43-49 (`lock_path` → `create_dir_all(parent)`, unconditional) runs at persist.rs:93, *before* the `load_state(...).ok_or_else(...)` guard at persist.rs:94-95. Nothing removes the directory on the error path.
  Failure scenario: with exactly one plan `myplan` on disk, `edda conduct skip --plan mypaln p3` creates `.edda/conductor/mypaln/state.lock` and then errors `no state for plan "mypaln"`. From that point on, `resolve_plan_name` (cmd_conduct.rs:567-587) counts two directories and every no-`--plan` verb fails with `Multiple plans found: mypaln, myplan. Use --plan to specify.`, and `edda conduct status` lists the phantom (cmd_conduct.rs:325-333, 364-367 → `Plan "mypaln": no state file found`). There is no CLI verb to remove it; the operator must delete the directory by hand. Before this diff the same typo produced a clean error with no side effect — `load_state` only checked `path.exists()`.
  Related consequence on the same lines: because `plan_name` is joined without validation, `edda conduct abort --plan ../../scratch` now runs `create_dir_all` on a path outside `.edda/conductor/`. Nothing is truncated (`create(true).truncate(false)`, persist.rs:52-53) so this is not data loss, but it is unvalidated directory creation from a CLI argument that the pre-diff code did not perform.
  Suggested shape: have `update_state` establish that the plan has state (or that its directory already exists) before acquiring, and return the existing `no state for plan "…"` error otherwise. That keeps the legitimate path fully race-free — the lock still wraps the real load-mutate-save — while removing the create-on-error-path. `save_state_reconciled` is unaffected: its `plan_name` comes from the runner's own `PlanState`, not from an argument.

- [P2-1] The workspace already has this exact primitive: `edda_store::lock_file` (edda-store/src/lib.rs:100-111) is `create_dir_all` + `OpenOptions` + fs2 `lock_exclusive` returning a `LockGuard`, and it is what `edda-store`, `edda-ledger` and `edda-transcript` use. `edda-conductor` already depends on `edda-store` (Cargo.toml:13, used at persist.rs:118 for `write_atomic`). The PR adds a second implementation plus a new `fs2` dependency (Cargo.toml:29) instead of reusing it — two locking primitives to keep in step.

- [P2-2] `try_acquire` is `pub` but has no production consumer (`wiring-scan.sh`: 0 outside references); its only reader is the unit test at persist.rs:408/415. `pub(crate)` or `#[cfg(test)]` would state that honestly. `lock_path` is likewise `pub` with only in-file readers.

- [P2-3] `is_lock_contention` (persist.rs:33-38) treats `ErrorKind::PermissionDenied` as contention, so a genuine permissions failure on `state.lock` would be reported as `Ok(None)` — "another process holds it" — rather than an error. fs2 ships its own canonical predicate (`fs2::lock_contended_error()`: `WouldBlock` on POSIX, `ERROR_LOCK_VIOLATION` on Windows); the raw-OS-error arms here match it, the `PermissionDenied` arm does not. Currently latent because `try_acquire` has no production caller (P2-2).

- [P2-4] `concurrent_update_state_serializes_without_clobbering` (persist.rs:423) spawns two threads with no start barrier and runs one iteration, so it would pass with high probability against the *pre-fix* code too — it does not fail without the change, which is what makes a test a regression test. Acceptance is still met because `plan_state_lock_mutual_exclusion` (persist.rs:400) proves the primitive deterministically, but as written this one is a smoke test. A `Barrier` plus a loop, or asserting on the lock rather than on the outcome, would make it load-bearing.

- [P3] `chore(conductor):` is a thin label for closing a state-loss race; `fix(conductor):` would have routed this diff through REVIEW.md C2 ("`fix:` without a regression test is P0"), which is exactly the check P2-4 is about.

- [P3] `save_state` (persist.rs:115) stays `pub` and unlocked, with nothing recording that callers must hold the plan lock. Every current production caller does, but the invariant lives only in this PR's head. A doc comment on `save_state` naming `update_state` / `save_state_reconciled` as the supported entry points would keep it true.

### FOLLOW-UP ISSUE

- **`conduct abort` issued while the runner is mid-phase is still reverted.** The reconcile loop folds only `Skipped` phases (persist.rs:141-154) and the runner never re-reads `plan_status` from disk — the abort check at sequential.rs:111-113 reads the in-memory copy. So the runner's next `save_state_reconciled` writes `plan_status: Running` back over the operator's `Aborted` and the run continues. Not blocking here: this is pre-existing (identical before the diff), and #714's acceptance names `skip` only. `retry` has the same shape but a much narrower window, since the non-interactive blocked branch breaks out of the loop (sequential.rs:248-257) before an operator would normally retry.
- **`save_state_reconciled` swallows a failed reconcile read.** `if let Ok(Some(disk)) = load_state(…)` (persist.rs:141) means a corrupt or unreadable `state.json` silently skips the fold and then gets overwritten by the in-memory copy. Pre-existing and unchanged by this diff, but it is the one remaining place where operator intent can be dropped without a signal.
- **Plan-name validation.** `state_path`/`lock_path` both trust `plan_name` from `--plan`. Validating it as a single path component would close P1-1's second consequence at the source rather than per-call-site.

### RAN vs READ

RAN (this session, Windows 11 / Git Bash, no Cargo build):
- REVIEW.md §3 classifier on the changed-file list → `classes=code-risk code-plain`, `canonical_class=code-risk`
- U2 / U3 / U4 / C3 / C4 / C5 / R1 / C2-count greps over `d185524..adfbec8` — outputs in the Rules table
- `sh scripts/lint-markdown-content.sh` → exit=0
- `sh scripts/wiring-scan.sh d185524 adfbec82cbd829c10d7aa7d4aac23c045c5be478` → 5 new pub surfaces, `try_acquire` and `PlanStateLock` at 0 outside references; "no swallow patterns on added lines"
- `gh pr checks 737`, `gh pr view 737`, `gh issue view 714`
- Source reads: persist.rs (whole file), cmd_conduct.rs (retry/skip/abort/status/resolve_plan_name), runner/sequential.rs (save call sites, abort/blocked branches), cmd_phase.rs (confirmed read-only), edda-store/src/lib.rs:100-111

READ:
- Exact-head CI @ adfbec8 — Format pass; Clippy pass on ubuntu/macos/windows (`--workspace --all-targets`); Test pass on ubuntu/macos/windows. Covers: `cargo fmt --all --check`, workspace clippy on all three OSes, `cargo test --workspace` on Linux and macOS, and the Windows 7-crate subset — which includes **both** crates this diff touches, so `plan_state_lock_mutual_exclusion` and `concurrent_update_state_serializes_without_clobbering` ran under `LockFileEx` as well as `flock`.
- Implementer's receipt in the PR body: `cargo test -p edda-conductor` 357 passed; `cargo test -p edda --bin edda -- cmd_conduct::` 15 passed; `cargo clippy -p edda-conductor -p edda --all-targets -- -D warnings` 0 warnings; `cargo fmt --all --check` clean. This is L0-shaped (per-crate), not a full L1 workspace receipt — the workspace gates are covered by exact-head CI instead, so the gap is closed, not open.
- **No local Cargo run and no lane used.** Ladder L2: the C5 selector is empty (no crate outside the CI Windows subset), CI is green at the exact head and not deterministically red, and the receipt is consistent with it — so none of the three stated reasons for a rerun applies. P1-1 is an error-path behaviour that no CI job exercises either way; it is evidenced by reading persist.rs:43-49/93-95 against cmd_conduct.rs:567-587, not by a build.

### Verdict

**Changes Requested, P0=0, P1=1** — the lock itself is correct, correctly scoped, correctly released, and it does close the window #714 named; but `acquire` creates a plan directory from an unvalidated `--plan` argument on the error path, which turns a typo into a persistent break of every no-`--plan` conduct verb.

fagemx:
## Review Response: Round 1

Thanks for the rigorous review! Addressed all findings in commit `82c26a99cbe550d761336f3dec2f6645ec414a16`:

### P1: Phantom directory on typo'd plan name
- **Root cause addressed**: In `update_state`, we now perform input validation (`validate_plan_name`) and check `state_path(cwd, plan_name).is_file()` *before* taking the lock or creating directories.
- If the plan state file does not exist or the plan name is invalid, `update_state` fails immediately without creating `.edda/conductor/{plan_name}`.
- Added tests `update_state_nonexistent_plan_does_not_create_directory` and `plan_name_with_traversal_is_rejected` to verify zero directory artifacts left on typos or traversal attempts.

### P2: Duplicate locking primitive & fs2 dependency
- Unified with `edda_store::lock_file` and added `edda_store::try_lock_file`.
- Removed redundant `fs2` direct dependency from `edda-conductor`.
- Made `try_acquire` `#[cfg(test)] pub(crate)` since it is only needed for unit testing lock exclusion.

### P2: Concurrency test barrier
- Added `std::sync::Barrier` synchronization between threads in `concurrent_update_state_serializes_without_clobbering` so both threads attempt `update_state` simultaneously.

fagemx:
## Code Review: Round 2 — PR #737 @ 82c26a99cbe550d761336f3dec2f6645ec414a16

- model_observed: claude-opus-5 (session file `model` field)
- spec: review-spec-v1
- escalations: none
- cost: 1 session, ~15 tool calls, no local Cargo build (see RAN vs READ). Wall clock dominated by waiting on `Test (windows-latest)` (7m56s).

### IN SCOPE

Delta round (REVIEW.md §2): target is `git diff adfbec8..82c26a9` plus a RAN confirmation that the Round 1 blocker is resolved. The Round 1 audit of `adfbec8` is not repeated.

- Changed behaviour in the delta: `update_state` now validates the plan name and requires `state.json` to already exist **before** any lock or directory is created; `PlanStateLock` is re-implemented on top of `edda_store::lock_file` / the new `edda_store::try_lock_file`; the direct `fs2` dependency is dropped from `edda-conductor`.
- Changed paths: `crates/edda-conductor/src/state/persist.rs`, `crates/edda-store/src/lib.rs`, `crates/edda-conductor/Cargo.toml`, `Cargo.lock`. `crates/edda-cli/src/cmd_conduct.rs` is unchanged in the delta.
- Direct consumers re-checked: `cmd_conduct::retry/skip/abort` (cmd_conduct.rs:378, 422, 471) and `save_state_reconciled` (persist.rs:121-122). No new call site.
- Acceptance ceiling: issue #714, unchanged — re-confirmed still MET below.
- Data-loss surface: the lock is still held across the whole load-mutate-save; moving the existence gate *outside* the lock does not widen any window (the authoritative `load_state` is still inside it, persist.rs:75-77).
- Current-base integration: base `d1855240`, no conflict, exact-head CI fully green.

### Round 1 findings — resolution (RAN)

| Round 1 finding | Status | Evidence at 82c26a9 |
|---|---|---|
| **[P1] phantom directory from an unvalidated `--plan`** | **RESOLVED** | `update_state` persist.rs:69-73: `validate_plan_name(plan_name)?` then `if !state_path(...).is_file() { bail!(...) }` — both **before** `PlanStateLock::acquire` at persist.rs:75. Regression test `update_state_nonexistent_plan_does_not_create_directory` persist.rs:404-412 asserts `!.edda.exists()`; it fails against the pre-fix code, where `acquire` → `create_dir_all` created `.edda/conductor/typo_plan/` unconditionally. Traversal arm covered by `plan_name_with_traversal_is_rejected` persist.rs:415-427. The second consequence (`--plan ../../scratch` running `create_dir_all` outside `.edda/conductor/`) is closed at the source: `validate_plan_name` persist.rs:6-11 rejects empty, `/`, `\` and `..`, and is called from `acquire` (persist.rs:42) and `try_acquire` (persist.rs:53) as well, so `save_state_reconciled` is covered too. |
| [P2-1] duplicate locking primitive + new `fs2` dep | **RESOLVED** | `PlanStateLock` now holds `edda_store::LockGuard` (persist.rs:35-36) and delegates to `edda_store::lock_file` (persist.rs:44) / `edda_store::try_lock_file` (persist.rs:55). `grep -rn fs2 crates/edda-conductor/` → no hits; `fs2.workspace` removed from Cargo.toml and from the `edda-conductor` block in Cargo.lock. One primitive again. |
| [P2-2] unearned `pub` on `try_acquire` | **RESOLVED** | `#[cfg(test)] pub(crate) fn try_acquire` persist.rs:51-52. (New `pub` surfaces introduced by the fix are P2-6 below.) |
| [P2-3] `PermissionDenied` misread as contention | **RESOLVED** | `is_lock_contention` is deleted; the new arm at edda-store/src/lib.rs:127-129 matches `WouldBlock`, 33 and 32 only. A real permissions failure now propagates as `Err` instead of `Ok(None)`. |
| [P2-4] concurrency test had no start barrier | **ADDRESSED** | `Barrier::new(2)` persist.rs:444, `barrier_a.wait()` persist.rs:450, `barrier_b.wait()` persist.rs:464. Still one iteration, so it remains probabilistic rather than deterministic — see P3-1; `plan_state_lock_mutual_exclusion` persist.rs:381 is still the deterministic proof. |
| [P3] `chore:` label | partly moot | the head commit is `fix(conductor): ...`, which routes this diff through C2 — and it passes (see the Rules table). The PR title is still `chore(...)`. |
| [P3] `save_state` `pub` and unlocked, invariant undocumented | unchanged | persist.rs:97; still no doc comment naming `update_state` / `save_state_reconciled` as the supported entry points. Non-blocking, carried. |

#### C1 — issue #714 acceptance, re-confirmed at 82c26a9

| Acceptance item | Code | Verdict |
|---|---|---|
| Plan-scoped lock across load-mutate-save in the runner | `save_state_reconciled` persist.rs:122 → reconcile read :123 → `save_state` :137 | MET |
| Plan-scoped lock across load-mutate-save in CLI verbs | `update_state` persist.rs:75-79, lock held across load → mutate → save | MET |
| `retry` / `skip` / `abort` serialized | cmd_conduct.rs:378, 422, 471 — unchanged in the delta | MET |
| Concurrent `conduct skip` cannot be lost by a later runner save | unchanged by the delta; the pre-lock existence gate is outside the lock but the authoritative `load_state` is inside it (persist.rs:76), so no window is reopened | MET |
| Tests demonstrate serialization | `plan_state_lock_mutual_exclusion` persist.rs:381 (deterministic); `concurrent_update_state_serializes_without_clobbering` persist.rs:430 (now barriered) | MET |
| No production writer bypasses the lock | re-grepped at this SHA: `save_state` (conductor) has no production caller outside `update_state` and `save_state_reconciled`; the `cmd_phase.rs:369/413/490` hits are inside `#[cfg(test)]` | MET |

### Rules

|---|---|---|---|---|
| U1 surface | any | P0 | PASS | 5 files across the PR; `crates/edda-store/src/lib.rs` is new to the surface this round and is there to serve the Round 1 P2-1 consolidation — inside the issue's locking surface, not a widening |
| U2 no closing keywords | any | P1 | PASS | grep over PR body and `origin/main..82c26a9` commit bodies → 0 lines |
| U3 `Issue: #N` line | any | P1 | PASS | `Issue: #714` |
| U4 conventional subjects | any | P1 | PASS | subject filter over `origin/main..82c26a9` → 0 offenders |
| U5 markdown content lint | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → exit=0 |
| U6 gates READ before RAN | any | P0 | PASS | `gh pr checks 737` @ 82c26a9: `CI Gate` pass, `Format` pass, `Clippy` pass on ubuntu/macos/windows, `Test` pass on ubuntu/macos/windows. Not deterministically red; no full local rerun spent. The Review Response carries no gate receipt (see P3-4), but exact-head CI covers the entire gate set for this SHA, so the gap is closed rather than open |
| U7 round complete | any | P1 | PASS | whole delta audited before posting; nothing blocking found, so nothing to batch |
| C1 doneWhen vs real code | code-plain | P0 | PASS | table above |
| C2 test for the change | code-plain | P0 | PASS | head commit is `fix(conductor): prevent phantom directory...`, so the `fix:` arm applies. `update_state_nonexistent_plan_does_not_create_directory` persist.rs:404 **fails without the change**: pre-fix, `update_state` → `acquire` → `create_dir_all` created `.edda/conductor/typo_plan/` before the load guard, so `!dir.path().join(".edda").exists()` would have tripped. Genuine regression test, and it ran on all three OSes at this SHA |
| C3 no unwrap/expect outside tests | code-plain | P1 | PASS | 2 candidates in the delta (persist.rs:406, 423), both `tempfile::tempdir().unwrap()` inside `#[cfg(test)] mod tests` (starts persist.rs:141), confirmed by opening the file |
| C4 no unsafe / blanket allow | code-plain | P1 | PASS | grep over the delta → 0 lines |
| C5 Windows coverage gap | code-plain | P1 | PASS | selector prints nothing: `edda-store`, `edda-conductor` and `edda-cli`→`edda` are all in the CI Windows 7-crate subset, and `Test (windows-latest)` passed at this SHA, so all four new tests ran under `LockFileEx`. No focused local rerun needed |
| R1 destructive commands | code-risk | P0 | PASS | grep over the delta → 0 lines |
| R2 shell operator precedence | code-risk | `[判斷]` | N.A.(no shell/PowerShell file in the diff) | — |
| R3 shell scripts parse | code-risk | P0 | N.A.(no `.sh`/`.bash` in the diff) | — |
| R4 resource release on every exit | code-risk | P1 | PASS | `PlanStateLock` holds `edda_store::LockGuard` (persist.rs:36), which holds the `File` (edda-store/src/lib.rs:96); guard bindings `let _lock = ...?` at persist.rs:75 and :122 drop on the error path as well as the success path. `try_lock_file` binds the `File` into the guard before returning `Ok(Some(..))` and drops it on both the `Ok(None)` and `Err` arms (edda-store/src/lib.rs:126-133) — no fd leak on contention |
| R5 boundary and injection, as properties | code-risk | P1 | **PASS** (was FAIL in Round 1) | *Two distinct plan names cannot encode to one lock file*: `validate_plan_name` (persist.rs:6) forces the name to a single path component, so `join` yields a distinct directory per distinct name. *A CLI-supplied value is validated before it reaches the filesystem*: persist.rs:69-73 validates and gates on existence before `create_dir_all` is reachable. *No directory is created for a name that has no state*: proven by persist.rs:404-412. The runner arm is safe by construction as well — plan names come from `plan.name`, which `parse_plan` already constrains to kebab-case (parser.rs:207-211), a strict subset of what `validate_plan_name` accepts, so no legitimate plan is newly rejected |
| §5.5 wiring verdict | any | — | filled | table below; `sh scripts/wiring-scan.sh adfbec8 82c26a9` → 3 new pub surfaces, "no swallow patterns on added lines" |

### Wiring

|---|---|---|---|---|
| `pub fn validate_plan_name` persist.rs:6 | pure predicate → `Err` on empty / `/` / `\` / `..` | in-crate only: `acquire` persist.rs:42, `try_acquire` persist.rs:53, `update_state` persist.rs:69 (`wiring-scan`: 0 outside references) | `Result`, `?`-propagated to the CLI verb — not swallowed | CLI `--plan` arg → `update_state` → rejected before any filesystem call |
| `edda_store::LockGuard` as `PlanStateLock`'s payload persist.rs:36 | replaces the crate-local `File` + `fs2` guard | `acquire` :44, `try_acquire` :55 | `anyhow` with `with_context`, propagated | CLI verb / runner save → `edda_store::lock_file` → OS lock. Behaviourally identical on all three OSes per exact-head CI |
| `pub fn try_lock_file` edda-store/src/lib.rs:115 | `create_dir_all` + `OpenOptions(create, !truncate, write)` + fs2 `try_lock_exclusive`; contention → `Ok(None)` | exactly one workspace consumer, `PlanStateLock::try_acquire` persist.rs:55, which is itself `#[cfg(test)]` — see P2-6 | contention mapped to `Ok(None)`; every other `io::Error` returned as `Err` (lib.rs:132) — the Round 1 `PermissionDenied` misclassification is gone | test-only reach today: `plan_state_lock_mutual_exclusion` persist.rs:381 exercises it on all three OSes |
| removed: `fs2` dep on `edda-conductor`, `is_lock_contention`, `try_acquire`'s `pub` | — | — | — | net surface reduction; nothing external referenced any of them (`wiring-scan` at Round 1 reported 0 outside references for `try_acquire`) |

### Findings

- [P2-5] `plan_name_with_traversal_is_rejected` persist.rs:426 asserts `!dir.path().join("escape").exists()`, which is a path the code could never have created even with validation removed: `cwd/.edda/conductor/` joined with `../escape` resolves to `cwd/.edda/escape`, not `cwd/escape`. The assertion passes vacuously. The five `validate_plan_name` asserts above it (persist.rs:417-421) and the `result.is_err()` check are what actually carry the test; asserting on `dir.path().join(".edda")` — as the sibling test at persist.rs:409 correctly does — would make the filesystem half load-bearing too.
- [P2-6] `edda_store::try_lock_file` (edda-store/src/lib.rs:115) is a new `pub` function on a risk-surface crate whose only workspace consumer is `PlanStateLock::try_acquire`, which is `#[cfg(test)]` — so in any non-test build it has zero consumers, and `edda-store` adds no unit test of its own for it. Not the §5.5 "no consumer → P1" cell under the adjudication already applied to `try_acquire` in Round 1 (a reader does exist in this PR), and it is a reasonable pairing with `lock_file`; but its contention arm is only ever reached from another crate's test module. A direct contention test in `edda-store` (next to the `lock_file` test at lib.rs:200) would earn the `pub`.
- [P3-1] `concurrent_update_state_serializes_without_clobbering` persist.rs:430 is materially better with the barrier, but it is still a single iteration: the barrier raises the collision probability, it does not guarantee one. It would still pass against the pre-fix code some of the time. `plan_state_lock_mutual_exclusion` persist.rs:381 remains the deterministic evidence for acceptance; a loop of N iterations around the barriered body would make this one deterministic in practice.
- [P3-2] `edda_store::try_lock_file` hardcodes `raw_os_error() == Some(33) || Some(32)` (lib.rs:128-129) rather than comparing against `fs2::lock_contended_error()`, which fs2 ships for exactly this purpose. The values are correct — this is a maintenance nit, not a defect, and it now matches `lock_file`'s neighbourhood in the same file.
- [P3-3] `PlanStateLock::acquire` (persist.rs:41) is still `pub` and still creates the plan directory as a side effect, via `edda_store::lock_file`'s `create_dir_all`. The "the plan must already exist" gate lives in `update_state`, not in `acquire`. Correct for both current callers — the runner legitimately needs the directory created on first save — but a future third caller would reintroduce the Round 1 P1. A doc comment on `acquire` stating that it creates the plan directory, and that callers taking a name from operator input must gate on existence first, would keep the invariant readable.
- [P3-4] `Review Response: Round 1` names the new full SHA and answers every blocking finding, but reports no ran gates, which `loop` item 4 asks for. Non-blocking here only because exact-head CI at 82c26a9 is fully green across `Format`, `Clippy` x3 and `Test` x3 and therefore covers the whole gate set independently. Worth correcting in the next response so the reviewer has a receipt to READ rather than a CI result to infer one from.

### FOLLOW-UP ISSUE

Carried from Round 1, still valid, still out of this PR's scope (both are pre-existing and unchanged by the delta):

- **`conduct abort` issued mid-phase is still reverted.** The reconcile loop folds only `Skipped` phases (persist.rs:123-136) and the runner never re-reads `plan_status` from disk (sequential.rs:111-113 reads the in-memory copy), so the next `save_state_reconciled` writes `plan_status: Running` back over the operator's `Aborted`. #714's acceptance names `skip` only.
- **`save_state_reconciled` swallows a failed reconcile read.** `if let Ok(Some(disk)) = load_state(...)` persist.rs:123 — a corrupt or unreadable `state.json` silently skips the fold and is then overwritten by the in-memory copy.

Round 1's third follow-up — plan-name validation at the source — is **done by this PR** (`validate_plan_name`, persist.rs:6) and needs no issue.

### RAN vs READ

RAN (this session, Windows 11 / Git Bash, no Cargo build, no lane — `n/a`):
- REVIEW.md §3 classifier on the changed-file list → `classes=code-risk code-plain`, `canonical_class=code-risk`
- `git diff adfbec8..82c26a9` (the delta), plus U2 / U3 / U4 / C3 / C4 / C5 / R1 greps — outputs in the Rules table
- `sh scripts/lint-markdown-content.sh` → exit=0
- `sh scripts/wiring-scan.sh adfbec8 82c26a9` → `try_acquire` 0 outside refs, `validate_plan_name` 0 outside refs, `try_lock_file` 1 outside ref; "no swallow patterns on added lines"
- `grep -rn fs2 crates/edda-conductor/` → no hits (P2-1 confirmation); `grep -rn "try_lock_file|lock_file(" crates/` (P2-6 consumer census); `grep -rn "save_state(" crates/` (no unlocked production writer)
- `gh pr checks 737` (polled to completion), `gh pr view 737`, `gh issue view 714`, `gh api .../issues/737/comments`
- Source reads at 82c26a9: persist.rs (whole file), edda-store/src/lib.rs:85-140, plan/parser.rs name validation, cmd_conduct.rs retry/skip/abort/status/`resolve_plan_name`, cmd_phase.rs `save_state` call sites
- `model_observed` read from the session file's `model` field

READ:
- **Exact-head CI @ 82c26a9 — all 9 checks green**: `CI Gate` pass, `Detect code changes` pass, `Format` pass, `Clippy` pass on ubuntu / macos / windows (`--workspace --all-targets -D warnings`), `Test` pass on ubuntu (2m30s) / macos (2m19s) / windows (7m56s). Covers `cargo fmt --all --check`, workspace clippy on all three OSes, `cargo test --workspace` on Linux and macOS, and the Windows 7-crate subset — which includes **all three** crates this PR touches (`edda-store`, `edda-conductor`, `edda`), so `update_state_nonexistent_plan_does_not_create_directory`, `plan_name_with_traversal_is_rejected`, `plan_state_lock_mutual_exclusion` and the barriered concurrency test all ran under `LockFileEx` as well as `flock`.
- Implementer's `Review Response: Round 1` — answers each finding and names the SHA; **no gate receipt** (P3-4). The Round 1 PR-body receipt is per-crate (L0-shaped) and predates this SHA.
- **No local Cargo run.** Ladder L2: the C5 selector is empty, exact-head CI is green and not deterministically red, and the C5 crate set fully contains this diff — so none of the three stated reasons for a rerun applies. The absent receipt for 82c26a9 would ordinarily be one, but the gate set it would have carried is exactly what exact-head CI ran on this SHA, so a local rerun would be over-verification (`ladder`, "Verification cost").

### Verdict

**LGTM (P0=0, P1=0)** — the Round 1 P1 is fixed at the root, not papered over: `update_state` validates the plan name and requires existing state before any lock or directory is created (persist.rs:69-75), and the regression test at persist.rs:404 fails without the change. All four Round 1 P2s are addressed, the duplicate locking primitive and its `fs2` dependency are gone in favour of `edda_store::lock_file` / `try_lock_file`, and `validate_plan_name` closes the traversal consequence at the source for the runner path as well as the CLI path. Remaining findings are two P2 and four P3 quality items, none blocking. Merge remains the operator's action (GATE-01).

PR #740, opened 2026-09-03
fix(fleet): fix process snapshot order and child PID indexing in lane-stop.ps1 (GH-706)

fagemx:
## Summary

Issue: #706

Fixes `scripts/fleet/lane-stop.ps1` where processes spawned during task shutdown escaped termination and false-positively reported `residual=0`.

### Root Cause
1. **Snapshot Timing**: `lane-stop.ps1` took its process snapshot before calling `Stop-ScheduledTask`. Any child process spawned while the task wrapper was shutting down was missing from the snapshot and target set.
2. **Type Mismatch in Child Indexing**: `Win32_Process.ProcessId` and `Win32_Process.ParentProcessId` return `uint32`, whereas the queue and lookup keys used `[int]`. In .NET hashtables, `[int].Equals([uint32])` is `false`, which caused `$childrenOf.ContainsKey($currentPid)` to always evaluate to `false` and miss all child process descendants.
3. **Weak Residual Check**: The residual check only verified if remaining processes matched `$Wrapper` or `$brief` in `CommandLine`, failing to catch arbitrary children (e.g. `cargo`, `git`, `node`) whose command lines do not contain either string.

### Changes
- **Pre- and Post-Stop Snapshots**:
  - Takes a pre-stop snapshot to record live wrapper PIDs before Task Scheduler terminates them.
  - Takes a post-stop snapshot after `Stop-ScheduledTask` to capture all children and grandchildren spawned before or during task shutdown.
- **Type-Safe `[int]` Indexing**:
  - Explicitly casts all `ProcessId` and `ParentProcessId` to `[int]` so `.NET` hashtables accurately match keys during BFS descendant tree traversal.
- **Comprehensive Residual Verification**:
  - Verifies whether any PID in `$targets` survived the kill.
  - Re-snapshots the process tree to verify no remaining processes match the lane's command line or are descendants of surviving lane processes.
- **Reproduction Evidence**:
  - Replicated the bug on a test task where a delayed child process survived pre-fix `lane-stop.ps1` and continued running despite reporting `residual=0`.
  - Verified that post-fix `lane-stop.ps1` successfully identifies, terminates, and confirms zero residual processes for the exact same scenario.

PR #720, opened 2026-09-02
fix(fleet): review transport pi → edda dispatch claude — Opus pinned, UUID sessions, narrowed closing-keyword rule (GH-708)

fagemx:
Fixes the shipped review path so it actually runs Claude Opus, per operator ruling 审查一律用 Claude Opus (`fleet.review-engine-model` / `fleet.review-backend`, superseding the review half of `fleet.agent-model-split`).

Issue: #708

Why `Issue:` and not a closing keyword: at this PR's base SHA the in-force spec is `review-spec-v1`, whose U2 still blanket-bans closing keywords — the exact rule this PR narrows to `pr.closing-keyword=only-when-all-donewhen-delivered` (#699). doneWhen items 1–2 and 4–6 are delivered in this diff; item 3's FAIL half is delivered verbatim (receipts below). Its remaining half — 修改後同一條指令產出真實裁決 — is now delivered: `scripts/review-pr.sh 720 4 --sha 7571e68694ea00f56646a1db8f30e91a4cd6a969` was run end to end (no `--dry-run`) and produced a real verdict, [posted on this PR](https://github.com/fagemx/edda/pull/720#issuecomment-5518986059) — 「Round 4 review」, LGTM (P0=0, P1=0), provisional per §6.4. Receipts: `.done` = `TRANSPORT=edda-dispatch` / `DISPATCH_EXIT=0`; brief 4162 chars → the `edda dispatch --agent claude` arm; model `claude-opus-5` requested **and** observed; cost $2.57; session `aa03b77b-18d2-40f7-ac2b-7a443d7ce452`. The comment was posted by hand in the exact `pr-review-watch.sh` format (watcher daemon not running); full receipts in `Review Response: Round 2`. Once the new U2 is in force this PR qualifies for `closes #708`. Using the closing keyword now would be flagged P1 by the very rule being fixed — the same mis-blocking that hit #687/#695/#696/#698/#700.

## What changed

[...]

PR #719, opened 2026-09-02
fix(tests): make workspace- and store-dependent tests hermetic (GH-646)

fagemx:
Issue: #646

Rebased onto current `main` (`b5dfb21`) for round 3 (pinned to `ba2cc73642ec883af569f1dc645029ff9f44dfcb`), resolving the conflict the
round-1 review found, fixing the Windows ceiling defect, adding `#[doc(hidden)]` on `find_root_bounded`, and canonicalizing test comparison for macOS symlink paths. See the
`Review Response: Round 2` comment for the per-finding answers.

## doneWhen × 3 — all delivered

1. **The 11 environment-dependent tests pass in both home-directory states** — see verification matrix below.
2. **Tests no longer write into workspaces they did not create** — every premise is established by the test itself (anchored `.edda/`, bounded walk, own pid dir, redirected store).
3. **Regression guards** so the next env-dependent test gets blocked:
   - `edda-ledger`: `EddaPaths::find_root_bounded(start, ceiling)` — `#[doc(hidden)]` test-facing bounded walk, plus guard tests (`find_root_bounded_stops_at_ceiling_and_ignores_workspace_above_it` and `find_root_bounded_handles_canonical_start_with_raw_ceiling`) that plant a hostile workspace *above* the ceiling and prove it stays invisible, and prove canonical start with raw ceiling behaves correctly.
   - `edda-cli`: `cmd_claim::e2e_repo()` — doc-comment-mandated helper; every e2e test MUST go through it (anchors `.edda` + `.git` at the fixture so the spawned `edda` resolves the test's own repo). `cmd_reconcile::canonical_main_repo_bounded` bounds the workspace walk at the fixture root.
   - `edda-bridge-claude`: `dispatch::tests::hermetic_unopenable_workspace()` — mandated helper for any future `render_workspace_section` test.

## 修改前的紅清單（issue 所列 11 + base 實測）

Issue（家目錄有 `.edda` 時）:

- `edda-ledger`: `paths::find_root_non_git_no_edda_returns_none`
- `edda-cli`: `cmd_claim` ×4、`cmd_reconcile` ×3
- `edda-bridge-claude`: `dispatch::render_workspace_section…`

[...]

PR #754, opened 2026-09-03
fix(fleet): one resumable reviewer conversation per PR (GH-708)

fagemx:
Issue: #708

Second tranche of #708. #720 moved the shipped review path onto Claude Opus; the operator's scope-addition comments on the issue (2026-09-03 02:25Z / 02:37Z) are what remained: **the per-PR reviewer session must resume**, and the review worktree must not accumulate.

## Why #720 could not keep the resumable session

`fleet.reviewer-agent=pi-with-per-pr-resumable-session` chose pi for one measured property — round 2+ reads only the delta instead of the whole PR. The claude transport as merged gives that up: `review-pr.sh` mints a **fresh UUID v4 per launch**, and its comment says so outright ("rounds are SHA-pinned independent reviews, not resumed conversations").

It had no alternative, because two things were missing:

```
$ edda dispatch --agent claude --session-id 0dc98ad1-…-000000000001 --prompt-file p2.txt --json
{"outcome":"crash","error":"agent exited with code 1 without result", …}
```

Claude Code refuses a `--session-id` that already exists (`Session ID <id> is already in use`), and `edda dispatch` had no `--resume`. So a stable per-PR id under the merged code would have crashed **every round after the first** — the pre-fix FAIL receipt for this half of the issue. (The first half's receipt still reproduces verbatim: `pi -p --model openrouter/anthropic/claude-opus-5 …` → `404 … quantization: fp8`, rc=1.)

The `--session-id` help text also claimed all three backends resume on a repeated id. They do not; claude is the exception.

## What changed

**`edda dispatch --resume`** (claude only, requires `--session-id`) spawns `claude --resume <id>` in place of `--session-id <id>`. pi and codex resume by repeating the id, so they **refuse** the flag rather than accept a switch that does nothing — the GH-574 honesty gate, one more row. `edda conduct` passes `resume: false`: its ids are per plan/phase/attempt, and a retry changes the attempt.

**`review-pr.sh` derives the reviewer session id from the PR number** — `SHA-1("edda-review-pr<N>")` in the RFC 4122 §4.3 v5 layout, e.g. PR #740 → `435cc101-9d69-56ca-878d-e454d6725f1a`. Stable across machines, rounds and reruns; cannot collide with an implementer's session (`review.independence-policy`).

**New-vs-resume is decided from the on-disk transcript, not the round number.** Round 2 is round 2 both when round 1 produced a session and when round 1 died before creating one; only `~/.claude/projects/*/<uuid>.jsonl` distinguishes them. Both spellings are generated correctly, and they differ per arm — `edda dispatch` keeps `--session-id` and *adds* `--resume`; `claude -p` *replaces* it.

**The reviewer session is observed, not echoed.** `MonitorResult` now carries the `session_id` from claude's stream-json `system/init` (the message was already parsed; the field just never surfaced), `AgentLauncher::last_observed_session()` exposes it on the same terms as `last_observed_model()`, and `edda dispatch` prints `Session observed:` / `session_observed` beside the requested id. Both lane arms record both lines and the watcher compares them, so a `--resume` that forked — `claude --resume` "starts a copy and says so when the session is already running" — is reported (`BACKEND REPORTED …`) instead of being invisible.

**Worktree lifecycle**: the lane removes `wt-review-pr<N>` once the round's verdict is in the log, and `review-pr.sh` prunes stale registrations before adding one at the same path. Resume is cwd-independent (below), so recreating it next round is safe.

## Verification (RAN, this workstation)

| Claim | Command | Result |
|---|---|---|
| pre-fix: repeating `--session-id` crashes | `edda dispatch --agent claude --session-id <id> …` (2nd turn) | `outcome=crash`, `agent exited with code 1 without result` |
| the fix resumes | same + `--resume` | `outcome=done`, `result_text="DISPATCH-708"` — round 1's codeword recalled |
| same conversation, not a fork | `find ~/.claude/projects -name '<id>.jsonl'` | one file, grown to 23 lines |
| delta rounds are cheaper | round 1 vs round 2 `cost_usd` | **$0.2154 → $0.0245** (and $0.22 → $0.02 on the second pair) |
| model pinned, requested + observed | `--json` | `model_requested=claude-opus-5`, `model_observed=claude-opus-5` |
| the observed session is a real observation | `claude -p --resume <A> --fork-session --output-format stream-json` | `system/init` reported `0aa32ab5-…` while `<A>` was `0dc98ad1-…` — the field diverges from the argument, so the fork guard is reachable |
| resume survives worktree deletion | delete + recreate the cwd, then `--resume` | codeword recalled |
| resume is cwd-independent | `--resume` from an unrelated directory | codeword recalled, same transcript grew |
| worktree removal takes the Windows path form | `git worktree remove --force "$(cygpath -w …)"` against a `/`-registered worktree | rc=0, worktree gone (round 1 P2 disproved by measurement) |
| pi/codex refuse the flag | `edda dispatch --agent pi --resume …` | `agent "pi" does not support: --resume true` |

Gates on `d6701ab`: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (`CARGO_INCREMENTAL=0`), `sh scripts/lint-markdown-content.sh`, `sh scripts/test-review-pr.sh` (19 cases), `sh scripts/test-pr-review-watch.sh`. Baseline for the two shell suites was recorded green on `df3bd4d` before the first edit. `edda-bridge-claude`'s env-var tests fail on this machine regardless of this diff (GH-646 / #719, non-hermetic under concurrent agents; the crate depends on neither touched crate, and the failure count varies run to run).

New offline fixtures: `D9a`–`D9j` in `test-review-pr.sh` (derivation pinned to a literal so a changed name string cannot silently fork every PR's conversation; round-1 opens, round-2 resumes, both arms' spellings, the `.done` receipts, the brief's `reviewer_session`, both session lines, worktree removal + an anchored prune assertion) and four watcher cases (resumed header, fork detection driven by the observed line, no-observation, missing receipt). Rust: 4 new tests for the flag, the argv selection and the requested-vs-observed split.

## Round 1 review (isolated sub-agent, `Changes Requested`, P0=0 P1=1) — answered in `d6701ab`

- **P1** the `reviewer_session` cross-check compared the launched id against an echo of itself on the `edda dispatch` arm. Confirmed at `cmd_dispatch.rs:430 → :590 → :264`. Fixed by making the observed id a real in-band reading (above), with the fork case demonstrated against a live backend.
- **P2** the installed `edda` has no `--resume` → runbook now states the lane needs an `edda` built at or after this commit; an older one fails the resume round loudly (`review:unreviewed`), never with a false verdict.
- **P2** `D9i` asserted text that a comment could satisfy → anchored to the command line. The related worry that `git worktree remove` might silently no-op on the `cygpath -w` path was measured and disproved.
- **P2** `exit 1` inside `SID=$(review_session_uuid …)` only left the subshell → `|| exit 1`.

## Out of scope, resolved by the operator

1. **`fleet.review-provider-overload`** — ruled 2026-09-03: not a contradiction, an outdated reading. Opus is the *default* engine; `fleet.review-engine-pool`'s anchor is still `gpt-5.6-sol` via pi, and codex has not been a review transport since `fleet.reviewer-agent`. Recorded as `opus-default-sol-via-pi-fallback-no-codex`; the watcher itself still never changes model — dropping to the anchor is an operator action.
2. **`REVIEW.md` §7 does not name `reviewer_session`** — correctly not bundled here; tracked as #756 (docs lane: add the field, bump the spec version, assert it in fixtures).
3. **The 14 stale `wt-review-pr*` worktrees** predating this change — swept: 12 merged removed, the one belonging to open #717 left untouched.

### DIFF
--- a/crates/edda-cli/src/agent_kind.rs
+++ b/crates/edda-cli/src/agent_kind.rs
@@ -82,6 +82,17 @@ impl AgentKind {
     pub fn supports_model_listing(self) -> bool {
         matches!(self, AgentKind::Pi)
     }
+
+    /// Whether the backend needs a distinct "continue this conversation"
+    /// spelling (`claude --resume <id>`). Only claude does: it refuses a
+    /// `--session-id` that already exists ("Session ID <id> is already in
+    /// use", exit 1), so repeating the id is a hard failure rather than a
+    /// resume. pi and codex resume by repeating `--session-id` — codex
+    /// through the persisted session→thread map (GH-535) — so `--resume`
+    /// there would claim a switch that does not exist (GH-708).
+    pub fn supports_resume(self) -> bool {
+        matches!(self, AgentKind::Claude)
+    }
 }
 
 // ── Backend × option support matrix (GH-574) ──
@@ -99,6 +110,9 @@ pub(crate) struct DispatchOptions<'a> {
     /// absent — there is no clap default masking explicitness, so nothing
     /// is claimed or dropped for backends that ignore the concept.
     pub permission_mode: Option<&'a str>,
+    /// `--resume`: continue the conversation the session id already names
+    /// (claude only; GH-708).
+    pub resume: bool,
 }
 
 /// Reject unsupported backend/option combinations with an explicit error.
@@ -139,6 +153,11 @@ pub(crate) fn validate_dispatch_options(
             options.permission_mode.map(|m| format!("{m:?}")),
             agent.supports_permission_mode(),
         ),
+        (
+            "--resume",
+            options.resume.then(|| "true".to_owned()),
+            agent.supports_resume(),
+        ),
     ];
     let refused: Vec<String> = unsupported
         .into_iter()
@@ -178,6 +197,10 @@ pub(crate) struct LauncherOptions {
     /// pi `--session-dir` (GH-574). Other backends manage their own session
     /// storage; dispatch rejects the flag for them before reaching here.
     pub session_dir: Option<PathBuf>,
+    /// Continue the conversation the session id already names, instead of
+    /// starting a new one (claude `--resume`; GH-708). Refused for the
+    /// other backends before reaching here.
+    pub resume: bool,
 }
 
 /// Construct and probe (`verify_available`) the launcher for `agent`.
@@ -194,7 +217,9 @@ pub(crate) fn build_launcher(
 ) -> Result<Box<dyn AgentLauncher>> {
     Ok(match agent {
         AgentKind::Claude => {
-            let mut launcher = ClaudeCodeLauncher::new().with_verbose(options.verbose);
+            let mut launcher = ClaudeCodeLauncher::new()
+                .with_verbose(options.verbose)
+                .with_resume(options.resume);
             launcher.transcript_dir = options.transcript_dir;
             launcher.verify_available()?;
             Box::new(launcher)
@@ -251,6 +276,7 @@ mod tests {
             exclude_tools,
             session_dir,
             permission_mode,
+            resume: false,
         }
     }
 
--- a/crates/edda-cli/src/cmd_conduct.rs
+++ b/crates/edda-cli/src/cmd_conduct.rs
@@ -241,6 +241,10 @@ pub fn run(
             // Conduct has no session-dir surface (GH-574); pi uses its own
             // default session storage under conduct.
             session_dir: None,
+            // Conduct's session ids are deterministic per plan/phase/attempt
+            // and each attempt is a fresh conversation, so it never resumes
+            // (GH-708). Retries change the attempt, and therefore the id.
+            resume: false,
         },
     )?;
     let engine = CheckEngine::new(cwd.clone());
--- a/crates/edda-cli/src/cmd_dispatch.rs
+++ b/crates/edda-cli/src/cmd_dispatch.rs
@@ -163,6 +163,13 @@ pub struct DispatchOutput {
     /// when it reported nothing. Observed, never inferred from config or
     /// session files (GH-574 honesty rule).
     pub model_observed: String,
+    /// The session id the backend reported in-band, or "unknown" when it
+    /// reported nothing. `session_id` above is what edda ASKED for; this is
+    /// what the backend says it ran. Keeping them apart is the whole point:
+    /// `claude --resume <id>` "starts a copy and says so when the session is
+    /// already running", so echoing the requested id back would report every
+    /// fork as a clean resume (GH-708).
+    pub session_observed: String,
 }
 
 impl DispatchOutput {
@@ -171,6 +178,7 @@ impl DispatchOutput {
         session_id: String,
         model_requested: String,
         model_observed: String,
+        session_observed: String,
     ) -> Self {
         match result {
             PhaseResult::AgentDone {
@@ -184,6 +192,7 @@ impl DispatchOutput {
                 error: None,
                 model_requested,
                 model_observed,
+                session_observed,
             },
             PhaseResult::AgentCrash { error } => Self {
                 outcome: Outcome::Crash,
@@ -193,6 +202,7 @@ impl DispatchOutput {
                 error: Some(error),
                 model_requested,
                 model_observed,
+                session_observed,
             },
             PhaseResult::Timeout => Self {
                 outcome: Outcome::Timeout,
@@ -202,6 +212,7 @@ impl DispatchOutput {
                 error: None,
                 model_requested,
                 model_observed,
+                session_observed,
             },
             PhaseResult::MaxTurns { cost_usd } => Self {
                 outcome: Outcome::MaxTurns,
@@ -211,6 +222,7 @@ impl DispatchOutput {
                 error: None,
                 model_requested,
                 model_observed,
+                session_observed,
             },
             PhaseResult::BudgetExceeded { cost_usd } => Self {
                 outcome: Outcome::BudgetExceeded,
@@ -220,6 +232,7 @@ impl DispatchOutput {
                 error: None,
                 model_requested,
                 model_observed,
+                session_observed,
             },
         }
     }
@@ -240,7 +253,8 @@ impl DispatchOutput {
 
     /// Text-mode stdout: an `Outcome:` line whenever the turn did not
     /// finish, then result text, then `Cost:`, then the model story
-    /// (GH-574), then `Session:`.
+    /// (GH-574), then the session story — `Session:` (asked for) and
+    /// `Session observed:` (what the backend reported, GH-708).
     ///
     /// A failed turn used to render exactly like a successful one — same
     /// `Cost:`/`Session:` summary, nothing naming the failure (GH-669) — so
@@ -262,6 +276,7 @@ impl DispatchOutput {
         out.push_str(&format!("Model requested: {}\n", self.model_requested));
         out.push_str(&format!("Model observed: {}\n", self.model_observed));
         out.push_str(&format!("Session: {}\n", self.session_id));
+        out.push_str(&format!("Session observed: {}\n", self.session_observed));
         out
     }
 
@@ -275,6 +290,7 @@ impl DispatchOutput {
             "error": self.error,
             "model_requested": self.model_requested,
             "model_observed": self.model_observed,
+            "session_observed": self.session_observed,
         })
         .to_string()
     }
@@ -585,11 +601,15 @@ pub async fn run_with_launcher(
     let model_observed = launcher
         .last_observed_model()
         .unwrap_or_else(|| "unknown".to_owned());
+    let session_observed = launcher
+        .last_observed_session()
+        .unwrap_or_else(|| "unknown".to_owned());
     Ok(DispatchOutput::from_result(
         result,
         session_id.to_owned(),
         model_requested,
         model_observed,
+        session_observed,
     ))
 }
 
@@ -739,6 +759,7 @@ mod tests {
             "s".into(),
             "inherited".into(),
             "unknown".into(),
+            "unknown".into(),
         );
         let text = out.render_text();
         assert!(text.contains("Model requested: inherited"), "{text}");
@@ -758,6 +779,7 @@ mod tests {
             "s".into(),
             "openai-codex/gpt-5.6-sol".into(),
             "openai-codex/gpt-5.6-sol".into(),
+            "unknown".into(),
         );
         let text = out.render_text();
         assert!(
@@ -787,9 +809,15 @@ mod tests {
             "s".into(),
             "openai-codex/gpt-5.6-sol".into(),
             "unknown".into(),
+            "unknown".into(),
         );
-        let without_model =
-            DispatchOutput::from_result(result, "s".into(), "inherited".into(), "unknown".into());
+        let without_model = DispatchOutput::from_result(
+            result,
+            "s".into(),
+            "inherited".into(),
+            "unknown".into(),
+            "unknown".into(),
+        );
         assert_ne!(
             with_model.render_text(),
             without_model.render_text(),
@@ -932,12 +960,14 @@ mod tests {
             }),
             result_text: Some(reason.into()),
             model: Some("claude-opus-5[1m]".into()),
+            session_id: None,
         };
         let out = DispatchOutput::from_result(
             classify_result(&monitor, Some(1)),
             "0e92629d-1f2e-597e-90ec-662e206efcde".into(),
             "inherited".into(),
             "claude-opus-5[1m]".into(),
+            "unknown".into(),
         );
 
         assert_eq!(out.exit_code(), 1, "{out:?}");
@@ -986,6 +1016,7 @@ mod tests {
                 "s".into(),
                 "inherited".into(),
                 "unknown".into(),
+                "unknown".into(),
             );
             assert_eq!(
                 exit_code_for(out.outcome),
@@ -1006,6 +1037,7 @@ mod tests {
             },
             "s".into(),
             "inherited".into(),
+            "unknown".into(),
             "unknown".into(),
         );
         assert_eq!(out.exit_code(), 0);
@@ -1043,6 +1075,7 @@ mod tests {
                 "sess-1".into(),
                 "inherited".into(),
                 "unknown".into(),
+                "unknown".into(),
             );
             let value: serde_json::Value =
                 serde_json::from_str(&out.to_json()).expect("json parses");
@@ -1067,7 +1100,8 @@ mod tests {
                     "model_requested",
                     "outcome",
                     "result_text",
-                    "session_id"
+                    "session_id",
+                    "session_observed"
                 ]
             );
         }
@@ -1081,6 +1115,7 @@ mod tests {
             },
             "s".into(),
             "inherited".into(),
+            "unknown".into(),
             "unknown".into(),
         );
         let value: serde_json::Value = serde_json::from_str(&crash.to_json()).unwrap();
@@ -1096,6 +1131,7 @@ mod tests {
             },
             "s".into(),
             "inherited".into(),
+            "unknown".into(),
             "unknown".into(),
         );
         let value: serde_json::Value = serde_json::from_str(&done.to_json()).expect("json parses");
@@ -1117,6 +1153,7 @@ mod tests {
             "abc-123".into(),
             "inherited".into(),
             "unknown".into(),
+            "unknown".into(),
         );
         let text = out.render_text();
         assert!(text.contains("done deal\n"), "{text}");
@@ -1135,6 +1172,7 @@ mod tests {
             PhaseResult::MaxTurns { cost_usd: None },
             "s".into(),
             "inherited".into(),
+            "unknown".into(),
             "unknown".into(),
         );
         let text = out.render_text();
@@ -1165,9 +1203,62 @@ mod tests {
             "my-session-42".into(),
             "inherited".into(),
             "unknown".into(),
+            "unknown".into(),
         );
         assert!(out.render_text().contains("Session: my-session-42"));
         assert!(out.to_json().contains("my-session-42"));
+    }
+
+    /// GH-708: the requested id and the observed one are separate fields
+    /// with separate sources, so a `--resume` that forked instead of
+    /// continuing is visible. Echoing the request into `session_observed`
+    /// would report every fork as a clean resume.
+    #[test]
+    fn requested_and_observed_session_ids_are_reported_separately() {
+        let forked = DispatchOutput::from_result(
+            PhaseResult::AgentDone {
+                cost_usd: None,
+                result_text: None,
+            },
+            "asked-for-42".into(),
+            "inherited".into(),
+            "unknown".into(),
+            "actually-ran-99".into(),
+        );
+        let text = forked.render_text();
+        assert!(text.contains("Session: asked-for-42"), "{text}");
+        assert!(text.contains("Session observed: actually-ran-99"), "{text}");
+        assert!(forked
+            .to_json()
+            .contains("\"session_observed\":\"actually-ran-99\""));
+    }
+
+    /// A backend that reports no session id renders "unknown", never the
+    /// requested id — the same honesty rule `model_observed` follows.
+    #[tokio::test]
+    async fn a_silent_backend_reports_an_unknown_observed_session() {
+        let launcher = MockLauncher::new();
+        let phase = build_phase(
+            "prompt",
+            None,
+            None,
+            "bypassPermissions",
+            CapabilityOptions::default(),
+        );
+        let out = run_with_launcher(
+            &launcher,
+            &phase,
+            "asked-for-42",
+            Path::new("."),
+            CancellationToken::new(),
+        )
+        .await
+        .expect("mock run");
+        assert_eq!(out.session_id, "asked-for-42");
+        assert_eq!(
+            out.session_observed, "unknown",
+            "a launcher that observed nothing must not echo the requested id"
+        );
     }
 
     /// Records the session ids it receives. This proves verbatim delivery
--- a/crates/edda-cli/src/cmd_dispatch.rs
+++ b/crates/edda-cli/src/cmd_dispatch.rs
@@ -35,12 +35,22 @@ pub struct DispatchArgs {
     pub prompt_file: Option<String>,
     /// Session id passed to the backend verbatim. Continuity semantics are
     /// per-backend, and all three persist conversations across invocations:
-    /// claude and pi delegate to their backends, and codex's session→thread
-    /// map is persisted in the per-user edda store (GH-535), so the same id
-    /// resumes the prior codex conversation. Generated and printed when
-    /// omitted so the caller can reuse it on the next call.
+    /// pi delegates to its backend, and codex's session→thread map is
+    /// persisted in the per-user edda store (GH-535), so for both the same
+    /// id resumes the prior conversation. claude is the exception — a
+    /// `--session-id` that already exists is refused ("Session ID <id> is
+    /// already in use"), so a second turn on the same conversation adds
+    /// `--resume` (GH-708). Generated and printed when omitted so the
+    /// caller can reuse it on the next call.
     #[arg(long)]
     pub session_id: Option<String>,
+    /// Continue the conversation `--session-id` names instead of starting a
+    /// new one (claude only, `claude --resume <id>`; GH-708). pi and codex
+    /// resume by repeating `--session-id` alone and refuse this flag rather
+    /// than accept a switch that does nothing. Requires --session-id: there
+    /// is nothing to resume without one.
+    #[arg(long, requires = "session_id")]
+    pub resume: bool,
     /// Working directory for the agent (default: current directory)
     #[arg(long)]
     pub cwd: Option<String>,
@@ -381,6 +391,7 @@ fn run_inner(args: DispatchArgs) -> Result<i32> {
             exclude_tools: args.exclude_tools.as_deref(),
             session_dir: args.session_dir.as_deref(),
             permission_mode: args.permission_mode.as_deref(),
+            resume: args.resume,
         },
     )?;
 
@@ -451,6 +462,7 @@ fn run_inner(args: DispatchArgs) -> Result<i32> {
             // recorded.
             persistent_codex_threads: true,
             session_dir: args.session_dir.as_ref().map(std::path::PathBuf::from),
+            resume: args.resume,
         },
     )?;
     let phase = build_phase(
@@ -1373,6 +1385,68 @@ mod tests {
             .expect_err("codex has no permission-mode concept; the value must be refused");
         assert!(error.to_string().contains("--permission-mode"), "{error}");
         let _ = std::fs::remove_dir_all(&dir);
+    }
+
+    // ── GH-708: --resume ──
+
+    #[test]
+    fn resume_requires_a_session_id_to_resume() {
+        // Without an id there is nothing to continue, and claude would pick
+        // "the most recent conversation in this directory" — which for a
+        // review lane is whatever ran there last, not this PR's reviewer.
+        assert!(
+            TestCli::try_parse_from([
+                "edda",
+                "--agent",
+                "claude",
+                "--prompt-file",
+                "p.txt",
+                "--resume"
+            ])
+            .is_err(),
+            "--resume without --session-id must not parse"
+        );
+        let args = parse(&[
+            "edda",
+            "--agent",
+            "claude",
+            "--prompt-file",
+            "p.txt",
+            "--session-id",
+            "7a9c6b1e-0000-4708-8000-000000000001",
+            "--resume",
+        ]);
+        assert!(args.resume);
+        assert_eq!(
+            args.session_id.as_deref(),
+            Some("7a9c6b1e-0000-4708-8000-000000000001")
+        );
+    }
+
+    #[test]
+    fn resume_is_refused_on_backends_that_resume_by_session_id_alone() {
+        for agent in ["pi", "codex"] {
+            let error = validate_dispatch_options(
+                match agent {
+                    "pi" => AgentKind::Pi,
+                    _ => AgentKind::Codex,
+                },
+                &DispatchOptions {
+                    resume: true,
+                    ..Default::default()
+                },
+            )
+            .expect_err("only claude needs a distinct resume spelling");
+            assert!(error.to_string().contains("--resume"), "{error}");
+        }
+        validate_dispatch_options(
+            AgentKind::Claude,
+            &DispatchOptions {
+                resume: true,
+                ..Default::default()
+            },
+        )
+        .expect("claude supports --resume");
     }
 
     // ── End-to-end-ish run through MockLauncher ──
--- a/crates/edda-cli/src/main.rs
+++ b/crates/edda-cli/src/main.rs
@@ -484,10 +484,14 @@ enum Command {
     ///   {"outcome":"done|crash|timeout|max_turns|budget_exceeded",
     ///    "result_text":string|null,"cost_usd":number|null,
     ///    "session_id":string,"error":string|null,
-    ///    "model_requested":string,"model_observed":string}
+    ///    "model_requested":string,"model_observed":string,
+    ///    "session_observed":string}
     /// model_requested is the --model value or "inherited"; model_observed
-    /// is what the backend reported in-band or "unknown". --json cannot be
-    /// combined with --list-models (the listing is text) and is refused.
+    /// is what the backend reported in-band or "unknown". session_id is the
+    /// id edda asked for; session_observed is the one the backend reported
+    /// in-band, or "unknown" — they differ when a --resume forked instead of
+    /// continuing. --json cannot be combined with --list-models (the listing
+    /// is text) and is refused.
     Dispatch {
         #[command(flatten)]
         args: cmd_dispatch::DispatchArgs,
--- a/crates/edda-conductor/src/agent/launcher.rs
+++ b/crates/edda-conductor/src/agent/launcher.rs
@@ -51,6 +51,18 @@ pub trait AgentLauncher: Send + Sync {
     fn last_observed_model(&self) -> Option<String> {
         None
     }
+
+    /// The session id the backend reported **in-band** during the most
+    /// recent turn, on the same terms as [`Self::last_observed_model`]: an
+    /// observation, never the id the caller asked for. It matters because
+    /// asking is not getting — `claude --resume <id>` "starts a copy and
+    /// says so when the session is already running", so only the reported
+    /// id says whether a resume continued the conversation or forked it.
+    /// `None` means the backend reported nothing and callers must render
+    /// `"unknown"` (GH-708).
+    fn last_observed_session(&self) -> Option<String> {
+        None
+    }
 }
 
 /// Fixed namespace UUID for conductor sessions.
@@ -88,6 +100,9 @@ pub struct ClaudeCodeLauncher {
     /// In-band model report from the most recent turn (stream-json
     /// `system/init`), for [`AgentLauncher::last_observed_model`].
     observed_model: std::sync::Mutex<Option<String>>,
+    /// In-band session id from the same `system/init` message, for
+    /// [`AgentLauncher::last_observed_session`] (GH-708).
+    observed_session: std::sync::Mutex<Option<String>>,
 }
 
 impl Default for ClaudeCodeLauncher {
@@ -194,6 +209,7 @@ impl ClaudeCodeLauncher {
             transcript_dir: None,
             resume: false,
             observed_model: std::sync::Mutex::new(None),
+            observed_session: std::sync::Mutex::new(None),
         }
     }
 
@@ -204,6 +220,7 @@ impl ClaudeCodeLauncher {
             transcript_dir: None,
             resume: false,
             observed_model: std::sync::Mutex::new(None),
+            observed_session: std::sync::Mutex::new(None),
         }
     }
 
@@ -217,6 +234,12 @@ impl ClaudeCodeLauncher {
     fn record_observed_model(&self, model: Option<String>) {
         if let Ok(mut slot) = self.observed_model.lock() {
             *slot = model;
+        }
+    }
+
+    fn record_observed_session(&self, session: Option<String>) {
+        if let Ok(mut slot) = self.observed_session.lock() {
+            *slot = session;
         }
     }
 
@@ -288,6 +311,10 @@ impl AgentLauncher for ClaudeCodeLauncher {
                 // In-band model observation: whatever the backend itself
                 // reported, or nothing. Never inferred from config.
                 self.record_observed_model(monitor_result.model.clone());
+                // Same terms for the session id: what claude says it ran,
+                // which is the only way to see a --resume that forked
+                // instead of continuing (GH-708).
+                self.record_observed_session(monitor_result.session_id.clone());
                 let exit = child.wait().await?;
                 Ok(classify_result(&monitor_result, exit.code()))
             }
@@ -305,6 +332,13 @@ impl AgentLauncher for ClaudeCodeLauncher {
 
     fn last_observed_model(&self) -> Option<String> {
         self.observed_model
+            .lock()
+            .ok()
+            .and_then(|slot| slot.clone())
+    }
+
+    fn last_observed_session(&self) -> Option<String> {
+        self.observed_session
             .lock()
             .ok()
             .and_then(|slot| slot.clone())
@@ -621,6 +655,17 @@ mod tests {
 
     /// GH-574: the in-band observation starts empty and is only filled by
     /// what the backend itself reports — never from config or session files.
+    /// GH-708: the same rule for the session id. A launcher that has run
+    /// nothing has observed nothing — it must not report the id a caller
+    /// would have passed, or a forked `--resume` would look like a clean one.
+    #[test]
+    fn claude_observed_session_starts_unknown() {
+        let launcher = ClaudeCodeLauncher::with_bin(PathBuf::from("claude"));
+        assert_eq!(launcher.last_observed_session(), None);
+        let resuming = ClaudeCodeLauncher::with_bin(PathBuf::from("claude")).with_resume(true);
+        assert_eq!(resuming.last_observed_session(), None);
+    }
+
     #[test]
     fn claude_observed_model_starts_unknown() {
         let launcher = ClaudeCodeLauncher::with_bin(PathBuf::from("claude"));
--- a/crates/edda-conductor/src/agent/launcher.rs
+++ b/crates/edda-conductor/src/agent/launcher.rs
@@ -78,6 +78,13 @@ pub struct ClaudeCodeLauncher {
     pub verbose: bool,
     /// If set, raw agent stdout is captured to `{transcript_dir}/{phase_id}-{session_id_prefix}.jsonl`.
     pub transcript_dir: Option<PathBuf>,
+    /// Continue the conversation the session id already names instead of
+    /// starting a new one: `claude --resume <id>` rather than
+    /// `claude --session-id <id>` (GH-708). Claude Code refuses a
+    /// `--session-id` that already exists ("Session ID <id> is already in
+    /// use", exit 1), so a caller that wants a second turn on the same
+    /// conversation has no other spelling.
+    pub resume: bool,
     /// In-band model report from the most recent turn (stream-json
     /// `system/init`), for [`AgentLauncher::last_observed_model`].
     observed_model: std::sync::Mutex<Option<String>>,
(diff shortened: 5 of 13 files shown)


fagemx:
> **Round 1 review** — **not** the shipped watcher: `edda-pr-review-watcher` is not registered on this workstation, so this round ran as an **isolated sub-agent** dispatched from the implementer's Claude Code session (Claude Opus, its own git worktree, read-only by instruction — REVIEW.md §0 — with no edit/push/GitHub-post performed). Provenance is stated rather than dressed up as the automated path: `model_observed` is therefore `unverified`, and `reviewer_session` is `subagent-isolated`, not a UUID from the lane. Reviewed head SHA: `879b6611739deff3b20709e5a074918dd0b98d56`.

## Code Review: Round 1 — PR #754 @ 879b6611739deff3b20709e5a074918dd0b98d56

- model_observed: unverified
- reviewer_session: subagent-isolated
- spec: review-spec-v1.2
- escalations: D5 (wording/structure — `[判斷]`, brief does not name me qualified for this class)
- cost: ~28 tool calls, ~12 min elapsed; no cargo run (CI at exact head READ instead)

**Worktree caveat, stated before the review:** my worktree HEAD is `df3bd4d1` — the PR's **base**, not the head. The head commit object is present in the object store (branch `fix/gh708-resumable-review-session` is checked out in the main worktree), so every file, diff and `sh -n` below was taken at `879b661` via `git show 879b661:<path>` into a scratch dir. No conclusion below rests on the base tree; where a check could only run against the base tree (U5) I say so and cite CI instead.

### IN SCOPE
Changed behaviour: a new `edda dispatch --resume` (claude-only, `requires = "session_id"`) that flips `ClaudeCodeLauncher` from `--session-id <id>` to `--resume <id>`; a PR-number-derived v5 reviewer session id in `review-pr.sh`; new-vs-resume decided from the on-disk transcript glob; per-round worktree removal plus `git worktree prune`; new `SESSION=`/`SESSION_MODE=` `.done` receipts and the watcher's `reviewer_session` header. Direct consumers: `cmd_dispatch.rs`, `cmd_conduct.rs`, `agent_kind.rs` validation table, both generated lane arms, `pr-review-watch.sh`. Acceptance ceiling: issue #708 body doneWhen plus the operator's 2026-09-03 scope-addition comments. Current-base integration: exact-head CI.

### Rules
|---|---|---|---|---|
| U1 | any | P0 | PASS | `git diff --name-only` → 11 files, exactly the brief's allowed surface; no file outside it |
| U2 | any | P1 | PASS | body grep → 0 hits; `git log --format=%B base..head` grep → 0 hits. Correct: #708 is partly delivered (#720 was tranche 1) and item 6 stays open for the operator, so no closing keyword is right |
| U3 | any | P1 | PASS | `Issue: #708` present |
| U4 | any | P1 | PASS | one subject, `fix(fleet): one resumable reviewer conversation per PR (GH-708)`; offender grep empty |
| U5 | any | P1 | PASS | CI `Format` job (which runs `lint-markdown-content.sh`) = success at `879b661`. My local run `exit=0` was against the base tree — cited as corroboration only |
| U6 | any | P0 | PASS | `check-runs` at `879b661`: 7 completed success (Format, Detect code changes, Clippy ×3, Test ubuntu, Test macos); `Test (windows-latest)` **in_progress**. Not deterministically red → audit, no full local rerun. See RAN vs READ for the coverage this leaves |
| U7 | any | P1 | PASS | whole scoped audit finished before this verdict; the single blocker is batched here |
| D1 | docs | P1 | PASS | added invocations: `edda dispatch --help` → exit=0 (want 0); `git worktree -h` → exit=129 (want 129) |
| D2 | docs | P1 | N.A.(ledger unreachable from this worktree) | `edda ask fleet.reviewer-agent` → `No results found. [repo] could not be checked`; `--fleet` → only "repo not on this machine" rows. Per D2 that is unreachable, not false. Checked against the durable carrier instead: issue #708's operator comment quotes `fleet.reviewer-agent=pi-with-per-pr-resumable-session` and `review.independence-policy` verbatim, and the PR's use of both matches |
| D3 | docs | P1 | PASS | 1 repo-anchored path added (`scripts/pr-review-watch.sh`) → `OK` |
| D4 | docs | P0 | PASS | authority grep → zero candidates |
| D5 | docs | `[判斷]` | 需升級 | judgement item; brief does not name me qualified for `code-risk`. Escalated per §6.1 |
| C1 | code-plain | P0 | PASS | every doneWhen mapped to code + test — see the doneWhen block below |
| C2 | code-plain | P0 | PASS | `fix:` with regression tests that fail without the change: `resume_requires_a_session_id_to_resume`, `resume_is_refused_on_backends_that_resume_by_session_id_alone`, `claude_default_names_a_new_session`, `claude_resume_continues_the_session_instead_of_naming_it` (4 `#[test]` adds), plus fixtures `D9a`–`D9i` and 3 watcher cases. `D9a` pins the derived id to a literal, so a changed name string fails loudly |
| C3 | code-plain | P1 | PASS | 3 `.expect(` candidates, all inside `#[cfg(test)] mod tests` (confirmed by opening both files) |
| C4 | code-plain | P1 | PASS | `unsafe`/blanket-allow grep → empty |
| C5 | code-plain | P1 | PASS | selector printed nothing: changed crates are `edda-cli` and `edda-conductor`, both inside the CI Windows 7-crate subset. No uncovered crate → no focused Windows check owed |
| R1 | code-risk | P0 | PASS | 2 real hits (+1 comment), triggers stated below |
| R2 | code-risk | `[判斷]` | N.A.(zero candidates) | mechanical enumeration of added lines mixing `\|\|` and `&&` returned nothing, so there is no parse tree to adjudicate — not an escalation, an empty precondition |
| R3 | code-risk | P0 | PASS | `sh -n` at head: `review-pr.sh` 0, `pr-review-watch.sh` 0, `test-review-pr.sh` 0, `test-pr-review-watch.sh` 0 |
| R4 | code-risk | P1 | PASS | no `mktemp`, so no `trap … 0` owed; `$SPEC` is a deliberate per-round file in `$SCRATCH`. The worktree is the resource, and this PR *adds* its release path. Kill-path leak (task killed between dispatch and removal) is covered by the new `git worktree prune` at the next round's start. Note: removal is not in a `try/finally`, but `.done` is written first and the next round prunes |
| R5 | code-risk | P1 | PASS | `$SID` is `[0-9a-f-]` only by construction (SHA-1 digest + fixed layout), so it cannot escape the single quotes in either lane — strictly safer than the base's random-hex id. Caller SHA still validated `^[0-9a-f]{40}$`. Distinct PR/round → distinct `.done`/`.log`/lane filenames. The `$MODEL`/path single-quote interpolation is pre-existing at base (`df3bd4d:scripts/review-pr.sh:61,378,392`) and routed to FOLLOW-UP, not charged here |
| §5.5 | any | — | PASS w/ 1 defect | table below; the defect is Finding P1-1 |

**R1 trigger conditions (required by the rule):**
1. `git worktree remove --force '$WTW'` (PowerShell `Remove-ReviewWorktree`) — fires unconditionally at the end of **both** transport arms, after `.done` is written and after `$code = $LASTEXITCODE` is captured; guarded by `if ('$ROOTW' -eq '') { return }`. Target is the fixed derived path `$SCRATCH/wt-review-pr<N>`; no glob, no variable path segment. `git worktree remove` refuses a path that is not a registered worktree ("is not a working tree") and removes nothing, so it cannot delete an arbitrary directory. Output `2>&1 | Out-Null`; the round's exit code is the pre-captured `$code`, so a failure here cannot change it.
2. `git -C '$ROOT' worktree remove --force '$WT' >/dev/null 2>&1 || true` (POSIX runner) — same trigger, guarded by `[ -n '$ROOT' ] || return 0`, failure swallowed by `|| true`, `code=$?` captured immediately after dispatch.
3. `--force` justification: the launcher copies untracked `.edda-review-spec.md` into the tree. The tree is detached and read-only by contract, so `--force` cannot discard operator work.

**doneWhen (issue body + the operator's 2026-09-03 02:25Z / 02:37Z comments):** id derived deterministically from the PR number as a v5 UUID, identical every round, never the implementer's → `review_session_uuid()`, test `D9a`/`D9e`, and I recomputed it independently: `SHA-1("edda-review-pr9999")` → `0dc98ad1-d4ae-5321-b81b-b249644368c6` (matches the pinned literal), PR #740 → `435cc101-9d69-56ca-878d-e454d6725f1a` (matches the doc), PR #754 → `4dd35987-26e4-5499-bea4-30e21f839b0f` (matches the id in my own brief). Version nibble is the literal `5`; the variant `case` covers all 16 hex digits and computes `(d & 0x3) | 0x8` correctly in every arm, with a non-hex `*)` that errors. Round 1 opens / later rounds resume → verified by generating both lanes offline. Resume verified not assumed → PR-body receipts (one transcript grown to 23 lines; $0.2154 → $0.0245). Worktree path stable / cwd-independence recorded → PR body, both recorded. Header carries `reviewer_session` → brief line + watcher. Preferred transport `edda dispatch --agent claude` → used. Lifecycle nit → both removal *and* prune implemented. All met. The one deliberate divergence — deciding new-vs-resume from the on-disk transcript rather than from `[prev-sha]` as the comment's wording suggested — is documented in code and PR body and is strictly more correct (round 1 can die before creating a session); not a finding.

### Wiring
|---|---|---|---|---|
| `--resume` CLI flag | `cmd_dispatch.rs:52` `#[arg(long, requires = "session_id")]` | `cmd_dispatch.rs:394,465` → `DispatchOptions.resume` / `LauncherOptions.resume` | clap hard-errors without `--session-id`; `validate_dispatch_options` errors for pi/codex | flag→validate→builder→spawn, asserted on the argv by `claude_resume_continues_the_session_instead_of_naming_it` |
| `AgentKind::supports_resume` | `agent_kind.rs:92` | `agent_kind.rs:158` unsupported table (same file — the wiring-scan "0 outside references" is a false positive) | refusal message contains `--resume`, asserted by test | enum→matrix→error |
| `ClaudeCodeLauncher.resume` + `with_resume` | `launcher.rs:81,211` | `build_command` `launcher.rs:113-117` picks `--resume` vs `--session-id` | none needed — pure argv selection | field→build_command→argv, both branches asserted |
| `LauncherOptions.resume` | `agent_kind.rs:200` | `build_launcher` Claude arm only | pi/codex refused earlier, so no silent drop | struct→builder; `cmd_conduct.rs:244` explicitly `false` with a reason |
| `SESSION=` / `SESSION_MODE=` `.done` receipts | both lane arms, both platforms | `reviewer_session_desc` `pr-review-watch.sh:279-280` | missing receipt → literal `mode unknown — no SESSION_MODE receipt in .done`, never a guess | script→.done→watcher→PR comment; asserted by `D9c` and watcher case 3 |
| `reviewer_session_desc` → header field | `pr-review-watch.sh:278-294`, consumed at `:447` | the posted PR comment | **defective — see P1-1**: the mismatch branch cannot fire on the primary transport | .done + log→header |
| `session=` / `session_mode=` dry-run keys | `review-pr.sh:394-395` | `test-review-pr.sh` `field()` | test-only consumer, adequate | script→stdout→fixtures |
| `Remove-ReviewWorktree` / `remove_review_worktree` + `git worktree prune` | lane both arms; `review-pr.sh:601` | git worktree admin state | deliberately swallowed (`Out-Null` / `\|\| true`) so it cannot change the round's exit code — correct for cleanup, but it also means a silent no-op is invisible; see P2-2 | lane→git; `D9i` asserts the *text*, not the effect |

### Findings

- **[P1] The "backend reported" half of the new `reviewer_session` cross-check is an echo of the launched id, not an observation, on the transport that actually runs — so the fork detector the PR advertises is inert on the primary arm.** `scripts/pr-review-watch.sh:272-276` states `$2` "is the id the backend REPORTED in its own receipt", and `docs/guides/pr-review-watcher.md` states the header is compared against "log 裡後端自己回報的 `Session:` 行". `$2` is `SIDO`, read by `session_model_cost` from the log's `Session:` line. On `TRANSPORT=edda-dispatch` that line comes from `crates/edda-cli/src/cmd_dispatch.rs:264` printing `DispatchOutput.session_id`, which is set at `cmd_dispatch.rs:430` — `let session_id = args.session_id.clone().unwrap_or_else(generate_session_id);` — and passed through unchanged into `DispatchOutput::from_result(result, session_id.to_owned(), …)` at `cmd_dispatch.rs:590`. That is verbatim the `$SID` the lane also wrote to `.done` as `SESSION=`, so the guard at `pr-review-watch.sh:287` (`[ "$2" != "$sid" ]`) is false by construction and the `BACKEND REPORTED …` branch is unreachable whenever the primary arm runs. Contrast `model_observed`, which the same function derives from a genuine in-band report (`launcher.last_observed_model()`, `cmd_dispatch.rs:~578`) with an `unknown` fallback — there is no `last_observed_session()` counterpart. The check is real only on the `claude-stdin` fallback, where the lane writes `Session: $($r.session_id)` from claude's own JSON. The new fixture that certifies fork detection hand-writes a divergence (`SESSION=99999999-…` in `.done` against a fixture log pinned to `Session: 11111111-…`, `scripts/test-pr-review-watch.sh:482,545-555`) that the primary transport cannot produce, so it passes without exercising any reachable state. Two ways out, both small: give `edda dispatch` a real in-band session observation parallel to `last_observed_model()` and print *that* as `Session:`; or narrow the comment, the doc paragraph and the PR body to say the cross-check is meaningful only on the fallback arm and that the dispatch arm's value is launched-only.

- **[P2] The `edda` currently installed on this workstation has no `--resume`, so every resume round would die until it is rebuilt.** `edda --version` → `edda 0.4.0` at `/c/Users/fagem/.cargo/bin/edda`; `edda dispatch --help | grep -- '--resume'` → empty. The lane resolves bare `edda` off PATH, so after this PR merges, round 1 (`--session-id` only) still works and round 2+ (`--session-id … --resume`) exits 2 on an unknown flag. It fails loudly — no verdict block, `review:unreviewed` — so it cannot produce a false LGTM, which is why this is not a blocker. Worth one line in the runbook saying the lane now requires an `edda` built at or after this commit.

- **[P2] `D9i` asserts the removal's *text*, never its *effect*, and one of its three greps would match a commented-out line.** `grep -q 'worktree prune' "$root/scripts/review-pr.sh"` (`scripts/test-review-pr.sh`) matches the string anywhere in the file, including inside the comment block directly above the real call. The other two greps are sound — `grep -c 'Remove-ReviewWorktree$'` correctly counts the two bare call lines and excludes the `function Remove-ReviewWorktree {` definition. Nothing in the suite proves the removal actually removes anything; in particular nothing confirms that `git worktree remove` matches a worktree registered with forward slashes (`git worktree list` shows `C:/Users/fagem/.edda/fleet/wt-review-pr675`) when handed the backslash form `cygpath -w` produces. Because the call is `2>&1 | Out-Null` / `|| true`, a path-matching failure is completely silent and the only symptom is the stale worktrees this change exists to prevent. A one-shot manual check on a real round would settle it.

- **[P2] `review_session_uuid`'s three error branches `exit 1` inside a command substitution, which under `set -u` and no `set -e` yields an empty `SID` rather than aborting.** `review-pr.sh:65` sets `set -u` only; `SID=$(review_session_uuid "$PR")` runs the function in a subshell, so `exit 1` in the no-`sha1sum`/`shasum`, short-digest and non-hex arms terminates only that subshell and the script continues to build a lane with `--session-id ''`. This is the identical pattern the removed `gen_uuid` had, so it is carried over rather than introduced, and the trigger is remote on the target platform (`sha1sum` is present in Git Bash). Flagging it because the PR touched this exact function; `SID=$(…) || exit 1` would close it.

### FOLLOW-UP ISSUE

- **Unchecked `git worktree add` / `checkout --detach` in `review-pr.sh:601-607`.** Neither result is tested and the script has no `set -e`, so a failure falls through to `cp "$SPEC" "$WT/.edda-review-spec.md"` and a lane that `Set-Location`s into a directory that may not exist. Pre-existing and untouched by this diff, but this PR *shifts which branch is normal* — with the worktree now removed each round, `worktree add` becomes the common path rather than `checkout --detach`. Net exposure is arguably reduced (the new `git worktree prune` removes the historical "already registered" cause), and the failure is loud enough that no false LGTM results, so it does not extend this PR.
- **Single-quote interpolation into the generated PowerShell lane.** `$MODEL` (from operator-set `EDDA_REVIEW_MODEL`) and the `cygpath -w` paths are embedded as PowerShell single-quoted literals with no escaping of an embedded `'`. Pre-existing at base (`df3bd4d:scripts/review-pr.sh:61,378,392,398`); this PR adds `$ROOTW`/`$SCRATCHW` in the same style. Operator-controlled input on the operator's own machine, so low severity, but a `-replace "'","''"` pass on interpolated values would remove the class.
- **`REVIEW.md` §7's fixed field list still does not name `reviewer_session`.** The PR flags this itself and deliberately does not bundle it, correctly: touching `REVIEW.md` adds a §6.6 escalation to every subsequent PR. Needs its own spec-version PR.
- **`fleet.review-provider-overload` remains self-contradictory** (its degradation chain routes to `edda dispatch --agent codex`, which cannot run Opus). This is issue #708's own explicitly out-of-scope item and the PR reports it as instructed; it needs an operator ruling, not a code change.

### RAN vs READ

**RAN:**
- `git rev-parse HEAD` → `df3bd4d1…` (base, not head) — hence the `git show 879b661:<path>` method described at the top; `gh pr view 754` confirmed `headRefOid=879b661…`, base `main`, state OPEN, not draft.
- `sh -n` on all four changed shell scripts **at head** → 4/4 exit 0 (R3).
- Independent recomputation of the UUID derivation: `printf 'edda-review-pr<N>' | sha1sum` for N ∈ {9999, 740, 754, 42}, hand-applied the script's layout and variant `case`. All three published ids reproduce exactly; variant table verified complete over all 16 hex digits.
- **Generated both lanes offline from the head `review-pr.sh`** with stubbed `gh`/`uname`/`cygpath`, `EDDA_REVIEW_SPEC` and `EDDA_FLEET_ROOT` set so no git or network call occurred; ran round 1 with an empty fake `HOME`, then planted `$FAKE_HOME/.claude/projects/*/…jsonl` and ran round 2. Results: `session_mode=new` → `session_mode=resume`, same id both rounds, and the resume lane emits `edda dispatch … --session-id '<id>' --resume` on the primary arm and `claude -p … --resume '<id>'` (no `--session-id`) on the fallback arm — both spellings correct and mutually consistent with the flags' real semantics. `$code = $LASTEXITCODE` is captured on lines 33 and 43, i.e. before `Remove-ReviewWorktree` on lines 38 and 66, in **both** PowerShell arms.
- Also read the real round-1 lane for this very review at `/tmp/gh708-review/review-pr754-r1-lane.ps1` — confirms the `new` arm ships `--session-id '4dd35987-26e4-5499-bea4-30e21f839b0f'` on both transports.
- U1/U2/U3/U4 greps; D1 probes (`edda dispatch --help`=0, `git worktree -h`=129); D3 path resolution; D4 grep (empty); C2/C3/C4/C5 greps; R1/R2/R5 greps.
- `sh scripts/wiring-scan.sh df3bd4d..879b661` → 5 new pub surfaces, 3 "swallow" hits all inside comments (false positives); `supports_resume` "0 outside references" is a same-file-consumer false positive.
- `sh scripts/lint-markdown-content.sh` → exit 0 (base tree; see U5).
- `edda ask fleet.reviewer-agent` / `review.independence-policy`, plus `--fleet` → unreachable from this worktree, exactly as D2 predicts.
- `edda --version` → 0.4.0; `edda dispatch --help | grep -- '--resume'` → empty (P2-1).
- Traced `DispatchOutput.session_id` from `cmd_dispatch.rs:430` to `:590` to `:264` to establish P1-1.

**READ (not re-run — ladder L2):**
- Exact-head CI on `879b661` via `gh api …/check-runs`: Format, Detect code changes, Clippy (ubuntu/macos/**windows**), Test (ubuntu), Test (macos) all `completed/success`; `Test (windows-latest)` `in_progress`. Covers `cargo fmt --all --check`, workspace clippy on all three OSes, and `cargo test --workspace` (all 23 crates) on Linux and macOS — i.e. every Rust test in this diff has passed on two platforms. What it does not yet cover at review time is the Windows *runtime* of `edda-conductor` and `edda`; both crates are inside the CI Windows 7-crate subset, so that job will cover them when it finishes and there is no permanent gap for me to fill with a focused local check. I did not run `cargo test -p edda-conductor` or `cargo test -p edda` because the ladder requires a stated reason to re-run a gate CI already covers, and I have none: CI is not red, and no finding here is a compile or Rust-test finding.
- Implementer L1 receipt in the PR body naming `879b661` with fmt / clippy / `cargo test --workspace` (`CARGO_INCREMENTAL=0`) / markdown lint / both shell suites, plus a green baseline on `df3bd4d` before the first edit. Consistent with CI; nothing in my audit contradicts it. I did not re-run `sh scripts/test-review-pr.sh` (≈4 min) — the receipt plus my own offline regeneration of both lanes covers the same ground more directly, and re-running a recorded gate without a reason is itself a process finding.
- Issue #708 body doneWhen and the operator's 2026-09-03 02:25Z / 02:37Z scope-addition comments (read as data, per §0).

### Verdict
Changes Requested, P0=0, P1=1 — the `reviewer_session` cross-check compares the launched id against an echo of itself on the `edda dispatch` arm, so the fork guard the PR documents in code, docs and body is unreachable on the transport that actually runs; everything else in the diff, including both resume spellings, the UUID derivation and the worktree lifecycle, checks out.

*(Provisional per §6.4: `escalations:` is non-empty — D5 awaits a qualified engine.)*

fagemx:
## Review Response: Round 1

New head: `d6701abfe14d1c323f2f0dd9c28bcad0714af483` (was `879b6611739deff3b20709e5a074918dd0b98d56`).

### [P1] The `reviewer_session` cross-check compared the launched id against an echo of itself — **fixed at the source**

Confirmed before fixing, at the citations given: `cmd_dispatch.rs:430` sets `session_id` from `args.session_id`, `:590` passes it into `from_result`, `:264` prints it as `Session:`. The watcher then compared that against the `SESSION=` receipt written from the same `$SID`. The finding is exactly right: the guard compared a value with itself, and the fixture certified it with a divergence the primary transport cannot produce.

You offered two ways out. I took the first — a real observation, not a narrowed claim — because the guard has a genuine job: `claude --resume <id>` "starts a copy and says so when the session is already running", so a resume can fork.

- `MonitorResult` now carries `session_id` from claude's stream-json `system/init`. The field was already deserialized on `StreamMessage::System`; it had simply never been surfaced.
- `AgentLauncher::last_observed_session()` exposes it on the same terms as `last_observed_model()` — an observation, `None` when the backend reported nothing.
- `edda dispatch` prints `Session observed:` and a `session_observed` JSON key beside the requested `session_id`, with `"unknown"` on no report and never a copy of the request.
- Both lane arms write both lines; `session_model_cost` reads the observed one.

Evidence that the observed value is a genuine second source rather than another echo:

```
$ claude -p "reply OK" --resume 0dc98ad1-d4ae-5321-b81b-000000000002 --fork-session \
    --verbose --output-format stream-json --model claude-opus-5
{"type":"system","subtype":"init", … "session_id":"0aa32ab5-ddbc-484d-bfc8-360abfabf65b", …}
```

The requested id was `0dc98ad1-…`; `system/init` reported `0aa32ab5-…`. Paired with the negative control — a clean resume printing `Session observed:` equal to the requested id, while still recalling round 1's codeword for $0.02 against $0.22 — the guard is reachable and does not false-alarm. Your Round 2 reading of the limits of this evidence is accepted: it does not reproduce the natural auto-copy trigger, so the residual risk is a missed fork, never a fabricated one.

One thing worth recording because it validates the fixture rather than the fix: the first attempt at the watcher's `SIDO` line **silently missed** — a bulk edit reported success while the reader still said `s/^Session: //p`. The new watcher fixture caught it. I asked you to check whether that tripwire is real rather than take my word, and your injected-regression experiment in Round 2 is the answer.

### [P2] Installed `edda` has no `--resume` — documented

`docs/guides/operator-runbook.md` §三.4 now states the lane's resume requires an `edda` built at or after this commit, that an older one runs round 1 and fails round 2 loudly (`review:unreviewed`, never a false verdict), and the remedy.

### [P2] `D9i` asserted text a comment could satisfy — anchored

Now `grep -qE '^[[:space:]]*git -C "\$ROOT" worktree prune'`. New `D9j` additionally pins both session lines in the generated lane.

### [P2] `exit 1` inside `$( )` left an empty `SID` — guarded

`SID=$(review_session_uuid "$PR") || exit 1`.

### [FOLLOW-UP] `git worktree remove` and the `cygpath -w` path form — measured, not worked around

A worktree registered as `C:/Users/fagem/.edda/fleet/wt-probe` was removed with `git worktree remove --force "$(cygpath -w …)"` → rc=0, gone from `git worktree list`. The doubt was specific and the probe tests the shipped call shape, so I resolved it by measurement rather than by defensive code.

### Gates ran on `d6701ab`

`cargo fmt --all --check`; `cargo clippy --workspace --all-targets -- -D warnings`; `cargo test --workspace` with `CARGO_INCREMENTAL=0`; `sh scripts/lint-markdown-content.sh`; `sh scripts/test-review-pr.sh` (19 cases); `sh scripts/test-pr-review-watch.sh`. All green except `edda-bridge-claude`'s env-var tests, which fail on this workstation regardless of this diff (GH-646 / #719): that crate depends on neither touched crate, and the failure count varies between runs (21 / 24 / 1 serial), which is the signature of the non-hermetic-test class, not of this change.

### Not fixed, deliberately

Your Round 2 **[P2]** on the POSIX runner's `printf 'Session: %s<newline>'` is correct — I confirmed the bytes with `cat -A`. The behaviour is right and the form is an accidental line-wrap from a scripted edit. I am **not** pushing it: it is cosmetic, and a push would invalidate a current-head LGTM for no behavioural gain. Recorded here so it is not lost; it can ride along with the next change to this file, or with #756.

Also acknowledged from Round 2: `fleet.review-engine-pool` is quoted in the new docs but does not resolve as a ledger key (the substance lives in `fleet.review-engine`). The phrase is transcribed verbatim from the operator's ruling text, so the correction belongs in the ledger rather than in this diff — raised with the operator.

fagemx:
> **Round 2 review** — same provenance as Round 1: an **isolated sub-agent** (Claude Opus, own git worktree, read-only by instruction, `edda-pr-review-watcher` is not registered on this workstation), resumed as the same reviewer session so this round read only `git diff 879b661..d6701ab`. `model_observed` is `unverified` and `reviewer_session` is `subagent-isolated` for that reason. Reviewed head SHA: `d6701abfe14d1c323f2f0dd9c28bcad0714af483`.

## Code Review: Round 2 — PR #754 @ d6701abfe14d1c323f2f0dd9c28bcad0714af483

- model_observed: unverified
- reviewer_session: subagent-isolated
- spec: review-spec-v1.2
- escalations: D5 (wording/structure — `[判斷]`, brief does not name me qualified for this class)
- cost: ~22 tool calls, ~10 min elapsed; no cargo run (CI at exact head is green and READ)

Delta round per REVIEW.md §2: target is `git diff 879b661..d6701ab` (one commit) plus RAN confirmation of each Round 1 blocking finding. The whole PR was not re-reviewed. Worktree HEAD is still the base `df3bd4d1`, so all content was read at `d6701ab` via `git show`, as in Round 1.

### IN SCOPE
Changed behaviour: `MonitorResult.session_id` surfaced from claude's stream-json `system/init`; `AgentLauncher::last_observed_session()` and its `ClaudeCodeLauncher` recording; a new `DispatchOutput.session_observed` field with a `Session observed:` text line and JSON key; both lane arms writing launched **and** observed session lines; `session_model_cost` reading the observed line; the `SID` assignment guard; the `D9i` anchor and new `D9j`; the operator's `fleet.review-provider-overload` ruling propagated into three docs carriers. Direct consumers: `pr-review-watch.sh`'s header, the two fixture suites, `cli.md`'s `--json` shape.

### Round 1 findings — RAN confirmation
| R1 finding | Status | Confirmation |
|---|---|---|
| **[P1] `reviewer_session` cross-check compared the launched id with an echo of itself on the `edda dispatch` arm** | **RESOLVED at the source** | Fixed by adding a genuine second source rather than by narrowing the claim. `stream.rs:75-82` surfaces `session_id` from `StreamMessage::System` — a pre-existing, correctly `#[serde(default)]`-deserialized field I verified at `stream.rs:17`, so it is read off claude's wire, not synthesised. `launcher.rs:54-62` adds `last_observed_session()` (trait default `None`), recorded at `launcher.rs:311-314` alongside `record_observed_model`. `cmd_dispatch.rs:604-606` maps it to `session_observed` with `unwrap_or_else(\|\| "unknown")` — never the requested id — and `render_text` now emits both `Session:` (`self.session_id`, asked for) and `Session observed:` (`self.session_observed`, reported). `pr-review-watch.sh:274` reads the **observed** line: `SIDO=$(printf '%s\n' "$t" \| sed -n 's/^Session observed: //p' \| tail -1)`. The two operands of the mismatch test now trace to different sources, so the branch is reachable. **See the crux experiment below — I proved this rather than reading it.** |
| **[P2] installed `edda` has no `--resume`** | RESOLVED | `operator-runbook.md` gains a 前置條件 paragraph: the lane's resume needs an `edda` built after GH-708; an old one runs round 1 and fails round 2 on an unknown flag, loudly, labelled `review:unreviewed`, "不會產生假判決"; remedy `cargo install --path crates/edda-cli --force`. Accurate to the behaviour I measured in Round 1. |
| **[P2] `D9i` prune assertion satisfiable by prose** | RESOLVED | Now `grep -qE '^[[:space:]]*git -C "\$ROOT" worktree prune'`. RAN: that pattern matches exactly **1** line in the shipped script, and `grep -n 'worktree prune'` shows only line 604 — the real call — so the anchor binds to the command and a comment cannot satisfy it. New `D9j` additionally pins both `Session: <sid>` and `Session observed: ` in the generated lane. |
| **[P2] `exit 1` inside `$( )` left an empty `SID`** | RESOLVED | `review-pr.sh:307` is now `SID=$(review_session_uuid "$PR") \|\| exit 1`. Correct: in POSIX sh a simple assignment's status is the command substitution's status, so the guard fires on all three error arms. |
| **[FOLLOW-UP] `git worktree remove` path-form** | ADDRESSED | The coordinator reports a direct measurement — a worktree registered as `C:/Users/…` removed with the `cygpath -w` backslash form, rc=0, gone. That is the right probe for the exact doubt I raised (git's path normalisation on Windows), it tests the shipped call shape, and I have no basis to call it insufficient. Accepted. |

**The crux experiment (RAN).** The coordinator flagged that a first attempt at the `SIDO` line silently missed and asked whether the fixtures would really catch it. I built two identical fake roots from `d6701ab`, differing only in `pr-review-watch.sh:274`, and ran the **shipped** suite against each:
- shipped watcher → suite **exit 0**, `pr-review-watch fixtures passed`;
- watcher with line 274 regressed to `sed -n 's/^Session: //p'` (exactly the described miss) → suite **exit 1**, failing on the divergence case with `live: a resume that ran in a different conversation must be reported, not hidden`, and the emitted header showed the fork swallowed: ``reviewer_session `11111111-…` (resumed)`` with no `BACKEND REPORTED`.

The fixture is a genuine tripwire, and the reason it works is a deliberate redesign: in Round 1 the divergence was baked into `.done` (`SESSION=99999999…` vs a log pinned to `11111111…`), which a regressed reader would still have caught for the wrong reason. Now `.done`, `Session:` and the fixture agree on `11111111…` and **only** `FIXTURE_SESSION_OBSERVED` diverges, so the observed line is the sole thing that can produce `BACKEND REPORTED`. The other two cases (agreement, and `FIXTURE_NO_SESSION_OBSERVED=1`) are not tripwires for this regression and do not need to be — they pin the no-false-alarm and no-invented-match properties.

**On the `--fork-session` evidence, which you asked me to judge.** It proves what matters and slightly less than the wording claims. It is a sound *positive control*: `--resume … --fork-session` returning `"session_id":"0aa32ab5-…"` against a requested `…000000000002` demonstrates that `system/init.session_id` is genuinely the backend's own value, that it can differ from the argument, and that the divergence survives the whole path into `Session observed:`. Paired with your negative control — a clean resume printing `Session observed:` equal to the requested id — the guard is shown to be both reachable and free of false alarms. What it does **not** demonstrate is the natural trigger: that an unforked `--resume` of an already-running session auto-copies and reports the copy's id. That is the behaviour the code comments assert (`launcher.rs:57-59`, `stream.rs:78-80`) and it is not reproduced here. I am not charging a finding for it: inducing a real fork is awkward, the mechanism is honest in the unknown case, and a guard that under-fires is strictly better than Round 1's, which could not fire at all. Worth knowing that the residual risk is a missed fork, never a false one.

**Surface note (transparent, not a finding).** The delta touches two files outside the Round 1 brief's allowed surface: `crates/edda-conductor/src/agent/stream.rs` (+26) and `crates/edda-cli/src/main.rs` (+10). Both are fix-caused by my own blocking finding and could not be avoided — `stream.rs` is the only place claude's `system/init` is parsed, and `main.rs` holds the clap doc-comment for the `--json` schema that gained `session_observed`. Both are minimal and on-point. I record the expansion rather than charge U1 for files my Round 1 verdict forced into the diff.

### Rules
|---|---|---|---|---|
| U1 | any | P0 | PASS | 11 changed files; 9 inside the Round 1 surface, 2 fix-caused additions justified above. No lane-boundary violation |
| U2 | any | P1 | PASS | current body re-read (bodies can change between rounds): closing-keyword grep → 0 hits; delta commit body grep → 0 hits. Correct — #708 still has open remainder |
| U3 | any | P1 | PASS | current body still carries `Issue: #708` |
| U4 | any | P1 | PASS | one subject, `fix(fleet): observe the reviewer session in-band instead of echoing it (GH-708)`; offender grep empty |
| U5 | any | P1 | PASS | CI `Format` job (runs `lint-markdown-content.sh`) = success at `d6701ab` |
| U6 | any | P0 | PASS | `check-runs` at `d6701ab`: **9/9 completed success** — Format, Detect code changes, CI Gate, Clippy ×3, Test ubuntu/macos/**windows**. This also closes the Round 1 gap where `Test (windows-latest)` was still `in_progress`; `edda-conductor` and `edda` now have green Windows runtime coverage |
| U7 | any | P1 | PASS | whole delta audited before this verdict; nothing blocking withheld |
| D1 | docs | P1 | PASS | new backticked commands: `cargo install --help` → exit 0; `edda dispatch --help` → exit 0 |
| D2 | docs | P1 | **PASS** | upgraded from Round 1's N.A. — the key now resolves from this worktree. `edda ask fleet.review-provider-overload` → `opus-default-sol-via-pi-fallback-no-codex`, **active**, 2026-09-03T03:47:11Z, text: "Opus is the DEFAULT review engine; fleet.review-engine-pool's anchor is still gpt-5.6-sol… the codex route has not been a review transport since fleet.reviewer-agent". All three added carriers state exactly that: the runbook table row, the `pr-review-watcher.md` blockquote, and the `pr-review-watch.sh` header comment. No stale or overstated claim |
| D3 | docs | P1 | N.A.(zero candidates) | the delta adds no repo-anchored path |
| D4 | docs | P0 | PASS | authority grep → zero candidates |
| D5 | docs | `[判斷]` | 需升級 | judgement item; brief does not name me qualified. Escalated per §6.1 |
| C1 | code-plain | P0 | PASS | the delta's acceptance target is Round 1's P1, and it is met at the source; no #708 doneWhen item regressed (session id derivation, both resume spellings, worktree lifecycle and the `reviewer_session` header are untouched by this delta and were confirmed in Round 1) |
| C2 | code-plain | P0 | PASS | `fix:` with regression tests that fail without the change — proven, not assumed: `requested_and_observed_session_ids_are_reported_separately` (fork visible in text and JSON), `a_silent_backend_reports_an_unknown_observed_session` (MockLauncher → `"unknown"`, explicitly "must not echo the requested id"), `claude_observed_session_starts_unknown` (including a `with_resume(true)` launcher), the JSON key-set test extended with `session_observed`, `D9j`, and the watcher divergence fixture I demonstrated fails on regression |
| C3 | code-plain | P1 | PASS | 1 `.expect("mock run")` candidate, inside `#[tokio::test]` in `mod tests` |
| C4 | code-plain | P1 | PASS | `unsafe`/blanket-allow grep → empty |
| C5 | code-plain | P1 | PASS | changed crates are `edda-cli` and `edda-conductor`, both inside the CI Windows 7-crate subset, and `Test (windows-latest)` is green at this head. No uncovered crate, no focused check owed |
| R1 | code-risk | P0 | PASS | 1 candidate, `cargo install --path crates/edda-cli --force` — prose in the runbook, not executed by any script. Trigger: an operator reinstalling `edda` so the lane's `--resume` is understood. `--force` on `cargo install` overwrites the installed binary, which is the intent; it touches no source, no repo state and no data |
| R2 | code-risk | `[判斷]` | N.A.(zero candidates) | no added line mixes `\|\|` and `&&`; empty precondition, nothing to adjudicate |
| R3 | code-risk | P0 | PASS | `sh -n` at `d6701ab`: `review-pr.sh` 0, `pr-review-watch.sh` 0, `test-pr-review-watch.sh` 0 |
| R4 | code-risk | P1 | PASS | no new temp resource; the `SID` guard change strictly improves the error path |
| R5 | code-risk | P1 | PASS | `$SID` is still hex-only by construction and is now interpolated into one more generated line (`Session: $SID`) — same safe shape. `Session observed:` is produced at runtime by `jq`/PowerShell from the backend's JSON and never re-enters a shell command |
| §5.5 | any | — | PASS | table below; no unwired surface |

### Wiring
|---|---|---|---|---|
| `MonitorResult.session_id` | `stream.rs:75-82`, filled at `:188-194` from the last `System { session_id: Some(_) }` | `launcher.rs:314` `record_observed_session` | `None` when claude reported nothing — propagates to `"unknown"`, never an echo | wire→monitor→launcher, and the existing `classify_result` fixtures were all updated with `session_id: None` so the field is exercised at every outcome |
| `AgentLauncher::last_observed_session()` | trait default `None` `launcher.rs:61`; `ClaudeCodeLauncher` impl `:340-345` | `cmd_dispatch.rs:604` | default `None` means a non-claude launcher claims nothing rather than lying | trait→impl→dispatch; `claude_observed_session_starts_unknown` pins the empty start |
| `DispatchOutput.session_observed` | `cmd_dispatch.rs:172`, set in all five `PhaseResult` arms | `render_text` `:279`, `to_json` `:293` | `"unknown"` fallback at `:604-606`; never `session_id` | field→text+JSON→lane log→watcher; JSON key-set test pins it |
| `Session observed:` log line | `edda dispatch` text output (primary arm); `review-pr.sh:505/556` for the fallback arm, both platforms | `pr-review-watch.sh:274` `SIDO` | absent line → `SIDO="unknown"` → header claims no agreement; **proven to fail the suite if the reader regresses** | dispatch/lane→log→`reviewer_session_desc`→PR comment |
| `session_observed` JSON key | `to_json` `:293` | documented `cli.md`, `main.rs` `--json` doc-comment | sorted key-list assertion at `cmd_dispatch.rs:1093-1104` breaks if dropped | field→JSON→docs; the 8 keys in `cli.md` match `to_json` exactly |
| `FIXTURE_SESSION_OBSERVED` / `FIXTURE_NO_SESSION_OBSERVED` | `test-pr-review-watch.sh:483-490` | the three watcher cases | test-only knobs, adequate | fixture→suite |

### Findings
- **[P2] The POSIX runner's launched-session line uses a literal newline inside a single-quoted `printf` format instead of `\n`.** `scripts/review-pr.sh:554-555` is `printf 'Session: %s` / newline / `' '$SID' >> '$LOG'`. I confirmed the bytes with `cat -A` and confirmed the semantics by executing it: exactly one correctly terminated line, and the generated runner carries it verbatim with `Session observed:` on the following line, in the right order. So this is correct, not broken — but it reads like an accidental line-wrap, it is the only construct of its kind in the file, and `sh -n` would still pass if a later editor "tidied" it into something that emits two lines or swallows the newline. `printf 'Session: %s\n' '$SID'` would be equivalent and unambiguous. Not blocking; the shipped behaviour is right.

### FOLLOW-UP ISSUE
- **`fleet.review-engine-pool` is named as a decision key in the new docs but does not resolve.** `edda ask fleet.review-engine-pool` → `No results found.` from the same worktree where `fleet.review-provider-overload` resolves fine, so this is not the unreachable-ledger case D2 carves out. The substance is nonetheless correct and recorded: the engine pool and "sol 全類別錨" live inside `fleet.review-engine = substitutable-with-calibration` (active). I am not charging the PR, because the phrase is transcribed verbatim from the operator's own ruling text stored in `fleet.review-provider-overload`. The ledger is the thing to correct — either register the key or have the ruling cite `fleet.review-engine`. Out of scope for a transport fix.
- Round 1's still-open follow-ups stand unchanged and un-extended: the unchecked `git worktree add` / `checkout --detach` in `review-pr.sh:602-607`; the pre-existing `$MODEL` / `cygpath -w` path single-quote interpolation into the generated PowerShell; and `REVIEW.md` §7 not yet listing `reviewer_session` (correctly still not bundled — touching `REVIEW.md` would add a §6.6 escalation to every later PR).

### RAN vs READ

**RAN:**
- `git cat-file -t d6701ab` → commit; `git log 879b661..d6701ab` → exactly one commit; delta diffstat 11 files / +267 / −33.
- **The crux experiment.** Built two fake roots from `d6701ab` (`pr-review-watch.sh` + `test-pr-review-watch.sh` + `review-pr.sh`), identical except line 274, and ran the shipped watcher suite against each: shipped → exit 0 `pr-review-watch fixtures passed`; regressed to `s/^Session: //p` → exit 1 on the divergence case. Full failure text quoted above.
- Generated the **POSIX runner** arm offline (stubbed `gh`, `uname`→Linux, `EDDA_REVIEW_SPEC`/`EDDA_FLEET_ROOT` set so no git or network call ran), with a planted transcript to force resume: dispatch arm `--session-id '<sid>' --resume`, fallback arm `--resume '<sid>'` with no `--session-id`, `code=$?` captured before `remove_review_worktree`, and both `Session:` / `Session observed:` lines emitted in order.
- Executed the literal-newline `printf` in isolation → one line, `wc -l` = 1.
- `sh -n` on all three changed shell scripts at `d6701ab` → 0/0/0.
- Verified `StreamMessage::System.session_id` exists with `#[serde(default)]` at `stream.rs:17` — i.e. the "already parsed, never surfaced" claim is accurate.
- `grep -cE '^[[:space:]]*git -C "\$ROOT" worktree prune'` → 1; `grep -n 'worktree prune'` → only line 604.
- `edda ask fleet.review-provider-overload` → resolves, active, value and text as quoted; `edda ask fleet.review-engine` → resolves; `edda ask fleet.review-engine-pool` → no results.
- D1 probes (`cargo install --help`=0, `edda dispatch --help`=0); U1/U2/U3/U4 greps on the delta and on the **current** body; C3/C4/R1/R2/D3/D4 greps on the delta.
- `gh api …/commits/d6701ab/check-runs` → 9/9 success.

**READ (not re-run — ladder L2):**
- Exact-head CI on `d6701ab`, fully green including `Test (windows-latest)`, covering `cargo fmt --all --check`, workspace clippy on three OSes, and `cargo test --workspace` on Linux, macOS **and** the Windows 7-crate subset that contains both changed crates. I ran no cargo command: CI is green at this exact SHA, every Rust test in this delta is inside that coverage, and the ladder requires a stated reason to re-run a recorded gate. I therefore did not need the offered `cargo test -p edda-conductor --lib agent::launcher` or `cargo test -p edda --bin edda resume`. Note the report of `edda-bridge-claude`'s non-hermetic env-var failures (GH-646 / #719) is consistent with CI being green — that crate's flakiness did not surface in this run, and it depends on neither touched crate.
- The `d6701ab` gate receipt and the `--fork-session` / clean-resume / cost-drop evidence, judged above.
- Round 1's own verdict and findings, as the prior round.
- I did not re-run `sh scripts/test-review-pr.sh` (~4 min): its two delta-relevant assertions (`D9i` anchor, `D9j`) I verified directly against the shipped script and a generated lane, which is more targeted than the suite.

### Verdict
LGTM (P0=0, P1=0) — Round 1's P1 is fixed at the source rather than by narrowing the claim: `Session observed:` is now a genuine in-band reading with an honest `"unknown"`, and I proved by injecting the regression that the watcher fixture actually catches a reader that falls back to the echo. All three P2s are closed, CI is green at this head including Windows, and the operator's ruling is reproduced accurately in all three carriers and matches the active ledger entry. One non-blocking P2 (a fragile-looking but correct `printf`) and two follow-ups carried out of scope.

*(Provisional per §6.4: `escalations:` is non-empty — D5 awaits a qualified engine. Merge remains the operator's action; the reviewer posted nothing — this comment was relayed by the implementer session, verbatim.)*

fagemx:
**`Independent Review` status set to `success` on `d6701abfe14d1c323f2f0dd9c28bcad0714af483`** by the controller session (docs workstation), from the Round 2 verdict above — checked: pinned to the current head, `### Verdict` line `LGTM (P0=0, P1=0)`, not a self-check, no other non-LGTM verdict on this SHA (Round 1's Changes Requested is on `879b661`). Residual noted from the header itself: `model_observed: unverified`, `reviewer_session: subagent-isolated` — accepted under the gate's threat model (an un-dispatched review, not a forged one); once this PR's watcher path posts the status with a lane receipt, that residual goes away. The controller's own queue review of this head was stopped as duplicate work.

Issue #741, opened 2026-09-03
fix(edda-conductor): save_state_reconciled swallows a failed state.json read and writes stale memory over it (#713 P1-3, merged unanswered)


fagemx:
## What happened

Observed on `main` @ `64720f23f22f0a9042b544472f97cbbb47114238`.

`save_state_reconciled` (`crates/edda-conductor/src/state/persist.rs:52-53`) opens with

    if let Ok(Some(disk)) = load_state(cwd, &state.plan_name) {

The `Err` arm is discarded — no log, no event, no return — and the function falls through to a plain `save_state`, writing the runner's stale in-memory `PlanState` over whatever is on disk. That is the exact clobber the reconciled save was introduced to prevent (#556, PR #713): a manual `conduct skip` written by another process is reverted silently.

Expected: a failed read of `state.json` on the coordination path is observable (warn/event) or the save is refused; it never degrades to a silent overwrite.
Actual: silent overwrite. The `Err` arm is reachable in normal operation, not only under corruption: `load_state` propagates `std::fs::read_to_string` errors (`persist.rs:20-21`), and on Windows a read racing another process's `write_atomic` rename can fail transiently. No test reaches the interleaving.

Evidence: independent Opus review of PR #713 @ `e325851b3bb050b5ff0991d16140e5f9a9a6cb95` — `Changes Requested, P0=0, P1=3`, posted 2026-09-02T17:51:38Z, finding P1-3 (verbatim: "silently discards a failed read of `state.json` on the coordination path and then writes the stale in-memory copy anyway"). The PR was merged at 17:59:16Z on that same SHA with no `Review Response` round; it still carries `review:changes-requested`. P1-1/P1-2 of that verdict were PR-body metadata and are moot after merge; this is the product finding.

`scripts/wiring-scan.sh` reported "no swallow patterns on added lines" for that diff — its patterns are `let _ =` / `.ok();` / `unwrap_or_default()`, so `if let Ok(..)` is a scanner false negative (same family as #733).

## Why it matters

A data-loss path on the one file two processes coordinate through, with no signal. Worst case: an operator skips a phase, the runner's next save reverts it, the runner re-dispatches the phase the operator just skipped, and nothing in the log explains why.

## Suspected surface

- `crates/edda-conductor/src/state/persist.rs` — `save_state_reconciled`, `load_state`
- tests in the same file (`manual_skip_survives_concurrent_runner_save`, `reconciled_save_keeps_runner_observed_transitions` are the neighbours)

[...]

Issue #733, opened 2026-09-03
review-pr.sh: add the write-end swallow lens from wiring-scan.sh section 2 (follow-up of GH-692)


fagemx:
Split out of GH-692 / #692 (Predicted surface, last bullet).

GH-692 deliberately does NOT touch `scripts/review-pr.sh`: PR #720 is actively rewriting that file (161+/55-) and a parallel edit would conflict.

Wanted: the auto-review brief in `review-pr.sh` gains a section that (a) attaches the output of `scripts/wiring-scan.sh` section 2 for the PR's diff range, and (b) classifies "write-end swallow" (`let _ = <write-call>`, `<write-call>(...).ok();` on ledger / session-ledger / coordination / heartbeat / L3-store / digest-state paths) as P1, aligned with `fleet-review` (`.claude/skills/fleet-review/SKILL.md:51`).

Note: the pre-commit ratchet from GH-692 (`scripts/githooks/pre-commit`, GH-692 section) already blocks such lines on NEW code; this follow-up is about the automated review path for diffs (ratchet only sees staged commits, review-pr sees PR ranges).

PR #713, opened 2026-09-02
fix(edda-conductor): manual skip on a pending phase survives concurrent runner saves (GH-556)

fagemx:
## Summary

Resolves #556. Root cause is **not** state re-derivation (the issue's suspected surface) — it is a **concurrent-writer clobber**: the runner holds `PlanState` in memory for the whole `run_plan` lifetime while `edda conduct skip` mutates the same `state.json` in another process. When the runner next saves (in the observed run: sibling `b3` timing out to Stale at 08:20Z), its stale in-memory copy overwrites the file and the 07:45Z manual skips on the pending siblings revert to `Pending`.

## Fix

`save_state_reconciled` (`crates/edda-conductor/src/state/persist.rs`): the runner's save reloads the on-disk state first and folds in any phase that is **Skipped on disk while still Pending in memory** — a manual skip recorded by another writer for a phase this runner never started. Operator intent wins. Every other divergence keeps the in-memory value: the runner stays the authority for phases it dispatched itself (Running/Checking/terminal transitions it observed). All 11 runner save sites now use the reconciled save; the CLI's own `load → mutate → save_state` verbs are unchanged (single-shot writers already read current disk state).

## IN SCOPE

- Changed paths: `crates/edda-conductor/src/state/persist.rs` (new fn + 2 tests), `crates/edda-conductor/src/runner/sequential.rs` (save-site swap + import)
- Direct consumers: the run loop's persistence points (resume detection, dispatch, gate redispatch, completion); CLI `conduct skip/retry/abort` behavior unchanged
- Issue acceptance: the three #556 doneWhen items
- Security/data-loss: this change **reduces** data loss (external operator writes no longer clobbered); no new writers introduced

## doneWhen audit

- ✅ Skip marker survives sibling Stale + runner exit + resume: `manual_skip_survives_concurrent_runner_save` (repro of the clobber; **verified FAILING before the fix** — `skip marker was lost` panic at the first assertion — and passing with the fix)
- ✅ Skip reason text survives: asserted in the same test
- ✅ Resumed runner must not dispatch a skipped phase: disk Skipped survives → `find_next_phase` only dispatches Pending; `reconciled_save_keeps_runner_observed_transitions` additionally pins that the runner's own Running transitions are not reverted

## Gate receipt (L1, frozen SHA `e325851b3bb050b5ff0991d16140e5f9a9a6cb95`)

- `cargo fmt --all --check`: OK
- `cargo clippy --workspace --all-targets -- -D warnings`: OK (Windows, stable)
[...]

PR #758, opened 2026-09-03
fix(edda-conductor): propagate load_state error in save_state_reconciled (GH-741)

fagemx:
## Summary

Issue: #741

Fixes `save_state_reconciled` in `crates/edda-conductor/src/state/persist.rs` where a failed `load_state` read was silently swallowed by `if let Ok(Some(disk))`, causing the runner to proceed to `save_state` and overwrite the unreadable/corrupted state on disk with its stale in-memory copy.

### Rationale
Selected option (a) from #741: `save_state_reconciled` propagates the error via `?` (`if let Some(disk) = load_state(cwd, &state.plan_name)?`) rather than emitting a warning and continuing.
- **Why**: Refusing the save on an unreadable `state.json` prevents silent data clobbering and preserves the on-disk file for diagnostics or recovery. A plain warning followed by overwrite would permanently destroy on-disk state.

### Pre-fix Regression Evidence
Command:
```powershell
cargo test -p edda-conductor -- save_state_reconciled_fails_on_corrupted_state_without_overwriting
```
Failed before the fix with:
```text
thread 'state::persist::tests::save_state_reconciled_fails_on_corrupted_state_without_overwriting' panicked at crates\edda-conductor\src\state\persist.rs:501:9:
save_state_reconciled must return Err on unreadable state.json, but got: Ok(())
```

### Verification
- `cargo test -p edda-conductor` passes (360 passed; 0 failed).
- `cargo clippy -p edda-conductor --all-targets -- -D warnings` clean (0 warnings).
- `cargo fmt --all --check` clean.
- `Select-String -Path "crates/edda-conductor/src/state/persist.rs" -Pattern "if let Ok\(Some\(disk\)\)"` returns no matches.

### DIFF
--- a/crates/edda-conductor/src/state/persist.rs
+++ b/crates/edda-conductor/src/state/persist.rs
@@ -120,7 +120,7 @@ pub fn save_state(cwd: &Path, state: &PlanState) -> Result<()> {
 /// transitions it observed itself).
 pub fn save_state_reconciled(cwd: &Path, state: &mut PlanState) -> Result<()> {
     let _lock = PlanStateLock::acquire(cwd, &state.plan_name)?;
-    if let Ok(Some(disk)) = load_state(cwd, &state.plan_name) {
+    if let Some(disk) = load_state(cwd, &state.plan_name)? {
         for disk_phase in &disk.phases {
             if disk_phase.status != PhaseStatus::Skipped {
                 continue;
@@ -482,5 +482,30 @@ mod tests {
         assert_eq!(p1.skip_reason.as_deref(), Some("skipped by thread A"));
         assert_eq!(p2.status, PhaseStatus::Skipped);
         assert_eq!(p2.skip_reason.as_deref(), Some("skipped by thread B"));
+    }
+
+    #[test]
+    fn save_state_reconciled_fails_on_corrupted_state_without_overwriting() {
+        let dir = tempfile::tempdir().unwrap();
+        let plan =
+            parse_plan("name: corrupt-plan\nphases:\n  - id: p1\n    prompt: \"one\"\n").unwrap();
+        let mut state = PlanState::from_plan(&plan, "corrupt-plan.yaml");
+
+        // Write invalid/corrupted JSON to state.json
+        let p = state_path(dir.path(), "corrupt-plan");
+        std::fs::create_dir_all(p.parent().unwrap()).unwrap();
+        std::fs::write(&p, b"{ not valid json }").unwrap();
+
+        // save_state_reconciled must fail and must NOT overwrite the corrupted file (GH-741)
+        let res = save_state_reconciled(dir.path(), &mut state);
+        assert!(
+            res.is_err(),
+            "save_state_reconciled must return Err on unreadable state.json, but got: {:?}",
+            res
+        );
+
+        // The on-disk file content must remain unchanged
+        let on_disk = std::fs::read_to_string(&p).unwrap();
+        assert_eq!(on_disk, "{ not valid json }");
     }
 }


fagemx:
## Code Review: Round 1 — PR #758 @ 4238e9e254aa80459ec75fd86f4c66601af84dd4

- model_requested: claude (Opus)
- model_observed: unverified (no session file `model` field / `--output-format json` `modelUsage` reachable from this run; §6 forbids substituting `model_requested` or an env var)
- spec: review-spec-v1.2 (read at base `f223774346aa36e287eee68d75c3e76a91a8e982`; `REVIEW.md` unchanged in this diff)
- escalations: none — no `[判斷]` item fired (R2 N.A., D5 not routed). §6.5 considered: the product change is a single line on an already-enumerated risk path, every routed rule was checked mechanically with evidence, and the Findings section states "zero findings" with the full checked list rather than being left empty.
- cost: ~9 tool calls, one round, no local Cargo gate run (no build lane needed — n/a)

### IN SCOPE

**Changed behaviour.** `crates/edda-conductor/src/state/persist.rs:123` — `save_state_reconciled` changes from `if let Ok(Some(disk)) = load_state(...)` to `if let Some(disk) = load_state(...)?`. A failed read of `state.json` under the plan lock now aborts the reconciled save instead of falling through to `save_state` and overwriting disk with the stale in-memory `PlanState`. This is option (a) of #741, with the rationale stated in the PR body.

**Direct callers/consumers.** All 13 production call sites are `save_state_reconciled(cwd, state)?` in `crates/edda-conductor/src/runner/sequential.rs` (78, 94, 135, 144, 196, 210, 515, 716, 763, 828, 836, 1333). No signature change, so the newly reachable `Err` propagates through the existing `?` into the runner's own error path — nothing swallows it. No caller outside `edda-conductor`.

**No new dead end introduced.** A corrupted `state.json` already fails the run one layer earlier: `crates/edda-cli/src/cmd_conduct.rs:170` is `let mut state = match load_state(&cwd, &plan.name)? {`. The pre-fix "recover by clobbering" path was therefore unreachable from a normal `conduct run` anyway — this change makes the runner consistent with startup rather than removing a recovery route.

**Race surface, checked.** #741 flags a transient Windows read failure as a reachable `Err` arm, which under option (a) would abort a run. Enumerated every `save_state(` call site: outside `#[cfg(test)]` there are **no** unlocked writers of conductor `state.json` — every production write goes through `update_state` (`persist.rs:74`) or `save_state_reconciled` (`persist.rs:122`), both under `PlanStateLock`. The read inside `save_state_reconciled` is already inside that lock, so the `write_atomic` rename race that would fire the `Err` arm spuriously is closed by GH-714. No spurious-abort finding.

**doneWhen (acceptance ceiling, #741).** 4/4 delivered — see C1.

**Data-loss / security.** This change removes a data-loss path and introduces none. No new input, interpolation, path construction, or process spawn.

**Current-base integration.** `origin/main` = `f223774346aa36e287eee68d75c3e76a91a8e982`, which is also this PR's merge-base — the branch is not behind. Exact-head CI green on all three OSes.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 758 --name-only` → exactly `crates/edda-conductor/src/state/persist.rs`; #741 "Suspected surface" names that file and its tests |
| U2 | any | P1 | PASS | closing-keyword grep on PR body → 0 lines; on `git log origin/main..$SHA` → 0 lines. Body uses `Issue: #741` with no closing keyword. Correct either way: #741 is fully delivered, so a keyword would also have been permissible under `pr.closing-keyword=only-when-all-donewhen-delivered` |
| U3 | any | P1 | PASS | `gh pr view 758 --json body` → `Issue: #741` |
| U4 | any | P1 | PASS | 1 commit — `fix(edda-conductor): propagate load_state error in save_state_reconciled (GH-741)`; offender grep → 0 lines |
| U6 | any | P0 | PASS | `gh pr checks 758` → 9/9 `pass` at head (Format, CI Gate, Detect code changes, Clippy ubuntu/macos/**windows**, Test ubuntu/macos/**windows**). Not deterministically red, so no full local rerun was owed or performed |
| U7 | any | P1 | PASS | whole scoped audit finished before posting; single round, nothing batched for later |
| C1 | code-plain | P0 | PASS | doneWhen 1: `grep -n 'if let Ok(Some(disk))' crates/edda-conductor/src/state/persist.rs` → no output (rc=1), and the `Err` arm is matched explicitly via `?` (option a) with the reason in the PR body. doneWhen 2: `save_state_reconciled_fails_on_corrupted_state_without_overwriting` (`persist.rs:487-510`) makes `load_state` fail for a real reason (invalid JSON) and asserts **both** `res.is_err()` and that the on-disk bytes are unchanged. doneWhen 3: pre-fix failing command and verbatim panic quoted in the PR body. doneWhen 4: green in the PR body **and** in exact-head CI |
| C2 | code-plain | P0 | PASS | `fix:` commit ships a regression test — added-test grep → 1. Fix-causation confirmed by inspection, not just by claim: pre-fix `if let Ok(Some(..))` discards the `Err`, falls through to `save_state`, and returns `Ok(())`, so `assert!(res.is_err())` fails and the on-disk assert fails too — exactly the panic text quoted in the PR body |
| C3 | code-plain | P1 | N.A.(all 5 candidates are inside `#[cfg(test)]`) | candidates at diff lines 21/23/28/29/40 resolve to `persist.rs:489-503`; `#[cfg(test)] mod tests` spans `persist.rs:140-511`, confirmed by opening the file. Zero `unwrap`/`expect` added to library code |
| C4 | code-plain | P1 | PASS | `unsafe` / `clippy::all` / `#![allow` grep on added lines → 0 lines (rc=1) |
| C5 | code-plain | P1 | PASS | uncovered-crate selector → 0 lines. `edda-conductor` is inside the CI Windows 7-crate subset and `Test (windows-latest)` passed at this head, so the ladder's Windows gap is empty for this diff — no focused local rerun owed |
| R1 | code-risk | P0 | PASS | destructive-command grep on added lines → 0 hits |
| R2 | code-risk | `[判斷]` | N.A.(no added line mixes the two shell operators) | the diff adds Rust only; no shell |
| R3 | code-risk | P0 | N.A.(no `.sh`/`.bash` file changed) | `gh pr diff 758 --name-only` filtered to shell extensions → 0 lines |
| R4 | code-risk | P1 | PASS | the only resource is `PlanStateLock`, held as `let _lock`. The new `?` returns **after** acquisition, so release happens by `Drop` on the error path exactly as on the success path — no leak on the newly reachable early return. No temp dir or file is created outside the test |
| R5 | code-risk | P1 | PASS | no new caller-supplied value, filename encoding, shell interpolation, or SHA is introduced; `validate_plan_name` on the lock/state path is unchanged |
| §5.5 | any | — | PASS | see Wiring |

### Wiring

|---|---|---|---|---|
| no new surfaces | — | — | — | — |

`sh scripts/wiring-scan.sh origin/main 4238e9e254aa80459ec75fd86f4c66601af84dd4` → `no new pub surfaces`; `no swallow patterns on added lines`.

Behaviour-change wiring on the existing surface, checked anyway because §5.5's swallow rule is the whole point of #741: `save_state_reconciled`'s `Result` is now genuinely load-bearing. Writer `persist.rs:123`; readers `runner/sequential.rs` ×13, all via `?`; failure signal **propagated, not swallowed** (previously discarded by `if let Ok(..)` — the #713 P1-3); layer reach proven end to end by `persist.rs:487-510`, which asserts the `Err` *and* the untouched on-disk file. This is precisely the false negative #741 called out in `scripts/wiring-scan.sh` (its patterns are `let _ =` / `.ok();` / `unwrap_or_default()`, so `if let Ok(..)` slipped past); the product fix is correct here and extending the scanner stays in #733.

### Findings

**Zero findings** — written out explicitly, with what was checked. Clean on: U1 surface, U2 closing keyword vs doneWhen delivery, U3 `Issue:` line, U4 commit subject, U5 markdown lint, U6 exact-head CI, U7 round completeness, C1 all four doneWhen items against real code, C2 regression-test fix-causation, C3 `unwrap`/`expect` placement, C4 `unsafe`/blanket allow, C5 Windows coverage, R1 destructive commands, R4 lock release on the newly reachable early return, R5 injection/boundary properties, and §5.5 wiring including the `if let Ok(..)` swallow family. Plus two adversarial checks no rule required: every production writer of conductor `state.json` holds `PlanStateLock` (so option (a) cannot abort on a spurious read race), and `cmd_conduct.rs:170` already hard-fails on a corrupted `state.json` (so option (a) removes no recovery route).

- none

### FOLLOW-UP ISSUE

- none. `scripts/wiring-scan.sh` not matching `if let Ok(..)` is already tracked in #733; the lock that makes option (a) safe landed in #714. Nothing new to route.

### RAN vs READ

RAN (this session, Windows 11 / Git Bash, read-only; no Cargo build, no lane):

- `gh pr view 758 …`, `gh pr diff 758 [--name-only]`, `gh issue view 741`, `git fetch origin main`, `git merge-base`, `git log --format=%s|%B origin/main..$SHA`
- REVIEW.md §3 classifier block on the changed-file list → `classes=code-risk`, `canonical_class=code-risk`
- U2 / U3 / U4 greps → 0 hits / `Issue: #741` / 0 offenders
- `gh pr checks 758` → 9/9 pass
- C2 / C3 / C4 / C5 → 1 added test / 5 candidates all in `#[cfg(test)]` / 0 / 0
- R1 / R3 → 0 / 0
- `sh scripts/wiring-scan.sh origin/main 4238e9e254aa80459ec75fd86f4c66601af84dd4` → no new pub surfaces, no swallow patterns
- `grep -n 'if let Ok(Some(disk))' crates/edda-conductor/src/state/persist.rs` → empty (doneWhen 1)
- caller and writer enumeration across `crates/` for `save_state_reconciled`, `save_state(`, `load_state(`
- opened `persist.rs:1-175` and `cmd_conduct.rs:165-180` to confirm lock/`Drop` behaviour and startup handling

READ (not re-run — `ladder` L2: exact-head CI plus the implementer's receipt cover the gate set, and this diff leaves no Windows coverage gap):

- Implementer L1 receipt in the PR body: `cargo test -p edda-conductor` 360 passed / 0 failed; `cargo clippy -p edda-conductor --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean; plus the pre-fix failing run of the new test with its verbatim panic at `persist.rs:501`.
- Exact-head CI @ `4238e9e254aa80459ec75fd86f4c66601af84dd4`: Format, CI Gate, Detect code changes, Clippy (ubuntu/macos/windows), Test (ubuntu/macos/windows) — all pass. Covers `cargo fmt --all --check`, workspace clippy on all three OSes, and `cargo test --workspace` on Linux and macOS.
- What that does not cover: the Windows job runs only 7 crates' test targets. `edda-conductor` is one of the 7 and this diff touches no other crate, so the usual Windows gap is empty here and no focused local run was owed. The pre-fix failure is a READ, but it is fix-caused by inspection (see C2) rather than an unverified claim.

### Verdict

LGTM (P0=0, P1=0)

Issue #712, opened 2026-09-02
fix(fleet): lane-stop.ps1 matches only edda-lane-*, so review lanes survive a stop and keep running


fagemx:
`scripts/fleet/lane-stop.ps1` (GH-672, merged as `fca3b4d`) kills a lane's whole process tree instead of just the wrapper. It selects tasks by the name pattern `edda-lane-*`. **Review lanes register as `edda-review-pr<N>-r<R>`**, so none of them are covered, and stopping one leaves its `pi` / `claude` child alive.

## Confirmed twice, independently

**Live, by the controller, 2026-09-02 ~23:05.** A review lane on PR #689 was opening browser windows on the operator's desktop (GH-691). Stopping it:

```
Stop-ScheduledTask -TaskName 'edda-review-pr689-r1'
(Get-ScheduledTask ...).State   ->  Ready
```

The task reported `Ready`, and the `node` child was still running and still opening windows. It had to be killed by PID:

```
Get-Process -Id 17928   ->  node, started 23:00:31
Stop-Process -Id 17928 -Force
```

Only then did the popups stop. The operator saw `git-diff.html` reappear for the whole interval between "stopped" and the PID kill.

**Independently, by the lane implementing PR #709**, which was checking GH-691's doneWhen item "the stop path verifies child termination". It reproduced the same thing and reported it rather than silently skipping the item — its words: after `Stop-ScheduledTask` **and** `Unregister-ScheduledTask`, the `node` child was still running and had to be killed by PID.

## Why the gap exists

GH-672 was written from the worker-lane experience, where every task is `edda-lane-<something>`. Review lanes are launched by `scripts/review-pr.sh`, which names its task `edda-review-pr$PR-r$ROUND`. The naming divergence is invisible until you try to stop one — the script reports success on an empty match.

This is the same shape as several defects found today: **a failure that looks like a success.** `Stop-ScheduledTask` returning and the state reading `Ready` is not evidence the lane died, any more than an empty process list is evidence it failed rather than finished. Only the actual child process, or a durable artifact, settles it.

## doneWhen

- `lane-stop.ps1` covers review lanes as well as worker lanes. Match the task-name patterns the fleet actually registers rather than one prefix — today that is `edda-lane-*` and `edda-review-pr*-r*`; deriving the set from the registered tasks would survive the next naming change.
- Stopping a review lane leaves no surviving descendant. Prove it: start one, stop it, and show `Get-Process` empty for its child, with the PID that existed before the stop.
[...]

PR #759, opened 2026-09-03
fix(fleet): support review lanes and action-based wrapper resolution in lane-stop and lane-status (GH-712)

fagemx:
## Summary

Issue: #712

Fixes `scripts/fleet/lane-stop.ps1` and `scripts/fleet/lane-status.ps1` to cover review lanes (`edda-review-pr*-r*`) in addition to worker lanes (`edda-lane-*`).

### Root Cause
- `lane-stop.ps1` hardcoded `$TaskName = "edda-lane-$Name"`, failing to match review lane tasks registered as `edda-review-pr<PR>-r<ROUND>`. Review lanes could not be targeted, leaving their process trees running.
- `lane-stop.ps1` assumed wrappers were always located at `$LogDir\$Name.wrapper.ps1`, whereas review lanes place wrappers under `$EDDA_FLEET_SCRATCH` with names like `review-pr<PR>-r<ROUND>.ps1`.
- `lane-status.ps1` defaulted to querying only `edda-lane-*`, rendering review lanes invisible to fleet status polling.

### Changes
- **Flexible Task Resolution**:
  - `lane-stop.ps1` and `lane-status.ps1` now resolve task names across candidates: `$Name`, `edda-lane-$Name`, and `edda-$Name`, cleanly matching both worker lanes and review lanes (`review-pr<PR>-r<ROUND>` or `edda-review-pr<PR>-r<ROUND>`).
  - When no `-Name` is supplied, `lane-status.ps1` enumerates both `edda-lane-*` and `edda-review-*`.
  - Non-matching names fail loudly with clear diagnostics listing the checked candidate task names.
- **Action-Based Wrapper Resolution**:
  - Extracts the `-File` parameter from the scheduled task's registered action arguments, locating the true wrapper script regardless of directory or naming convention (`.wrapper.ps1` vs `.ps1`).
  - Derives log and done file paths directly from the resolved wrapper.
- **Stopped-Lane Live Process Check**:
  - `lane-status.ps1` adds `liveProcs=<N>`, allowing the controller to immediately observe in one command whether any processes matching the lane's wrapper are still running.

### Pre-fix Regression Evidence
```text
Task: edda-review-pr999-r1 (Running, child PID 40588)
Attempting stop with -Name review-pr999-r1:
lane-stop: no scheduled task edda-lane-review-pr999-r1 (scripts/fleet/lane-status.ps1 lists what exists)
Attempting stop with -Name edda-review-pr999-r1:
lane-stop: no scheduled task edda-lane-edda-review-pr999-r1 (scripts/fleet/lane-status.ps1 lists what exists)
VULNERABILITY REPRODUCED: Review child PID 40588 is STILL RUNNING! lane-stop cannot target review lanes!
```

### Post-fix Verification
```text
--- 1. lane-status before stop ---
edda-review-pr999-r1 state=Running lastTaskResult=267009 logBytes=36 done=False head=- cwd=C:\Users\fagem\AppData\Local\Temp\edda-review-test liveProcs=1
Extracted Child PID: 43388
--- 2. lane-stop ---
task=edda-review-pr999-r1 state=Ready
terminated=2 pids=43388,42492
residual=0 (CommandLine match + process tree verification)
--- 3. verify child PID is dead ---
SUCCESS: Child PID 43388 is COMPLETELY DEAD!
--- 4. lane-status after stop ---
edda-review-pr999-r1 state=Ready lastTaskResult=267014 logBytes=103 done=True head=- cwd=C:\Users\fagem\AppData\Local\Temp\edda-review-test liveProcs=0
--- 5. test non-existent task fails loudly ---
lane-stop on non-existent exit code: 1, output: lane-stop: no scheduled task matching 'non-existent-lane-xyz' (checked: non-existent-lane-xyz, edda-lane-non-existent-lane-xyz, edda-non-existent-lane-xyz; scripts/fleet/lane-status.ps1 lists what exists)
```
- Worker lane backward compatibility verified on `test-worker` (`edda-lane-test-worker` stopped cleanly, child killed, `liveProcs=0`).
- PowerShell AST parse: 0 errors on both `lane-stop.ps1` and `lane-status.ps1`.

### DIFF
--- a/scripts/fleet/lane-status.ps1
+++ b/scripts/fleet/lane-status.ps1
@@ -15,7 +15,7 @@ param(
 )
 
 $tasks = if ($Name) {
-  $cand = @($Name, "edda-lane-$Name", "edda-$Name")
+  $cand = @("edda-$Name", "edda-lane-$Name", $Name)
   $found = @()
   foreach ($c in $cand) {
     $t = @(Get-ScheduledTask -TaskName $c -ErrorAction SilentlyContinue)
@@ -34,24 +34,38 @@ if ($tasks.Count -eq 0) {
   exit 1
 }
 
-# Snapshot once to check live processes (GH-712: "A stopped-lane check is available to the controller as one command")
+# Snapshot once and index children for live process tree traversal (GH-712 P1-3)
 $allProcs = @(Get-CimInstance Win32_Process)
+$childrenOf = @{}
+foreach ($p in $allProcs) {
+  $ppid = [int]$p.ParentProcessId
+  $procId = [int]$p.ProcessId
+  if (-not $childrenOf.ContainsKey($ppid)) {
+    $childrenOf[$ppid] = New-Object System.Collections.Generic.List[int]
+  }
+  $childrenOf[$ppid].Add($procId)
+}
 
 foreach ($t in $tasks) {
   $info = Get-ScheduledTaskInfo -TaskName $t.TaskName
 
   $wrapper = $null
   $actionArg = if ($t.Actions -and $t.Actions.Count -gt 0) { $t.Actions[0].Arguments } else { $null }
-  if ($actionArg -and $actionArg -match '-File\s+"?([^"\s]+)"?') {
-    $wrapper = $Matches[1]
+  if ($actionArg -and $actionArg -match '-File\s+("([^"]+)"|''([^'']+)''|([^\s]+))') {
+    $extracted = if ($Matches[2]) { $Matches[2] } elseif ($Matches[3]) { $Matches[3] } else { $Matches[4] }
+    if (Test-Path -LiteralPath $extracted -PathType Leaf) {
+      $wrapper = $extracted
+    }
   }
 
   $log = $null
   $done = $null
   if ($wrapper -and (Test-Path -LiteralPath $wrapper)) {
+    # Strip -lane suffix for review lanes where wrapper is review-pr<N>-r<R>-lane.ps1 (GH-712 P1-2)
     $base = $wrapper -replace '\.(wrapper\.)?ps1$', ''
-    $log = "$base.log"
-    $done = "$base.done"
+    $cleanBase = $base -replace '-lane$', ''
+    $log = if (Test-Path -LiteralPath "$cleanBase.log") { "$cleanBase.log" } elseif (Test-Path -LiteralPath "$base.log") { "$base.log" } else { "$cleanBase.log" }
+    $done = if (Test-Path -LiteralPath "$cleanBase.done") { "$cleanBase.done" } elseif (Test-Path -LiteralPath "$base.done") { "$base.done" } else { "$cleanBase.done" }
   } else {
     $lane = $t.TaskName -replace '^edda-lane-', '' -replace '^edda-', ''
     $log = Join-Path $LogDir "$lane.log"
@@ -67,10 +81,35 @@ foreach ($t in $tasks) {
     if ($LASTEXITCODE -eq 0 -and $h) { $head = $h }
   }
 
-  # Live process count: check if any process currently on system names this wrapper
+  # Live process count: traverse wrapper, brief, and their full descendant trees (GH-712 P1-3)
   $liveProcs = 0
   if ($wrapper) {
-    $liveProcs = @($allProcs | Where-Object { $_.CommandLine -and $_.CommandLine.Contains($wrapper) }).Count
+    $brief = if (Test-Path -LiteralPath $wrapper) {
+      $wb = Get-Content -LiteralPath $wrapper -Raw
+      if ($wb -match '--prompt-file\s+[''"]([^''"]+)[''"]') {
+        $b = $Matches[1].Trim()
+        if ($b.Length -gt 4) { $b } else { $null }
+      } else { $null }
+    } else { $null }
+
+    $seeds = @($allProcs | Where-Object {
+      $_.ProcessId -ne $PID -and $_.CommandLine -and (
+        $_.CommandLine.Contains($wrapper) -or
+        ($brief -and $_.CommandLine.Contains($brief)))
+    })
+    $liveSet = New-Object System.Collections.Generic.HashSet[int]
+    $q = New-Object System.Collections.Generic.Queue[int]
+    foreach ($s in $seeds) { [void]$liveSet.Add([int]$s.ProcessId); $q.Enqueue([int]$s.ProcessId) }
+    while ($q.Count -gt 0) {
+      $curr = $q.Dequeue()
+      if (-not $childrenOf.ContainsKey($curr)) { continue }
+      foreach ($c in $childrenOf[$curr]) {
+        if ($c -ne $PID -and $liveSet.Add($c)) {
+          $q.Enqueue($c)
+        }
+      }
+    }
+    $liveProcs = $liveSet.Count
   }
 
   "{0} state={1} lastTaskResult={2} logBytes={3} done={4} head={5} cwd={6} liveProcs={7}" -f `
--- a/scripts/fleet/lane-status.ps1
+++ b/scripts/fleet/lane-status.ps1
@@ -1,10 +1,10 @@
 # lane-status.ps1 — one-line-per-lane status for fleet lanes launched by
-# lane-launch.ps1. Replaces "check the log file's timestamp by hand" when
+# lane-launch.ps1 or review-pr.sh. Replaces "check the log file's timestamp by hand" when
 # asking whether a lane is alive (fleet.lane-launch / fleet.lane-dispatch).
 #
 # usage:
-#   pwsh -NoProfile -File scripts/fleet/lane-status.ps1                 # all edda-lane-* tasks
-#   pwsh -NoProfile -File scripts/fleet/lane-status.ps1 -Name <lane>    # one lane
+#   pwsh -NoProfile -File scripts/fleet/lane-status.ps1                 # all edda-lane-* and edda-review-* tasks
+#   pwsh -NoProfile -File scripts/fleet/lane-status.ps1 -Name <lane>    # one lane (worker or review)
 #
 # Per lane: task state, LastTaskResult, log size, done-file presence, and the
 # worktree's short HEAD (read from the task's WorkingDirectory).
@@ -14,27 +14,66 @@ param(
   [string]$LogDir = "$env:TEMP\edda-lanes"
 )
 
-$pattern = if ($Name) { "edda-lane-$Name" } else { 'edda-lane-*' }
-$tasks = @(Get-ScheduledTask -TaskName $pattern -ErrorAction SilentlyContinue | Sort-Object TaskName)
+$tasks = if ($Name) {
+  $cand = @($Name, "edda-lane-$Name", "edda-$Name")
+  $found = @()
+  foreach ($c in $cand) {
+    $t = @(Get-ScheduledTask -TaskName $c -ErrorAction SilentlyContinue)
+    if ($t.Count -gt 0) { $found = $t; break }
+  }
+  $found
+} else {
+  @(Get-ScheduledTask -TaskName 'edda-*' -ErrorAction SilentlyContinue | Where-Object {
+    $_.TaskName -like 'edda-lane-*' -or $_.TaskName -like 'edda-review-*'
+  } | Sort-Object TaskName)
+}
+
 if ($tasks.Count -eq 0) {
-  [Console]::Error.WriteLine("lane-status: no scheduled task matching $pattern")
+  $target = if ($Name) { "matching '$Name'" } else { 'matching edda-lane-* or edda-review-*' }
+  [Console]::Error.WriteLine("lane-status: no scheduled task $target")
   exit 1
 }
 
+# Snapshot once to check live processes (GH-712: "A stopped-lane check is available to the controller as one command")
+$allProcs = @(Get-CimInstance Win32_Process)
+
 foreach ($t in $tasks) {
-  $lane = $t.TaskName -replace '^edda-lane-', ''
   $info = Get-ScheduledTaskInfo -TaskName $t.TaskName
-  $log = Join-Path $LogDir "$lane.log"
-  $done = Join-Path $LogDir "$lane.done"
-  $logBytes = if (Test-Path -LiteralPath $log) { (Get-Item -LiteralPath $log).Length } else { 0 }
-  $doneExists = Test-Path -LiteralPath $done
-  $cwd = $t.Actions[0].WorkingDirectory
+
+  $wrapper = $null
+  $actionArg = if ($t.Actions -and $t.Actions.Count -gt 0) { $t.Actions[0].Arguments } else { $null }
+  if ($actionArg -and $actionArg -match '-File\s+"?([^"\s]+)"?') {
+    $wrapper = $Matches[1]
+  }
+
+  $log = $null
+  $done = $null
+  if ($wrapper -and (Test-Path -LiteralPath $wrapper)) {
+    $base = $wrapper -replace '\.(wrapper\.)?ps1$', ''
+    $log = "$base.log"
+    $done = "$base.done"
+  } else {
+    $lane = $t.TaskName -replace '^edda-lane-', '' -replace '^edda-', ''
+    $log = Join-Path $LogDir "$lane.log"
+    $done = Join-Path $LogDir "$lane.done"
+  }
+
+  $logBytes = if ($log -and (Test-Path -LiteralPath $log)) { (Get-Item -LiteralPath $log).Length } else { 0 }
+  $doneExists = if ($done) { Test-Path -LiteralPath $done } else { $false }
+  $cwd = if ($t.Actions -and $t.Actions.Count -gt 0) { $t.Actions[0].WorkingDirectory } else { '' }
   $head = '-'
   if ($cwd -and (Test-Path -LiteralPath $cwd)) {
     $h = & git -C $cwd rev-parse --short HEAD 2>$null
     if ($LASTEXITCODE -eq 0 -and $h) { $head = $h }
   }
-  "{0} state={1} lastTaskResult={2} logBytes={3} done={4} head={5} cwd={6}" -f `
-    $t.TaskName, $t.State, $info.LastTaskResult, $logBytes, $doneExists, $head, $cwd
+
+  # Live process count: check if any process currently on system names this wrapper
+  $liveProcs = 0
+  if ($wrapper) {
+    $liveProcs = @($allProcs | Where-Object { $_.CommandLine -and $_.CommandLine.Contains($wrapper) }).Count
+  }
+
+  "{0} state={1} lastTaskResult={2} logBytes={3} done={4} head={5} cwd={6} liveProcs={7}" -f `
+    $t.TaskName, $t.State, $info.LastTaskResult, $logBytes, $doneExists, $head, $cwd, $liveProcs
 }
 exit 0
--- a/scripts/fleet/lane-stop.ps1
+++ b/scripts/fleet/lane-stop.ps1
@@ -27,9 +27,7 @@
 # lane-status.ps1): 0 = stopped (N processes terminated, or the lane was not
 # running); 1 = error: no matching task, wrapper missing, or processes
 # survived the kill.
-#
-# Never points at another lane: -Name selects exactly one edda-lane-* task,
-# and the CommandLine matches are the lane's own wrapper and brief paths.
+# Matches tasks registered across the fleet (edda-lane-*, edda-review-pr*-r*, GH-712).
 param(
   [Parameter(Mandatory = $true)][string]$Name,
   [string]$LogDir = "$env:TEMP\edda-lanes"
@@ -45,24 +43,64 @@ function Fail([string]$Msg) {
 if ($Name -notmatch '^[A-Za-z0-9][A-Za-z0-9._-]*$') {
   Fail "-Name '$Name' may contain only letters, digits, dot, underscore, hyphen"
 }
-$TaskName = "edda-lane-$Name"
-$task = Get-ScheduledTask -TaskName $TaskName -ErrorAction SilentlyContinue
-if (-not $task) { Fail "no scheduled task $TaskName (scripts/fleet/lane-status.ps1 lists what exists)" }
 
-$Wrapper = Join-Path $LogDir "$Name.wrapper.ps1"
-$Log = Join-Path $LogDir "$Name.log"
-$Done = Join-Path $LogDir "$Name.done"
-if (-not (Test-Path -LiteralPath $Wrapper -PathType Leaf)) {
-  Fail "wrapper not found: $Wrapper; cannot verify the lane's process tree by CommandLine"
+# Resolve scheduled task: support exact name, edda-lane-$Name, and edda-$Name (covers edda-review-pr*-r*) (GH-712)
+$task = $null
+$TaskName = $null
+$candidateTaskNames = @($Name, "edda-lane-$Name", "edda-$Name")
+foreach ($cand in $candidateTaskNames) {
+  $t = Get-ScheduledTask -TaskName $cand -ErrorAction SilentlyContinue
+  if ($t) {
+    $TaskName = $cand
+    $task = $t
+    break
+  }
+}
+if (-not $task) {
+  Fail "no scheduled task matching '$Name' (checked: $($candidateTaskNames -join ', '); scripts/fleet/lane-status.ps1 lists what exists)"
 }
 
+# Resolve wrapper script:
+# 1. From the task's registered action argument line (works for both worker and review lanes)
+$Wrapper = $null
+$actionArg = if ($task.Actions -and $task.Actions.Count -gt 0) { $task.Actions[0].Arguments } else { $null }
+if ($actionArg -and $actionArg -match '-File\s+"?([^"\s]+)"?') {
+  $extracted = $Matches[1]
+  if (Test-Path -LiteralPath $extracted -PathType Leaf) {
+    $Wrapper = $extracted
+  }
+}
+
+# 2. Fallback to LogDir candidate paths if not in action arguments
+if (-not $Wrapper) {
+  $lane = $TaskName -replace '^edda-lane-', '' -replace '^edda-', ''
+  $candidates = @(
+    (Join-Path $LogDir "$Name.wrapper.ps1"),
+    (Join-Path $LogDir "$lane.wrapper.ps1"),
+    (Join-Path $LogDir "$Name.ps1"),
+    (Join-Path $LogDir "$lane.ps1")
+  )
+  foreach ($c in $candidates) {
+    if (Test-Path -LiteralPath $c -PathType Leaf) {
+      $Wrapper = $c
+      break
+    }
+  }
+}
+
+if (-not $Wrapper -or -not (Test-Path -LiteralPath $Wrapper -PathType Leaf)) {
+  Fail "wrapper not found for task $TaskName; cannot verify the lane's process tree by CommandLine"
+}
+
+$base = $Wrapper -replace '\.(wrapper\.)?ps1$', ''
+$Log = "$base.log"
+$Done = "$base.done"
+
 # What the lane's processes look like on the command line. The wrapper path
-# matches the wrapper itself; the brief path (extracted from the wrapper the
-# same way the #606 clobber-guard does) matches the `edda dispatch` child and
-# stays identifiable after the wrapper is already dead — exactly the
-# State=Ready-but-still-working orphan Stop-ScheduledTask leaves behind.
+# matches the wrapper itself; the brief path matches the child process and
+# stays identifiable after the wrapper is already dead.
 $wrapperBody = Get-Content -LiteralPath $Wrapper -Raw
-$brief = if ($wrapperBody -match "--prompt-file '([^']+)'") { $Matches[1] } else { $null }
+$brief = if ($wrapperBody -match "--prompt-file\s+['""]?([^'""]+)['""]?") { $Matches[1] } else { $null }
 
 function Get-ProcessSnapshot {
   # Index children by parent PID using explicit [int] keys (GH-706: Win32_Process
--- a/scripts/fleet/lane-stop.ps1
+++ b/scripts/fleet/lane-stop.ps1
@@ -47,7 +47,7 @@ if ($Name -notmatch '^[A-Za-z0-9][A-Za-z0-9._-]*$') {
 # Resolve scheduled task: support exact name, edda-lane-$Name, and edda-$Name (covers edda-review-pr*-r*) (GH-712)
 $task = $null
 $TaskName = $null
-$candidateTaskNames = @($Name, "edda-lane-$Name", "edda-$Name")
+$candidateTaskNames = @("edda-$Name", "edda-lane-$Name", $Name)
 foreach ($cand in $candidateTaskNames) {
   $t = Get-ScheduledTask -TaskName $cand -ErrorAction SilentlyContinue
   if ($t) {
@@ -64,8 +64,8 @@ if (-not $task) {
 # 1. From the task's registered action argument line (works for both worker and review lanes)
 $Wrapper = $null
 $actionArg = if ($task.Actions -and $task.Actions.Count -gt 0) { $task.Actions[0].Arguments } else { $null }
-if ($actionArg -and $actionArg -match '-File\s+"?([^"\s]+)"?') {
-  $extracted = $Matches[1]
+if ($actionArg -and $actionArg -match '-File\s+("([^"]+)"|''([^'']+)''|([^\s]+))') {
+  $extracted = if ($Matches[2]) { $Matches[2] } elseif ($Matches[3]) { $Matches[3] } else { $Matches[4] }
   if (Test-Path -LiteralPath $extracted -PathType Leaf) {
     $Wrapper = $extracted
   }
@@ -78,7 +78,9 @@ if (-not $Wrapper) {
     (Join-Path $LogDir "$Name.wrapper.ps1"),
     (Join-Path $LogDir "$lane.wrapper.ps1"),
     (Join-Path $LogDir "$Name.ps1"),
-    (Join-Path $LogDir "$lane.ps1")
+    (Join-Path $LogDir "$lane.ps1"),
+    (Join-Path $LogDir "$Name-lane.ps1"),
+    (Join-Path $LogDir "$lane-lane.ps1")
   )
   foreach ($c in $candidates) {
     if (Test-Path -LiteralPath $c -PathType Leaf) {
@@ -92,15 +94,21 @@ if (-not $Wrapper -or -not (Test-Path -LiteralPath $Wrapper -PathType Leaf)) {
   Fail "wrapper not found for task $TaskName; cannot verify the lane's process tree by CommandLine"
 }
 
+# Derive .log and .done paths: review lanes name the wrapper review-pr<N>-r<R>-lane.ps1
+# while .log and .done are review-pr<N>-r<R>.log/.done (without the -lane suffix, GH-712 P1-1).
 $base = $Wrapper -replace '\.(wrapper\.)?ps1$', ''
-$Log = "$base.log"
-$Done = "$base.done"
+$cleanBase = $base -replace '-lane$', ''
+$Log = if (Test-Path -LiteralPath "$cleanBase.log") { "$cleanBase.log" } elseif (Test-Path -LiteralPath "$base.log") { "$base.log" } else { "$cleanBase.log" }
+$Done = if (Test-Path -LiteralPath "$cleanBase.done") { "$cleanBase.done" } elseif (Test-Path -LiteralPath "$base.done") { "$base.done" } else { "$cleanBase.done" }
 
 # What the lane's processes look like on the command line. The wrapper path
 # matches the wrapper itself; the brief path matches the child process and
 # stays identifiable after the wrapper is already dead.
 $wrapperBody = Get-Content -LiteralPath $Wrapper -Raw
-$brief = if ($wrapperBody -match "--prompt-file\s+['""]?([^'""]+)['""]?") { $Matches[1] } else { $null }
+$brief = if ($wrapperBody -match '--prompt-file\s+[''"]([^''"]+)[''"]') {
+  $b = $Matches[1].Trim()
+  if ($b.Length -gt 4) { $b } else { $null }
+} else { $null }
 
 function Get-ProcessSnapshot {
   # Index children by parent PID using explicit [int] keys (GH-706: Win32_Process


fagemx:
## Code Review: Round 1 — PR #759 @ 4ba71c37cbe0ba1fed93bbe8913de46ea3229a1c

- model_observed: unverified (no session file / `modelUsage` key readable from this worktree)
- spec: review-spec-v1.2
- escalations: none (R2 has no candidate lines; D5 not routed — no docs/skills paths in the diff)
- cost: ~15 tool calls, single round, no Cargo gate run (no Rust/product change in diff)

### IN SCOPE

Changed paths: `scripts/fleet/lane-stop.ps1`, `scripts/fleet/lane-status.ps1` (2 files, +107/−30).

Changed behaviour: (a) scheduled-task resolution by candidate list `$Name` / `edda-lane-$Name` / `edda-$Name` in both scripts; (b) default enumeration in `lane-status.ps1` widened to `edda-lane-*` + `edda-review-*`; (c) wrapper resolved from the registered task action's `-File` argument, with a `LogDir` fallback; (d) `$Log`/`$Done` derived from the resolved wrapper path instead of `$LogDir\$Name.*`; (e) new `liveProcs=<N>` field on the status line; (f) `--prompt-file` brief regex widened to accept double quotes.

Direct callers/consumers checked: `scripts/review-pr.sh` (registers `edda-review-pr<N>-r<R>`, generates the lane wrapper and the log/done paths — lines 346–348, 512–519), `scripts/pr-review-watch.sh:343-344` (polls the review lane's `.done` and `.log`), `scripts/fleet/lane-launch.ps1` (worker-lane wrapper/log/done shape), `docs/guides/operator-runbook.md:41`, `docs/guides/pr-review-watcher.md:102,168`.

Acceptance ceiling: GH-712 doneWhen (4 items). Current-base integration: base `662ade2` is the merge base; no conflict surface.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 759 --name-only` → `scripts/fleet/lane-status.ps1`, `scripts/fleet/lane-stop.ps1`; both are the surface GH-712 names (item 1 → `lane-stop.ps1`, item 4 → a controller status command) |
| U2 | any | P1 | PASS | body grep → 0 lines; `git log --format=%B origin/main..$SHA` grep → 0 lines. GH-712 is only partially delivered (see P1-1/P1-3), so the absence of a closing keyword is correct |
| U3 | any | P1 | PASS | `gh pr view 759 --json body --jq .body \| awk …` → `Issue: #712` |
| U4 | any | P1 | PASS | subject grep → empty (`fix(fleet): support review lanes and action-based wrapper resolution in lane-stop and lane-status (GH-712)`) |
| U6 | any | P0 | PASS | `gh pr checks 759`: Format pass, Clippy ubuntu/macos/windows pass, Test ubuntu/macos pass, Detect code changes pass, Test (windows) pending. No Rust/`Cargo.*` change in the diff, so no Cargo gate is load-bearing here. No L1 receipt is required or claimed for a PowerShell-only diff; the PR body carries manual pre-fix/post-fix evidence instead (see P1-1 on its validity) |
| U7 | any | P1 | PASS | whole scoped audit finished before posting; all blocking findings batched in this round |
| C1 | code-plain | P0 | **FAIL** | doneWhen item 4 ("a stopped-lane check … so 'is it actually dead' does not depend on remembering to look up a PID") is not met in substance — see P1-3. doneWhen items 1 and 3 met; item 2's proof is not against the real review-lane shape — see P1-1 |
| C2 | code-plain | P0 | N.A.(no `crates/**` in the diff; the repo has no PowerShell test harness — `scripts/test-pr-review-watch.sh` is the only script-level harness and covers the watcher, not `lane-*.ps1`) | `git diff … -- 'crates/**' \| grep -c '#\[test\]\|fn test_'` → 0 files. See FOLLOW-UP ISSUE |
| C3 | code-plain | P1 | N.A.(no Rust in the diff) | `git diff … -- 'crates/**'` empty |
| C4 | code-plain | P1 | PASS | `grep -nE '^\+.*(unsafe[[:space:]]*\{\|#\[allow\(clippy::all\)\]\|#!\[allow)'` → no lines |
| C5 | code-plain | P1 | N.A.(no `crates/` paths changed) | `gh pr diff 759 --name-only \| grep '^crates/'` → empty |
| R1 | code-risk | P0 | PASS | `grep -E 'rm -rf\|git clean\|reset --hard\|git rm\|--delete-branch\|--force\|Remove-Item'` on added lines → no hits. The pre-existing `Stop-Process -Force` / `taskkill /T /F` at `lane-stop.ps1:191,198` are unchanged by this diff; their trigger is the target set built from the wrapper/brief CommandLine seeds and their descendants — see P2-2 for the one way that set can be mis-seeded |
| R2 | code-risk | `[判斷]` | N.A.(no added line mixes `\|\|`/`-or` with `&&`/`-and`) | `grep -nE '^\+.*(\|\|\|-or).*(&&\|-and)'` → no lines |
| R3 | code-risk | P0 | N.A.(no `.sh`/`.bash` changed) — focused substitute RAN | PowerShell AST parse of both changed files → `parseErrors=0` each |
| R4 | code-risk | P1 | PASS | Neither script creates a temp directory or lock; `lane-stop.ps1` writes `$Log`/`$Done` only on the success path, which is pre-existing GH-672 behaviour. `$ErrorActionPreference = 'Stop'` unchanged |
| R5 | code-risk | P1 | **FAIL** | Two interpolation/parse properties do not hold: the `-File` extractor does not accept a quoted path containing a space (P2-1), and the `--prompt-file` extractor's capture class does not exclude whitespace (P2-3). Property that does hold: `-Name` is still validated against `^[A-Za-z0-9][A-Za-z0-9._-]*$` at `lane-stop.ps1:43`, so no wildcard reaches `Get-ScheduledTask -TaskName`. Property that no longer holds: "two distinct inputs must not resolve to the same target" — see P2-2 |
| §5.5 | any | — | see Wiring table | `sh scripts/wiring-scan.sh origin/main $SHA` → "no new pub surfaces", "no swallow patterns on added lines" (Rust-oriented aid; the PowerShell surfaces below were enumerated by hand) |

### Wiring

|---|---|---|---|---|
| `liveProcs=<N>` status field | `lane-status.ps1:71-77` — count of `Win32_Process` rows whose `CommandLine` contains the wrapper path | Controller by eye (`docs/guides/operator-runbook.md:41`); no script parses it | **Success-only / under-covering**: counts wrapper matches only, never the brief-matched or descendant processes `lane-stop.ps1:131-135,218-222` treats as the lane. Prints `liveProcs=0` for a live orphan → P1-3 | field → stdout → operator. No test, no consumer asserts it |
| `$Log` / `$Done` derived from wrapper base (`lane-stop.ps1:95-97`) | `Add-Content $Log` / `Set-Content $Done` at `lane-stop.ps1:270-274` | `scripts/pr-review-watch.sh:343-344` polls `$SCRATCH/review-pr<N>-r<R>.done` and `.log` | **Wrong write end**: for the real review lane the derived name is `review-pr<N>-r<R>-lane.done`, which nothing reads; the polled `.done` is never written → P1-1 | write end → watcher poll loop. Reach is broken; no test covers it |
| `$log` / `$done` derived from wrapper base (`lane-status.ps1:52-54`) | read-only `Test-Path` / `Get-Item` | the `logBytes=` and `done=` columns of the status line | **Silent zero**: derived path never exists for a real review lane → `logBytes=0 done=False` forever → P1-2 | read end → stdout → operator |
| `-Name` now accepts a full task name (`lane-stop.ps1:50`, `lane-status.ps1:19`) | candidate list `@($Name, "edda-lane-$Name", "edda-$Name")`, first match wins | both scripts' own task resolution | First candidate is the bare `$Name`, matched against **all** scheduled tasks in every folder; the removed `lane-stop.ps1` comment ("Never points at another lane") is no longer true → P2-2 | flag → `Get-ScheduledTask` → `Stop-ScheduledTask`. No test |
| `edda-review-*` in the default enumeration (`lane-status.ps1:27-30`) | `Get-ScheduledTask -TaskName 'edda-*'` + `Where-Object` filter | operator runbook §START HERE poll | none swallowed; empty set still fails loudly at `lane-status.ps1:32-35` | flag-less path → stdout. Delivers GH-712 doneWhen item 4's "one command" shape |

### Findings

- **[P1] The wrapper→log/done derivation is wrong for the real review lane, so `lane-stop` writes its end record to files nothing reads and never writes the `.done` the watcher polls.** `review-pr.sh` names the wrapper `review-pr$PR-r$ROUND-**lane**.ps1` (`scripts/review-pr.sh:348`) while the log and done are `review-pr$PR-r$ROUND.log` / `.done` (`scripts/review-pr.sh:346-347`) — the `-lane` suffix is on the wrapper only. `lane-stop.ps1:95` computes `$base = $Wrapper -replace '\.(wrapper\.)?ps1$', ''`, which strips only `.ps1`, so `$Done` becomes `review-pr<N>-r<R>-lane.done`. Measured: `('C:\Users\fagem\.edda\fleet\review-pr759-r1-lane.ps1' -replace '\.(wrapper\.)?ps1$','')` → `C:\Users\fagem\.edda\fleet\review-pr759-r1-lane`. On-disk confirmation of the real name: `ls ~/.edda/fleet | grep review-pr709` → `review-pr709-r1-lane.ps1`. The consumer that loses its signal is `scripts/pr-review-watch.sh:343` (`DONE="$SCRATCH/review-pr$pr-r$round.done"`), whose documented behaviour when `.done` never appears is the 45-minute dead-verdict timeout (`docs/guides/pr-review-watcher.md:168`). This breaks GH-672's stated property — "a lane log with START but no EXIT is never left ambiguous" (`lane-stop.ps1:19-21`) — for exactly the lane class this PR adds. Note the PR body's post-fix evidence used a synthetic task whose wrapper was `review-pr999-r1.ps1`, and the body states review lane wrappers are "named like `review-pr<PR>-r<ROUND>.ps1`" — that name does not exist in `review-pr.sh`, so the evidence does not exercise the real review lane and doneWhen item 2 is not proven against it.
- **[P1] `lane-status` reports `logBytes=0 done=False` permanently for a real review lane, from the same root cause.** `lane-status.ps1:52-54` derives `$log`/`$done` from the same wrapper base, so for `review-pr<N>-r<R>-lane.ps1` it probes `review-pr<N>-r<R>-lane.log`/`.done`, which `review-pr.sh` never creates. The pre-PR code path (`Join-Path $LogDir "$lane.log"`) was also wrong for review lanes, but this PR is what claims to make review lanes visible, and the two columns the controller reads to answer "did it finish" are the two that stay constant. The `else` branch at `lane-status.ps1:56-60` is unreachable for a review lane because the `-File` extraction succeeds.
- **[P1] `liveProcs` misses the orphan class GH-712 exists for, so it reports `liveProcs=0` while the lane's child is alive.** `lane-status.ps1:73` counts only processes whose `CommandLine.Contains($wrapper)`. The lane's `edda dispatch` child is launched as `& edda dispatch … --prompt-file "<brief>"` (`scripts/review-pr.sh:392`); its command line contains the brief path, not the wrapper path, and the `claude`/`node` grandchildren contain neither. `lane-stop.ps1:131-135` and `218-222` deliberately seed on **both** the wrapper and the brief, and then traverse descendants, precisely because — in that file's own words (`lane-stop.ps1:99-101`) — the brief "matches the child process and stays identifiable after the wrapper is already dead". After a bare `Stop-ScheduledTask` (the incident in GH-712's own report: task `Ready`, `node` child still opening windows), the wrapper is gone and `liveProcs` prints `0`. That is the same "failure that looks like a success" GH-712 was filed about, and it means doneWhen item 4 — "so 'is it actually dead' does not depend on remembering to look up a PID" — is not delivered. Fix shape: seed on wrapper **or** brief and traverse descendants, i.e. reuse `lane-stop.ps1`'s seed/traverse logic rather than a substring count.
- **[P2] The `-File` extractor cannot read a quoted wrapper path containing a space.** `lane-stop.ps1:67` / `lane-status.ps1:47` use `-File\s+"?([^"\s]+)"?`; the capture class excludes whitespace even inside the quotes. Measured: `'-NoProfile -File "C:\Program Files\x\review-lane.ps1"' -match '-File\s+"?([^"\s]+)"?'` → `$Matches[1]` = `C:\Program`. Both generators emit a quoted path (`review-pr.sh:517`, `lane-launch.ps1:228`), so any user whose `$HOME`/`$env:TEMP` contains a space loses wrapper resolution. A worker lane is rescued by the `LogDir` fallback (`lane-stop.ps1:75-89`); a review lane is not, and hard-fails with "wrapper not found". Not triggered on this workstation (`C:\Users\fagem`). Suggested class: `-File\s+(?:"([^"]+)"|(\S+))`.
- **[P2] The candidate list tries the bare `$Name` first against every scheduled task on the system, so `-Name` can resolve a non-fleet task.** `lane-stop.ps1:50` / `lane-status.ps1:19` order the candidates `$Name`, `edda-lane-$Name`, `edda-$Name`, and `Get-ScheduledTask -TaskName <name>` searches all task folders. If a task literally named `<lane>` exists alongside `edda-lane-<lane>`, the unrelated one wins; `lane-stop.ps1:142` then issues `Stop-ScheduledTask` against it. Damage is bounded because wrapper resolution must still succeed, but the guarantee the diff deleted from the header ("Never points at another lane: `-Name` selects exactly one `edda-lane-*` task", old `lane-stop.ps1:31-32`) no longer holds and nothing replaced it. Suggested fix: put the `edda-`-prefixed candidates first, or require the resolved task name to match `^edda-(lane|review)-`.
- **[P2] The widened `--prompt-file` regex can over-capture on an unquoted argument.** `lane-stop.ps1:103` uses `--prompt-file\s+['"]?([^'"]+)['"]?`; `[^'"]+` excludes neither whitespace nor newlines, so an unquoted `--prompt-file <path> --session-id …` would capture everything up to the next quote character anywhere in the wrapper body, and the resulting `$brief` would match no `CommandLine` — silently disabling the orphan seed. Today both generators quote the argument (`review-pr.sh:392` double, `lane-launch.ps1:114` via `PsQuote` single), so this is latent, not live. Suggested class: `--prompt-file\s+(?:'([^']+)'|"([^"]+)"|(\S+))`.
- **[P2] `$task` from `Get-ScheduledTask` is not coerced to a single object in `lane-stop.ps1:52-57`.** If two folders hold tasks with the same name, `$task` is an array and `$task.Actions[0]`, `$task.State -eq 'Running'`, and `Get-ScheduledTask -TaskName $TaskName` at line 265 all behave on the collection. `lane-status.ps1:22` already wraps its result in `@(…)`; `lane-stop.ps1` does not.

### FOLLOW-UP ISSUE

- **No test harness covers `scripts/fleet/lane-*.ps1`.** `scripts/test-pr-review-watch.sh` shows the repo does test scripts, but nothing references `lane-stop`/`lane-status` (`grep -rln 'lane-stop\|lane-status' --include=*.rs --include=*.sh .` → no test files). Every fix to these scripts since GH-672 — PR #696, #740, and this one — has been verified by hand-built scheduled tasks in the PR body, which is exactly how the `-lane` suffix mismatch in P1-1 survived. A fixture-based harness that asserts wrapper→log/done derivation for both the `lane-launch.ps1` and the `review-pr.sh` shapes would be cheap and would have caught P1-1 and P1-2 mechanically. Out of GH-712's doneWhen; does not extend this PR.
- **`lane-status.ps1` fallback derivation for a dry-run lane.** `lane-launch.ps1:156` names its dry-run wrapper `<Name>.dryrun-wrapper.ps1`, which `\.(wrapper\.)?ps1$` reduces to `<Name>.dryrun-wrapper`, not `<Name>`. Harmless today because the dry-run path unregisters its own task (`lane-launch.ps1:201`), so no live task carries that action. Out of scope.
- **Log/done naming divergence in `review-pr.sh` itself.** The cleanest fix for P1-1/P1-2 may be to make `review-pr.sh` name its wrapper `review-pr<N>-r<R>.ps1` (matching its log/done), rather than teaching two consumers about the `-lane` suffix — but that touches a file outside this PR's surface and would need its own round.

### RAN vs READ

RAN (this worktree, Windows 11, Git Bash + pwsh 7):
- `gh pr view 759 --json headRefOid,…` → head `4ba71c37cbe0ba1fed93bbe8913de46ea3229a1c`, base `main`, state OPEN, not draft
- `git merge-base origin/main $SHA` → `662ade23ecd4640f4fd959bc51294208e12fec54` (matches the stated base)
- `gh pr diff 759 --name-only` → 2 paths, both `scripts/fleet/*.ps1` → classes `code-risk`, canonical `code-risk` (§3.1 rows `**/*.ps1` and `scripts/**`)
- U2/U3/U4 greps → results in the Rules table
- `gh pr checks 759` → 7 pass, `Test (windows-latest)` pending
- R1/R2/C3/C4/C5 greps → no hits (results in the Rules table)
- PowerShell AST `Parser::ParseFile` on both changed files → `parseErrors=0`
- PowerShell regex probes of the two derivations under review: `-replace '\.(wrapper\.)?ps1$',''` on the real review wrapper name → `…-r1-lane`; `-match '-File\s+"?([^"\s]+)"?'` on a space-bearing quoted path → `C:\Program`
- `ls ~/.edda/fleet | grep '^review-pr'` → real wrapper artefacts named `review-pr<N>-r<R>-lane.ps1`
- `sh scripts/wiring-scan.sh origin/main $SHA` → no new pub surfaces, no swallow patterns
- Source reads for consumer confirmation: `scripts/review-pr.sh:346-348,392,512-519`, `scripts/pr-review-watch.sh:343-344`, `scripts/fleet/lane-launch.ps1:84,106,114,228`

READ:
- Exact-head CI (run 33715859444): Format, Clippy ×3, Test (ubuntu/macos), Detect code changes all pass; `Test (windows-latest)` pending at review time. Covers the Rust workspace, which this diff does not touch. Covers **nothing** in `scripts/fleet/*.ps1` — no CI job lints, parses or executes these scripts, which is why the AST parse and the regex probes above were RAN locally.
- No L1 gate receipt is claimed in the PR body and none is required: the diff contains no Rust, no `Cargo.toml`/`Cargo.lock`, and no toolchain change, so no workspace Cargo gate is load-bearing for this SHA (`ladder`). The PR body's manual pre-fix/post-fix transcripts were read and are the only behavioural evidence offered; their limitation is stated in P1-1.
- GH-712 doneWhen (4 items) read as the acceptance ceiling; GH-672 / PR #696 and GH-706 / PR #740 read for the properties this change must preserve.

### Verdict

Changes Requested, P0=0, P1=3 — the wrapper→log/done derivation misses the real review lane's `-lane.ps1` suffix, so the stop path writes its end record where nothing reads it and `lane-status` reports `logBytes=0 done=False` forever; and `liveProcs` counts only wrapper matches, so it prints `0` for exactly the orphaned child GH-712 was filed about.

fagemx:
## Review Response: Round 1 (Commit `c4b1056`)

Addressing all P1 and P2 findings from Opus Round 1 review:

### P1-1 & P1-2: Review lane `-lane` suffix handling for `.log` and `.done`
- In both `scripts/fleet/lane-stop.ps1` and `scripts/fleet/lane-status.ps1`, added clean base resolution (`$cleanBase = $base -replace '-lane$', ''`).
- Resolves `review-pr<PR>-r<ROUND>-lane.ps1` to `review-pr<PR>-r<ROUND>.log` and `review-pr<PR>-r<ROUND>.done`.
- Fallbacks preserved for worker lanes (`worker-1.wrapper.ps1` -> `worker-1.log`/`worker-1.done`).

### P1-3: `liveProcs` in `lane-status.ps1` now traverses full descendant tree
- `lane-status.ps1` now indexes processes by `ParentProcessId` with explicit `[int]` keys.
- Extracts `$brief` from wrapper script using robust quote-matching regex (`--prompt-file\s+['"]([^'"]+)['"]`).
- Seeds include both wrapper and brief paths, then traverses all children/grandchildren via BFS. Surviving orphan descendants are accurately counted in `liveProcs`.

### P2 Improvements
- Action `-File` argument regex now handles single-quoted, double-quoted, and unquoted paths: `-match '-File\s+("([^"]+)"|''([^'']+)''|([^\s]+))'`.
- Candidate task name resolution prioritizes `"edda-$Name"` and `"edda-lane-$Name"` before bare `$Name`.

### Verification with Authentic Review Lane Naming
Tested with task `edda-review-pr709-r1`, wrapper `review-pr709-r1-lane.ps1`, log `review-pr709-r1.log`, done `review-pr709-r1.done`, brief `review-pr709-r1-brief.md`:
```text
--- 1. lane-status while running ---
edda-review-pr709-r1 state=Running lastTaskResult=267009 logBytes=36 done=False head=- cwd=C:\Users\fagem\AppData\Local\Temp\edda-review-test-712 liveProcs=4
PIDs: wrapper=20364, child=21096
Is child alive before stop? Yes
--- 2. lane-stop targeting review-pr709-r1 ---
task=edda-review-pr709-r1 state=Ready
terminated=2 pids=21096,3268
residual=0 (CommandLine match + process tree verification)
--- 3. verify child is dead ---
Is child alive after lane-stop? No
--- 4. lane-status after stop ---
edda-review-pr709-r1 state=Ready lastTaskResult=267014 logBytes=103 done=True head=- cwd=C:\Users\fagem\AppData\Local\Temp\edda-review-test-712 liveProcs=0
--- 5. verify .done and .log files ---
Done file exists: True Content: stopped
Log file exists: True Last line: === EXIT (stopped by lane-stop.ps1; terminated 2 process(es)) ===
```

fagemx:
## Code Review: Round 2 — PR #759 @ c4b1056bf4d44e9571acbe486f9517f30e4759d6

- model_observed: unverified (no session file / `modelUsage` key readable from this worktree)
- spec: review-spec-v1.2 (read at base `662ade23ecd4640f4fd959bc51294208e12fec54`; `REVIEW.md` is not in this diff)
- escalations: none (R2 has no candidate lines; D5 not routed — no docs/skills paths in the diff)
- cost: ~14 tool calls, delta round, no Cargo gate run (no Rust/product change in diff)

### IN SCOPE

**Delta round.** Target is `git diff 4ba71c37cbe0ba1fed93bbe8913de46ea3229a1c..c4b1056` (the Round 1 SHA → this SHA) plus a RAN confirmation that each Round 1 blocking finding is resolved (REVIEW.md §2). Not a re-review of the whole PR.

Delta: `scripts/fleet/lane-status.ps1` (+47/−8), `scripts/fleet/lane-stop.ps1` (+15/−7). Changed behaviour: (a) `.log`/`.done` derived from the wrapper base with the review lane's `-lane` suffix stripped and both spellings existence-probed (`lane-stop.ps1:99-102`, `lane-status.ps1:64-68`); (b) `liveProcs` replaced by a wrapper-**or**-brief seed set plus a full descendant BFS (`lane-status.ps1:84-113`); (c) `-File` extractor widened to double-quoted / single-quoted / unquoted with a `Test-Path -PathType Leaf` guard (`lane-stop.ps1:67-72`, `lane-status.ps1:54-59`); (d) task-name candidate order changed to `edda-$Name`, `edda-lane-$Name`, `$Name`; (e) two `-lane.ps1` entries added to the `LogDir` wrapper fallback (`lane-stop.ps1:82-83`); (f) `--prompt-file` extractor tightened to require quotes and to reject a ≤4-char capture.

Direct consumers re-checked at the new write end: `scripts/review-pr.sh:346-348` (LOG/DONE/LANE naming), `scripts/review-pr.sh:392` (the `--prompt-file "<brief>"` the brief seed matches), `scripts/pr-review-watch.sh:343-344,405-421,447-476` (what the watcher does once `.done` appears), `scripts/fleet/lane-launch.ps1:104-106` (worker-lane naming, must not regress).

Acceptance ceiling: GH-712 doneWhen (4 items).

### Round 1 blocking findings — resolution confirmed

| R1 finding | Status | RAN evidence |
|---|---|---|
| **P1-1** `lane-stop` derives `.log`/`.done` as `…-lane.log`/`…-lane.done`, so it writes where nothing reads and never writes the `.done` the watcher polls | **RESOLVED** | Generated a current-shape review lane with `sh scripts/review-pr.sh 759 99 --dry-run` (writes the launcher, starts nothing), then ran the shipped derivation on the real wrapper `…\review-pr759-r99-lane.ps1` → `log=…\review-pr759-r99.log`, `done=…\review-pr759-r99.done` — byte-identical to `review-pr.sh:346-347` and to the watcher's `DONE`/`LOG` at `pr-review-watch.sh:343-344` |
| **P1-2** `lane-status` reports `logBytes=0 done=False` forever for a review lane | **RESOLVED** | Same derivation, same code path (`lane-status.ps1:65-68`). The `else` default is `$cleanBase`, so the pre-launch window (log not yet created — `review-pr.sh:505,515` delete both before `Start-ScheduledTask`) also resolves to the correct path, not the `-lane` one |
| **P1-3** `liveProcs` counts only wrapper substring matches, so it prints `0` for the orphaned child GH-712 was filed about | **RESOLVED for the reported incident** | `lane-status.ps1:95-112` now seeds on wrapper **or** brief and BFS-traverses descendants — the seed/traverse shape Round 1 prescribed. Verified the brief seed actually fires on the current wrapper: the shipped `--prompt-file` regex against the generated lane → `brief=C:\…\review-pr759-r99-brief.md` (len 52 > the 4-char guard). In the GH-712 incident shape (`Stop-ScheduledTask` kills the wrapper; the `edda dispatch` child survives with `--prompt-file <brief>` on its command line) the child is a seed and `claude`/`node` grandchildren are reached by BFS. One narrower residual case remains — see P2-4, routed as FOLLOW-UP, not a blocker |

Round 1 P2s: **P2-1 (`-File` quoted path with a space) fixed** — probed all three shapes, `"C:\Program Files\x\review-lane.ps1"` → `C:\Program Files\x\review-lane.ps1`, `'C:\Program Files\y.ps1'` → `C:\Program Files\y.ps1`, bare `C:\a\b.ps1` → `C:\a\b.ps1`. **P2-2 (bare `$Name` first) mitigated** by reordering (`edda-` candidates now first); bare `$Name` remains a last-resort candidate. **P2-3 (`--prompt-file` over-capture) fixed** by requiring quotes; see P2-3 below for the narrowing that introduces. **P2-4 (`$task` not coerced with `@()`) not addressed** — still P2, still non-blocking.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 759 --name-only` → `scripts/fleet/lane-status.ps1`, `scripts/fleet/lane-stop.ps1`; unchanged surface from Round 1, both inside GH-712's |
| U2 | any | P1 | PASS | body grep → 0 lines; `git log --format=%B origin/main..$SHA` grep → 0 lines. All four doneWhen items are now delivered, so a closing keyword would be permitted — its absence is not a defect (`pr.closing-keyword=only-when-all-donewhen-delivered` makes it optional, not mandatory) |
| U3 | any | P1 | PASS | `Issue: #712` |
| U4 | any | P1 | PASS | subject grep → empty; new subject `fix(fleet): fix review lane log/done suffix and full process tree liveProcs count (GH-712)` |
| U6 | any | P0 | PASS | `gh pr checks 759` (run 33717488421): Format, Clippy ×3, Test ubuntu/macos, Detect code changes all **pass**; `Test (windows-latest)` pending. Not red, and not load-bearing: the delta contains no Rust, no `Cargo.*`, no toolchain change, so no Cargo gate applies and no L1 receipt is required (`ladder`) |
| U7 | any | P1 | PASS | Whole delta audited before posting; no blocking finding withheld. The one newly-noticed gap (P2-4) is neither fix-caused nor previously unobservable, so it is routed as FOLLOW-UP per `loop` item 2, not raised as a late blocker |
| C1 | code-plain | P0 | **PASS** | doneWhen 1 ✓ `lane-stop.ps1:50`, `lane-status.ps1:18,26-28` cover `edda-lane-*` and `edda-review-*`. doneWhen 2 ✓ implementer's Round-2 evidence uses the authentic shape (task `edda-review-pr709-r1`, wrapper `review-pr709-r1-lane.ps1`, log/done without `-lane`, brief `…-brief.md`), child PID 21096 alive before / dead after, `liveProcs 4→0`, `.done`=`stopped`, log carries `=== EXIT ===`; the load-bearing derivations in that claim were re-derived mechanically here. doneWhen 3 ✓ `lane-stop.ps1:59-61` (names every candidate tried), `lane-status.ps1:31-35`. doneWhen 4 ✓ `lane-status.ps1 -Name <lane>` prints `liveProcs` over wrapper + brief + full descendant tree |
| C2 | code-plain | P0 | N.A.(no `crates/**` in the diff; the repo has no PowerShell test harness — `scripts/test-review-pr.sh` covers `review-pr.sh`'s generated artefacts, not `lane-*.ps1`) | `gh pr diff --name-only \| grep '^crates/'` → empty. Same call as Round 1; the gap is carried as the standing FOLLOW-UP below |
| C3 | code-plain | P1 | N.A.(no Rust in the diff) | `git diff … -- 'crates/**'` empty |
| C4 | code-plain | P1 | PASS | blanket-allow / `unsafe` grep on the delta → no lines |
| C5 | code-plain | P1 | N.A.(no `crates/` paths changed) | selector → empty |
| R1 | code-risk | P0 | PASS | destructive grep on the delta's added lines → no hits. `Stop-Process -Force` / `taskkill /T /F` (`lane-stop.ps1:199,206`) are untouched by this delta; their trigger set is the wrapper/brief CommandLine seeds plus creation-date-guarded descendants, and this delta only makes the brief seed **stricter**, never wider |
| R2 | code-risk | `[判斷]` | N.A.(no added line mixes `\|\|`/`-or` with `&&`/`-and`) | grep → no lines |
| R3 | code-risk | P0 | N.A.(no `.sh`/`.bash` changed) — focused substitute RAN | PowerShell AST `Parser::ParseFile` → `lane-stop.ps1 parseErrors=0`, `lane-status.ps1 parseErrors=0` |
| R4 | code-risk | P1 | PASS | The delta creates no temp directory, file handle or lock that needs a release path. `Get-Content -Raw` is self-closing; `$ErrorActionPreference = 'Stop'` unchanged; the end-record write (`lane-stop.ps1:278-283`) is unchanged and now targets the correct path |
| R5 | code-risk | P1 | PASS | Properties confirmed: (i) the `-File` extractor round-trips all three real argument shapes including a path with a space (probed); (ii) `$Matches[2]`/`[3]` are **absent**, not empty-string-truthy, for non-participating alternation groups, so the `if/elseif/else` selection cannot pick a wrong group (`keys=4,1,0` measured on the unquoted arm); (iii) the extracted path is gated by `Test-Path -PathType Leaf` before use in both scripts; (iv) `-Name` is still validated against `^[A-Za-z0-9][A-Za-z0-9._-]*$` in `lane-stop.ps1:43`, so no wildcard reaches `Get-ScheduledTask`; (v) the two distinct wrapper shapes (`worker-1.wrapper.ps1`, `review-pr<N>-r<R>-lane.ps1`) encode to two distinct log/done pairs — verified both, no collision |
| §5.5 | any | — | see Wiring table | `sh scripts/wiring-scan.sh 4ba71c3 c4b1056` → "no new pub surfaces", "no swallow patterns on added lines" (Rust-oriented aid; the PowerShell write ends below were enumerated by hand) |

### Wiring

|---|---|---|---|---|
| `.done` / `.log` write end **re-pointed** at the watcher-polled path for review lanes | `lane-stop.ps1:99-102` derives `$cleanBase`; `lane-stop.ps1:279-281` writes `=== EXIT … ===` to `$Log` and `stopped` to `$Done` | `scripts/pr-review-watch.sh:343-344` (`DONE`/`LOG`), then `:405-421` | Not swallowed. Newly reachable path traced end to end: a `.done` containing `stopped` makes `is_done=1`, `extract_verdict` finds no `<<<VERDICT` block, and the watcher takes the retry-then-`mark_unreviewed` branch (`:447-476`) — loud, no bogus verdict, no silent drop. This is the GH-672 "START with no EXIT is never ambiguous" property finally holding for review lanes | write end → watcher poll → labelled outcome. Reach verified by derivation + source read; still no automated test (FOLLOW-UP) |
| `.log`/`.done` **read** end in `lane-status.ps1:65-68` | existence-probed `$cleanBase` / `$base` | the `logBytes=` and `done=` columns | Correct for both lane shapes; the `else` default is `$cleanBase`, so the pre-launch window resolves correctly rather than silently to the `-lane` spelling | read end → stdout → operator |
| `liveProcs` **recomputed** (`lane-status.ps1:84-113`) | wrapper-or-brief seed set + descendant BFS over an `[int]`-keyed `ParentProcessId` index | operator by eye (`docs/guides/operator-runbook.md:41`); no script parses the line | No longer success-only for the reported incident. Residual: single-snapshot, so it cannot see an orphan whose every CommandLine-matching ancestor is already dead — P2-4, FOLLOW-UP | field → stdout → operator. No consumer asserts it; no test |
| `-lane.ps1` entries in the `LogDir` wrapper fallback (`lane-stop.ps1:82-83`) | `"$Name-lane.ps1"` / `"$lane-lane.ps1"` — verified PowerShell interpolates these as `<name>-lane.ps1` (the hyphen terminates the variable token) | `lane-stop.ps1:85-90` | Dead-but-harmless today: review lanes live in `$EDDA_FLEET_SCRATCH`, not `$LogDir`, and the action-arg path resolves first. Reached only if a `-lane.ps1` wrapper is ever placed in `$LogDir` | fallback → wrapper resolution. No consumer, no test — a defensive branch of an existing surface, not a new one, so §5.5's "no consumer → P1" rule does not fire |
| task-name candidate **order** (`lane-stop.ps1:50`, `lane-status.ps1:18`) | `@("edda-$Name", "edda-lane-$Name", $Name)`, first match wins | both scripts' own resolution | Improves Round 1 P2-2; bare `$Name` is now last-resort. Still unanchored — P2-2 below | flag → `Get-ScheduledTask`. No test |

### Findings

- **[P2] The `-lane` strip mis-derives `.done` for a worker lane whose name ends in `-lane`.** `lane-stop.ps1:100` / `lane-status.ps1:66` strip `-lane$` unconditionally, so `<LogDir>\build-lane.wrapper.ps1` → `$cleanBase = …\build`. Measured: the shipped two-step replace on `C:\tmp\edda-lanes\build-lane.wrapper.ps1` → `C:\tmp\edda-lanes\build`. `$Log` is rescued by the `elseif (Test-Path "$base.log")` arm once the wrapper has started writing, but `$Done` never exists before the stop, so both arms miss and the `else` default writes the sentinel to `build.done` while `lane-launch.ps1:105` polls `build-lane.done`. Not live today — the sanctioned lane names are `worker-1|worker-2|verifier|verifier-2` — but the guard is a name-shape assumption, not a lane-class test. Suggested fix: strip `-lane` only when the resolved wrapper is outside `$LogDir` (i.e. the review-lane shape), or match the review-lane name pattern explicitly.
- **[P2] Bare `$Name` is still a candidate and is still matched against every scheduled task in every folder.** `lane-stop.ps1:50` / `lane-status.ps1:18` now try `edda-$Name` and `edda-lane-$Name` first, which removes the realistic collision, but a task literally named `<lane>` with no `edda-*` sibling still resolves and `lane-stop.ps1:150` then issues `Stop-ScheduledTask` against it. Damage stays bounded by the wrapper-resolution requirement at `:93-95`. The guarantee the diff deleted in Round 1 ("Never points at another lane") is still unreplaced. Suggested fix: assert the resolved `$TaskName` matches `^edda-(lane|review)-`.
- **[P2] The `--prompt-file` extractor now requires quotes on both sides, so an unquoted argument silently yields `$brief = $null` and loses the orphan seed.** `lane-stop.ps1:108` / `lane-status.ps1:89`. Measured: the shipped regex against `& edda dispatch --prompt-file C:\a\brief.md --session-id x` → NO MATCH. Latent, not live: both generators quote (`review-pr.sh:392` double, `lane-launch.ps1:114` via `PsQuote` single). The character class also accepts a mismatched pair (opening `'`, closing `"`). Suggested class: alternation over `'…'`, `"…"`, `\S+` — the Round 1 suggestion, which keeps the unquoted arm.
- **[P2] `liveProcs` still cannot see an orphan whose every CommandLine-matching ancestor is already dead.** `lane-status.ps1:95-99` seeds only from **live** processes, and `$childrenOf` is built from the same single snapshot, so a `claude`/`node` grandchild that outlives its `edda dispatch` parent has no reachable seed: the grandchild's own command line carries the brief's *contents*, not its path (confirmed — the generated brief contains no self-reference to its own path), and its `ParentProcessId` points at a PID no longer in the snapshot. `lane-stop.ps1` does not have this hole because it enqueues the pre-stop `$seedPids` into the verify traversal (`:242-244`). This is inherent to a one-shot status command with no persisted PID list, it was equally true of the Round 1 code, and Round 1's prescribed fix shape was implemented as written — so it is routed as FOLLOW-UP rather than a late blocker (`loop` item 2). It does not defeat doneWhen item 4 for the incident GH-712 records, where the `edda dispatch` / `pi` child survives the `Stop-ScheduledTask` and is therefore a seed.
- **[P2] Round 1 P2-4 is unaddressed: `$task` is not coerced to a single object in `lane-stop.ps1:52`.** `lane-status.ps1:21` wraps its result in `@(…)`; `lane-stop.ps1` does not, so with same-named tasks in two folders `$task.Actions[0]` and `$task.State -eq 'Running'` operate on a collection. Non-blocking, unchanged severity.

### FOLLOW-UP ISSUE

- **No test harness covers `scripts/fleet/lane-*.ps1`** (carried from Round 1, now with a concrete cheap target). `scripts/test-review-pr.sh:303,368` already asserts the exact wrapper path `review-pr<N>-r1-lane.ps1` that `review-pr.sh` generates, so the fixture is half-built: a case that feeds that path plus `lane-launch.ps1`'s `<Name>.wrapper.ps1` through the shipped derivation and asserts the resulting `.log`/`.done` pair would have caught Round 1's P1-1/P1-2 mismatch mechanically, and would pin P2-1 above. Out of GH-712's doneWhen.
- **`liveProcs` for the fully-orphaned grandchild** (P2-4). Needs either a persisted PID list written at launch or a second identifying token on the grandchild's command line. Deserves its own issue with the incident shape spelled out; do not bolt it onto this PR.
- **`lane-status.ps1` does not validate `-Name`** while `lane-stop.ps1:43` does. `Get-ScheduledTask -TaskName '*'` would enumerate every task. Read-only, so no destruction, and pre-existing — but the two scripts disagree on their own input contract.
- **`lane-status.ps1` fallback derivation for a dry-run lane** (carried from Round 1, unchanged): `<Name>.dryrun-wrapper.ps1` reduces to `<Name>.dryrun-wrapper`, not `<Name>`. Measured again this round. Harmless because `lane-launch.ps1:201` unregisters its own dry-run task.
- **Log/done naming divergence in `review-pr.sh` itself** (carried from Round 1): naming the wrapper `review-pr<N>-r<R>.ps1` would remove the `-lane` special case from two consumers instead of teaching it to both. Outside this PR's surface.

### RAN vs READ

RAN (this worktree, Windows 11, Git Bash + pwsh 7):
- `gh pr view 759 …` → head `c4b1056bf4d44e9571acbe486f9517f30e4759d6`, base `main`, OPEN, not draft; Round 1 verdict pinned to `4ba71c3` → delta target confirmed
- `git diff --stat 4ba71c3..c4b1056` → 2 files, +62/−15; `gh pr diff 759 --name-only` → the same 2 paths
- `sh scripts/review-pr.sh 759 99 --dry-run` → `classes=code-risk`, `canonical_class=code-risk`, `spec=REVIEW.md@662ade2 (base of main) (review-spec-v1.2)`; it generated a **current-shape** review lane wrapper without starting anything (scratch artefacts removed afterwards)
- PowerShell probes of the shipped expressions against that real wrapper: derivation → `review-pr759-r99.log` / `.done`; `--prompt-file` extractor → the real brief path, len 52; `-File` extractor → all three argument shapes incl. a space-bearing quoted path; `$Matches` key set on the unquoted arm → `4,1,0`
- PowerShell probes of the worker-lane shape (`worker-1.wrapper.ps1` → `worker-1.log`/`.done`) and the two edge shapes (`build-lane.wrapper.ps1`, `worker-1.dryrun-wrapper.ps1`)
- PowerShell AST `Parser::ParseFile` on both changed files → `parseErrors=0` each
- U2/U3/U4 greps, `sh scripts/lint-markdown-content.sh` → `exit=0`, `gh pr checks 759`, R1/R2/C4/C5 greps on the delta — all results in the Rules table
- `sh scripts/wiring-scan.sh 4ba71c3 c4b1056` → no new pub surfaces, no swallow patterns
- Source reads for consumer confirmation: `scripts/review-pr.sh:276,346-348,373-395,505-520`, `scripts/pr-review-watch.sh:343-344,405-421,447-476`, `scripts/fleet/lane-launch.ps1:104-106,114,156,228`, `scripts/test-review-pr.sh:303,368`
- `git status --porcelain` → clean; nothing edited, committed or pushed in this round

READ:
- Exact-head CI (run 33717488421): Format, Clippy ×3, Test (ubuntu/macos), Detect code changes **pass**; `Test (windows-latest)` pending. Covers the Rust workspace, which this delta does not touch; covers **nothing** in `scripts/fleet/*.ps1` — no CI job lints, parses or executes these scripts, which is why the AST parse and the derivation probes above were RAN locally rather than READ
- No L1 gate receipt is claimed and none is required for this SHA: no Rust, no `Cargo.toml`/`Cargo.lock`, no toolchain change (`ladder`)
- `Review Response: Round 1 (Commit c4b1056)` — read as the implementer's answer to each blocking finding. Its behavioural transcript (authentic review-lane naming, `liveProcs 4→0`, child PID dead, `.done`/`.log` written at the un-suffixed paths) is consistent with every derivation independently re-measured above
- GH-712 doneWhen (4 items) as the acceptance ceiling; the Round 1 verdict as the delta baseline

### Verdict

LGTM (P0=0, P1=0) — all three Round 1 P1s are fixed and independently re-measured against the real `review-pr.sh` lane shape rather than a synthetic one; the remaining items are P2 or evidenced FOLLOW-UPs. Merge remains the operator's action (GATE-01).

Issue #797, opened 2026-09-03
fix(fleet): the GH-715 guard covers .git/config only — the same interrupted write NUL-ed a branch ref today and every pre-flight gate passed


fagemx:
GH-715 shipped a guard for one file. The interruption class it defends against destroys others, and today it destroyed one for real.

## What happened

A Claude Code session on this workstation was interrupted mid-git-operation. `refs/heads/feat/gh730-status-json` came back as **41 bytes, all 41 NUL** — the same corruption shape as the two `.git/config` incidents that produced GH-715 (27328/27328 and 2561/2561), just a different file. The worktree became unusable:

```
$ git -C C:/ai_agent/edda-wt-gh730 log --oneline -3
fatal: your current branch appears to be broken

$ git show-ref | grep gh730
fatal: git show-ref: bad ref refs/heads/feat/gh730-status-json (0000000000000000000000000000000000000000)

$ python -c "d=open('.git/refs/heads/feat/gh730-status-json','rb').read(); print(len(d), d.count(b'\0'))"
41 41
```

Recovered by hand: the objects were intact, and the worktree's own reflog (`.git/worktrees/<wt>/logs/HEAD`) still held the SHA, so `rm` the dead ref file + `git update-ref` restored it. No work was lost. The corrupt ref is preserved for forensics.

Note `git update-ref` alone is not enough — it refuses to lock a broken ref (`cannot lock ref …: reference broken`), so the dead file has to be removed first. Worth having in whatever recovery this produces.

## The gap, measured

Everything below was measured in a throwaway repo whose current branch ref was overwritten with NULs.

| Check | Result | So |
|---|---|---|
| `scripts/fleet/git-config-guard.ps1 -Verify` | `config=… healthy (parses, no NUL bytes)`, **exit 0** | the GH-715 guard calls a repo with a destroyed branch ref healthy |
| `git rev-parse --is-inside-work-tree` | `true`, **exit 0** | `lane-launch.ps1:93`'s pre-flight gate **passes**, so a lane is dispatched into a broken worktree |
| `git status` | **exit 0** | the obvious health check misses it entirely |
| `git log` | **exit 128**, `fatal: your current branch appears to be broken` | the failure only surfaces once real work starts |
| `git fsck` | **exit 10**, `error: refs/heads/main: invalid sha1 pointer 0000…`, `error: invalid HEAD` | **a detector already exists** |
| `.git/logs/refs/heads/<branch>` | still holds the SHA | **a recovery source already exists** |

[...]

Issue #786, opened 2026-09-03
fix(fleet): test-git-config-guard.sh leaks a live scheduled task and its process tree when an E2E case fails


fagemx:
Routed to follow-up from the PR #773 review (raised round 3, carried through round 5, merged as `94f227f` with this open). Not a regression in shipped behaviour — it is a defect in the new test suite's teardown.

## Defect

`scripts/test-git-config-guard.sh` starts real Windows scheduled tasks for its opt-in end-to-end cases (`GIT_CONFIG_GUARD_E2E=1`). It drops each lane only on the success path, so any `fail` between starting a lane and dropping it leaks:

```sh
# scripts/test-git-config-guard.sh:353-355
pwsh -NoProfile -NonInteractive -File "$tmp/throw.ps1" >"$tmp/out14" 2>&1 ||
  fail "locked-config lane-stop probe did not run: $(cat "$tmp/out14")"
drop_fake_lane "$lane4"
```

`start_fake_lane` has the same hole — it calls `fail` itself, *after* registering and starting the task.

The leak is not just an untidy task. The lane's `cmd.exe /c ping` child keeps the suite's temp dir open, so the `trap 'rm -rf "$tmp"'` cannot succeed either.

## Reproduction (measured on `bfb15b5`, the merged head)

Inject one forced failure immediately after a lane starts, then run with `GIT_CONFIG_GUARD_E2E=1`:

```
rm: cannot remove '/tmp/tmp.b4QkC3UqtP/e2e': Device or resource busy
leftover tasks = 1 : edda-lane-guard-e2e-1047
leftover procs = 1
stale suite temp dirs = 1
```

The reviewer independently hit this during round 3 and cleaned up by hand.

## Why reordering the two lines is not the fix

`start_fake_lane`'s own `fail` fires after the task exists, and every future `fail` added between a start and its drop reopens the hole. Teardown has to be driven from the trap.

## Suggested fix (written and verified before the branch changed hands)

```sh
tmp=$(mktemp -d)

# Every fake lane this run starts is torn down from the trap, not only from
# the success path: a `fail` between starting a lane and dropping it would
# otherwise leave a live scheduled task and its process tree behind, and that
# tree holds the temp dir open so even `rm -rf` fails.
started_lanes=''

e2e_cleanup() {
  [ -n "$started_lanes" ] || return 0
  for cl_lane in $started_lanes; do
[...]

PR #804, opened 2026-09-03
feat(cli): include build identity in edda --version (GH-746)

fagemx:
Implement GH-746: include git commit and build date in `edda --version` output, with dirty flag when tracked files are changed and fallback to `unknown` when git metadata is unavailable. Adds a build script, CLI version binding, documentation, reviewer checklist note, and regression coverage.

Issue: #746

PR #803, opened 2026-09-03
fix(fleet): tear down fake lanes from trap in test-git-config-guard.sh (GH-786)

fagemx:
Issue: #786

## Summary

In \scripts/test-git-config-guard.sh\, fake lanes were only dropped on the success path via explicit \drop_fake_lane\. Any failure occurring after registering a scheduled task leaked both the live scheduled task in Windows Task Scheduler and its background process tree (e.g. \cmd.exe /c ping\). The leaked process held the suite's temporary directory open, causing \	rap 'rm -rf \"$tmp\"' ...\ to fail with \Device or resource busy\.

This change:
1. Tracks all active lanes in \started_lanes\.
2. Registers lanes in \started_lanes\ inside \start_fake_lane\ *before* starting the scheduled task.
3. Implements \2e_cleanup\ in the exit/signal trap to unregister all remaining scheduled tasks and kill their process trees before directory removal.
4. Uses a standalone \.ps1\ file for process matching to prevent self-matching and exit-255 traps.
5. Keeps \drop_fake_lane\ removing dropped lanes from \started_lanes\.

### Pre-fix Fail-First (RED)
Forced failure after starting a lane resulted in:
\\\	ext
FAIL (after case 10): forced failure for fail-first
rm: cannot remove '/tmp/tmp.1DqHD7L5uk/e2e': Device or resource busy
\\\
Leftover scheduled task:
\\\	ext
TaskName: edda-lane-guard-e2e-934    State: Running
\\\

### Post-fix Verification (GREEN)
- Forced failure test: \2e_cleanup\ successfully reaped the task and process tree; \
m -rf\ succeeded without errors; 0 scheduled tasks leaked; 0 processes leaked.
- Offline tests: 10/10 passed.
- Full E2E suite: \GIT_CONFIG_GUARD_E2E=1 scripts/test-git-config-guard.sh\ → 15/15 passed.
- Pre-commit & file-length ratchets: passed.

PR #805, opened 2026-09-03
fix(fleet): extend git-config-guard to detect and recover corrupt refs (GH-797)

fagemx:
## Summary

Issue: #797

Extends `git-config-guard.ps1` and the fleet lane lifecycle (`lane-launch.ps1`, `lane-stop.ps1`) to detect and recover from corrupted git branch refs caused by interrupted writes during lane hard-kills.

### Details
- **`scripts/fleet/git-config-guard.ps1`**:
  - Adds `-VerifyRefs` and `-RestoreRefs` switches.
  - Detects broken refs via `git fsck --connectivity-only` and direct inspection for NUL bytes in loose refs.
  - Automatically recovers broken refs from reflog sources (common dir reflog, worktree HEAD reflog, common HEAD reflog), validating commit objects with `git cat-file -e`.
  - Consumes `CanRepair` across `-Backup`, `-Verify`, `-VerifyRefs`, `-RestoreRefs`, and `-Restore` to distinguish repairable issues from unrepairable ones and provide accurate hints.
  - Preserves dead ref files outside `.git/refs/` (`.git/ref.CORRUPT.*.bak`) so `git fsck` is not tripped by the backup, and restores using `git update-ref`. Emits all partially restored refs even if subsequent operations fail.
  - In `-Backup`, rejects dispatch (exit 2) if branch refs are corrupt.
  - In `-VerifyOrRestore` and `-Restore`, recovers both `.git/config` and broken branch refs.
- **`scripts/fleet/lane-launch.ps1`**:
  - Verifies git metadata via `git-config-guard.ps1 -Backup` and blocks launch if metadata is corrupt.
- **`scripts/fleet/lane-stop.ps1`**:
  - Evaluates and restores both `.git/config` and refs in step 6 post-kill verification.
  - Reports ground truth for both `gitconfig=` and `gitrefs=` (preserving healthy verdict if only refs failed, and reporting UNVERIFIED if config check failed before refs could be checked).
  - Records both verdicts in the durable `=== EXIT ===` record and final report.
- **`scripts/test-git-config-guard.sh`**:
  - Adds offline test cases 11-15 exercising ref corruption verification, backup refusal, reflog restore, unrecoverable missing reflog error, and linked worktree ref repair.
  - Adds safe trap cleanup (GH-786) and robust scheduled task startup polling.
  - Adds explicit assertions for `gitrefs=` across E2E test cases 16, 17, 18, 19.

## Validation (doneWhen 4 Receipt)
- Measured on a 27-worktree workstation repository (`edda`):
  * `git fsck --connectivity-only`: 2016 ms (cold), 1482 ms / 1400 ms (warm, avg ~1.44 s)
  * `git fsck` (full): 2381 ms (~65% slower)
  * `git-config-guard.ps1 -Verify` (full invocation): 2323 ms / 2379 ms / 2349 ms (~2.35 s)
- Ran offline tests: all 15 cases passed.
- Ran end-to-end tests (`GIT_CONFIG_GUARD_E2E=1`): all 20 cases passed.
- `git diff --check` and `lint-markdown-content.sh` clean.

### DIFF
--- a/scripts/fleet/git-config-guard.ps1
+++ b/scripts/fleet/git-config-guard.ps1
@@ -367,7 +367,12 @@ function Restore-BrokenRefs([string]$Repo, [string]$CommonDir, $BrokenRefs) {
 if ($VerifyRefs) {
   $refsHealth = Get-RefsHealth $RepoPath $commonDir
   if (-not $refsHealth.Healthy) {
-    Fail "refs UNHEALTHY: $($refsHealth.Reason)" 2
+    $repairHint = if ($refsHealth.CanRepair) {
+      "repair available from reflog: run scripts/fleet/git-config-guard.ps1 -RepoPath '$RepoPath' -RestoreRefs"
+    } else {
+      "automatic reflog repair unavailable; inspect repo manually"
+    }
+    Fail "refs UNHEALTHY: $($refsHealth.Reason) ($repairHint)" 2
   }
   "refs=healthy ($($refsHealth.Reason))"
   exit 0
@@ -382,12 +387,15 @@ if ($RestoreRefs) {
     exit 0
   }
   [Console]::Error.WriteLine("git-config-guard: refs UNHEALTHY: $($refsHealth.Reason)")
-  $restoreResult = Restore-BrokenRefs $RepoPath $commonDir $refsHealth.BrokenRefs
-  if (-not $restoreResult.Success) {
-    Fail "RESTORE FAILED: $($restoreResult.Reason)" 2
+  if (-not $refsHealth.CanRepair) {
+    Fail "cannot repair refs automatically: $($refsHealth.Reason)" 2
   }
+  $restoreResult = Restore-BrokenRefs $RepoPath $commonDir $refsHealth.BrokenRefs
   foreach ($r in $restoreResult.Restored) {
     "$r"
+  }
+  if (-not $restoreResult.Success) {
+    Fail "RESTORE FAILED: $($restoreResult.Reason)" 2
   }
   exit 0
 }
@@ -405,7 +413,12 @@ if ($Verify) {
   }
   $refsHealth = Get-RefsHealth $RepoPath $commonDir
   if (-not $refsHealth.Healthy) {
-    Fail "refs UNHEALTHY: $($refsHealth.Reason)" 2
+    $repairHint = if ($refsHealth.CanRepair) {
+      "repair available from reflog: run scripts/fleet/git-config-guard.ps1 -RepoPath '$RepoPath' -RestoreRefs"
+    } else {
+      "automatic reflog repair unavailable; inspect repo manually"
+    }
+    Fail "refs UNHEALTHY: $($refsHealth.Reason) ($repairHint)" 2
   }
   "config=$ConfigPath healthy ($($configHealth.Reason))"
   "refs=healthy ($($refsHealth.Reason))"
@@ -423,7 +436,12 @@ if ($Backup) {
   }
   $refsHealth = Get-RefsHealth $RepoPath $commonDir
   if (-not $refsHealth.Healthy) {
-    Fail "ref(s) unhealthy ($($refsHealth.Reason)); cannot guarantee repo consistency — repair with scripts/fleet/git-config-guard.ps1 -RepoPath '$RepoPath' -RestoreRefs" 2
+    $repairHint = if ($refsHealth.CanRepair) {
+      "repair with scripts/fleet/git-config-guard.ps1 -RepoPath '$RepoPath' -RestoreRefs"
+    } else {
+      "automatic repair from reflog unavailable; manual inspection of .git required"
+    }
+    Fail "ref(s) unhealthy ($($refsHealth.Reason)); cannot guarantee repo consistency — $repairHint" 2
   }
   try {
     if ((Get-ConfigHealth $BackupPath).Healthy) { Copy-Atomic $BackupPath $PrevBackupPath }
@@ -474,12 +492,15 @@ if (-not $configHealth.Healthy) {
 $refsHealth = Get-RefsHealth $RepoPath $commonDir
 if (-not $refsHealth.Healthy) {
   [Console]::Error.WriteLine("git-config-guard: refs UNHEALTHY: $($refsHealth.Reason)")
-  $restoreResult = Restore-BrokenRefs $RepoPath $commonDir $refsHealth.BrokenRefs
-  if (-not $restoreResult.Success) {
-    Fail "RESTORE FAILED: $($restoreResult.Reason)" 2
+  if (-not $refsHealth.CanRepair) {
+    Fail "cannot repair refs automatically: $($refsHealth.Reason)" 2
   }
+  $restoreResult = Restore-BrokenRefs $RepoPath $commonDir $refsHealth.BrokenRefs
   foreach ($r in $restoreResult.Restored) {
     "$r"
+  }
+  if (-not $restoreResult.Success) {
+    Fail "RESTORE FAILED: $($restoreResult.Reason)" 2
   }
 } else {
   "refs healthy ($($refsHealth.Reason)); nothing restored"
--- a/scripts/fleet/git-config-guard.ps1
+++ b/scripts/fleet/git-config-guard.ps1
@@ -1,45 +1,41 @@
-# git-config-guard.ps1 — keep the shared .git/config recoverable (GH-715).
+# git-config-guard.ps1 — keep the shared .git/config and refs recoverable (GH-715, GH-797).
 #
 # The main checkout and all 30+ linked worktrees read ONE .git/config to find
-# the remote. On 2026-09-02 17:16 and 2026-09-03 01:33 that file became a run
-# of NUL bytes (27328/27328 and 2561/2561) after lanes were hard-killed while
-# git was extending it — `fatal: bad config line 1` for every worktree at once.
-# There was no restore point: the .bak sitting next to it had been copied
-# AFTER the corruption, so it was all NULs too.
+# the remote, and shared git metadata under .git/refs to resolve branches.
+# On 2026-09-02 and 2026-09-03 hard-killing lanes mid-write resulted in NUL-corrupted
+# metadata files:
+#   * .git/config became runs of NUL bytes (27328/27328 and 2561/2561), breaking
+#     every worktree at once (GH-715).
+#   * refs/heads/<branch> became 41 NUL bytes, breaking branch checkout and log
+#     while rev-parse --is-inside-work-tree silently passed (GH-797).
 #
-# The fix that makes a backup worth having is validating BEFORE copying. This
-# script never writes a backup it has not parsed, and never reports a repair it
-# did not make.
+# The defense:
+#   * .git/config has no recovery source, so it requires a validated backup taken
+#     BEFORE the lane can write, never backing up an unparsed file.
+#   * Branch refs DO have a recovery source: git's own write-ahead reflog.
+#     The detector is `git fsck --connectivity-only` (and loose ref inspection).
+#     The repair is preserving the dead ref for forensics, removing it, and
+#     restoring the SHA from the reflog via `git update-ref`.
 #
 # usage:
 #   pwsh -NoProfile -File scripts/fleet/git-config-guard.ps1 -RepoPath <repo> -Backup
 #   pwsh -NoProfile -File scripts/fleet/git-config-guard.ps1 -RepoPath <repo> -Verify
 #   pwsh -NoProfile -File scripts/fleet/git-config-guard.ps1 -RepoPath <repo> -Restore
 #   pwsh -NoProfile -File scripts/fleet/git-config-guard.ps1 -RepoPath <repo> -VerifyOrRestore
+#   pwsh -NoProfile -File scripts/fleet/git-config-guard.ps1 -RepoPath <repo> -VerifyRefs
+#   pwsh -NoProfile -File scripts/fleet/git-config-guard.ps1 -RepoPath <repo> -RestoreRefs
 #
 # -RepoPath may be the main checkout or any linked worktree: the target is
-# always the SHARED config (`git rev-parse --git-common-dir`), because that is
-# the single file whose loss takes down every worktree.
-#
-# -Backup keeps one generation: the outgoing backup moves to <backup>.prev
-# before the new one is written. Be precise about what that buys. The
-# AUTOMATIC fallback only fires when the newest backup is DETECTABLY broken
-# (case 10). It cannot fire for the case that motivates keeping a generation
-# at all — a torn write ending on a boundary that still parses — because
-# nothing can tell that file from a good one; there, .prev is the copy an
-# operator points -BackupPath at by hand once they know the newest is wrong.
-#
-# A config that cannot be READ (held open by whatever is writing it) is a third
-# outcome, not a synonym for corrupt: nothing is backed up from it and nothing
-# is restored over it, because it may be perfectly good.
+# always the SHARED config and refs (`git rev-parse --git-common-dir`), because that
+# is the single directory whose loss takes down every worktree.
 #
 # Exit codes:
-#   0  the config is healthy (backed up / verified / restored)
+#   0  metadata is healthy (backed up / verified / restored)
 #   1  usage or resolution error (no mode, not a git repo, git unavailable)
-#   2  the config is unhealthy — and for -Restore/-VerifyOrRestore, could not
-#      be repaired (no usable backup, or the restored file still does not parse)
+#   2  metadata is unhealthy — and for -Restore/-VerifyOrRestore, could not
+#      be repaired (no usable backup/reflog, or the restored state still does not parse)
 #   3  no backup was taken: the config is healthy but the copy failed
-#   4  the config could not be judged at all, because it could not be read.
+#   4  config could not be judged at all, because it could not be read.
 #      Distinct from 2 on purpose: nothing is known to be wrong with it
 param(
   [string]$RepoPath = '.',
@@ -47,6 +43,8 @@ param(
   [switch]$Verify,
   [switch]$Restore,
   [switch]$VerifyOrRestore,
+  [switch]$VerifyRefs,
+  [switch]$RestoreRefs,
   [string]$BackupPath = ''
 )
 
@@ -62,28 +60,30 @@ function Fail([string]$Msg, [int]$Code = 1) {
   exit $Code
 }
 
-$modes = @($Backup, $Verify, $Restore, $VerifyOrRestore) | Where-Object { $_ }
+$modes = @($Backup, $Verify, $Restore, $VerifyOrRestore, $VerifyRefs, $RestoreRefs) | Where-Object { $_ }
 if ($modes.Count -ne 1) {
-  Fail 'give exactly one of -Backup, -Verify, -Restore, -VerifyOrRestore'
+  Fail 'give exactly one of -Backup, -Verify, -Restore, -VerifyOrRestore, -VerifyRefs, -RestoreRefs'
 }
 
-# --- resolve the shared config ----------------------------------------------
+# --- resolve the shared config & worktree gitdir ----------------------------
 
 if (-not (Test-Path -LiteralPath $RepoPath)) { Fail "-RepoPath '$RepoPath' does not exist" }
 $RepoPath = (Resolve-Path -LiteralPath $RepoPath).Path
 
-# Ask git first — it is authoritative and honours GIT_DIR and friends. But the
-# case this tool exists for is exactly the one where git refuses to answer:
-# once .git/config is a run of NULs, EVERY git command in the repo fails,
-# rev-parse included. So fall back to walking the same links git would.
+$script:WorktreeGitDir = $null
+
 function Resolve-CommonDir([string]$Start) {
   # Collect the whole stream before reading $LASTEXITCODE: `Select-Object
   # -First` stops the pipeline early and leaves the exit code unset.
-  $out = @(& git -C $Start rev-parse --path-format=absolute --git-common-dir 2>$null)
-  if ($LASTEXITCODE -eq 0 -and $out.Count -gt 0) {
-    $fromGit = "$($out[0])".Trim()
-    if ($fromGit -and (Test-Path -LiteralPath $fromGit -PathType Container)) {
-      return (Resolve-Path -LiteralPath $fromGit).Path
+  $out = @(& git -C $Start rev-parse --path-format=absolute --git-common-dir --git-dir 2>$null)
+  if ($LASTEXITCODE -eq 0 -and $out.Count -ge 2) {
+    $fromGitCommon = "$($out[0])".Trim()
+    $fromGitDir = "$($out[1])".Trim()
+    if ($fromGitDir -and (Test-Path -LiteralPath $fromGitDir -PathType Container)) {
+      $script:WorktreeGitDir = (Resolve-Path -LiteralPath $fromGitDir).Path
+    }
+    if ($fromGitCommon -and (Test-Path -LiteralPath $fromGitCommon -PathType Container)) {
+      return (Resolve-Path -LiteralPath $fromGitCommon).Path
     }
   }
 
@@ -109,6 +109,7 @@ function Resolve-CommonDir([string]$Start) {
   }
   if (-not $gitDir -or -not (Test-Path -LiteralPath $gitDir -PathType Container)) { return $null }
   $gitDir = (Resolve-Path -LiteralPath $gitDir).Path
+  $script:WorktreeGitDir = $gitDir
 
   $commonFile = Join-Path $gitDir 'commondir'
   if (Test-Path -LiteralPath $commonFile -PathType Leaf) {
@@ -127,16 +128,10 @@ if (-not $commonDir) {
 }
 $ConfigPath = Join-Path $commonDir 'config'
 if (-not $BackupPath) { $BackupPath = Join-Path $commonDir 'config.guard.bak' }
+$PrevBackupPath = "$BackupPath.prev"
 
-# --- health ------------------------------------------------------------------
+# --- config health -----------------------------------------------------------
 
-# Healthy means all three: the file has content, holds no NUL byte, and git
-# itself can parse it. The NUL check is what names the failure seen twice on
-# 2026-09-03; `git config --list` is the criterion GH-715 asks a backup to meet.
-#
-# "Unreadable" is a THIRD outcome, not a synonym for unhealthy: a config held
-# open by a process that is still writing it cannot be judged, and must not be
-# overwritten from a backup on the strength of a failed read.
 function Get-ConfigHealth([string]$Path) {
   if (-not (Test-Path -LiteralPath $Path -PathType Leaf)) {
     return @{ Healthy = $false; Reason = 'file does not exist' }
@@ -161,10 +156,7 @@ function Get-ConfigHealth([string]$Path) {
   return @{ Healthy = $true; Reason = 'parses, no NUL bytes' }
 }
 
-# Copy through a sibling temp file so a reader never observes a half-written
-# destination — the same class of hazard that produced the NUL config. The temp
-# file is never left behind: a half-written .guard-tmp is the very thing this
-# script exists to keep out of the git directory.
+# Copy through a sibling temp file so a reader never observes a half-written destination.
 function Copy-Atomic([string]$From, [string]$To) {
   $tmp = "$To.guard-tmp"
   try {
@@ -177,8 +169,6 @@ function Copy-Atomic([string]$From, [string]$To) {
   }
 }
 
-# A name no existing file holds, so a second restore inside the same second
-# cannot overwrite the first one's evidence.
 function New-UniquePath([string]$Path) {
   if (-not (Test-Path -LiteralPath $Path)) { return $Path }
   $n = 2
@@ -186,39 +176,254 @@ function New-UniquePath([string]$Path) {
   return "$Path.$n"
 }
 
-$health = Get-ConfigHealth $ConfigPath
+# --- refs health & reflog recovery (GH-797) ----------------------------------
+
+function Resolve-RefRepairSha([string]$Repo, [string]$CommonDir, [string]$Ref) {
+  $normalizedRef = $Ref -replace '\\', '/'
+  $candidates = New-Object System.Collections.Generic.List[string]
+
+  # 1. Primary reflog: <commonDir>/logs/<ref>
+  $primaryReflog = Join-Path $CommonDir "logs\$($normalizedRef -replace '/', '\')"
+  if (Test-Path -LiteralPath $primaryReflog -PathType Leaf) {
+    $candidates.Add($primaryReflog)
+  }
+
+  # 2. Worktree HEAD reflog (if this worktree's HEAD points to this ref)
+  if ($script:WorktreeGitDir) {
+    $wtHeadPath = Join-Path $script:WorktreeGitDir 'HEAD'
+    if (Test-Path -LiteralPath $wtHeadPath -PathType Leaf) {
+      $wtHeadContent = "$(Get-Content -LiteralPath $wtHeadPath -TotalCount 1)".Trim()
+      if ($wtHeadContent -match '^\s*ref:\s*(.+?)\s*$' -and ($Matches[1] -replace '\\', '/') -eq $normalizedRef) {
+        $wtHeadLog = Join-Path $script:WorktreeGitDir 'logs\HEAD'
+        if (Test-Path -LiteralPath $wtHeadLog -PathType Leaf) {
+          $candidates.Add($wtHeadLog)
+        }
+      }
+    }
+  }
+
+  # 3. Common HEAD reflog (if common HEAD points to this ref)
+  $commonHeadPath = Join-Path $CommonDir 'HEAD'
+  if (Test-Path -LiteralPath $commonHeadPath -PathType Leaf) {
+    $commonHeadContent = "$(Get-Content -LiteralPath $commonHeadPath -TotalCount 1)".Trim()
+    if ($commonHeadContent -match '^\s*ref:\s*(.+?)\s*$' -and ($Matches[1] -replace '\\', '/') -eq $normalizedRef) {
+      $commonHeadLog = Join-Path $CommonDir 'logs\HEAD'
+      if (Test-Path -LiteralPath $commonHeadLog -PathType Leaf) {
+        $candidates.Add($commonHeadLog)
+      }
+    }
+  }
+
+  foreach ($logPath in $candidates) {
+    $lines = @(Get-Content -LiteralPath $logPath -ErrorAction SilentlyContinue)
+    for ($i = $lines.Count - 1; $i -ge 0; $i--) {
+      $line = $lines[$i]
+      $parts = $line -split '\s+'
+      if ($parts.Count -ge 2) {
+        $candidateSha = $parts[1]
+        if ($candidateSha -match '^[0-9a-fA-F]{40}$' -and $candidateSha -ne '0000000000000000000000000000000000000000') {
+          & git -C $Repo cat-file -e $candidateSha 2>&1 | Out-Null
+          if ($LASTEXITCODE -eq 0) {
+            return @{ Sha = $candidateSha; Source = $logPath }
+          }
+        }
+      }
+    }
+  }
+
+  return @{ Sha = $null; Source = $null }
+}
+
+function Get-RefsHealth([string]$Repo, [string]$CommonDir) {
+  $fsckRaw = @(& git -C $Repo fsck --connectivity-only 2>&1)
+  $fsckExit = $LASTEXITCODE
+  if ($fsckExit -eq 0) {
+    return @{ Healthy = $true; Reason = 'connectivity ok' }
+  }
+
+  # Parse fsck output for broken refs
+  $brokenMap = @{}
+  foreach ($line in $fsckRaw) {
+    if ($line -match 'error:\s+(refs/[^\s:]+):\s+(.*)') {
+      $r = ($Matches[1].Trim() -replace '\\', '/')
+      $msg = $Matches[2].Trim()
+      if (-not $brokenMap.ContainsKey($r)) {
+        $brokenMap[$r] = $msg
+      }
+    }
+  }
+
+  # Also inspect loose ref files under $CommonDir/refs/heads for NULs or bad length
+  $headsDir = Join-Path $CommonDir 'refs\heads'
+  if (Test-Path -LiteralPath $headsDir -PathType Container) {
+    $refFiles = Get-ChildItem -LiteralPath $headsDir -Recurse -File -ErrorAction SilentlyContinue
+    foreach ($rf in $refFiles) {
+      $relPath = ($rf.FullName.Substring($CommonDir.Length).TrimStart('\', '/') -replace '\\', '/')
+      if (-not $brokenMap.ContainsKey($relPath)) {
+        try {
+          $bytes = [System.IO.File]::ReadAllBytes($rf.FullName)
+          $hasNul = $false
+          foreach ($b in $bytes) { if ($b -eq 0) { $hasNul = $true; break } }
+          if ($hasNul) {
+            $brokenMap[$relPath] = "$($bytes.Length) bytes, contains NUL bytes (GH-797)"
+          }
+        } catch {}
+      }
+    }
+  }
+
+  if ($brokenMap.Count -eq 0) {
+    $firstLine = if ($fsckRaw.Count -gt 0) { "$($fsckRaw[0])" } else { "git fsck exit $fsckExit" }
+    return @{
+      Healthy = $false
+      BrokenRefs = @()
+      Reason = "git fsck failed: $firstLine"
+      CanRepair = $false
+    }
+  }
+
+  $brokenList = New-Object System.Collections.Generic.List[hashtable]
+  $allHaveRepairs = $true
+  $reasons = @()
+  foreach ($r in $brokenMap.Keys) {
+    $repairInfo = Resolve-RefRepairSha $Repo $CommonDir $r
+    $item = @{
+      Ref = $r
+      Error = $brokenMap[$r]
+      RepairSha = $repairInfo.Sha
+      RepairSource = $repairInfo.Source
+    }
+    $brokenList.Add($item)
+    if ($repairInfo.Sha) {
+      $reasons += "$r is broken ($($brokenMap[$r])); repair available from $($repairInfo.Source): $($repairInfo.Sha)"
+    } else {
+      $allHaveRepairs = $false
+      $reasons += "$r is broken ($($brokenMap[$r])); no repair SHA found in reflog"
+    }
+  }
+
+  return @{
+    Healthy = $false
+    BrokenRefs = @($brokenList)
+    Reason = ($reasons -join '; ')
+    CanRepair = $allHaveRepairs
+  }
+}
+
+function Restore-BrokenRefs([string]$Repo, [string]$CommonDir, $BrokenRefs) {
+  $restored = @()
+  foreach ($b in $BrokenRefs) {
+    if (-not $b.RepairSha) {
+      return @{
+        Success = $false
+        Restored = $restored
+        Reason = "no repair SHA available for ref $($b.Ref)"
+      }
+    }
+    $refName = $b.Ref
+    $sha = $b.RepairSha
+
+    # Preserve corrupt file outside .git/refs to avoid tripping git fsck
+    $targetFile = Join-Path $CommonDir ($refName -replace '/', '\')
+    $sanitized = $refName -replace '[/\\:]', '_'
+    $stamp = Get-Date -Format 'yyyy-MM-ddTHH-mm-ss'
+    $preserved = Join-Path $CommonDir "ref.CORRUPT.$sanitized.$stamp.bak"
+    $preserved = New-UniquePath $preserved
+
+    if (Test-Path -LiteralPath $targetFile) {
+      Copy-Item -LiteralPath $targetFile -Destination $preserved -Force
+      Remove-Item -LiteralPath $targetFile -Force
+    }
+
+    & git -C $Repo update-ref $refName $sha
+    if ($LASTEXITCODE -ne 0) {
+      return @{
+        Success = $false
+        Restored = $restored
+        Reason = "git update-ref $refName $sha failed with exit $LASTEXITCODE"
+      }
+    }
+    $restored += "RESTORED ref=$refName to sha=$sha from $($b.RepairSource); corrupt file preserved at $preserved"
+  }
+
+  $after = Get-RefsHealth $Repo $CommonDir
+  if (-not $after.Healthy) {
+    return @{
+      Success = $false
+      Restored = $restored
+      Reason = "refs still unhealthy after restoration: $($after.Reason)"
+    }
+  }
+
+  return @{
+    Success = $true
+    Restored = $restored
+    Reason = 'all refs restored and verified healthy'
+  }
+}
+
+# --- -VerifyRefs -------------------------------------------------------------
+
+if ($VerifyRefs) {
+  $refsHealth = Get-RefsHealth $RepoPath $commonDir
+  if (-not $refsHealth.Healthy) {
+    Fail "refs UNHEALTHY: $($refsHealth.Reason)" 2
+  }
+  "refs=healthy ($($refsHealth.Reason))"
+  exit 0
+}
+
+# --- -RestoreRefs ------------------------------------------------------------
+
+if ($RestoreRefs) {
+  $refsHealth = Get-RefsHealth $RepoPath $commonDir
+  if ($refsHealth.Healthy) {
+    "refs healthy ($($refsHealth.Reason)); nothing restored"
+    exit 0
+  }
+  [Console]::Error.WriteLine("git-config-guard: refs UNHEALTHY: $($refsHealth.Reason)")
+  $restoreResult = Restore-BrokenRefs $RepoPath $commonDir $refsHealth.BrokenRefs
+  if (-not $restoreResult.Success) {
+    Fail "RESTORE FAILED: $($restoreResult.Reason)" 2
+  }
+  foreach ($r in $restoreResult.Restored) {
+    "$r"
+  }
+  exit 0
+}
 
 # --- -Verify -----------------------------------------------------------------
 
+$configHealth = Get-ConfigHealth $ConfigPath
+
 if ($Verify) {
-  if ($health.Healthy) {
-    "config=$ConfigPath healthy ($($health.Reason))"
-    exit 0
+  if ($configHealth.Unreadable) {
+    Fail "config=$ConfigPath COULD NOT BE JUDGED: $($configHealth.Reason)" 4
   }
-  if ($health.Unreadable) {
-    Fail "config=$ConfigPath COULD NOT BE JUDGED: $($health.Reason)" 4
+  if (-not $configHealth.Healthy) {
+    Fail "config=$ConfigPath UNHEALTHY: $($configHealth.Reason)" 2
   }
-  Fail "config=$ConfigPath UNHEALTHY: $($health.Reason)" 2
+  $refsHealth = Get-RefsHealth $RepoPath $commonDir
+  if (-not $refsHealth.Healthy) {
+    Fail "refs UNHEALTHY: $($refsHealth.Reason)" 2
+  }
+  "config=$ConfigPath healthy ($($configHealth.Reason))"
+  "refs=healthy ($($refsHealth.Reason))"
+  exit 0
 }
 
 # --- -Backup -----------------------------------------------------------------
 
-# One generation back. The automatic fallback below only fires when the newest
-# backup is DETECTABLY broken; for a torn-but-parsing one nothing can tell it
-# from a good copy, and .prev is then the file an operator restores by hand
-# with -BackupPath. See the header.
-$PrevBackupPath = "$BackupPath.prev"
-
 if ($Backup) {
-  if ($health.Unreadable) {
-    # Nothing is known about the config, so nothing may be concluded from it:
-    # do not overwrite the backup, and do not claim the config is broken.
-    Fail "config could not be read, so no backup was taken ($($health.Reason)); $BackupPath left as it was" 4
+  if ($configHealth.Unreadable) {
+    Fail "config could not be read, so no backup was taken ($($configHealth.Reason)); $BackupPath left as it was" 4
   }
-  if (-not $health.Healthy) {
-    # Refusing here is the whole point: the real .bak files on this machine were
-    # copied after the corruption and were therefore useless.
-    Fail "refusing to back up an unhealthy config ($($health.Reason)); $BackupPath left as it was" 2
+  if (-not $configHealth.Healthy) {
+    Fail "refusing to back up an unhealthy config ($($configHealth.Reason)); $BackupPath left as it was" 2
+  }
+  $refsHealth = Get-RefsHealth $RepoPath $commonDir
+  if (-not $refsHealth.Healthy) {
+    Fail "ref(s) unhealthy ($($refsHealth.Reason)); cannot guarantee repo consistency — repair with scripts/fleet/git-config-guard.ps1 -RepoPath '$RepoPath' -RestoreRefs" 2
   }
   try {
     if ((Get-ConfigHealth $BackupPath).Healthy) { Copy-Atomic $BackupPath $PrevBackupPath }
@@ -227,46 +432,57 @@ if ($Backup) {
     Fail "config is healthy but the backup could not be written to ${BackupPath}: $($_.Exception.Message)" 3
   }
   "config=$ConfigPath healthy; backup=$BackupPath"
+  "refs=healthy ($($refsHealth.Reason))"
   exit 0
 }
 
 # --- -Restore / -VerifyOrRestore ---------------------------------------------
 
-if ($health.Healthy) {
-  "config=$ConfigPath healthy ($($health.Reason)); nothing restored"
-  exit 0
+$configRestored = $false
+if (-not $configHealth.Healthy) {
+  if ($configHealth.Unreadable) {
+    Fail "config=$ConfigPath could not be read, so it was NOT restored ($($configHealth.Reason)); check what holds it open" 4
+  }
+  [Console]::Error.WriteLine("git-config-guard: config=$ConfigPath UNHEALTHY: $($configHealth.Reason)")
+
+  $source = $null
+  foreach ($candidate in @($BackupPath, $PrevBackupPath)) {
+    $candidateHealth = Get-ConfigHealth $candidate
+    if ($candidateHealth.Healthy) { $source = $candidate; break }
+    [Console]::Error.WriteLine("git-config-guard: unusable backup ${candidate}: $($candidateHealth.Reason)")
+  }
+  if (-not $source) {
+    Fail "no usable backup to restore from (tried $BackupPath, $PrevBackupPath); config left untouched for forensics" 2
+  }
+
+  $stamp = Get-Date -Format 'yyyy-MM-ddTHH-mm-ss'
+  $preserved = New-UniquePath (Join-Path $commonDir "config.CORRUPT.$stamp.bak")
+  Copy-Item -LiteralPath $ConfigPath -Destination $preserved -Force
+  Copy-Atomic $source $ConfigPath
+
+  $after = Get-ConfigHealth $ConfigPath
+  if (-not $after.Healthy) {
+    Fail "RESTORE FAILED: config still unhealthy after copying $source ($($after.Reason))" 2
+  }
+  "RESTORED config=$ConfigPath from backup=$source; corrupt file preserved at $preserved"
+  $configRestored = $true
+} else {
+  "config=$ConfigPath healthy ($($configHealth.Reason)); nothing restored"
 }
 
-if ($health.Unreadable) {
-  # A locked config is not a corrupt one. Overwriting it here would destroy a
-  # file that may be perfectly good and is currently being written.
-  Fail "config=$ConfigPath could not be read, so it was NOT restored ($($health.Reason)); check what holds it open" 4
+# Check and restore refs
+$refsHealth = Get-RefsHealth $RepoPath $commonDir
+if (-not $refsHealth.Healthy) {
+  [Console]::Error.WriteLine("git-config-guard: refs UNHEALTHY: $($refsHealth.Reason)")
+  $restoreResult = Restore-BrokenRefs $RepoPath $commonDir $refsHealth.BrokenRefs
+  if (-not $restoreResult.Success) {
+    Fail "RESTORE FAILED: $($restoreResult.Reason)" 2
+  }
+  foreach ($r in $restoreResult.Restored) {
+    "$r"
+  }
+} else {
+  "refs healthy ($($refsHealth.Reason)); nothing restored"
 }
 
-[Console]::Error.WriteLine("git-config-guard: config=$ConfigPath UNHEALTHY: $($health.Reason)")
-
-$source = $null
-foreach ($candidate in @($BackupPath, $PrevBackupPath)) {
-  $candidateHealth = Get-ConfigHealth $candidate
-  if ($candidateHealth.Healthy) { $source = $candidate; break }
-  [Console]::Error.WriteLine("git-config-guard: unusable backup ${candidate}: $($candidateHealth.Reason)")
-}
-if (-not $source) {
-  Fail "no usable backup to restore from (tried $BackupPath, $PrevBackupPath); config left untouched for forensics" 2
-}
-
-# Keep the corrupt file: it is the only evidence of what the kill did, and the
-# name matches the ones already archived for the two 2026-09 incidents.
-$stamp = Get-Date -Format 'yyyy-MM-ddTHH-mm-ss'
-$preserved = New-UniquePath (Join-Path $commonDir "config.CORRUPT.$stamp.bak")
-Copy-Item -LiteralPath $ConfigPath -Destination $preserved -Force
-
-Copy-Atomic $source $ConfigPath
-
-$after = Get-ConfigHealth $ConfigPath
-if (-not $after.Healthy) {
-  Fail "RESTORE FAILED: config still unhealthy after copying $source ($($after.Reason))" 2
-}
-
-"RESTORED config=$ConfigPath from backup=$source; corrupt file preserved at $preserved"
 exit 0
--- a/scripts/fleet/git-config-guard.ps1
+++ b/scripts/fleet/git-config-guard.ps1
@@ -279,11 +279,13 @@ function Get-RefsHealth([string]$Repo, [string]$CommonDir) {
       BrokenRefs = @()
       Reason = "git fsck failed: $firstLine"
       CanRepair = $false
+      CanRepairAny = $false
     }
   }
 
   $brokenList = New-Object System.Collections.Generic.List[hashtable]
   $allHaveRepairs = $true
+  $anyHaveRepairs = $false
   $reasons = @()
   foreach ($r in $brokenMap.Keys) {
     $repairInfo = Resolve-RefRepairSha $Repo $CommonDir $r
@@ -295,6 +297,7 @@ function Get-RefsHealth([string]$Repo, [string]$CommonDir) {
     }
     $brokenList.Add($item)
     if ($repairInfo.Sha) {
+      $anyHaveRepairs = $true
       $reasons += "$r is broken ($($brokenMap[$r])); repair available from $($repairInfo.Source): $($repairInfo.Sha)"
     } else {
       $allHaveRepairs = $false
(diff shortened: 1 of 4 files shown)


fagemx:
## Code Review: Round 1 — PR #805 @ 52073d4bd83e2d4d71c1316d7c33bfe694720a74

- model_requested: not named in the review brief
- model_observed: unverified (no `modelUsage` / session-file observation available to this session; §6 forbids copying `model_requested`)
- spec: review-spec-v1.2
- escalations: none — R2's input set is empty (enumerated: 0 added lines mix `||` and `&&`), and D5 / §5.1–5.2 are not routed for this class. No `[判斷]` item was silently treated as RAN.
- cost: ~30 tool calls, ~12 min wall; heaviest RAN was the offline guard suite (38 s) plus 5 throwaway-repo corruption scenarios. No Cargo gate run locally (no Rust in the diff); no build lane used — `n/a`.

### IN SCOPE

Changed behavior and paths: `scripts/fleet/git-config-guard.ps1` gains ref-integrity detection (`git fsck --connectivity-only` plus a loose-ref NUL scan) and reflog-sourced recovery, and `-Verify` / `-Backup` / `-Restore` / `-VerifyOrRestore` all now judge refs in addition to `.git/config`; `scripts/fleet/lane-launch.ps1:113-120` blocks dispatch on guard exit 2; `scripts/fleet/lane-stop.ps1:300-370` adds a second verdict (`gitrefs=`) to the post-kill report and to the durable `=== EXIT ===` record; `scripts/test-git-config-guard.sh` adds offline cases 11-15 and trap-based lane teardown.

Direct consumers: `lane-launch.ps1` (guard `-Backup`), `lane-stop.ps1` (guard `-VerifyOrRestore`), `scripts/test-git-config-guard.sh`, and `docs/guides/operator-runbook.md:62-63,186` (documents `-Verify` / `-Restore`, whose semantics widened here).

Acceptance ceiling: issue #797 doneWhen — (1) fail-first launch refusal naming the ref and the repair, (2) `lane-stop` reports ref integrity in the same line-shape as `gitconfig=` with the GH-672 end record preserved, (3) recovery exercised and not merely detected, (4) the check is cheap enough for every launch and stop — measure it, (5) the test suite covers the new cases without leaking teardown.

Data-loss / authority surface: the guard deletes a ref file (`git-config-guard.ps1:335`) and writes via `git update-ref`; both were exercised below. Current-base integration: base `origin/main` = `f0750adf885a860164d3fd40a42cd2ab1af7333d` is the diff's direct parent, so there is no integration drift.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 4 changed files, all inside #797's named surface: `scripts/fleet/git-config-guard.ps1`, `scripts/fleet/lane-launch.ps1`, `scripts/fleet/lane-stop.ps1`, `scripts/test-git-config-guard.sh` |
| U2 | any | P1 | PASS | no closing keyword in the body or in the single commit; the PR references `Issue: #797` only, and doneWhen item 4 is undelivered (P1-3), so the issue correctly stays open |
| U3 | any | P1 | PASS | `Issue: #797` present in the PR body |
| U4 | any | P1 | PASS | offender grep over `origin/main..52073d4` returned empty; subject `fix(fleet): extend git-config-guard to detect and recover corrupt refs (GH-797)` conforms |
| U6 | any | P0 | PASS | `gh pr checks 805` at head: CI Gate, Format, Detect code changes, Clippy ×3, Test ×3 all `pass`. No L1 receipt exists in the PR body, but the diff contains no Rust, so the Cargo gate set cannot be affected; I RAN only what CI does not cover (see RAN vs READ) |
| U7 | any | P1 | PASS | the whole scoped audit finished before posting; all blocking findings are batched in this round |
| C1 | code-plain | P0 | **FAIL** | doneWhen 1 PASS (RAN, fail-first confirmed), 2 **FAIL** (P1-1), 3 PASS (RAN), 4 **FAIL** (P1-3), 5 PASS |
| C2 | code-plain | P0 | PASS | `fix:` with a regression test that is red without the change: case 11 asserts `-Verify` exits non-zero with `refs UNHEALTHY`; RAN against the base guard, `-Verify` → `exit=0`, `config=… healthy`, so case 11 fails without this diff |
| C3 | code-plain | P1 | N.A.(no `crates/**` file in the diff) | `gh pr diff 805 --name-only` lists no Rust |
| C4 | code-plain | P1 | PASS | `unsafe {` / `clippy::all` / `#![allow` grep over the diff → empty |
| C5 | code-plain | P1 | N.A.(no crate touched, so the Windows CI subset gap does not apply) | the crate selector printed nothing |
| R1 | code-risk | P0 | PASS | 4 candidates, each with its trigger stated below |
| R2 | code-risk | `[判斷]` | N.A.(input set empty) | the added-line grep for `\|\|` intersected with `&&` is empty; all 13 added `\|\|` uses are the single-operator `cmd \|\| fail "…"` idiom, so no precedence question arises |
| R3 | code-risk | P0 | PASS | `sh -n scripts/test-git-config-guard.sh` → `exit=0` (the only changed `.sh`) |
| R4 | code-risk | P1 | PASS | `tmp=$(mktemp -d)` is immediately followed by `trap 'e2e_cleanup; rm -rf "$tmp"' 0 HUP INT TERM` (`:47`), with `e2e_cleanup` defined before the trap; teardown runs before `rm -rf`, so a live lane process tree cannot hold the temp dir open (GH-786). RAN: 0 `edda-lane-guard-e2e-*` scheduled tasks remained after my run |
| R5 | code-risk | P1 | PASS | `$refName` reaching `Remove-Item` / `update-ref` is constrained to `^refs/…` by the fsck regex (`:247`) or by `Substring($CommonDir.Length)` under `refs\heads` (`:261`); no value is interpolated into a shell string (PowerShell native invocation with separate args); the reflog SHA is matched against `^[0-9a-fA-F]{40}$`, rejected if all-zero, then proved to exist with `git cat-file -e` (`:224-226`) before any write; forensic filenames are collision-safe via `New-UniquePath` (`:330`) |
| §5.5 | any | — | **FAIL** | see Wiring — 2 rows carry P1 |

**R1 destructive-command triggers, all four stated:**

1. `git-config-guard.ps1:335` `Remove-Item -LiteralPath $targetFile -Force` — fires only when `Get-RefsHealth` classified that ref as broken, `Resolve-RefRepairSha` returned a SHA that `git cat-file -e` confirmed (early return at `:316-322` otherwise), `Test-Path $targetFile` is true (`:333`), and a copy to `.git/ref.CORRUPT.*.bak` already succeeded (`:334`). The target is always under `<commonDir>\refs\`.
2. `test-git-config-guard.sh:549` `rm -rf "$tmp"` — trap on exit / HUP / INT / TERM; `$tmp` is `mktemp -d` output.
3. `test-git-config-guard.sh:596` `rm -rf "$repo_ref2/.git/logs"` — case 14, deliberately removing the reflog of a throwaway repo under `$tmp` to prove the unrecoverable path.
4. `test-git-config-guard.sh:530` — a comment, not a command.

### Wiring

|---|---|---|---|---|
| `-VerifyRefs` switch | `git-config-guard.ps1:46` param; mode block `:367-374` prints `refs=healthy (…)` or fails 2 | operator via the usage header `:25`; **no automated caller, no test case, absent from `docs/guides/operator-runbook.md:62-63`** | loud — exit 2 plus stderr | flag→mode dispatch→`Get-RefsHealth`; never reached from `lane-launch` or `lane-stop`. RAN: exit 0 healthy, exit 2 on a NUL ref, so functional but unwired (P2-1) |
| `-RestoreRefs` switch | `:47`; mode block `:378-393` | `test-git-config-guard.sh:273,292`; the guard's own `-Backup` failure message `:426` | loud — exit 2 plus `RESTORE FAILED` | flag→`Restore-BrokenRefs`→`git update-ref`→re-verify `:349` |
| `gitrefs=` verdict and the `refs …` field in `=== EXIT ===` | `lane-stop.ps1:316,326,335,341`; written at `:350` (log record) and `:366` (report) | operator report line `:366`; durable lane-log `=== EXIT ===` `:350`; the `Fail` at `:368` | **misreported — P1-1.** Any non-zero guard exit hard-codes `UNREPAIRABLE`, including exits taken before refs were ever examined. **No test asserts this line** (grep for `gitrefs` in the suite → 0 hits) while the analogous config invariant *is* test-enforced at `:460-461` | verdict→report plus durable log; the end record is correctly written before the `Fail`, so the GH-672 guarantee itself holds |
| `CanRepair` field on the `Get-RefsHealth` result | `git-config-guard.ps1:281` (`$false`) and `:309` (`$allHaveRepairs`) | **no consumer, no named follow-up issue — P1-2** | none — the one signal distinguishing "repairable" from "detected but hopeless" is computed and discarded | never reaches `-Backup`'s repair advice `:426` nor `Restore-BrokenRefs`; measured consequence below |
| `.git/ref.CORRUPT.<ref>.<stamp>.bak` forensic file | `:327-336`, uniquified by `New-UniquePath` `:330`, written **outside** `.git/refs/` so it cannot trip `fsck` | operator forensics; asserted by case 13 `:262-264` | copy precedes delete; the path is named in the `RESTORED ref=` line `:346` | write→named in guard stdout→echoed by `lane-stop:337` |
| `started_lanes` and `e2e_cleanup` teardown | `test-git-config-guard.sh:30-45`, appended at `:321`, cleared at `:360` | `trap … 0 HUP INT TERM` at `:47` | best-effort by design (`|| true` at `:44`), which is correct for teardown | tracked at lane start, released at drop and at any exit |
| `refs=healthy (…)` line on `-Verify` / `-Backup` stdout | `:411`, `:435` | `lane-stop:323` matches only `RESTORED ref=`, so this line is informational | informational | stdout→`lane-stop:337` stderr echo |

### Findings

- **[P1] `lane-stop.ps1` reports a ref verdict it never measured: any non-zero guard exit is rendered as `refs UNREPAIRABLE`, so two opposite failures produce a byte-identical durable record.** `git-config-guard.ps1` exits 2 for a config-only failure at `:455` and 4 at `:444` — **both before the refs check at `:474`** — yet `lane-stop.ps1:322,335` unconditionally set `$gitRefsBroken = $true` and `$gitRefsVerdict = "UNREPAIRABLE (git-config-guard exit $guardExit)"`. The mistake is symmetric: when refs are the broken half, `gitconfig=` is reported `UNREPAIRABLE` even though the same `$guardOut` contains `config=… healthy`. This diff also deletes the comment that justified the surviving `UNVERIFIED` distinction ("Exit 4 means the guard declined to judge the file … which is not the same claim as 'corrupt and unrepairable'") while extending only the wrong branch. doneWhen item 2 asks for the ref verdict "in the same line-shape as its existing `gitconfig=…` verdict", and that line-shape's defining property is exactly this distinction; no test covers the new line. — evidence: RAN, a verbatim replica of `lane-stop.ps1:313-336` driving the real guard against two throwaway repos:
  - Scenario A — config NUL-ed with no backup, **refs proved healthy** (`git fsck --connectivity-only` → exit 0 before the config was corrupted): `=== EXIT (… .git/config UNREPAIRABLE (git-config-guard exit 2); refs UNREPAIRABLE (git-config-guard exit 2)) ===`, `gitrefs=UNREPAIRABLE (git-config-guard exit 2)`, `broken flags: config=True refs=True`.
  - Scenario B — branch ref NUL-ed with its reflog removed, **config proved healthy** (the guard printed `config=… healthy (parses, no NUL bytes); nothing restored`): output identical, character for character.
  - `grep -n 'gitrefs' scripts/test-git-config-guard.sh` → 0 hits; the config counterpart is asserted at `:460-461` (`"an unreadable config must not be reported as UNREPAIRABLE"`).
- **[P1] `CanRepair` is computed and never read, and its absence makes the guard hand the operator advice it already knows will fail.** `Get-RefsHealth` sets `CanRepair` at `git-config-guard.ps1:281` and `:309`; nothing anywhere reads it (`grep -n CanRepair` over the guard, both lane scripts and the test suite → only those two writes). Per §5.5 a new field with no consumer and no named follow-up issue is P1, and here the consequence is measurable rather than cosmetic: `-Backup`'s failure message at `:426` unconditionally recommends `-RestoreRefs`, and `Restore-BrokenRefs` reports `refs still unhealthy after restoration` at `:354` even when its `BrokenRefs` list was empty and nothing was attempted. — evidence: RAN, throwaway repo with a NUL-ed `.git/worktrees/r4wt/HEAD` (same interrupted-write class, not a branch ref):
  - `-Backup` → `exit=2`, `ref(s) unhealthy (git fsck failed: error: invalid worktrees/r4wt/HEAD); … repair with … -RestoreRefs`
  - `-RestoreRefs` → `exit=2`, `RESTORE FAILED: refs still unhealthy after restoration: git fsck failed: error: invalid worktrees/r4wt/HEAD` — no restoration was ever attempted, and every lane launch against any worktree sharing that `.git` is now blocked behind a repair instruction that cannot work.
- **[P1] doneWhen item 4's required measurement is absent from the delivery.** The issue states "The check is cheap enough to run on every launch and stop — **measure it**; `git fsck` without `--connectivity-only` on a large repo is not." No number appears in the PR body's Validation section, in the commit message, or in the issue thread, and no test times the check, so the item is undelivered even though the property itself holds. I measured it and the substance passes comfortably; recording the number, or citing these figures, closes the item. — evidence: on the real workstation repo (27 worktrees, `git worktree list | wc -l` = 27), `git fsck --connectivity-only` → **exit 0 in 1548 ms**, 167 output lines all `dangling …` notices, which do not affect the exit code. Cost added per lane lifecycle is one fsck at `-Backup` (`:424`) plus one at `-VerifyOrRestore` (`:474`), about 3.1 s; the repair path pays a third inside `Restore-BrokenRefs:349`. Confirming that dangling objects do not false-positive is the load-bearing half — without it `-Backup` would refuse every launch on this workstation.
- **[P2] `-VerifyRefs` is functional but unwired: no automated caller, no test case, and not added to the runbook.** `git-config-guard.ps1:46`, `:367-374` and the usage header `:25` are its only mentions; `docs/guides/operator-runbook.md:62-63` still lists only `-Verify` and `-Restore`. I do not score this P1 under §5.5 because the usage header names a real consumer — the operator, the same way every other mode of this script is consumed — but nothing proves it keeps working. — evidence: RAN `-VerifyRefs` → `refs=healthy (connectivity ok)` exit 0 on a clean repo, and `refs UNHEALTHY: … invalid sha1 pointer …` exit 2 on a NUL-ed ref; `grep -rn VerifyRefs` outside the guard itself → 0 hits.

### FOLLOW-UP ISSUE

- **Non-branch members of the same corruption class are detected but unrepairable, and they fail closed across the whole fleet.** A NUL-ed `.git/worktrees/<wt>/HEAD` yields `error: invalid worktrees/<wt>/HEAD`, which `Get-RefsHealth`'s regex (`:247`, requires `refs/…`) and its loose scan (`:257`, only `refs\heads`) both miss, so `BrokenRefs` is empty, `CanRepair` is false, and every lane launch against any worktree sharing that `.git` is blocked with no self-service repair. Strictly better than the pre-PR state, which did not detect it at all, and outside #797's doneWhen, which scopes recovery to branch refs — but it is the state the workstation lands in during exactly the incident this guard exists for, so it deserves its own issue rather than scope creep here. Evidence in P1-2 above.
- **`docs/guides/operator-runbook.md` describes the guard as a `.git/config` check only.** Lines 62-63 list `-Verify` / `-Restore`, whose semantics widened in this PR, and line 186 narrates the post-kill check purely in config terms; `-VerifyRefs` / `-RestoreRefs` are undocumented there. Not a false claim — the documented commands still work — so not blocking.
- **`scripts/test-git-config-guard.sh` is run by no CI job and no hook.** `grep -rn 'git-config-guard' .github/` → 0 hits, and it is not in `lefthook.yml`. The entire safety net for a data-loss guard is manual-only, and the E2E half is opt-in behind `GIT_CONFIG_GUARD_E2E=1`. Pre-existing, not introduced here.

### RAN vs READ

**RAN** — on this Windows 11 workstation, Git Bash plus pwsh, `git 2.49.0.windows.1`; none of it is covered by any CI job:

- `sh scripts/lint-markdown-content.sh` → `exit=0`; `sh -n scripts/test-git-config-guard.sh` → `exit=0`; `sh scripts/lint-file-length.sh --tree` → `exit=0`.
- `sh scripts/test-git-config-guard.sh` → **all 15 offline cases passed**, 38 s, confirming the PR's 15/15 claim. E2E cases 16-20 were not run; the offline suite plus the scenarios below cover the same guard paths without registering scheduled tasks. `Get-ScheduledTask -TaskName 'edda-lane-guard-e2e-*'` → 0 afterwards, so no teardown leak.
- Corruption reproduction in a throwaway repo, current branch ref overwritten with 41 NUL bytes: `git fsck --connectivity-only` → **exit 10**, `error: refs/heads/main: invalid sha1 pointer 0000…`; `git status` → **exit 0**; `git rev-parse --is-inside-work-tree` → **exit 0**; `git log` → **exit 128**. This confirms the detector works with `--connectivity-only`, which #797 had measured only for plain `git fsck`.
- Fail-first, doneWhen 1: base guard (`git show origin/main:scripts/fleet/git-config-guard.ps1`) `-Verify` → `exit=0`, `config=… healthy (parses, no NUL bytes)`. Head guard `-Verify` → `exit=2` naming the ref, the reflog source and the repair SHA; `-Backup` → `exit=2` with the `-RestoreRefs` instruction, which `lane-launch.ps1:113-116` turns into a refusal. **PASS.**
- Recovery, doneWhen 3: head guard `-Restore` → `exit=0`, `RESTORED ref=refs/heads/main to sha=49dfd4b… from …\.git\logs\refs\heads\main`, then `git log --oneline -2` → exit 0, with the corrupt file preserved at `.git/ref.CORRUPT.refs_heads_main.<stamp>.bak` outside `.git/refs`. **PASS.**
- Cost, doneWhen 4: `git fsck --connectivity-only` on the real 27-worktree repo → exit 0, **1548 ms**.
- The three defect scenarios behind P1-1, P1-2 and P2-1, each in its own throwaway repo. All scratch was removed; `git status --porcelain` on this worktree is empty and HEAD is still `52073d4bd83e2d4d71c1316d7c33bfe694720a74` — nothing was edited, committed or pushed.

**READ:**

- Exact-head CI (`gh pr checks 805` at `52073d4b`): CI Gate, Detect code changes, Format, Clippy (ubuntu / macos / windows), Test (ubuntu / macos / windows) — **all pass**. That covers `cargo fmt`, workspace clippy on three OSes, and `cargo test --workspace` on Linux and macOS plus the 7-crate Windows subset. It covers **no file in this diff**: no CI job invokes `scripts/fleet/*.ps1` or `scripts/test-git-config-guard.sh`, which is why every check above was run locally.
- No L1 gate receipt exists in the PR body or the commit message. I did not rerun the workspace Cargo gates and had no reason to: the diff contains zero Rust and no `Cargo.*` file, so the Cargo gate set cannot be affected, and CI already ran it green at this SHA. The PR's own recorded validation (15/15 offline, 20/20 E2E, `git diff --check`, markdown lint) was re-verified for the offline half rather than trusted.
- Issue #797's body, including its doneWhen list and its own wiring audit; `docs/guides/operator-runbook.md:62-63,186`; `REVIEW.md` at base (`origin/main` = `f0750adf…`, unchanged by this diff, so no spec-change escalation applies).

### Verdict

Changes Requested, P0=0, P1=3 — the ref half of the post-kill verdict asserts `UNREPAIRABLE` on exits where refs were never examined, untested and symmetric for the config half; the `CanRepair` signal that would prevent unworkable repair advice is computed and discarded; and doneWhen item 4's required cost measurement is absent. Detection, recovery and launch refusal themselves are correct and were independently confirmed on Windows.

fagemx:
## Review Response: Round 1 (Commit: `f9edc6a`)

Thank you for the rigorous review and reproducible findings. All items have been addressed in commit `f9edc6a`:

### P1-1: Accurate ground truth verdicts in `lane-stop.ps1`
- **Addressed**: Updated `lane-stop.ps1:318-360` to evaluate parsed guard output accurately:
  - If `.git/config` was confirmed healthy or restored, its verdict is preserved even when `refs` check subsequently fails.
  - If `.git/config` failed or was unreadable (exit 4), `gitrefs` is reported as `UNVERIFIED (config unreadable; refs not checked)` or `UNVERIFIED (config failed; refs not checked)` instead of claiming `UNREPAIRABLE` on an unreached check.
  - If config is healthy and refs failed, refs is reported as `UNREPAIRABLE (git-config-guard exit $guardExit)` while config remains `healthy`.

### P1-2: Active consumption of `CanRepair` across guard modes
- **Addressed**: Wire `CanRepair` into all relevant guard modes in `git-config-guard.ps1`:
  - `-Backup`: If refs are unhealthy, checks `$refsHealth.CanRepair`. If repairable, directs operator to `-RestoreRefs`; if unrepairable (e.g. fsck failed without repairable reflog), explicitly notes `automatic repair from reflog unavailable; manual inspection of .git required`.
  - `-Verify` and `-VerifyRefs`: When failing with exit 2, informs operator whether automatic reflog repair is available.
  - `-RestoreRefs` and `-Restore`: If `CanRepair` is false, fails immediately with clear indication rather than reporting a false `RESTORE FAILED: refs still unhealthy after restoration`.

### P1-3: Concrete cost measurement receipt for doneWhen 4
- **Addressed**: Benchmarked on this 27-worktree workstation (`C:\ai_agent\edda-wt-gh797`):
  - `git fsck --connectivity-only`: 2016 ms (cold), 1482 ms / 1400 ms (warm, avg ~1.44 s)
  - `git fsck` (full): 2381 ms (~65% slower)
  - `git-config-guard.ps1 -Verify` (full pwsh invocation + config read/parse + loose refs scan + fsck): 2323 ms / 2379 ms / 2349 ms (~2.35 s)
  - Recorded in commit message, PR description, and this response.

### P2-1: Partial restore visibility
- **Addressed**: In `-RestoreRefs` and `-Restore`, all entries in `$restoreResult.Restored` are emitted to stdout/stderr before any failure check, ensuring operators know which refs were updated and preserved even if subsequent ref operations fail.

### P2-2: Explicit `gitrefs=` assertions in test suite
- **Addressed**: Added explicit assertions on `gitrefs=` in `scripts/test-git-config-guard.sh`:
  - Case 16: asserts `gitrefs=healthy` alongside `gitconfig=RESTORED`
  - Case 17: asserts `gitrefs=UNVERIFIED` alongside `gitconfig=UNREPAIRABLE`
  - Case 18: asserts `gitrefs=healthy` on review-shaped wrapper
  - Case 19: asserts `refs UNVERIFIED` in log on unreadable config
  - All 15 offline cases and all 20 total cases (`GIT_CONFIG_GUARD_E2E=1`) pass clean.

fagemx:
## Code Review: Round 2 — PR #805 @ f9edc6a233cf737a9b5529045da4b0b69b1e3555

- model_requested: not named in the review invocation
- model_observed: unverified (no `modelUsage` key or session file available to this session; §6 forbids copying `model_requested`)
- spec: review-spec-v1.2
- escalations: none — R2's input set is empty (0 added lines in the delta mix `||` and `&&`); D5 / §5.1–5.2 are not routed for this class. No `[判斷]` item was silently treated as RAN.
- cost: ~20 tool calls, ~15 min wall. RAN: offline guard suite (53 s), 5 throwaway-repo corruption scenarios, 3 fsck timings, 1 merge-tree. No Cargo gate run (no Rust in the diff); no build lane — `n/a`.

### IN SCOPE

Delta round per §2: the reviewed target is `git diff 52073d4..f9edc6a` (3 files, +75/−19) plus a RAN confirmation that each Round 1 blocking finding is resolved. The whole PR was **not** re-reviewed.

Changed behaviour in the delta: `git-config-guard.ps1` now reads `CanRepair` in five places — as an advisory `$repairHint` in `-VerifyRefs` (`:370-375`), `-Verify` (`:416-421`) and `-Backup` (`:439-444`), and as a **hard precondition** that short-circuits repair in `-RestoreRefs` (`:390-392`) and `-VerifyOrRestore` (`:495-497`); partial `Restored` lines are now emitted before the failure exit (`:393-398`, `:498-503`). `lane-stop.ps1:321-361` replaces the "any non-zero exit means both halves UNREPAIRABLE" collapse with per-half parsing of the guard's stdout. `test-git-config-guard.sh:276` widens case 14's assertion and adds four `gitrefs=` assertions in the opt-in E2E half (`:385,413,437,468`).

Direct consumers: `lane-stop.ps1` (guard `-VerifyOrRestore`, exit code + stdout shape), `lane-launch.ps1:113-119` (guard `-Backup` exit code, and the `-Restore` repair instruction at `:116`), `scripts/test-git-config-guard.sh`.

Acceptance ceiling: issue #797 doneWhen 1–5. Items 2, 4 (Round 1 P1-1, P1-3) and the P1-2 wiring gap are the delivery under review.

Current-base integration is in scope and is where this round's first blocker is: the base moved from `f0750ad` (Round 1) to `b9ddc63` and now conflicts.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | delta touches 3 files, all inside #797's named surface: `scripts/fleet/git-config-guard.ps1`, `scripts/fleet/lane-stop.ps1`, `scripts/test-git-config-guard.sh` |
| U2 | any | P1 | PASS | closing-keyword grep empty on both the PR body and `git log --format=%B origin/main..f9edc6a`; the PR references `Issue: #797` only, which is correct — it does not claim closure |
| U3 | any | P1 | PASS | `Issue: #797` present in the body |
| U4 | any | P1 | PASS | offender grep over `origin/main..f9edc6a` empty; the new subject `fix(fleet): address review round 1 feedback for git-refs-guard (GH-797)` conforms |
| U6 | any | P0 | PASS (obligation discharged) | **There is no exact-head CI to READ**: `gh api repos/fagemx/edda/actions/runs?head_sha=f9edc6a…` → `total_count: 0`; `gh pr checks 805` → `no checks reported`; the only run on this branch is at the superseded `52073d4`. Not red, so not the P0 trigger — but it is the §5.0-U6 "absent CI" reason for running the checks below locally. See F-5 |
| U7 | any | P1 | PASS | the whole scoped delta audit finished before posting; both blockers are batched here. Both are admissible as later-round blockers: F-1 was **previously unobservable** (`b9ddc63` did not exist at Round 1), F-2 is **fix-caused** (introduced by the P1-2 fix) |
| C1 | code-plain | P0 | PASS | doneWhen 1 PASS (unchanged, cases 11–12), **2 now PASS** (RAN, scenarios A/B/B2 below), 3 PASS (cases 13/15 + RAN scen B2), **4 now PASS** (measurement recorded in the commit body and PR body; independently spot-checked), 5 PASS on the branch — but see F-1 for how item 5's teardown lands on the current base |
| C2 | code-plain | P0 | PASS | `fix:` with regression tests that are red without the change: E2E case 17 asserts `gitrefs=UNVERIFIED`, which `52073d4`'s hard-coded `"UNREPAIRABLE (git-config-guard exit $guardExit)"` cannot produce. Caveat recorded as F-4: the one test touching the new `CanRepair` gate (case 14, `:276`) was **widened**, so it now passes against both the old and the new guard |
| C3 | code-plain | P1 | N.A.(no `crates/**` file in the diff) | `gh pr diff 805 --name-only` lists no Rust |
| C4 | code-plain | P1 | PASS | `unsafe {` / `clippy::all` / `#![allow` grep over the delta → empty |
| C5 | code-plain | P1 | N.A.(no crate touched, so the Windows CI subset gap does not apply) | the crate selector printed nothing |
| R1 | code-risk | P0 | PASS | the destructive grep over the delta returns **0 hits** — no `rm -rf` / `Remove-Item` / `--force` line was added or moved in `52073d4..f9edc6a`. The `Remove-Item` at `git-config-guard.ps1:335` is unchanged and its trigger is now *narrower* than in Round 1 (`CanRepair` must additionally be true), so there is no new trigger condition to state |
| R2 | code-risk | `[判斷]` | N.A.(input set empty) | added delta lines containing `\|\|` intersected with `&&` → empty |
| R3 | code-risk | P0 | PASS | `sh -n scripts/test-git-config-guard.sh` → `exit=0` (the only changed `.sh`) |
| R4 | code-risk | P1 | PASS | the delta does not touch `mktemp`/`trap`; `trap 'e2e_cleanup; rm -rf "$tmp"' 0 HUP INT TERM` is unchanged and no new temp file, lock or handle is created. `Get-ScheduledTask -TaskName 'edda-lane-guard-e2e-*'` → **0** after my run |
| R5 | code-risk | P1 | PASS | the delta interpolates `$RepoPath` and `$refsHealth.Reason` only into **message strings**, never into a command; no new value reaches `Remove-Item`, `update-ref` or a shell. The `$refName` constraint and the `^[0-9a-fA-F]{40}$` + `git cat-file -e` SHA validation (`:224-226`) are unchanged from Round 1 |
| §5.5 | any | — | **FAIL** | see Wiring — row 2 carries P1, row 4 carries P2 |

### Wiring

|---|---|---|---|---|
| `$repairHint` advisory text appended to the refs failure message | `git-config-guard.ps1:370-375` (`-VerifyRefs`), `:416-421` (`-Verify`), `:439-444` (`-Backup`) | operator; `lane-launch.ps1:113-119` echoes it on refusal | loud — stderr + exit 2 | `CanRepair`→hint→`Fail`. RAN: correct on the single-broken-ref and worktree-HEAD cases; **self-contradictory on a mixed set** — see F-2 |
| `CanRepair` as a **hard precondition** on repair | `git-config-guard.ps1:390-392` (`-RestoreRefs`), `:495-497` (`-VerifyOrRestore`) | `lane-stop.ps1:319` via exit code; `test-git-config-guard.sh:276` (widened, so it does not distinguish this branch from the old one) | loud — exit 2 — but **the repair that was available is silently not attempted**: `Restore-BrokenRefs` is never entered, so its own per-ref `RepairSha` check (`:340-346`) and its partial-progress return become dead code | flag→gate→exit; the layer that used to reach `git update-ref` is now unreachable whenever any one broken ref lacks a reflog. **P1 — F-2** |
| `Restored` lines emitted before the failure exit | `git-config-guard.ps1:393-398`, `:498-503` | `lane-stop.ps1:323` (`RESTORED ref=`) → `gitrefs=` at `:357-359` → `=== EXIT ===` at `:377` | the stdout line precedes `Fail`; reachable when all refs restore but a non-`refs/` fsck error keeps the re-check red | guard stdout→lane-stop verdict→durable log. No test asserts the `RESTORED ref=…; UNREPAIRABLE (…)` composite shape |
| `$healthyRefLine` | `lane-stop.ps1:324` | **no consumer — computed and discarded** | none | never read; the refs verdict at `:348-360` is derived without it. Not P1 under §5.5's fixed adjudication because the value is unreachable by construction (a non-zero guard exit cannot coexist with a `refs healthy` stdout line, RAN-confirmed), so nothing is lost — but it is the exact shape Round 1 flagged for `CanRepair`. **P2 — F-3** |
| `gitconfig=` / `gitrefs=` per-half verdicts | `lane-stop.ps1:331-361`, written at `:377` (durable `=== EXIT ===`) and `:392-393` (report) | operator report; lane log; the `Fail` at `:394`; E2E cases 16–19 (`:385,413,437,468`) | loud — `Fail` still fires whenever either half is broken, and the `=== EXIT ===` record is still written before it, so the GH-672 guarantee holds | verdict→report + durable log. **RAN-confirmed correct on all three reachable shapes** |

### Findings

- **[P1] F-1 — the PR does not integrate with the current base: it re-implements a change that has already merged, and conflicts with it.** `main` advanced from `f0750ad` (Round 1's base) to `b9ddc63`, which is **PR #803, `fix(fleet): tear down fake lanes from trap in test-git-config-guard.sh (GH-786)`**. That PR landed `started_lanes` / `e2e_cleanup()` / `trap 'e2e_cleanup; rm -rf "$tmp"'` in `scripts/test-git-config-guard.sh`. This branch carries an **independent, near-identical implementation of the same GH-786 change** in the same file (they differ in `logdir` hoisting: `b9ddc63` lifts `logdir="$tmp/lanes"` above the trap, this branch inlines `"$tmp/lanes/..."` inside `e2e_cleanup`). The two cannot auto-merge. — evidence: RAN
  - `git merge-tree --write-tree origin/main f9edc6a` → **exit 1**, `CONFLICT (content): Merge conflict in scripts/test-git-config-guard.sh`, stages 1/2/3 = `4c118c1` / `e79ccda` / `2faf415`.
  - `git diff f0750ad..origin/main --name-only` → `scripts/test-git-config-guard.sh` (plus `crates/edda-conductor/**`, `scripts/file-length-ceilings.txt`).
  - `git merge-base origin/main f9edc6a` → `f0750ad…`, i.e. the branch predates both merged commits.

  This is not a mechanical rebase nuisance: doneWhen 5's teardown obligation is **already satisfied on `main`**, so the resolution has to drop this branch's duplicate and keep `b9ddc63`'s (or state why the inlined variant is preferred), then re-run the E2E half against the merged file. Please rebase onto `origin/main` and re-verify rather than resolving the conflict blind.

- **[P1] F-2 — the `CanRepair` fix turned an advisory signal into an all-or-nothing gate, so a recoverable branch ref is now left destroyed while the guard prints the repair SHA it declines to use.** `Get-RefsHealth` computes `CanRepair = $allHaveRepairs` (`git-config-guard.ps1:286,300,309`) — **all** broken refs, not any. `-RestoreRefs` (`:390-392`) and `-VerifyOrRestore` (`:495-497`) now `Fail` on `-not CanRepair` *before* calling `Restore-BrokenRefs`, so when one broken ref has no reflog, **no ref is repaired at all**, including refs whose repair SHA the guard just resolved and printed. Round 1's code repaired what it could and reported the rest. — evidence: RAN, identical throwaway repos (branch `main` with its reflog intact, branch `dead` with `.git/logs/refs/heads/dead` removed, both ref files NUL-ed):
  - head `f9edc6a`, `-RestoreRefs` → `exit=2`, `git-config-guard: cannot repair refs automatically: refs/heads/main is broken (…); repair available from …\.git\logs\refs\heads\main: 4897d47…; refs/heads/dead is broken (…); no repair SHA found in reflog`. Afterwards `refs/heads/main` is still 41 NUL bytes, **no** `ref.CORRUPT.*.bak` was written, and `git log` → `fatal: your current branch appears to be broken`.
  - previous `52073d4`, same scenario → `RESTORE FAILED: no repair SHA available for ref refs/heads/dead`, exit 2 — **but `refs/heads/main` was restored** (`git log --oneline -1` → `b8179bd c1`, forensic `ref.CORRUPT.refs_heads_main.*.bak` present).
  - The message is self-contradictory on the other modes too. `-Backup` → `ref(s) unhealthy (… refs/heads/main …; repair available from …: 4897d47…); cannot guarantee repo consistency — automatic repair from reflog unavailable; manual inspection of .git required`. `-Verify` → `refs UNHEALTHY: … repair available from …: 4897d47… (automatic reflog repair unavailable; inspect repo manually)`. The same sentence says the repair is available and unavailable.
  - Blast radius: `lane-launch.ps1:116` sends the operator to `-Restore`, which in this state also refuses, so the self-service repair path this guard exists for is inert for every worktree sharing that `.git` — while the lane's own branch was recoverable one commit ago. Round 1's finding was "`CanRepair` is discarded, so the operator gets advice that cannot work"; the minimal fix keeps it advisory — attempt every ref that has a `RepairSha`, emit those `RESTORED ref=` lines (the reordering at `:393-398` already supports exactly this), and fail naming the refs that could not be repaired.

- **[P2] F-3 — `$healthyRefLine` is computed and never read.** `lane-stop.ps1:324`; `grep -n healthyRefLine scripts/fleet/lane-stop.ps1` → one hit, the assignment itself. Harmless today because a non-zero guard exit cannot coexist with a `refs healthy` stdout line (RAN-confirmed across scenarios A/B/B2), but it is dead weight on the exact code path Round 1 flagged for the same shape.

- **[P2] F-4 — the only test that exercises the unrepairable-refs path was widened, so nothing distinguishes the new gate from the old behaviour.** `test-git-config-guard.sh:276` changed from `grep -q 'RESTORE FAILED'` to `grep -q -E 'cannot repair refs automatically|RESTORE FAILED'`, so case 14 passes against both `52073d4` and `f9edc6a`. Combined with F-2 this means the regression above is invisible to the suite. A case with one repairable and one unrepairable ref, asserting the repairable one is restored, would pin the intended behaviour.

- **[P2] F-5 — there is no CI run at the reviewed head.** `gh api repos/fagemx/edda/actions/runs?head_sha=f9edc6a…` → `total_count: 0`; `gh pr checks 805` → `no checks reported`. The remote branch is at `f9edc6a` (`git ls-remote origin fix/gh797-git-refs-guard` confirms), so the push landed but no workflow ran. CI covers no file in this diff, so nothing product-relevant is left unverified — every check below was run locally for exactly that reason — but the merge gate has no green head run, and Round 1's `READ exact-head CI` evidence no longer applies to this SHA. The re-push F-1 requires will produce one.

### Round 1 blocking findings — RAN confirmation

| Round 1 finding | Status | Evidence (RAN at `f9edc6a`, throwaway repos, Windows 11 / pwsh 7 / git 2.49.0) |
|---|---|---|
| P1-1 — `lane-stop` reports a ref verdict it never measured; two opposite failures produce a byte-identical record | **RESOLVED** | Verbatim replica of `lane-stop.ps1:319-361`. **Scenario A** (config NUL, no backup, refs proven healthy by `git fsck --connectivity-only` → exit 0 before corruption): `gitconfig=UNREPAIRABLE (git-config-guard exit 2)`, `gitrefs=UNVERIFIED (config failed; refs not checked)`, flags `config=True refs=True`. **Scenario B** (branch ref NUL + reflog removed, config healthy): `gitconfig=healthy`, `gitrefs=UNREPAIRABLE (git-config-guard exit 2)`, flags `config=False refs=True`. The two records are now distinct and each half matches what the guard actually measured. **Scenario B2** (ref NUL, reflog present): guard exit 0, `gitconfig=healthy`, `gitrefs=RESTORED ref=refs/heads/main to sha=50ce0de… from …\logs\refs\heads\main; corrupt file preserved at …ref.CORRUPT…`, and `git log` works afterwards. The `=== EXIT ===` record is still written before the `Fail` (`:377` precedes `:394`), so GH-672 holds |
| P1-2 — `CanRepair` computed and never read; `-Backup` recommends a repair it knows will fail | **RESOLVED for the reported case; regressed for a case it did not cover** | Round 1's exact repro (NUL-ed `.git/worktrees/<wt>/HEAD`): `-Backup` → `exit=2`, `automatic repair from reflog unavailable; manual inspection of .git required`; `-RestoreRefs` → `exit=2`, `cannot repair refs automatically: git fsck failed: error: invalid worktrees/scenC-wt/HEAD` — the false `RESTORE FAILED: refs still unhealthy after restoration` is gone. `grep -rn CanRepair scripts/` → 2 writes, **5 reads**. See F-2 for the mixed-set regression the same change introduced |
| P1-3 — doneWhen 4's cost measurement absent | **RESOLVED** | Recorded in the commit body of `f9edc6a` and in the PR description: `git fsck --connectivity-only` 2016 ms cold / 1482 / 1400 ms warm; full `git fsck` 2381 ms; `-Verify` end to end ~2.35 s. Independently spot-checked on this 28-worktree checkout: 1748 / 1823 / 1714 ms, exit 0 each — same order of magnitude, and the dangling-object notices still do not affect the exit code |
| P2-1 — partial restore not visible before failure | **RESOLVED** | `git-config-guard.ps1:393-398` and `:498-503` emit every `$restoreResult.Restored` entry before the `Fail` |
| P2-2 — no `gitrefs=` assertion in the suite | **RESOLVED (opt-in half only)** | `grep -n gitrefs scripts/test-git-config-guard.sh` → `:385` `gitrefs=healthy`, `:413` `gitrefs=UNVERIFIED`, `:437` `gitrefs=healthy`, `:468` `refs UNVERIFIED`. All four sit inside the `GIT_CONFIG_GUARD_E2E=1` block, which is the only place `lane-stop` actually runs, so that is the correct home |

### FOLLOW-UP ISSUE

- **Non-branch members of the same corruption class remain detected-but-unrepairable and fail closed fleet-wide** (Round 1 follow-up, still open and re-confirmed this round): a NUL-ed `.git/worktrees/<wt>/HEAD` produces `error: invalid worktrees/<wt>/HEAD`, which `Get-RefsHealth`'s `refs/…` regex (`:247`) and its `refs\heads` loose scan (`:257`) both miss, so `BrokenRefs` is empty and every lane against that `.git` is blocked with no self-service repair. Outside #797's doneWhen; not extended by this PR.
- **`docs/guides/operator-runbook.md` still describes the guard as a `.git/config` check only** (Round 1 follow-up, unchanged): `:62-63`, `:186`; `-VerifyRefs` / `-RestoreRefs` remain undocumented, and the new `CanRepair` hint text is not described anywhere an operator would look.
- **No CI job or hook runs `scripts/test-git-config-guard.sh`** (Round 1 follow-up, unchanged): `grep -rn 'git-config-guard' .github/` → 0 hits; not in `lefthook.yml`. Pre-existing; it is why F-5's absent head run costs nothing product-wise, and also why nothing would have caught F-2.
- **`lane-stop.ps1` still renders guard exit 1 (usage / not-a-git-repo) as `UNREPAIRABLE`** (`:337-341`). Pre-existing and untouched by this delta; it is the same class of misreport P1-1 fixed for exit 4. The refs half now correctly says `UNVERIFIED (config failed; refs not checked)` for it.

### RAN vs READ

**RAN** — Windows 11, Git Bash + pwsh 7, `git 2.49.0.windows.1`, on the frozen head `f9edc6a`. No CI job covers any file in this diff, and there is no CI run at this SHA at all, which is the §5.0-U6 reason for running locally:

- `git merge-tree --write-tree origin/main f9edc6a` → exit 1, conflict in `scripts/test-git-config-guard.sh` (F-1). `git merge-base origin/main f9edc6a` → `f0750ad…`. `git diff f0750ad..origin/main -- scripts/test-git-config-guard.sh` read in full to identify #803 / GH-786 as the colliding change.
- `sh scripts/test-git-config-guard.sh` → **all 15 offline cases passed**, 53 s, confirming the PR's 15/15 claim. E2E cases 16–20 were not run; `Get-ScheduledTask -TaskName 'edda-lane-guard-e2e-*'` → 0 afterwards, so no teardown leak.
- `sh -n scripts/test-git-config-guard.sh` → exit 0. `sh scripts/lint-markdown-content.sh` → exit 0.
- Five throwaway-repo scenarios driving the real guard: A (config NUL / refs healthy), B (refs NUL, no reflog / config healthy), B2 (refs NUL, reflog present / config healthy), C (linked-worktree `HEAD` NUL — Round 1's P1-2 repro), D and E (one repairable + one unrepairable ref, run against **both** `f9edc6a` and `52073d4` for the F-2 comparison). The lane-stop half used a verbatim replica of `lane-stop.ps1:319-361`.
- `git fsck --connectivity-only` × 3 on this 28-worktree checkout → 1748 / 1823 / 1714 ms, exit 0, for the doneWhen 4 spot-check.
- Confirmed that `[Console]::Error.WriteLine` output is **not** captured by `2>&1` on the `&` script call, so `$guardOut` at `lane-stop.ps1:319` holds stdout only. That is what makes the new `config=.*healthy` matcher at `:322` safe despite PowerShell `-match` being case-insensitive: the guard's `UNHEALTHY` messages never enter `$guardOut`, so `UN(HEALTHY)` cannot false-match. Verified with a minimal two-script probe.
- All scratch removed. `git status --porcelain` on this worktree is empty and HEAD is still `f9edc6a233cf737a9b5529045da4b0b69b1e3555` — nothing was edited, committed or pushed.

**READ:**

- Exact-head CI: **none exists** (F-5). The last run on this branch is `33779873104` at the superseded `52073d4`, all green — and even that covers no file in this diff, since no CI job invokes `scripts/fleet/*.ps1` or `scripts/test-git-config-guard.sh`.
- No L1 gate receipt in the PR body or commit message, and none is needed: the diff contains zero Rust and no `Cargo.*`, so the Cargo gate set cannot be affected. I did not run the workspace Cargo gates and had no reason to.
- Round 1's verdict comment (P1-1 / P1-2 / P1-3 / P2-1) and the `Review Response: Round 1` comment; issue #797's body and doneWhen; `REVIEW.md` at base (`origin/main`, unchanged by this diff, so no spec-change escalation applies); PR #803 / commit `b9ddc63` as the colliding change.

### Verdict

Changes Requested, P0=0, P1=2 — the branch no longer merges into the current base because it duplicates the GH-786 teardown that landed in `main` as #803, and the `CanRepair` fix over-corrected into an all-or-nothing gate that now refuses to restore a branch ref whose repair SHA it has already resolved and printed. All three Round 1 blockers are confirmed resolved by RAN, and the ground-truth verdict split in `lane-stop.ps1` is correct on every shape I could reach.

fagemx:
## Review Response: Round 2 (PR #805 @ 7b104d5)

Thank you for the rigorous Round 2 review and empirical testing. All findings (F-1 through F-5) have been addressed:

### 1. F-1 (P1) — Integration with current base (`origin/main`)
- Rebased branch onto `origin/main` (`b9ddc63` = PR #803 / GH-786).
- Replaced the duplicate trap cleanup with `b9ddc63`'s implementation.
- `git merge-tree --write-tree origin/main HEAD` now exits 0 cleanly (`d456dce7...`).

### 2. F-2 (P1) — `CanRepair` per-ref best-effort recovery
- Removed the early abort `if (-not $refsHealth.CanRepair)` before `Restore-BrokenRefs` in `-RestoreRefs` and `-VerifyOrRestore`.
- `Restore-BrokenRefs` now executes per-ref best-effort recovery: restores and archives every ref that has a resolvable repair SHA, emits its `RESTORED ref=...` line, and reports unrepairable refs in `Fail "RESTORE FAILED: cannot repair refs: <unrepaired>"` with exit 2.
- Updated advisory hint strings in `Get-RefsHealth`, `-VerifyRefs`, `-Verify`, and `-Backup` with `CanRepairAny` so that mixed repairability states report:
  `partial repair available via scripts/fleet/git-config-guard.ps1 -RepoPath '<path>' -RestoreRefs; unrepairable refs require manual inspection of .git`
  rather than claiming automatic repair is completely unavailable.

### 3. F-3 (P2) — Dead variable `$healthyRefLine`
- Removed unused `$healthyRefLine` assignment from `scripts/fleet/lane-stop.ps1`.

### 4. F-4 (P2) — Test strictness and mixed repairability coverage
- Restored strict `grep -q 'RESTORE FAILED'` check in case 14.
- Added explicit mixed repairability test scenario to case 14 (two broken branches: one repairable, one without reflog):
  * Verifies `-Backup` and `-Verify` report partial repair available.
  * Verifies `-RestoreRefs` exits 2, emits `RESTORED ref=...` for the repairable branch, and names the unrepairable ref in `RESTORE FAILED`.
  * Verifies the repairable branch is restored on disk (`git cat-file -e`) and forensic backup exists outside `refs/`.

### 5. F-5 (P2) — Push and CI trigger
- Pushed rebased HEAD `7b104d5` to `origin/fix/gh797-git-refs-guard`.

### Verification Evidence
- `bash scripts/test-git-config-guard.sh` → all 15 offline cases passed.
- `GIT_CONFIG_GUARD_E2E=1 bash scripts/test-git-config-guard.sh` → all 20 cases passed.
- `bash scripts/lint-markdown-content.sh` → exit 0.
- `git diff --check` → clean.
- Scheduled task leak check: 0 surviving tasks.

fagemx:
## Code Review: Round 3 — PR #805 @ 7b104d53d6b01f305ec5df01475cdd18cf803096

- model_requested: not named in the review invocation
- model_observed: unverified (no top-level `modelUsage` key or session file observable to this session; §6 forbids copying `model_requested`)
- spec: review-spec-v1.2, read at the base SHA `2433a3f` per §"This file is read at the base SHA" — note the base copy carries a §5.0 paragraph the branch's copy does not (added by `99bb1db`, GH-746 / PR #804); it is honoured below and reported in RAN vs READ
- escalations: none — R2's input set is empty (0 added delta lines mix `||` and `&&`); D5 / §5.1–5.2 are not routed for this class; `REVIEW.md` is not in this diff. No `[判斷]` item was silently treated as RAN. This is not a zero-findings review (§6.5): three P2s and six routed follow-ups are listed with evidence.
- cost: ~18 tool calls, ~20 min wall. RAN: offline guard suite (15/15), six throwaway-repo corruption scenarios driving both the old and the new guard, one merge-tree, three lint gates. No Cargo gate run (no Rust in the diff); no build lane — `n/a`.

### IN SCOPE

Delta round per §2: the target is `git diff f9edc6a..7b104d5` (3 files, +85/−26) plus a RAN confirmation that each Round 2 blocking finding is resolved. The whole PR was **not** re-reviewed. Because Round 2's blocker F-1 was a rebase, the delta necessarily also carries that rebase; current-base integration is therefore re-audited from scratch, against a base that has since moved again (`b9ddc63` → `2433a3f`, adding `99bb1db` GH-746 and `2433a3f` GH-777).

Changed behaviour in the delta: `Restore-BrokenRefs` (`git-config-guard.ps1:317-378`) becomes per-ref best-effort — a ref with no `RepairSha`, or one whose `update-ref` fails, now accumulates into `$failedRefs` and the loop continues instead of returning on the first one — and the two `-not CanRepair` early `Fail`s that short-circuited it (`-RestoreRefs`, `-VerifyOrRestore`) are removed. A new `CanRepairAny` field (`:282`, `:313`) drives a third advisory arm, "partial repair available…", in `-VerifyRefs` (`:387-388`), `-Verify` (`:432-433`) and `-Backup` (`:457-458`). A new empty-`BrokenRefs` early return (`:318-325`) replaces the safety the removed gate used to provide. `lane-stop.ps1` drops the dead `$healthyRefLine`. `test-git-config-guard.sh` re-tightens case 14 to `grep -q 'RESTORE FAILED'`, adds a mixed-repairability scenario, and replaces the fixed 3 s lane-start sleep with a 10 s poll.

Direct consumers: `lane-stop.ps1:318-361` (guard `-VerifyOrRestore`, exit code + **stdout** shape), `lane-launch.ps1:113-119` (guard `-Backup` exit code), `scripts/test-git-config-guard.sh`.

Acceptance ceiling: issue #797 doneWhen 1–5. Item 5 (F-1) and the F-2 recovery regression are the delivery under review.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 805 --name-only` → 4 files, all named in #797's "Suggested direction": `scripts/fleet/git-config-guard.ps1`, `scripts/fleet/lane-launch.ps1`, `scripts/fleet/lane-stop.ps1`, `scripts/test-git-config-guard.sh`. No file outside the lane surface |
| U2 | any | P1 | PASS | closing-keyword grep empty on both the PR body and `git log --format=%B b9ddc63..7b104d5`. The body says `Issue: #797` only, which is correct — the issue carries out-of-scope and follow-up material beyond its doneWhen, so it stays open |
| U3 | any | P1 | PASS | `Issue: #797` present in the body |
| U4 | any | P1 | PASS | offender grep over `b9ddc63..7b104d5` empty; the new subject `fix(fleet): address review round 2 feedback for git-refs-guard (GH-797)` conforms |
| U6 | any | P0 | PASS | **Exact-head CI now exists and is green** — `gh api …/actions/runs?head_sha=7b104d5…` → `total_count: 1`, `CI status=completed conclusion=success`; `gh pr checks 805` → 9/9 pass including `Test (windows-latest)` 7m1s and `CI Gate`. That closes Round 2's F-5. No CI job covers any file in this diff (`grep -rn 'git-config-guard' .github/ lefthook.yml` → 0 hits), which is the §5.0-U6 coverage-gap reason for the focused local runs below. No recorded gate was rerun, and no workspace Cargo gate was run (zero Rust in the diff) |
| U7 | any | P1 | PASS | the whole scoped delta audit finished before posting. Nothing blocking survived; every later-round observation that was already observable at Round 2 (`lane-stop.ps1:327,356` first-line-only reporting) is routed as `FOLLOW-UP ISSUE` rather than raised as a new blocker, per `loop` item 2 |
| C1 | code-plain | P0 | PASS | doneWhen 1 PASS (cases 11–12 + `lane-launch.ps1:113-116`), 2 PASS (`lane-stop.ps1:392-394`; the `=== EXIT ===` write at `:376` still precedes the `Fail` at `:394`, so GH-672 holds), 3 PASS (cases 13/15 + RAN below), 4 PASS (measurement in the PR body, spot-checked at Round 2), **5 now PASS** — see F-1 |
| C2 | code-plain | P0 | PASS | `fix:` with a **RAN-verified fail-first** regression test. Case 14's new mixed-repairability assertions were executed against the previous head's guard and the new one on identical repos: at `f9edc6a` both `RESTORED ref=refs/heads/master` and `RESTORE FAILED: cannot repair refs: refs/heads/dead` are **absent** (red); at `7b104d5` both are present (green). Transcript under RAN |
| C3 | code-plain | P1 | N.A.(no `crates/**` file in the diff) | `gh pr diff 805 --name-only` lists no Rust |
| C4 | code-plain | P1 | PASS | `unsafe {` / `#[allow(clippy::all)]` / `#![allow` grep over the delta → empty |
| C5 | code-plain | P1 | N.A.(no crate touched, so the Windows CI subset gap does not apply) | the crate selector printed nothing. The Windows-behaviour obligation this rule exists for is still discharged: every RAN below is Windows 11 / pwsh 7 / git 2.49 against the real scripts |
| R1 | code-risk | P0 | PASS | destructive grep over the delta → 1 hit, `test-git-config-guard.sh:31`, which is **prose inside a comment** and executes nothing. Over the full PR diff the two hits are `git-config-guard.ps1:345` `Remove-Item -LiteralPath $targetFile -Force` and `test-git-config-guard.sh:275` `rm -rf "$repo_ref2/.git/logs"`. Triggers, stated: the `Remove-Item` fires only inside `Restore-BrokenRefs`, only **after** `Copy-Item` has archived the same file to `.git/ref.CORRUPT.<sanitized>.<stamp>.bak`, and only for a ref that already resolved a validated `RepairSha`; the `rm -rf` is confined to a `mktemp -d` throwaway repo. The delta neither widens nor narrows that trigger — it is reached for exactly the refs reached before, plus refs a now-continuing loop gets to instead of aborting before |
| R2 | code-risk | `[判斷]` | N.A.(input set empty) | added delta lines containing `\|\|` intersected with `&&` → empty. Not silently treated as RAN |
| R3 | code-risk | P0 | PASS | `sh -n scripts/test-git-config-guard.sh` → `exit=0` (the only changed `.sh`) |
| R4 | code-risk | P1 | PASS | the delta creates no new temp file, lock or handle. The `mktemp -d` / `trap 'e2e_cleanup; rm -rf "$tmp"' 0 HUP INT TERM` release path is now **`b9ddc63`'s**, not a duplicate (F-1). `Get-ScheduledTask -TaskName 'edda-lane-guard-e2e-*'` → **0** after my runs; all scratch removed; `git status --porcelain` empty. `sh scripts/lint-file-length.sh --tree` → `exit=0` despite the guard growing to 524 lines |
| R5 | code-risk | P1 | PASS | the delta interpolates `$($b.Ref)`, `$refName` and `$LASTEXITCODE` into **message strings** only; no new value reaches a command. `git update-ref $refName $sha` is unchanged, and `$sha` is still gated on `^[0-9a-fA-F]{40}$` plus a `git cat-file -e` object probe (`:224`) before use. Two-distinct-inputs-one-filename is held by `New-UniquePath` (`:172-177`) over the `[/\\:] → _` sanitisation — RAN-confirmed on a two-ref restore whose archives share a one-second stamp: `ref.CORRUPT.refs_heads_b2.…` and `ref.CORRUPT.refs_heads_master.…` are distinct files |
| §5.5 | any | — | PASS | see Wiring — 4 rows, no P1 adjudication triggered |

### Wiring

|---|---|---|---|---|
| `CanRepairAny` field on the `Get-RefsHealth` hashtable | `git-config-guard.ps1:282` (fsck-unparseable arm, `$false`), `:313` (`$anyHaveRepairs`) | `:387` `-VerifyRefs`, `:432` `-Verify`, `:457` `-Backup` — three reads, no dead end | loud — the `Fail … 2` it decorates still fires; only the hint text changes | field→hint→`Fail` message→`lane-launch.ps1:116` operator instruction. **Proved by test**: case 14 asserts `partial repair available` on `-Backup` and `-Verify`. RAN-confirmed on a fresh mixed repo for all three arms including `-VerifyRefs`, which the suite does not assert (routed follow-up; not required by doneWhen) |
| `RESTORE FAILED: cannot repair refs: <ref> (<why>)` — new aggregated failure message | `git-config-guard.ps1:358-365` (`$failedRefs` join) → `Fail` at `:412` / `:518` | `test-git-config-guard.sh` case 14 (`grep -q 'RESTORE FAILED: cannot repair refs: refs/heads/dead'`); `lane-stop.ps1:319` via exit code | loud — exit 2, and unlike Round 2's gate the refs that **could** be repaired are repaired first, with their `RESTORED ref=` lines preceding it on stdout | guard stdout/stderr→`lane-stop.ps1:323` `RESTORED ref=` matcher→`gitrefs=` at `:356-359`→durable `=== EXIT ===` at `:376`. RAN-confirmed end to end on the `-VerifyOrRestore` mixed path: stdout carries `config=… healthy` ×1 and `RESTORED ref=` ×1 with exit 2, so lane-stop lands on `gitrefs=RESTORED ref=…; UNREPAIRABLE (git-config-guard exit 2)` — accurate, and the `Fail` still fires |
| `no broken refs identified to restore` — new empty-`BrokenRefs` early return | `git-config-guard.ps1:318-325` | `Fail` at `:412` and `:518`; no test asserts the string | loud — exit 2, unchanged from the `CanRepair` gate it replaces. RAN parity check (worktree-`HEAD` NUL, old vs new guard): both exit 2, both print the same `refs UNHEALTHY: git fsck failed: …` stderr line first | this arm is what keeps the removal of the `-not CanRepair` `Fail` from falling through into a misleading "refs still unhealthy after restoration". Message specificity regressed — **P2-1** |
| `$healthyRefLine` (Round 2 F-3) | **removed** from `lane-stop.ps1` | n/a | n/a | `grep -rn 'healthyRefLine' scripts/` → 0 hits. The dead write end is gone, not re-homed |

No other new `pub` fn / field / CLI flag / config key / event payload field / written file or side-file is added by this delta. `sh scripts/wiring-scan.sh b9ddc63 7b104d5` → `no new pub surfaces`; its single swallow-pattern hit is a comment line, not code.

### Round 2 blocking findings — RAN confirmation

| Round 2 finding | Status | Evidence (RAN at `7b104d5`, Windows 11 / pwsh 7 / `git 2.49.0.windows.1`) |
|---|---|---|
| F-1 (P1) — branch does not integrate with the current base; duplicates the GH-786 teardown that landed as #803 | **RESOLVED, and re-verified against a base that has moved again** | `git merge-base origin/main 7b104d5` → `b9ddc63` (= #803). `git diff b9ddc63:scripts/test-git-config-guard.sh 7b104d5:scripts/test-git-config-guard.sh` shows the branch now **carries `b9ddc63`'s implementation** rather than its own: `logdir="$tmp/lanes"` is hoisted above the trap, `e2e_cleanup` uses `$logdir`, and the `drop_fake_lane` word-loop no longer appears as a branch change because `main` already owns it. `main` has since advanced to `2433a3f`; `git merge-tree --write-tree origin/main 7b104d5` → **exit 0**, tree `bac2635`. doneWhen 5's teardown obligation is satisfied by `main`'s code, not by a second copy of it |
| F-2 (P1) — `CanRepair` turned an advisory signal into an all-or-nothing gate, leaving a recoverable ref destroyed while the guard printed the repair SHA it declined to use | **RESOLVED** | Identical throwaway repos (branch `master` reflog intact, branch `dead` with `.git/logs/refs/heads/dead` removed, both ref files NUL-ed), the same scenario run against both guards. **`f9edc6a`**: `exit=2`, `cannot repair refs automatically: …`; afterwards `git rev-parse refs/heads/master` → `warning: ignoring broken ref` + `fatal: ambiguous argument`, `git log` → `fatal: your current branch appears to be broken`, and **no** `ref.CORRUPT.*.bak` was written. **`7b104d5`**: `exit=2`, `RESTORED ref=refs/heads/master to sha=ee9102c… ; corrupt file preserved at …ref.CORRUPT.refs_heads_master.…bak`, then `RESTORE FAILED: cannot repair refs: refs/heads/dead (no repair SHA available)`; afterwards `git rev-parse refs/heads/master` → `ee9102c…` (= expected) and `git log --oneline -1` → `ee9102c c1`. The recoverable ref is recovered; the unrecoverable one is named. The self-contradictory advisory is gone too: on a *genuinely* mixed set `-Backup`, `-Verify` and `-VerifyRefs` all now say `partial repair available via … -RestoreRefs; unrepairable refs require manual inspection of .git`, while on a set where nothing is repairable they still correctly say `automatic … unavailable` |
| F-3 (P2) — `$healthyRefLine` computed and never read | **RESOLVED** | `grep -rn 'healthyRefLine' scripts/` → 0 hits |
| F-4 (P2) — the one test on the unrepairable path was widened, so nothing distinguished the new gate from the old behaviour | **RESOLVED** | case 14 is back to strict `grep -q 'RESTORE FAILED'` (`:280`) and gains a mixed-repairability scenario (`:283-318`). Fail-first proven above: its two new assertions are red at `f9edc6a`, green at `7b104d5`. `sh scripts/test-git-config-guard.sh` → **all 15 offline cases pass**, confirming the PR's 15/15 claim |
| F-5 (P2) — no CI run at the reviewed head | **RESOLVED** | run `33821763334` at `head_sha=7b104d5…`, `conclusion=success`, 9/9 checks pass |

### Findings

- **[P2] P2-1 — removing the `CanRepair` early `Fail` dropped the underlying `git fsck` reason out of the terminal failure message on the no-parsable-broken-refs path.** When `fsck` fails for a reason that yields no `refs/…` line — the known NUL-ed `.git/worktrees/<wt>/HEAD` shape — `BrokenRefs` is empty and the new `:318-325` early return makes the last word `RESTORE FAILED: no broken refs identified to restore`, where `f9edc6a` said `cannot repair refs automatically: git fsck failed: fatal: not a git repository: …/worktrees/child`. — evidence: RAN, identical linked-worktree repos, `-RestoreRefs`: old → `exit=2` + `cannot repair refs automatically: git fsck failed: fatal: not a git repository: …/wt-OLD/.git/worktrees/child-OLD`; new → `exit=2` + `RESTORE FAILED: no broken refs identified to restore`. Non-blocking because the parity that matters holds: the exit code is unchanged, `lane-stop`'s verdict is unchanged, and the specific reason is still printed on the line immediately above by `[Console]::Error.WriteLine("git-config-guard: refs UNHEALTHY: $($refsHealth.Reason)")` (`git-config-guard.ps1:406`, `:512`). Carrying `$refsHealth.Reason` into the `:322` `Reason` string restores it in one line.

- **[P2] P2-2 — case 14's on-disk restoration assertion is vacuous as written.** `test-git-config-guard.sh:311` is `git -C "$repo_mixed" cat-file -e "$head_sha"`, which probes that the **commit object** exists — and it always did; neither the corruption nor the repair touches it. That line passes even if the ref were never restored. The case still has real teeth through `grep -q "RESTORED ref=$head_ref"` (`:307`) and the forensic-archive count (`:313-316`), both RAN-confirmed red against `f9edc6a`, so F-4's regression genuinely is pinned — but the Review Response's claim that it "Verifies the repairable branch is restored on disk (`git cat-file -e`)" overstates what that line proves. `[ "$(git -C "$repo_mixed" rev-parse "$head_ref")" = "$head_sha" ]` is the check the sentence describes.

- **[P2] P2-3 — whitespace-only churn inside `b9ddc63`'s code.** `scripts/test-git-config-guard.sh:40` adds a blank line inside `e2e_cleanup`, a function this branch otherwise correctly adopted verbatim from #803. It is the only remaining non-additive difference from `main` in that function; dropping it would make the F-1 resolution a pure adoption.

### FOLLOW-UP ISSUE

- **`lane-stop`'s durable record names only the first restored ref, and never names the unrepairable one.** `lane-stop.ps1:327` and `:356` take `$restoredRefLine[0]`, so a multi-ref restore reaches `=== EXIT ===` (`:376`) and `gitrefs=` (`:392`) as one line out of N; and `RESTORE FAILED: cannot repair refs: refs/heads/dead` never reaches the durable record at all, because the guard writes it with `[Console]::Error.WriteLine` and `$guardOut = & … 2>&1` (`:319`) does not capture that stream (Round 2 RAN-confirmed). The operator sees it on lane-stop's inherited stderr; the log does not. Both shapes were observable at Round 2 (`:356-359` was audited then), and doneWhen 2 only requires "the same line-shape as its existing `gitconfig=…` verdict", which is met — so this routes here rather than blocking, per `loop` item 2. The F-2 fix does make multi-ref restore materially more reachable, which is why it is worth an issue now. RAN evidence: two-repairable-ref repo → two `RESTORED ref=` lines on stdout, exit 0; mixed repo → one `RESTORED ref=` plus exit 2.

- **Non-branch members of the same corruption class remain detected-but-unrepairable and fail closed fleet-wide** (carried from Rounds 1–2, re-confirmed this round): a NUL-ed `.git/worktrees/<wt>/HEAD` yields an fsck error that `Get-RefsHealth`'s `refs/…` regex (`:248`) and its `refs\heads` loose scan (`:257`) both miss, so `BrokenRefs` is empty and every lane against that `.git` is blocked with no self-service repair. Outside #797's doneWhen; not extended by this PR. P2-1 above is the message half of the same gap.

- **`docs/guides/operator-runbook.md` still describes the guard as a `.git/config` check only** (carried): `:62-63` document `-Verify` / `-Restore`; `:186` describes the lane-stop post-kill check purely in `.git/config` terms. `grep -n 'VerifyRefs\|RestoreRefs' docs/guides/operator-runbook.md` → 0 hits, so `-VerifyRefs`, `-RestoreRefs` and the new `partial repair available` hint appear nowhere an operator would look.

- **No CI job or hook runs `scripts/test-git-config-guard.sh`** (carried): `grep -rn 'git-config-guard' .github/ lefthook.yml` → **0 hits**. This is why the green exact-head CI above, while required, proves nothing about this diff, and why every product claim in this review rests on local RAN.

- **`lane-stop.ps1` still renders guard exit 1 (usage / not-a-git-repo) as `UNREPAIRABLE`** (carried): `:336-342`. Pre-existing and untouched by this delta.

- **`edda --version` on this workstation cannot satisfy the base spec's new §5.0 U6 addendum.** The base copy of `REVIEW.md` at `2433a3f` carries a paragraph — added by `99bb1db` (GH-746, PR #804) — requiring a reviewer to confirm `edda --version` prints the checked-out build identity (commit + date), or `unknown`, before claiming `RAN on this workstation`. Here it prints `edda 0.4.0`: neither. The cause is benign and outside this PR — GH-746 is **not in this branch** (merge-base `b9ddc63` predates `99bb1db`), and the `edda` on PATH is an installed release binary, not built from any worktree. It costs this review nothing: the diff contains zero Rust and **no RAN claim above depends on the `edda` binary** — every check ran `git`, `sh`, `pwsh`, or the PR's own scripts by absolute path. Worth an issue so the addendum becomes satisfiable once the workstation binary is rebuilt from a post-GH-746 tree.

### RAN vs READ

**RAN** — Windows 11, Git Bash + pwsh 7, `git 2.49.0.windows.1`, on the frozen head `7b104d53d6b01f305ec5df01475cdd18cf803096`. Justification per §5.0-U6: no CI job covers any file in this diff (0 hits across `.github/` and `lefthook.yml`), so exact-head CI — though green — is a coverage gap for this change. Each run below is a focused check for that gap; no recorded gate was rerun.

- **Fail-first comparison, the core of this round.** `git show f9edc6a:scripts/fleet/git-config-guard.ps1` extracted to scratch and run side by side with the head guard on freshly built, identical repos. Mixed set (`master` repairable, `dead` reflog removed, both NUL-ed): old → no `RESTORED ref=` line, no archive, `git log` → `fatal: your current branch appears to be broken`; new → `master` restored to `ee9102c…` and archived, `dead` named in `RESTORE FAILED`. Case 14's two new assertions are therefore red without the change.
- `sh scripts/test-git-config-guard.sh` → **all 15 offline cases passed**, exit 0. E2E cases 16–20 were not run; `Get-ScheduledTask -TaskName 'edda-lane-guard-e2e-*'` → 0 afterwards, so no teardown leak.
- Six throwaway-repo scenarios driving the real guard: mixed repairability (against both guards), linked-worktree `HEAD` NUL parity (against both guards), `-VerifyOrRestore` on a mixed set with `lane-stop.ps1:321-359`'s own matchers applied to the captured stdout, a two-repairable-ref multi-restore exercising the same-second archive-collision path, and fresh-mixed advisory-hint checks across `-Backup` / `-Verify` / `-VerifyRefs`.
- `git merge-tree --write-tree origin/main 7b104d5` → exit 0. `git merge-base origin/main 7b104d5` → `b9ddc63`. `git diff b9ddc63:scripts/test-git-config-guard.sh 7b104d5:…` read in full to confirm the F-1 resolution adopts #803 rather than re-implementing it.
- `sh -n scripts/test-git-config-guard.sh` → exit 0. `sh scripts/lint-markdown-content.sh` → exit 0. `sh scripts/lint-file-length.sh --tree` → exit 0. `sh scripts/wiring-scan.sh b9ddc63 7b104d5` → no new pub surfaces.
- `gh pr checks 805`; `gh api …/actions/runs?head_sha=7b104d5…`.
- All scratch removed. `git status --porcelain` is empty and HEAD is still `7b104d53d6b01f305ec5df01475cdd18cf803096` — nothing was edited, committed, pushed or merged.

**READ:**

- **Exact-head CI**: run `33821763334` at `7b104d5…`, `conclusion=success`, 9/9 — `Format`, `Clippy` ×3, `Test` ×3 (including `windows-latest`, 7m1s), `Detect code changes`, `CI Gate`. What it covers: that the Rust workspace still builds and tests on all three OSes at this head. What it does **not** cover: any file in this diff.
- No L1 gate receipt in the PR body or commit message, and none is needed: zero Rust, no `Cargo.*`, no toolchain change, so the Cargo gate set cannot be affected. I did not run the workspace Cargo gates and had no reason to.
- Round 2's verdict comment (F-1…F-5) and the `Review Response: Round 2` comment; issue #797's body and doneWhen 1–5; `REVIEW.md` **at the base SHA `2433a3f`** (`git diff b9ddc63:REVIEW.md origin/main:REVIEW.md` read in full — one added §5.0 paragraph, honoured and reported above); PR #803 / commit `b9ddc63` as the change F-1 required this branch to adopt.

### Verdict

LGTM (P0=0, P1=0) — both Round 2 blockers are RAN-confirmed resolved: the branch now merges cleanly into a base that has moved twice since Round 1 and carries #803's teardown rather than a duplicate of it, and the `CanRepair` over-correction is replaced by genuine per-ref best-effort recovery that restores what it can, archives the corrupt file, and names what it cannot — proven red against the previous head and green here. All five of #797's doneWhen items are met against real code and a fail-first test. Three P2s and six follow-ups are recorded; none blocks. Merge remains the operator's action.

Issue #810, opened 2026-09-03
chore(repo): no [profile.dev] and no .cargo/config.toml — 8.1 GB of debug symbols across 84 executables, and the workstation lane tool is blocked on this


fagemx:
The workspace tunes `[profile.release]` only and has no `.cargo/config.toml`. Two thirds of a measured 40.9 GB `target/debug` is avoidable, and `fleet-workstation`'s lane-tool plan is explicitly blocked on this being fixed first.

## What happened

Measured on `main` `cbbc8b0`, 2026-09-04:

- Root `Cargo.toml` carries `[profile.release]` (`lto`, `codegen-units`, `strip`) and **no `[profile.dev]`** — so debug builds use the default `debug = 2`, full symbols.
- **No `.cargo/config.toml` exists** anywhere in the tree.
- `.claude/CLAUDE.md` "Build lanes" records the footprint of a long-lived `target/debug`: **40.9 GB across 88,789 files** — `incremental` 21.6 GB, `.rlib`+`.rmeta` 10.3 GB, **`.pdb` 8.1 GB**, `.exe` 0.9 GB. `edda.pdb` alone is 310 MB against a 51 MB `edda.exe`. There are 84 test and binary executables, each statically linked and each carrying full symbols.

`fagemx/fleet-workstation`, `docs/superpowers/plans/2026-08-18-verification-cost-discipline-pr2-lane-tool.md:20`, states the dependency in its own words:

> Footprint work comes first and is not part of this PR. … 8.1 GB was Windows debug symbols (`edda.pdb` 310 MB against a 51 MB `edda.exe`), because the edda workspace tunes `[profile.release]` only. **Reclaiming by age and tuning the debug profile are the difference between a ~40 GB and a ~10 GB steady state**, and neither needs this tool. Sizing a ceiling before they land enforces an inflated number.

So the lane tool cannot ship a defensible pool ceiling until the debug profile is tuned here. That plan owns the reclaim half; this issue owns the profile half.

## Why it matters

Beyond disk: on Windows MSVC, generating and linking debug info is a large fraction of link time, and this workspace links 84 executables. `verification.l1-owner` exists because three of five fix lanes died at a 45-minute timeout *after finishing the work but before pushing* — link time is directly on that path. The `.claude/CLAUDE.md` note that "any lane pool ceiling stated before that footprint is fixed is provisional" is the same conclusion from the other side.

[...]

Issue #809, opened 2026-09-03
chore(repo): no rust-toolchain.toml and no rust-version — CI runs floating stable under -Dwarnings, and local (1.93.1) already differs from CI


fagemx:
The workspace has no `rust-toolchain.toml` and no `rust-version`, CI installs floating `stable`, and the merge gate is `-D warnings`. Those three together mean a Rust release can turn `main` red with no commit, and local and CI are already on different compilers.

## What happened

Measured on `main` `cbbc8b0`, 2026-09-04.

- `.github/workflows/ci.yml:100,124,163` — all three jobs run `rustup toolchain install stable --profile minimal`. Floating: whatever stable is on the day the job runs.
- This workstation: `rustc 1.93.1 (2026-02-11)`.
- `xai-org/grok-build` pins `1.94.0` in its `rust-toolchain.toml`, so stable is at or past 1.94. **Local and CI are therefore already on different compilers**, and nothing reports that.
- Root `Cargo.toml` declares `edition = "2021"` and **no `rust-version`** — the workspace has no declared MSRV.
- `.github/workflows/ci.yml:82` — the path filter that decides whether a change is build-affecting already lists `rust-toolchain*`. The workflow anticipates a file that does not exist.

## Why it matters

**1. A red `main` with no commit.** `RUSTFLAGS: -Dwarnings` plus `cargo clippy --workspace --all-targets` means every new or widened clippy lint in a stable release is an immediate build break, on a tree nobody touched. The blast radius is every open PR at once, and the failure looks like the PR's fault. grok-build's own toolchain file documents exactly this discipline — bump one point version at a time, wait a couple of weeks after release, then run check and clippy — which is only possible because the version is pinned somewhere to bump.

**2. Silent local/CI divergence.** An L1 gate receipt says "clippy clean" under 1.93.1 while CI runs something newer. Under `verification.l1-owner` the receipt is load-bearing precisely for the crates outside the Windows CI subset — the ones CI never runs. A receipt from a different compiler than the one that will judge the code is weaker evidence than it appears, and today nothing in the receipt records which compiler produced it.

**3. No MSRV for consumers.** GH-648 publishes the workspace to crates.io and `product.direction` makes open source the distribution channel. Without `rust-version`, a downstream user on an older toolchain gets a wall of type errors instead of Cargo's one-line "requires rustc X". GH-611's SDK consumers hit the same wall.

## Suspected surface

[...]

Issue #757, opened 2026-09-03
test(bridge-claude): env-var tests race under parallel execution — failure set differs run to run


fagemx:
## Context

Found during PR #738's L1 gate. `cargo test -p edda-bridge-claude` fails with a **different set** of failures on every run (observed: 3–23 of `agent_phase`/`bg_detect`/`bg_digest`/`bg_scan`/`dispatch`/`render`/`signals`/`state` tests, verified on `main` @ `df3bd4d` with no local changes): the tests mutate process-global env vars (`set_var` in `digest/tests.rs`, `dispatch/tests.rs`, …) while Cargo runs them on parallel threads. Two tests touching the same env var race; scheduling decides who wins.

## Direction

- Prefer reading config from an injected parameter over `std::env::set_var` in tests; or
- serialize env-mutating tests behind a static Mutex; or
- use the unstable `#\[serial_test\]`-style crate (adds a dep — prefer the Mutex).

## Acceptance

- [ ] `cargo test -p edda-bridge-claude` passes 20 consecutive runs locally
- [ ] No test reads an env var another test in the same binary mutates, or all such tests share one lock

fagemx:
## The stated hypothesis does not hold — measured on `fix/gh646-hermetic-store-tests` (after GH-646)

### Reproduction (fail-first)

`cargo test -p edda-bridge-claude` at **default parallelism** (32-thread host), three consecutive runs:

| run | result |
|---|---|
| 1 | 621 passed; **13 failed** |
| 2 | 630 passed; **4 failed** |
| 3 | 629 passed; **5 failed** |

Same tree at `--test-threads=4` (the CI Windows runner shape): **634 passed; 0 failed**, in both home-directory states. So the flake is real, parallelism-dependent, and reproduces exactly as this issue describes.

### But the mechanism is not env-var mutation

This issue's Direction section says the tests race on `std::env::set_var`. Audited every `#[test]` in the crate whose body calls `set_var`/`remove_var`:

- **11** such tests, **11** of them already acquire the lock (via `env_guard()` / `with_env_guard` / `ENV_LOCK`)
- **0** unguarded

And none of the observed failures is one of those tests. A representative failing set:

```
agent_phase::tests::heartbeat_age_secs_reads_through_the_sanitized_funnel
bg_detect::tests::increment_session_count_works
bg_digest::tests::should_run_returns_false_when_already_digested
bg_scan::tests::draft_storage_roundtrip
```

[...]

PR #811, opened 2026-09-03
chore(repo): halve the debug build footprint locally and cut Windows link time 12-21% — [profile.dev] line-tables-only plus rust-lld

fagemx:
Issue: #810

Two build-configuration changes, both measured on this Windows workstation rather than adopted on reputation. One hypothesis in the issue held and one did not; both are reported.

## What changed

- Root `Cargo.toml` — a `[profile.dev]` section setting `debug = "line-tables-only"`. The workspace previously tuned `[profile.release]` only, so debug builds carried full symbols.
- New `.cargo/config.toml` — the workspace had none. It sets `rust-lld` as the linker for `x86_64-pc-windows-msvc` only; other targets are untouched.

## Measurements

All runs: clean `target/`, `CARGO_INCREMENTAL=0`, `cargo build --workspace --all-targets`, base `f0750ad`, rustc 1.93.1, this workstation.

| run | profile | linker | wall | target | `.pdb` | `edda.pdb` |
|---|---|---|---:|---:|---:|---:|
| baseline | `debug = 2` (default) | MSVC `link.exe` | 187s | 4.14 GB | 2.64 GB | 312 MB |
| A | `line-tables-only` | MSVC `link.exe` | 193s | **2.07 GB** | 1.04 GB | 141 MB |
| B | `line-tables-only` | `rust-lld` (RUSTFLAGS) | 149s | 2.08 GB | — | — |
| C | `line-tables-only` | `rust-lld` (committed config) | **153s** | 2.07 GB | — | — |

Baseline and A have identical artifact counts (3330 files, 143 `.pdb`, 120 `.exe`, 330 `.rlib`), so the two totals are directly comparable.

**Disk: −50% overall, −61% on debug symbols.** Debug info was 64% of a clean build, because all 120 executables link statically and each carried full symbols.

**Build time: the debug-info hypothesis failed, and CI never tested it anyway.** The issue predicted that cutting symbol volume would cut link time. It did not — 187s → 193s, inside the noise band.

Round 1 found the more important half, which this PR had wrong: **`ci.yml:16-17` has set `CARGO_PROFILE_DEV_DEBUG: line-tables-only` and `CARGO_PROFILE_TEST_DEBUG: line-tables-only` since #411 (`cd63a0c`)**, and environment settings beat the manifest. So on CI the new `[profile.dev]` key is inert, the green `Test (windows-latest)` validates only the linker half, and this body's original premise — "debug builds carried full symbols" — was false for CI. It was only ever true locally.

That also means the finding was not new. The comment above those two CI lines already says the quiet part: *"skips the heavy PDB generation and linking that dominate Windows build time"*. The conclusion this PR reports as a correction was recorded in `ci.yml` ten months ago.

What remains genuinely new is narrower, and is the part that matters for the problem #810 was opened about: **the manifest extends this to local and lane builds**, which the CI-only environment variables never touched, and that is exactly where the measured multi-gigabyte target directories live. The `[profile.dev]` change is worth making on that ground alone. A follow-up should decide whether the now-redundant CI environment variables should go.

One negative result worth recording so nobody repeats it: pointing `-Clinker` at `lib/rustlib/<target>/bin/gcc-ld/lld-link.exe` fails. That binary is the gcc-driver shim; it ignores the `-flavor link` rustc passes and then tries to open a file literally named `link`. The usable binary is `lib/rustlib/<target>/bin/rust-lld.exe`.

## Backtrace quality — verified, not assumed

`line-tables-only` drops the type and variable information a debugger uses, so the risk was losing usable test failures. A deliberately panicking test under `RUST_BACKTRACE=1` still reports its panic site and a fully symbolicated stack with `file:line` on every frame:

```
panicked at crates\edda-core\tests\tmp_backtrace_probe.rs:4:14
   5: tmp_backtrace_probe::probe_panics_to_show_backtrace_quality
             at .\tests\tmp_backtrace_probe.rs:4
```

The probe file was removed before committing.

## Gate receipt — not green, and the reason is not this change

Build-affecting change (`Cargo.toml`), so full L1. Clean tree, `CARGO_INCREMENTAL=0`, rustc 1.93.1, no build lane (solo work, worktree default `target/`).

| gate | result |
|---|---|
| `cargo fmt --all --check` | pass (4s) |
| `cargo clippy --workspace --all-targets -- -D warnings` | pass (92s) |
| `cargo test --workspace` | **pass — 2680 passed, 0 failed, 56 suites** (304s) |

The first run of this gate reported `1261 passed, 1 failed` and was **incomplete**: `cargo test` fails fast, so it stopped after 15 of 56 suites. A complete re-run at `0d1da1c` is green (2680 passed, 0 failed, 56 suites, 304s). See the correction comment for the full account.

The one failure in that first run was `bg_scan::tests::draft_storage_roundtrip` — **#757**, an open P1 flake, unassigned, whose body already lists `bg_scan` among the affected modules. It is non-deterministic on the same binary at the same SHA (run 1 failed, run 2 `ok. 635 passed`), passes in isolation and with its whole module, and its mechanism is `store_root()` re-reading `EDDA_STORE_ROOT` while `digest/tests.rs` redirects it to a `TempDir` that is then dropped. A debug-info setting cannot affect that. A sharper reader-side diagnosis is posted on #757.

## Risk to check in CI

The linker line is verified locally on rustc 1.93.1. CI installs floating `stable` (currently 1.98.1 — see #809), so **CI is the check that a bare `rust-lld.exe` resolves there too**. If either Windows job fails to link, the fix is to drop the `linker` line and keep `.cargo/config.toml` with its comment; the profile change stands on its own and is where the disk win lives.

Not addressed here: the 21.6 GB of orphan `incremental` session directories in long-lived lane target dirs. That is reclamation, owned by the fleet-workstation lane tool, and needs no change in this repo.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,9 +2,17 @@
 # and every CI runner, not just this workstation.
 
 [target.x86_64-pc-windows-msvc]
-# rust-lld ships inside the toolchain, so no external install is required.
-# Measured on this workstation at f0750ad (clean target, CARGO_INCREMENTAL=0,
-# debug = "line-tables-only"): a full `--workspace --all-targets` build takes
-# 149s with rust-lld against 193s with the default MSVC link.exe, for the same
-# 2.07 GB of output. See GH-810.
+# rust-lld ships inside the toolchain, so no external install is required, and
+# a bare name resolves because rustc adds the sysroot target `bin` directory.
+#
+# Measured on this workstation (clean target, CARGO_INCREMENTAL=0,
+# debug = "line-tables-only", `cargo build --workspace --all-targets`):
+# 153s with rust-lld, against an MSVC link.exe arm whose three runs spanned
+# 173-193s. So the honest range is -12% to -21%, not a single figure -- the
+# direction is solid because both rust-lld runs (149s via -Clinker, 153s via
+# this file) sit below that arm's floor, but the magnitude is a range.
+#
+# Consumers beyond `cargo build`: `cargo test`, and release.yml's tagged
+# Windows binary, which is built with --release and is first exercised at tag
+# time. See GH-810.
 linker = "rust-lld.exe"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,10 @@
+# Shared build configuration. Committed, so it must hold for every contributor
+# and every CI runner, not just this workstation.
+
+[target.x86_64-pc-windows-msvc]
+# rust-lld ships inside the toolchain, so no external install is required.
+# Measured on this workstation at f0750ad (clean target, CARGO_INCREMENTAL=0,
+# debug = "line-tables-only"): a full `--workspace --all-targets` build takes
+# 149s with rust-lld against 193s with the default MSVC link.exe, for the same
+# 2.07 GB of output. See GH-810.
+linker = "rust-lld.exe"
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -226,7 +226,8 @@ across 88,789 files:
 Two thirds of that is avoidable rather than intrinsic. Every one of the 509
 incremental session directories was older than 24 hours — orphans left by
 interrupted or killed builds, which nothing in Cargo ever reclaims. The
-workspace tunes `[profile.release]` only, so debug builds carry full symbols
+workspace sets `debug = "line-tables-only"` for dev builds (GH-810); before
+that it tuned `[profile.release]` only and debug builds carried full symbols
 into every statically linked test binary.
 
 **Any lane pool ceiling stated before that footprint is fixed is provisional.**
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,3 +64,12 @@ too_many_lines = "warn"
 lto = true
 codegen-units = 1
 strip = true
+
+[profile.dev]
+# Debug info is 64% of a clean `--all-targets` build here (2.64 GB of 4.14 GB,
+# `edda.pdb` alone 312 MB against a 53 MB `edda.exe`), because every one of the
+# workspace's test and binary executables links statically and carries full
+# symbols. `line-tables-only` keeps file and line numbers in backtraces — what
+# a failing test actually needs — and drops the type and variable information
+# that a debugger would use. Measured before/after in GH-810.
+debug = "line-tables-only"


fagemx:
## Correction: the gate receipt in the body was incomplete, not just flaky. Complete re-run is green.

The body reports `cargo test --workspace` as "1261 passed, 1 failed". That count is not the workspace. **`cargo test` fails fast** — it stops after the first failing test binary — so that run covered **15 of 56 suites** and the remaining 41 never executed. I explained the one failure and did not notice that four fifths of the suite had not run.

It surfaced by comparison: the sibling branch for #808 produced `2680 passed, 0 failed, 56 suites` from the same command, and 1262 against 2680 is not a rounding difference.

**Complete re-run on this branch at `0d1da1c`**, clean tree, `CARGO_INCREMENTAL=0`, rustc 1.93.1:

```
cargo test --workspace   exit=0   304s
2680 passed, 0 failed, 56 suites
```

Revised receipt:

| gate | result |
|---|---|
| `cargo fmt --all --check` | pass (4s) |
| `cargo clippy --workspace --all-targets -- -D warnings` | pass (92s) |
| `cargo test --workspace` | **pass — 2680 passed, 0 failed, 56 suites** (304s) |

`#757` did not fire this time. It is probabilistic and remains open and unassigned; a green run is not evidence it is fixed, and the reader-side diagnosis posted there still stands.

Two consequences:

- The doneWhen item "`cargo test --workspace` passes" is now met, so every doneWhen on #810 is delivered and the body carries a closing keyword.
- The general point is worth more than this PR: a failing `cargo test --workspace` receipt reports a **partial** result, and "N passed, 1 failed" reads as though the suite ran. Under #757's non-determinism that is how an incomplete gate gets recorded as a complete one. `--no-fail-fast` would make the count honest.

fagemx:
## Code Review: Round 1 — PR #811 @ 0d1da1ce28e4ac5868d84b73d65ad1a3e245d7df

- model_requested: not named in the brief
- model_observed: unverified (no session record readable from here; a system-prompt string is self-labelling, which §6 excludes)
- spec: review-spec-v1
- escalations: none — the two `[判斷]` rules are D5 (not routed for this class) and R2 (N.A., no shell in the diff)
- cost: ~20 tool calls, no workspace build. RAN two scratch-crate probes (hello-world, seconds each) plus the mechanical greps; everything else was READ.

### IN SCOPE

Changed behaviour: the debug-info density of every local build (`Cargo.toml:68-75`) and the linker used for every `x86_64-pc-windows-msvc` build, dev **and** release (`.cargo/config.toml:4-10`). Direct consumers: `.github/workflows/ci.yml` (which already sets the same debug value), `.github/workflows/release.yml:157,176` (the shipped Windows binary), `.claude/CLAUDE.md` §"Footprint and thresholds" (which documents the config being changed), and every contributor's `$CARGO_HOME/config.toml` linker preference, which a committed project config now overrides. Acceptance ceiling: #810 `doneWhen`. Integration: base `f0750ad`, `main` has moved on; `mergeable=MERGEABLE`, no textual conflict on either file.

The engineering here is real and I want to say so before the findings: the disk half is a clean A/B (identical artifact counts, 3330 files / 143 `.pdb` / 120 `.exe` / 330 `.rlib`, so 4.14 GB → 2.07 GB is directly comparable and needs no repetition — disk is deterministic). The failed prediction is reported rather than buried. The `gcc-ld/lld-link.exe` negative result is recorded so nobody repeats it. The incomplete-receipt correction comment is the right response to the right problem, and the `--no-fail-fast` observation in it is more valuable than this PR. The decision is in the ledger and active. Findings below are about how the result is *stated* and *reconciled*, not about whether the change works.

### Rules

|---|---|---|---|---|
| U1 surface | any | P0 | PASS | `.cargo/config.toml`, `Cargo.toml` — both named in #810 "Suspected surface" |
| U2 no closing keywords | any | P1 | **FAIL** | `gh pr view 811 --json body` → `1:Closes #810` |
| U3 `Issue: #N` line | any | P1 | **FAIL** | `awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → empty |
| U4 conventional subjects | any | P1 | PASS | both subjects `chore(repo): …`; grep for offenders empty |
| U5 markdown lint | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → `exit=0` |
| U6 gates READ before RAN | any | P0 | PASS | 9/9 checks pass incl. `Test (windows-latest)` 9m41s; L1 receipt READ (corrected: 2680 passed / 0 failed / 56 suites) |
| U7 round complete | any | P1 | PASS | whole scoped audit finished before posting; all blockers batched here |
| C1 doneWhen vs real code | code-plain | P0 | PASS | all six items met — see below |
| C2 feat/fix test rule | code-plain | P0 | N.A.(both commits are `chore:`; config-only, no code path to regress) |
| C3 unwrap/expect | code-plain | P1 | PASS | `git diff origin/main...$SHA -- 'crates/**'` → empty (no crate files touched) |
| C4 unsafe / blanket allow | code-plain | P1 | PASS | grep empty |
| C5 Windows coverage gap | code-plain | P1 | PASS | selector prints nothing (no `crates/` paths). Coverage is better than the selector implies: the L1 receipt's `cargo test --workspace` linked all 56 suites with rust-lld locally, and CI's Windows job links the 7-crate subset plus `edda` |
| R1 destructive commands | code-risk | P0 | PASS | grep empty |
| R2 shell precedence | code-risk | `[判斷]` | N.A.(no added line mixes `\|\|` and `&&`; grep empty) |
| R3 shell scripts parse | code-risk | P0 | N.A.(no `.sh`/`.bash` in diff) |
| R4 resource release | code-risk | P1 | N.A.(no script, no temp dir, no lock) |
| R5 boundary, as properties | code-risk | P1 | PASS | the linker value is consumed by cargo/rustc, never interpolated into a shell. Property confirmed: **it fails loudly, it does not silently fall back** — `linker = "definitely-not-a-linker.exe"` → `error: linker not found`, build fails. And `rust-lld.exe` is *not* on PATH here yet resolves, because rustc adds the sysroot target `bin` dir for the link step |
| §5.5 wiring | any | — | filled | table below |

`doneWhen` detail (C1): `[profile.dev]` set and justified against `debug = 0` ✓; before/after target size, `.pdb` size and wall clock all present ✓; panicking-test backtrace with `file:line` demonstrated ✓; `.cargo/config.toml` exists and its linker line carries a link-time measurement ✓; `cargo test --workspace` passes ✓ (via the correction comment); full L1 receipt ✓. **No P0.** But see F4 — the item "before and after numbers … on this Windows workstation" was written by the same session that wrote the code, and it sets no repetition or dispersion requirement for a wall-clock claim, which is the one quantity in the set that is not deterministic. The doneWhen is satisfiable by exactly the single-run-per-arm comparison that produced F4.

### Wiring

|---|---|---|---|---|
| `[profile.dev] debug` (`Cargo.toml:68,75`) | cargo manifest profile key, string `"line-tables-only"` | cargo profile resolution → rustc `-Cdebuginfo=line-tables-only`. **Local builds only** — `.github/workflows/ci.yml:16-17` sets `CARGO_PROFILE_DEV_DEBUG`/`_TEST_DEBUG` at workflow level and config/env beats the manifest, so on every CI job this key is inert (see F2) | declarative, no error path | verified to rustc: scratch crate, manifest `debug = 2` → `-C debuginfo=2`; add `CARGO_PROFILE_DEV_DEBUG=line-tables-only` → `-C debuginfo=line-tables-only` |
| `[target.x86_64-pc-windows-msvc] linker` (`.cargo/config.toml:4,10`) | cargo config, `"rust-lld.exe"` | rustc `-Clinker` for that target: CI Windows Clippy/Test jobs, **and `release.yml:176`'s `--release --target x86_64-pc-windows-msvc -p edda`** (see F5) | **hard error, never silent fallback** — verified | verified to the actual link step on 1.93.1, dev and `--release --target … ` with `lto=true, codegen-units=1, strip=true` |

### Findings

- **[P1] F1 — `Closes #810` auto-closes the issue on merge, against this repo's stated convention.** Evidence: `gh pr view 811 --json body | grep -Ein '…close[sd]?…'` → `1:Closes #810`. REVIEW.md §5.0 U2: this repo links with `(#N)` in the title and `Issue: #N` in the body precisely so merging does not close the issue. #810 itself asks for the opposite of an auto-close — "Link back once this lands so the ceiling can be sized against a measured steady state". The correction comment notes "the body carries a closing keyword" as an observation; it is a finding.
- **[P1] F2 — `Issue: #810` line is missing, so the next round loses its acceptance ceiling.** Evidence: `gh pr view 811 --json body --jq .body | awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → empty. `scripts/review-pr.sh` parses exactly this line to load `doneWhen` into the brief; I had to load #810 by hand. F1 and F2 share one fix: replace line 1 with `Issue: #810`.
- **[P1] F3 — `.claude/CLAUDE.md:223-224` becomes false on merge, and the PR does not address it.** Evidence: `.claude/CLAUDE.md:223` reads "workspace tunes `[profile.release]` only, so debug builds carry full symbols into every statically linked test binary." After this diff the workspace tunes both. Spec §5 lists "a contradiction with `.claude/CLAUDE.md`" as P1 outright, and this diff *creates* the contradiction rather than inheriting it. It is not a wish beyond `doneWhen`: that sentence sits three lines above "**Any lane pool ceiling stated before that footprint is fixed is provisional**", which is the exact blocker #810 exists to lift, so the downstream consumer keeps reading the stale claim. Your own ledger note records a second correction to the same paragraph ("CLAUDE.md's 40.9 GB figure is an ACCUMULATED long-lived target dir … a clean full workspace build with all targets is 4.14 GB and 187 seconds"), and that correction is likewise not in the tree. Either update the paragraph here or open and link a follow-up issue — the wiring rule already treats a named follow-up as clearing an equivalent finding. `docs/superpowers/specs/2026-08-18-…-design.md:178` says the same thing but is a dated design record; leave it.
- **[P1] F4 — the committed comment at `.cargo/config.toml:6-9` states a wall-clock comparison that contradicts the ledger decision and the PR's own noise band.** Evidence: the file says a build "takes 149s with rust-lld against 193s with the default MSVC link.exe". The ledger decision `build.dev-profile-and-linker` (active, `edda ask` from the checkout) says "took 193s to 153s, -21%". So the committed file reports −23% by pairing the *RUSTFLAGS proxy* run (149s) against the link.exe arm, while the configuration this file actually establishes measured 153s. Worse, both the file and the body use 193s as *the* link.exe value, and the body discloses two paragraphs earlier that "three runs at this configuration spanned 173–193s" — 193s is the maximum of the arm's own measured band. The direction is well supported and I am not disputing it: {149, 153} sits entirely below the band floor of 173, which is real evidence with no overlap. The *magnitude* is not: it is bounded by −12% (173→153) and −21% (193→153), and the PR reports only the upper bound — in the title, in the body headline, and most durably in the file a future reader consults when deciding whether to keep the linker line. Fix is text: state the band, or state the committed config's 153s against the arm's midpoint, and drop the 149s proxy number from the file since it is not what this config produces.
- **[P1] F5 — the new `[profile.dev] debug` key has no effect on any CI job, and the premise offered for it is false for CI.** Evidence: `.github/workflows/ci.yml:14-17` has set `CARGO_PROFILE_DEV_DEBUG: line-tables-only` and `CARGO_PROFILE_TEST_DEBUG: line-tables-only` at workflow level since `cd63a0c` (#411), with the same stated rationale. Cargo's config/env profile settings override the manifest — measured, not assumed: scratch crate with manifest `[profile.dev] debug = 2` builds as `-C debuginfo=2`; with `CARGO_PROFILE_DEV_DEBUG=line-tables-only` also set, the same manifest builds as `-C debuginfo=line-tables-only`. Three consequences. (a) The body's "The workspace previously tuned `[profile.release]` only, so debug builds carried full symbols" is true of local builds and false of CI, and it is the premise the change is argued from. (b) The green `Test (windows-latest)` is evidence for the linker half only; CI has been building this way since #411, so it validates nothing about the profile half. (c) The repo now carries the setting in two places where the less obvious one silently wins — a later contributor who sets `debug = 2` in `Cargo.toml` to chase a CI-only bug will get `line-tables-only` on CI and no indication why. Reconcile it: drop the two now-redundant `ci.yml` env lines so the manifest is the single source, or say in the `Cargo.toml` comment that CI overrides it. Also worth correcting in #810, whose "Dedupe ran" searched issue titles and so missed the prior art sitting in `ci.yml`.
- **[P2] F6 — the risk section scopes the linker risk to the two CI Windows jobs; `release.yml` is the third consumer and no gate on this PR covers it.** Evidence: `.github/workflows/release.yml:157,176` runs `cargo build --locked --release --target x86_64-pc-windows-msvc -p edda` on `windows-latest`, which this config also redirects to rust-lld, under `[profile.release] lto = true, codegen-units = 1, strip = true`. That job runs only on a `v*` tag, and `create-release` has already created the draft release by the time `build-release` runs, so a first failure surfaces mid-release. I probed the mechanism rather than assume it: that exact combination (`--release --target x86_64-pc-windows-msvc`, lto + codegen-units=1 + strip, `linker = "rust-lld.exe"`) links cleanly on 1.93.1 — on a trivial crate, which is not the 23-crate graph, but it does show the combination is supported. Not a blocker; the body's "if either Windows job fails to link, drop the linker line" should say which jobs, and the shipped binary is the one it omits.
- **[P2] F7 — the two halves are separately revertible in the branch but not after merge.** `b5058f5` is the profile, `0d1da1c` is the linker, which is good structure. `main`'s history (`… (#709)`) shows squash merges, so both land as one commit and the stated fallback is a one-line manual edit, not `git revert`. True and trivial — recording it because the body presents reversibility as a property of the change.

### FOLLOW-UP ISSUE

- `[profile.dev.package."*"]` is unaddressed and unmentioned. #810's own "Suspected surface" asked for it explicitly — "it should consider `[profile.dev.package."*"]` separately from workspace crates — dependency debug info is rarely wanted". Not a `doneWhen` item, so not blocking, but it is the larger remaining lever on the same 2.07 GB and the PR should not leave it silently dropped.
- `split-debuginfo` — same status: named in #810's suspected surface with "measure on Windows before adopting", neither measured nor declined.
- `.github/workflows/ci.yml:14-15` claims trimming debug info "skips the heavy PDB generation and linking that dominate Windows build time". This PR's measurement disproves the linking half (187s → 193s, no change from a 61% cut in debug info) — the body says so in as many words. The comment that motivated #411 should be corrected with this PR's data.
- `REVIEW.md` §3.1's enumerated risk surface lists `Cargo.toml`, `**/Cargo.toml` and `Cargo.lock` but not `.cargo/config.toml`, so the §3 router classes this PR's linker file as `code-plain`. A file that selects the linker for every build on a platform is at least as much "dependency and toolchain surface" as `Cargo.toml`. Out of scope here — `REVIEW.md` is not in this diff.

### RAN vs READ

RAN:
- `gh pr view/diff/checks 811`, `gh issue view 810`, `gh pr view 814`; `git fetch origin main`; `git merge-base` → base `f0750ad`. Note for the next reviewer: the branch is behind `main`, so `git diff origin/main..$SHA` returns 53 files; the three-dot form `origin/main...$SHA` returns the real 2 files / +19 / −0.
- REVIEW.md §3 classifier on the file list → `classes=code-risk code-plain`, `canonical_class=code-risk`.
- U2, U3, U4, U5, C3, C4, C5, R1, R2 greps as written in the spec (outputs in the table).
- `edda ask build.dev-profile-and-linker` from the checkout → active, quoted in F4.
- Two scratch-crate probes (hello-world, no deps, seconds — not a workspace build). Reason for running rather than reading: neither the L1 receipt nor exact-head CI answers either question. (1) Manifest-vs-env profile precedence, which decides whether the key this PR adds is live on CI → env wins (F5). (2) `rust-lld.exe` resolution given it is not on PATH, and whether the release-profile shape links → resolves, links dev and release, and hard-errors on a bad linker name rather than falling back (R5, F6).

READ:
- L1 gate receipt at `0d1da1c`, as corrected by the 2026-09-03 comment: `cargo fmt --all --check` pass 4s; `cargo clippy --workspace --all-targets -- -D warnings` pass 92s; `cargo test --workspace` pass, 2680 passed / 0 failed / 56 suites, 304s; clean tree, `CARGO_INCREMENTAL=0`, rustc 1.93.1, no lane. The original receipt was a fail-fast partial (15 of 56 suites) and the author caught and corrected it; I am relying on the corrected run.
- Exact-head CI: 9/9 pass — Format, CI Gate, Detect, Clippy ×3 OSes, Test ×3 OSes including `Test (windows-latest)` 9m41s, which is the real evidence that a bare `rust-lld.exe` resolves on CI's floating stable. What it does not cover: the release build path (F6), and anything about the profile half, which CI already had (F5).
- No full local workspace rerun. Nothing here gave grounds to distrust the corrected receipt, CI is green on three platforms, and re-deriving it would be the over-verification the ladder exists to prevent.
- Checked and found *not* to be problems, so they are not findings: `.cargo/*` is already in `ci.yml`'s code-detection whitelist (line 81), so a future `.cargo/config.toml`-only PR still triggers clippy and test — the path filter was pre-provisioned. `line-tables-only` costs debugger locals and expression evaluation, but this repo ships no `.vscode/`, no `launch.json`, and no lldb/gdb/WinDbg workflow, so it degrades nothing committed here; `RUST_BACKTRACE=full` adds frames, not variables, so the probe with `=1` is the right evidence and the `Cargo.toml:71-73` comment describes the trade-off accurately rather than overstating it. The disk claim needs no repetition — artifact counts are identical across the two arms and disk is deterministic.

### Verdict

Changes Requested, P0=0, P1=5 — the change itself is sound and measured, but two repo conventions are broken (`Closes #810`, missing `Issue:` line), the committed linker comment states a −23% best-vs-worst pairing that contradicts the ledger decision's own −21% and the PR's disclosed 173–193s band, the new profile key is inert on CI because `ci.yml` has set it since #411, and the diff leaves `.claude/CLAUDE.md:223` asserting the opposite of what it just did.

fagemx:
# Review Response: Round 1

**New head SHA:** `146fb7772a00a3467605d10418d6378b3136372d` (was `0d1da1ce`)
**Gates ran:** pre-commit `sh scripts/lint-markdown-content.sh` and `bash scripts/lint-file-length.sh --staged`, both pass. No workspace gate re-run: the code delta is a comment block in `.cargo/config.toml` (the `linker` line is byte-identical) and prose in `.claude/CLAUDE.md`. Exact-head CI is the gate for this push.

All five P1s accepted. Two of them were right about things I had checked badly, and F5 is the most useful finding of the round.

## F5 — `[profile.dev] debug` is inert on CI, and the finding was not new

Confirmed independently. `.github/workflows/ci.yml:16-17`, added by #411 (`cd63a0c`):

```yaml
  # Trim debug info in CI: keeps readable backtraces (function name + line) but
  # skips the heavy PDB generation and linking that dominate Windows build time.
  CARGO_PROFILE_DEV_DEBUG: line-tables-only
  CARGO_PROFILE_TEST_DEBUG: line-tables-only
```

So the green `Test (windows-latest)` validated only the linker half, and this PR's premise — "debug builds carried full symbols" — was false for CI and true only locally. Worse for my framing: that comment already states the conclusion I reported today as a correction, *"linking … dominate Windows build time"*, and it has been in the tree for ten months.

The body now says this outright. What survives as new is narrower and is still the point of #810: **the manifest extends the setting to local and lane builds**, which the CI-only environment variables never touched, and that is precisely where the multi-gigabyte target directories live. Whether the now-redundant CI variables should be removed is routed as a follow-up rather than smuggled into this PR, since it is a CI edit.

## F4 — the comment reported the best-looking pair

Also correct. It cited 149s (the `-Clinker` proxy, not what the committed file produces) against 193s (that arm's *maximum* of a disclosed 173–193s band). Restated in both the file and the body as **153s against a 173–193s arm, so −12% to −21%**. The direction is unaffected — both rust-lld runs sit below the other arm's floor with no overlap — but only the favourable bound was being reported, including in the title, which is now corrected too. The comment also names `release.yml`'s tagged Windows binary as the third consumer (your P2).

## F3 — `.claude/CLAUDE.md:223`

Correct, and the placement made it worse: it sits three lines above the lane-pool text #810 exists to unblock. Rewritten so the sentence describes the post-merge state and keeps the old claim as history.

## F1 / F2 — `Closes #810`, and one correction

`Closes` removed, `Issue: #810` added, auto-close link verified gone via `closingIssuesReferences`.

One factual correction to F2, offered as evidence rather than argument. The claim was that `scripts/review-pr.sh` parses `Issue: #N` exclusively, so the brief would have loaded an empty ceiling. It mines from three sources (`scripts/review-pr.sh:166-178`): lines matching `^issue:`, **a closing-keyword regex** (`CLOSING_KW='closes|closed|close|fixes|…'`), and **`gh pr view --json closingIssuesReferences`**. `Closes #810` is found by the latter two. The fix stands on U2 alone, which is sufficient.

Separately, U2's blanket ban conflicts with the operator ruling `pr.closing-keyword` (*"A PR body writes `closes #N` exactly when it delivers every doneWhen item"*), whose carrier is the ruling comment on PR #687. The same staleness is already recorded against `scripts/review-pr.sh` contract item 4. Complied here because compliance is free; routed as a `FOLLOW-UP ISSUE` rather than argued in this PR.

## P2 — squash collapses the two commits

Accepted as stated; no change. Worth knowing at merge time: "drop the linker line" is a manual edit afterwards, not a revert.

fagemx:
## Code Review: Round 2 — PR #811 @ 146fb7772a00a3467605d10418d6378b3136372d

- model_requested: anthropic/claude-opus-5
- model_observed: unverified — no session record readable from this worktree; a system-prompt string is self-labelling, which §6 excludes
- spec: review-spec-v1.2
- class: code-risk  (REVIEW.md classes: code-risk code-plain skills)
- escalations: **D5** (wording/structure of the `.claude/CLAUDE.md` prose edit and the rewritten `.cargo/config.toml` comment block) — the brief does not name me as qualified for this class, so per §6.1 I mark it 需升級 rather than adjudicate it. This LGTM is therefore **provisional** under §6.4.
- cost: ~6 min, 13 tool calls, no cargo build, no workspace gate re-run.

### IN SCOPE

Delta round per §2: target is `git diff 0d1da1c..146fb77` plus RAN confirmation that each Round 1 blocker is resolved. The delta is comment text in `.cargo/config.toml` (the `linker` line is byte-identical — confirmed in the diff) and one sentence in `.claude/CLAUDE.md:226-229`. `Cargo.toml` is untouched since Round 1.

`.claude/CLAUDE.md` is new to the changed-file list this round, so the `skills` class and §5.1/§5.2 are routed for the first time — that is also why D5 becomes an escalation this round and was legitimately "not routed" in Round 1.

Changed behaviour at head is unchanged from Round 1: debug-info density of every local build (`Cargo.toml:68-75`) and the linker for every `x86_64-pc-windows-msvc` build, dev and release (`.cargo/config.toml:4,18`). Consumers: `ci.yml` Windows Clippy/Test, `release.yml:157,176` (tagged Windows binary), `.claude/CLAUDE.md` §"Footprint and thresholds", and any contributor's `$CARGO_HOME/config.toml` linker preference that a committed project config now outranks. Acceptance ceiling: #810 `doneWhen`. Integration: base `f0750ad`, `mergeable=MERGEABLE`, three-dot file list is the same 3 files, no textual conflict.

### Rules

|---|---|---|---|---|
| U1 surface | any | P0 | PASS | `gh pr diff 811 --name-only` → `.cargo/config.toml`, `.claude/CLAUDE.md`, `Cargo.toml` — all three inside the brief's allowed surface |
| U2 closing keyword ⟺ doneWhen | any | P1 | PASS | body grep → empty; `git log --format=%B origin/main..$SHA` grep → empty; `closingIssuesReferences: []`. F1 resolved |
| U3 `Issue: #N` line | any | P1 | PASS | `awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → `Issue: #810` (body line 1, and a commit trailer). F2 resolved |
| U4 conventional subjects | any | P1 | PASS | offender grep over `origin/main..$SHA` → empty; new subject is `chore(repo): …` |
| U5 markdown content lint | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → `exit=0` |
| U6 gates READ before RAN | any | P0 | PASS | exact-head `gh pr checks 811` → 9/9 pass incl. Clippy ×3 OS and Test ×3 OS (`Test (windows-latest)` 6m56s). `Independent Review` is `pending` — that is the review-queue status this round feeds, not a red gate |
| U7 round complete | any | P1 | PASS | whole scoped audit run before posting; no blocker withheld |
| D1 commands exist | docs/skills | P1 | PASS | `cargo build --help` → exit=0 OK; `cargo test --help` → exit=0 OK (the only added backticked invocations) |
| D2 ledger claims | docs/skills | P1 | N.A.(ledger unreachable from this worktree) | `edda ask build.dev-profile-and-linker` → `No results found.`; `--fleet` → `[edda] repo not on this machine (C:\ai_agent\edda-wt-gh566)`. Per §5.1 that is unreachable, never a P1. Durable-carrier check below |
| D3 paths and links resolve | docs/skills | P1 | PASS | D3 enumerator over `origin/main...$SHA` → zero `MISSING` lines |
| D4 authority boundary | docs/skills | P0 | PASS | grep for `gh pr merge\|--delete-branch\|--admin\|--no-verify\|push --force` → empty |
| D5 wording and structure | docs/skills | `[判斷]` | **需升級** | brief does not name me as qualified for this class (§6.1) |
| S1 skill may not cross a gate | skills | P0 | PASS | `git diff … -- '.claude/**'` grep for `merge\|self-review\|skip review` → empty. The added sentence is a factual footprint statement, not an instruction |
| S2 skill's factual claims resolve | skills | P1 | PASS | `.claude/CLAUDE.md:226-229` asserts the workspace sets `debug = "line-tables-only"` for dev builds — matches `Cargo.toml:75` at this head. GH-810 open. Tense is correct: it describes post-merge state and keeps the old claim explicitly as history |
| S3 one source of truth | skills | P1 | PASS | the edit corrects a fact inside the file that owns it; it restates no rule owned by `REVIEW.md` |
| C1 doneWhen vs real code | code-plain | P0 | PASS | all six items delivered at this head — detail below |
| C2 feat/fix test rule | code-plain | P0 | N.A.(all three subjects are `chore:`; config- and comment-only, no code path to regress) |
| C3 unwrap/expect | code-plain | P1 | PASS | `git diff origin/main...$SHA -- 'crates/**'` grep → empty (no crate files touched) |
| C4 unsafe / blanket allow | code-plain | P1 | PASS | grep → empty |
| C5 Windows coverage gap | code-plain | P1 | PASS | selector prints nothing (no `crates/` paths). Coverage is stronger than the selector implies: the L1 receipt linked all 56 suites with rust-lld locally, and exact-head `Test (windows-latest)` is green |
| R1 destructive commands | code-risk | P0 | PASS | grep → empty |
| R2 shell operator precedence | code-risk | `[判斷]` | N.A.(no added line mixes `\|\|` and `&&`; grep empty — nothing to judge) |
| R3 shell scripts parse | code-risk | P0 | N.A.(no `.sh`/`.bash` in the diff) |
| R4 resource release / `set -e` | code-risk | P1 | N.A.(no script, no temp dir, no lock) |
| R5 boundary, as properties | code-risk | P1 | PASS | the linker value is consumed by cargo/rustc, never interpolated into a shell. Fail-loud property was RAN in Round 1 (`linker = "definitely-not-a-linker.exe"` → `error: linker not found`, no silent fallback) and I did not re-run it — the line is byte-identical this round. Resolution on CI's floating stable is proven by the green Windows Clippy and Test jobs at this exact head. Config is target-scoped, so Linux/macOS are untouched |
| §5.5 wiring | any | — | filled | table below |

**C1 detail, re-verified at this head.** `[profile.dev]` set and the level justified against the alternative (`Cargo.toml:69-74`) ✓. Before/after target size, `.pdb` size and wall clock, clean build, same SHA, this workstation (body table, 4 runs) ✓. Panicking-test backtrace showing `file:line` on every frame, demonstrated ✓. `.cargo/config.toml` exists and its linker line carries a link-time measurement ✓. `cargo test --workspace` passes (corrected L1 receipt at `0d1da1c`: 2680 passed / 0 failed / 56 suites) ✓. Full L1 with gate receipt ✓ — and the Round 2 push is comment/prose-only with the `linker` line byte-identical, so CLAUDE.md's docs-only clause applies and exact-head CI is the gate; it re-ran fmt, clippy ×3 and test ×3 regardless.

### Wiring

|---|---|---|---|---|
| no new surfaces this round | — | — | — | — |

The delta adds no `pub` item, flag, config key, payload field or written file. Both surfaces (`Cargo.toml:75` `[profile.dev] debug`; `.cargo/config.toml:18` `linker`) were adjudicated in Round 1 and are unchanged in value. One addition to that record, checked here: `[profile.test]` inherits from `[profile.dev]` in Cargo, so the manifest form reaches test binaries with a single key — which is why `ci.yml` needed two environment variables to do the same job.

### Findings

- **[P2] The 173s floor the committed comment depends on appears in no run of the PR's measurement table.** Evidence: `.cargo/config.toml:8-9` says "an MSVC link.exe arm whose three runs spanned 173-193s"; the body table lists exactly two `link.exe` runs, 187s and 193s. The 173s datapoint is reachable only inside the Round 1 review comment, which quotes the pre-edit body. Round 2 moved the band into the durable file and removed its support from the body. The direction is unaffected and the reported range is the conservative one, so this understates rather than oversells — but the file's own headline number cannot be reconstructed from the evidence table beside it. Adding the third run to the table would close it.
- **[P2] The comment's stated measurement conditions do not cover all three runs it pools.** Evidence: `.cargo/config.toml:6-7` scopes the measurement to `debug = "line-tables-only"`, but the 187s run in the band is the `debug = 2` baseline (body table row 1). Pooling them is defensible — this PR's own result is that debug density does not move wall clock (187s → 193s) — but the comment does not say so.
- **[P2] The body routes the `ci.yml` reconciliation to an unnamed follow-up when a named one exists.** Evidence: body says "A follow-up should decide whether the now-redundant CI environment variables should go"; open issue **#821** §2 already carries exactly this, quoting `ci.yml:16-17`, #411/`cd63a0c`, and the same trap wording. Citing the number turns a promise into a link. This is what resolves Round 1's F5 residual, so it is worth naming.

### FOLLOW-UP ISSUE

- **Ledger decision `build.dev-profile-and-linker` may still carry the superseded single figure.** Round 1 quotes it as "took 193s to 153s, -21%", which the tree now restates as −12% to −21%. I could not verify its current content — the `[edda]` project's recorded repo path is `C:\ai_agent\edda-wt-gh566`, not on this machine, so `edda ask` and `edda ask --fleet` both fail to reach it. Stated as unverified, not as a stale claim. Worth a re-record or supersede from the checkout after merge.
- Round 1's four follow-ups stand and are unaffected by this delta: `[profile.dev.package."*"]` and `split-debuginfo` (both named in #810's suspected surface, neither measured nor declined); the `ci.yml:14-15` comment claiming trimmed debug info cuts Windows link time, which this PR's own measurement disproves; and `REVIEW.md` §3.1's enumerated risk surface omitting `.cargo/config.toml`.

### RAN vs READ

RAN:
- `gh pr view/diff/checks 811`, `gh issue view 810`, `gh issue view 821`, `gh issue list --search`; `git fetch origin main`; `git merge-base` → `f0750ad`.
- REVIEW.md §3 classifier on the file list → `classes=code-risk code-plain skills`, `canonical_class=code-risk`.
- Delta diff `0d1da1c..146fb77`, and the Round 2 commit message.
- U2 (body + commit bodies), U3, U4, D1 probe loop, D3 path enumerator, D4, S1, C3, C4, C5 selector, R1, R2, R3 — all as written in the spec, outputs in the table. Note for the next round: `origin/main..$SHA` still returns 53 files because the branch is behind `main`; the three-dot form returns the real 3.
- `sh scripts/lint-markdown-content.sh` → exit=0. `sh scripts/lint-file-length.sh --tree` → exit=0 (run because `.claude/CLAUDE.md` grew a line and the GH-779 ratchet is a repo gate).
- `sh scripts/wiring-scan.sh origin/main $SHA` → all hits are in `crates/edda-conductor/src/runner/sequential.rs` and `scripts/githooks/pre-commit`, neither of which this PR touches. Known false-positive class; no new surface in the delta.
- `gh pr view --json closingIssuesReferences` → `[]`, and `mergeable=MERGEABLE` / `mergeStateStatus=BLOCKED` (blocked on the pending `Independent Review` status this round feeds).
- Round 1 blocker confirmation, one by one: F1 keyword gone and auto-close link empty; F2 `Issue: #810` present; F3 `.claude/CLAUDE.md:226-229` rewritten and now factually matches `Cargo.toml:75`; F4 comment restated as 153s against a 173-193s arm, −12% to −21%, with the 149s proxy retained only as corroborating direction and the title corrected; F5 body premise corrected and the reconciliation routed to open issue #821 §2, which quotes the exact `ci.yml` lines. All five resolved.

READ:
- L1 gate receipt at `0d1da1c`, as corrected by the 2026-09-03 comment: fmt pass 4s; clippy `--workspace --all-targets -D warnings` pass 92s; `cargo test --workspace` pass, 2680 passed / 0 failed / 56 suites, 304s; clean tree, `CARGO_INCREMENTAL=0`, rustc 1.93.1.
- Exact-head CI at `146fb77`: Format, CI Gate, Detect code changes, Clippy ×3 OS, Test ×3 OS — 9/9 pass. This covers the head SHA more completely than the receipt does.
- Round 1's two scratch-crate probes (env-beats-manifest precedence; `rust-lld.exe` resolution and hard-error-on-bad-linker). Not re-run: the `linker` line and `Cargo.toml` are byte-identical since that measurement.
- No full local workspace rerun and no cargo invocation of any kind. Reason for not running one: the delta is comments and prose, the L1 receipt is recorded and corrected, exact-head CI is green on three platforms, and re-deriving it is the over-verification the ladder exists to remove.
- Checked and found not to be findings: `.cargo/*` is already in `ci.yml:82`'s code-detection whitelist, so a future `.cargo/config.toml`-only PR still triggers clippy and test; `release.yml:157,176` does build `--release --target x86_64-pc-windows-msvc -p edda`, so the comment's third-consumer claim is accurate; `[profile.test]` inherits `[profile.dev]`, so no test-profile gap; the "lane pool ceiling is provisional" sentence three lines below the edit remains correct, because the 21.6 GB incremental half is explicitly out of scope here.

### Verdict
LGTM (P0=0, P1=0)

All five Round 1 blockers are resolved in the tree and confirmed individually above; the three findings this round are P2 and none of them blocks. Two are about making the committed comment's own evidence reconstructable from the PR beside it, and one is a missing issue number for a follow-up that already exists as #821. One caveat on the merge gate rather than on the code: the `escalations:` field is non-empty — D5 is routed for the first time this round because `.claude/CLAUDE.md` entered the diff — so under §6.4 this LGTM does not satisfy the merge gate until a qualified engine adjudicates the prose edit. Merge remains the operator's action.

---

*Relayed by the controller session (docs workstation). Reviewer: **Claude Opus** via Claude Code (`fleet.review-engine-model=opus-via-claude-code`), read-only (`--disallowedTools Edit,Write,NotebookEdit`), build lane `n/a`. It cannot post to GitHub; the text above is its verbatim output between its `<<<VERDICT` markers. Reviewed SHA `146fb7772a00a3467605d10418d6378b3136372d` - **any push voids this verdict**. Contract item 5 was overridden per GH-691 so the reviewer used `git <verb> -h` rather than opening browser manuals.*

Issue #808, opened 2026-09-03
chore(repo): no clippy.toml — dirs::home_dir ignores HOME on Windows at 10 call sites while the lane wrapper sets it


fagemx:
Three classes of Windows defect this project has already paid for are enforced today only by prose — a ledger decision, a runbook paragraph, a CLAUDE.md line. `cargo clippy` can enforce all three at compile time with the reason attached to the error, and the workspace has no `clippy.toml` at all.

## What happened

Comparison against `xai-org/grok-build` (local checkout `C:\ai_project\grok-build`, v1.0.16, Rust, Apache-2.0), 2026-09-04. Its root `clippy.toml` bans twelve methods, each with the failure it prevents written inline. Three of those bans map onto defects edda has actually hit:

| grok-build ban | Their stated reason | edda's matching incident |
|---|---|---|
| `std::process::Command::spawn`, `tokio::process::Command::spawn` | "an unenrolled child outlives the session that started it, while an enrolled one dies with its scope" | `fleet.lane-launch` — the Windows Job Object inversion: six lanes lost over 2026-09-01/02 because children inherit job membership from the creating process |
| `std::env::home_dir`, `dirs::home_dir` | "$HOME-only reads broke Windows GUI-spawned processes, which have USERPROFILE but no HOME"; `dirs::home_dir` "asks the known-folder API and ignores a redirected USERPROFILE" | `fleet.lane-launcher` — the generated lane wrapper must set `HOME` explicitly because the Task Scheduler environment has none |
| `std::fs::canonicalize`, `Path::canonicalize`, `tokio::fs::canonicalize` | "returns `\\?\` verbatim paths on Windows that break external tools, leak into prompts, and poison path-equality keys" | not yet a recorded incident — see scope below |

The bans would fire on current `main`:

- `.spawn()` — 9 sites, including `crates/edda-conductor/src/agent/launcher.rs`, the launcher behind the lane losses, plus `codex_app_server.rs`, `pi_rpc.rs`, `stream.rs`, `cmd_reconcile.rs`.
- `dirs::home_dir()` — 8 sites across five crates: `edda-bridge-claude/src/dispatch/helpers.rs:23`, `edda-bridge-codex/src/admin.rs:38`, `edda-bridge-cursor/src/admin.rs:18,188`, `edda-bridge-hermes/src/admin.rs:38,234`.
[...]

PR #787, opened 2026-09-03
chore(githooks): file- and function-length ratchet (GH-779)

fagemx:
## Summary

Issue: #779

Seed a file- and function-length ratchet so the split sweep can only move down. Nothing is shrunk in this PR.

### Root Cause
Nothing in the repo bounded file or function length. `clippy::too_many_lines` is pedantic and was off; pre-commit gated fmt, per-crate clippy, markdown lint and a 1 MB blob cap, not line counts; no CI job measured size. Every long file got there one reasonable commit at a time.

### Changes
1. **Function length (clippy-native)**
   - Added `clippy.toml` with `too-many-lines-threshold = 150`.
   - Turned the lint on for every crate: `[workspace.lints.clippy] too_many_lines = "warn"` plus `[lints] workspace = true` in all 23 crates.
   - Marked current production functions over 150 awk-lines (plus the one clippy-flagged test) with `#[allow(clippy::too_many_lines)] // <len> lines at #779; split tracked in #<sibling or none>`.
   - 45 markers. Clippy's body-only count flags a subset (18); the allow list is the awk debt ledger so it stays grep-able. Sibling split issues: `#776` sequential.rs, `#777` cmd_bridge.rs, `#778` cmd_reconcile.rs.
2. **File length (pre-commit + CI)**
   - `scripts/lint-file-length.sh --staged|--tree` (bash + awk, NUL-safe).
   - `scripts/file-length-ceilings.txt`: one `<path> <ceiling>` for every tracked `*.rs` currently over 1000 (32 files, seeded at HEAD sizes). Default ceiling 1000.
   - Pre-commit calls `--staged` on each staged `*.rs` blob (`git cat-file -p <newsha>`).
   - CI Format job runs `bash -n` plus `--tree`.
3. **Docs**: L0 row and Pre-commit Checklist in `.claude/CLAUDE.md` point at the script.

### Pre-fix Regression Evidence
`scripts/githooks/test.sh` against the base hook (no length ratchet):

```text
FAIL: reject: staged .rs one line over file-length ceiling (GH-779)
  detail: expected rejection, got success: pre-commit: cargo fmt --all --check
pre-commit: cargo clippy -p hooktest --all-targets -- -D warnings
[master 561c2fb] feat(test): over-ceiling rust file
 1 file changed, 1001 insertions(+)
 create mode 100644 crates/hooktest/src/long.rs
```

`--tree` with a listed ceiling lowered by one line:

```text
[...]

Issue #779, opened 2026-09-03
chore(githooks,ci): file- and function-length ratchet — clippy too_many_lines at 150 with the 42 existing offenders marked, plus a per-path line ceiling in pre-commit and CI so the split sweep can only move down


fagemx:
## What happened

Observed on `cbbc8b03a533aa854939847316fce789cc15b51d` (main).

Nothing in the repo bounds file or function length. There is no `clippy.toml`
(so `clippy::too_many_lines` is unconfigured and, being pedantic, off);
`scripts/githooks/pre-commit` gates fmt, per-crate clippy, markdown lint and a
1 MB blob cap (decision `repo.commit-gates`), not line counts; no CI job
measures size.

Current stock, workspace-wide over `git ls-files '*.rs'`:

- **30 Rust files exceed 1000 lines**; the top five are 5650 / 5616 / 4903 /
  4137 / 3479.
- **43 functions exceed 150 lines** (42 production, 1 test); the top five are
  `run_plan` 684, `process_phase_result` 400, `cmd_gc::execute` 376,
  `digest_one_session` 307, `main` 304.
- The three longest files are also the three most-edited files in the last 200
  commits (34 / 31 / 30 touches).

Measured with:

```
git ls-files '*.rs' | xargs wc -l | awk '$1>1000 && $2!="total"' | wc -l      # 30
awk '
  FNR==1 { if (start) flush(); f=FILENAME; start=0 }
  /^[[:space:]]*(pub(\([a-z]+\))? )?(async )?(unsafe )?fn [a-zA-Z_]/ { if (start) flush(); start=FNR; name=$0 }
  function flush(){ len=FNR-start; if (len>150) printf "%5d %s:%d\n", len, f, start }
  END { if (start) flush() }
' $(git ls-files '*.rs') | sort -rn | wc -l                                      # 43
```

The sibling issues reduce the stock. Without a ratchet it grows back — every
one of those files got there one reasonable commit at a time.

## Why it matters

Same three costs as the siblings — agent read cost per edit, `Edit` uniqueness,
lane conflicts — but as a rate, not a level. A ratchet costs nothing on a clean
commit and turns "this file is getting long" from a review opinion into a
mechanical check with no `[判斷]` slot.

## Suspected surface

Two mechanisms, both zero-dependency, matching
`ci.markdown-content-lint=tracked-shell-awk-fence-aware` and `repo.commit-gates`:

**1. Function length — clippy-native.**

[...]

PR #812, opened 2026-09-03
chore(repo): ban dirs::home_dir and resolve home through edda_core::paths — it ignores HOME on Windows while the lane wrapper sets it

fagemx:
Issue: #808

Ten call sites across eight crates each resolved the user's home directory by calling `dirs::home_dir()` directly, with no shared resolver. This routes them through one function and makes `clippy.toml` fail the build if a new call site bypasses it.

## The platform fact, read from the vendored source

Not repeated from another project — read out of the crate this workspace compiles:

```rust
// dirs 6.0.0, src/win.rs:5
pub fn home_dir() -> Option<PathBuf> { dirs_sys::known_folder_profile() }
// dirs-sys 0.5.0, src/lib.rs:172
pub fn known_folder_profile() -> Option<PathBuf> { known_folder(Shell::FOLDERID_Profile) }
```

On Windows it is `SHGetKnownFolderPath(FOLDERID_Profile)`. **It reads neither `HOME` nor `USERPROFILE`.**

Two consequences in this repo:

1. `fleet.lane-launcher` requires the generated lane wrapper to set `HOME` explicitly, because the Task Scheduler environment has none. Nothing reached through these calls observes that `HOME`.
2. The bridge-admin sites locate *another tool's* configuration — `~/.codex/hooks.json`, `~/.hermes/cli-config.yaml`, `~/.claude/plans`. A Node-based tool's `os.homedir()` reads `USERPROFILE`, so on a redirected profile edda writes hook config where that tool will not look for it.

## What changed

- New `crates/edda-core/src/paths.rs` — `home_dir()`, carrying the mechanism above in its doc comment with `file:line` citations, and the single `#[allow(clippy::disallowed_methods)]`. `edda-core` has no in-workspace dependencies, so it is a safe base with no cycle risk.
- `clippy.toml` — `disallowed-methods` for `dirs::home_dir` and `std::env::home_dir`, each `reason` naming the edda failure rather than a generic one.
- Ten call sites routed. `edda-bridge-cursor` and `edda-transcript` gain an `edda-core` dependency (one line each); the other six crates already had it.

**Behaviour is deliberately unchanged.** The resolver wraps the same call. Switching to an environment-first lookup would move paths for users who already have state on disk — that is a separate decision with a migration question attached, and it is not made here. What changes is that the fact is stated once instead of being rediscovered at ten sites.

## Fail-first evidence

With the ban in place and before routing, `cargo clippy --workspace --all-targets` reported exactly ten sites:

```
crates\edda-bridge-claude\src\dispatch\helpers.rs:23:19
crates\edda-bridge-codex\src\admin.rs:38:5
crates\edda-bridge-cursor\src\admin.rs:18:5      crates\edda-bridge-cursor\src\admin.rs:188:22
crates\edda-bridge-hermes\src\admin.rs:38:5      crates\edda-bridge-hermes\src\admin.rs:234:9
crates\edda-bridge-openclaw\src\admin.rs:150:5
crates\edda-ledger\src\paths.rs:192:20
crates\edda-store\src\lib.rs:59:32
crates\edda-transcript\src\pi.rs:56:16
```

After routing: zero.

## `clippy.toml` is edited additively — and nearly was not

`clippy.toml` already existed, added by #787 (the GH-779 function-length ratchet), containing `too-many-lines-threshold = 150`. I first wrote the new file with `cat >` and destroyed it, on a wrong premise from checking a worktree one day stale. The threshold fell back to clippy's built-in 100 and six functions in `edda-search-fts` and `edda-ledger` — crates this PR does not touch — began failing `-D warnings`. That is how the mistake surfaced.

It surfaced only because the deleted setting had a visible effect. A key whose loss did not immediately fail a gate would have vanished silently. The restored file now leads with the reason that is easy to get wrong: **clippy reads the nearest `clippy.toml` and does not merge**, so that one file is the only place any of this can live. The threshold is preserved with a `Do not remove` comment naming GH-779, and the diff is `+12` lines with no deletions.

## What was dropped from the issue

The issue also proposed banning `Command::spawn`. Reading the call sites killed it, and the review corrected my census — recorded here because the numbers in the original body were wrong.

Real inventory, 9 sites: **5 production** — `cmd_reconcile.rs:2314` and `:2549` (deliberately detached, `CREATE_NO_WINDOW`), `codex_app_server.rs:36` and `pi_rpc.rs:327` (`kill_on_drop(true)`), `launcher.rs:293` (managed and awaited) — and **4 test** — `codex_app_server.rs:998` and `:1023` (both inside the `#[cfg(test)]` module that opens at line 649; the original body counted `:998` as production) and `stream.rs:690` **and `:697`** (the second was enumerated nowhere).

The conclusion is unchanged: all 5 production sites are already correct, so a ban would need `#[allow]`s to preserve a correct status quo. The incident I originally cited, `fleet.lane-launch`, is the *opposite* failure — children dying too early through Windows Job Object inheritance — and was fixed in PowerShell with no Rust `spawn` implicated. Full reasoning is on the issue, which carries the same correction.

## Gate receipt

Clean tree, `CARGO_INCREMENTAL=0`, rustc 1.93.1, worktree default `target/`, no build lane.

| gate | result |
|---|---|
| `cargo fmt --all --check` | pass (4s) |
| `cargo clippy --workspace --all-targets -- -D warnings` | pass (48s) |
| `cargo test --workspace` | **pass — 2680 passed, 0 failed, 56 suites** (441s) |

`#757` did not fire on this run; it is probabilistic, so a green run is not evidence it is fixed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -818,6 +818,7 @@ dependencies = [
  "anyhow",
  "dirs",
  "edda-bridge-claude",
+ "edda-core",
  "edda-pack",
  "edda-postmortem",
  "edda-store",
@@ -915,6 +916,7 @@ name = "edda-core"
 version = "0.4.0"
 dependencies = [
  "anyhow",
+ "dirs",
  "globset",
  "hex",
  "regex",
@@ -1123,6 +1125,7 @@ version = "0.4.0"
 dependencies = [
  "anyhow",
  "dirs",
+ "edda-core",
  "edda-store",
  "fs2",
  "serde",
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,1 +1,13 @@
+# Clippy configuration for the workspace.
+#
+# Clippy reads the NEAREST clippy.toml and does not merge, so everything the
+# workspace needs has to live in this one file.
+
+# Function-length ratchet from GH-779 (#787). Do not remove.
 too-many-lines-threshold = 150
+
+# Methods banned workspace-wide, each with the failure it prevents.
+disallowed-methods = [
+    { path = "dirs::home_dir", reason = "on Windows this is SHGetKnownFolderPath(FOLDERID_Profile) (dirs 6.0.0 src/win.rs:5 -> dirs-sys 0.5.0 src/lib.rs:172) and reads neither HOME nor USERPROFILE, so it disagrees with the lane wrapper that fleet.lane-launcher requires to set HOME, and with any Node tool's os.homedir(); resolve through edda_core::paths::home_dir" },
+    { path = "std::env::home_dir", reason = "resolve through edda_core::paths::home_dir so every home-anchored path in the workspace agrees" },
+]
--- a/crates/edda-bridge-claude/src/dispatch/helpers.rs
+++ b/crates/edda-bridge-claude/src/dispatch/helpers.rs
@@ -20,7 +20,7 @@ const PLAN_EXCERPT_MAX_LINES: usize = 30;
 pub(crate) fn render_active_plan(project_id: Option<&str>) -> Option<String> {
     let plans_dir = match std::env::var("EDDA_PLANS_DIR") {
         Ok(dir) => PathBuf::from(dir),
-        Err(_) => dirs::home_dir()?.join(".claude").join("plans"),
+        Err(_) => edda_core::paths::home_dir()?.join(".claude").join("plans"),
     };
     render_active_plan_from_dir(&plans_dir, project_id)
 }
--- a/crates/edda-bridge-codex/src/admin.rs
+++ b/crates/edda-bridge-codex/src/admin.rs
@@ -35,7 +35,7 @@ const HOOK_EVENTS: &[&str] = &[
 ];
 
 fn default_hooks_path() -> Option<PathBuf> {
-    dirs::home_dir().map(|h| h.join(".codex").join("hooks.json"))
+    edda_core::paths::home_dir().map(|h| h.join(".codex").join("hooks.json"))
 }
 
 /// Write (or merge) edda hooks into the Codex hooks config.
--- a/crates/edda-bridge-cursor/Cargo.toml
+++ b/crates/edda-bridge-cursor/Cargo.toml
@@ -11,6 +11,7 @@ keywords.workspace = true
 
 [dependencies]
 edda-bridge-claude = { path = "../edda-bridge-claude", version = "0.4.0" }
+edda-core = { path = "../edda-core", version = "0.4.0" }
 edda-pack = { path = "../edda-pack", version = "0.4.0" }
 edda-postmortem = { path = "../edda-postmortem", version = "0.4.0" }
 edda-store = { path = "../edda-store", version = "0.4.0" }
--- a/crates/edda-bridge-cursor/src/admin.rs
+++ b/crates/edda-bridge-cursor/src/admin.rs
@@ -15,7 +15,7 @@ const HOOK_EVENTS: &[&str] = &[
 ];
 
 fn default_hooks_path() -> anyhow::Result<PathBuf> {
-    dirs::home_dir()
+    edda_core::paths::home_dir()
         .map(|home| home.join(".cursor").join("hooks.json"))
         .ok_or_else(|| anyhow::anyhow!("cannot determine home directory"))
 }
@@ -185,7 +185,7 @@ fn store_is_writable(store_root: &Path) -> bool {
 }
 
 fn claude_hook_detected() -> bool {
-    let Some(home) = dirs::home_dir() else {
+    let Some(home) = edda_core::paths::home_dir() else {
         return false;
     };
     [
--- a/crates/edda-bridge-hermes/src/admin.rs
+++ b/crates/edda-bridge-hermes/src/admin.rs
@@ -35,7 +35,7 @@ fn default_config_path() -> Option<PathBuf> {
     // equivalent on Windows. The primary config filename is `cli-config.yaml`
     // (the codebase writes both cli-config.yaml and config.yaml in different
     // eras; cli-config.yaml is the current name).
-    dirs::home_dir().map(|h| h.join(".hermes").join("cli-config.yaml"))
+    edda_core::paths::home_dir().map(|h| h.join(".hermes").join("cli-config.yaml"))
 }
 
 /// Write (or merge) the edda hook into Hermes' cli-config.yaml.
@@ -231,7 +231,7 @@ pub fn doctor() -> anyhow::Result<()> {
     // Check consent status. Even if hooks are installed, they don't fire
     // until approved once via TTY/env/config.
     let allowlist_path =
-        dirs::home_dir().map(|h| h.join(".hermes").join("shell-hooks-allowlist.json"));
+        edda_core::paths::home_dir().map(|h| h.join(".hermes").join("shell-hooks-allowlist.json"));
     let approved_here = allowlist_path
         .as_ref()
         .and_then(|p| fs::read_to_string(p).ok())
--- a/crates/edda-bridge-openclaw/src/admin.rs
+++ b/crates/edda-bridge-openclaw/src/admin.rs
@@ -147,7 +147,7 @@ module.exports = plugin;
 
 /// Default plugin directory.
 fn default_plugin_dir() -> Option<PathBuf> {
-    dirs::home_dir().map(|h| h.join(".openclaw").join("extensions").join("edda-bridge"))
+    edda_core::paths::home_dir().map(|h| h.join(".openclaw").join("extensions").join("edda-bridge"))
 }
 
 /// Install the OpenClaw plugin files.
--- a/crates/edda-core/Cargo.toml
+++ b/crates/edda-core/Cargo.toml
@@ -19,6 +19,7 @@ hex.workspace = true
 ulid.workspace = true
 time.workspace = true
 serde_yaml.workspace = true
+dirs.workspace = true
 globset.workspace = true
 regex.workspace = true
 
--- a/crates/edda-core/src/lib.rs
+++ b/crates/edda-core/src/lib.rs
@@ -6,6 +6,7 @@ pub mod decision;
 pub mod event;
 pub mod git;
 pub mod hash;
+pub mod paths;
 pub mod policy;
 pub mod secret_guard;
 pub mod tool_tier;
--- a/crates/edda-core/src/paths.rs
+++ b/crates/edda-core/src/paths.rs
@@ -1,0 +1,37 @@
+//! Home-anchored path resolution.
+//!
+//! Every home-anchored path in the workspace resolves here, so all of them
+//! agree. `dirs::home_dir` is banned in `clippy.toml`; this module holds the
+//! one sanctioned call.
+
+use std::path::PathBuf;
+
+/// The current user's home directory.
+///
+/// # Platform behaviour, and why this wrapper exists
+///
+/// On Windows this is **not** `%HOME%` and **not** `%USERPROFILE%`. It is a
+/// shell known-folder lookup — `dirs` 6.0.0 `src/win.rs:5` calls
+/// `dirs_sys::known_folder_profile()`, which is `dirs-sys` 0.5.0
+/// `src/lib.rs:172` calling `known_folder(FOLDERID_Profile)`. Neither
+/// environment variable is read.
+///
+/// That matters here for two reasons:
+///
+/// 1. `fleet.lane-launcher` requires the generated lane wrapper to set `HOME`
+///    explicitly, because the Task Scheduler environment has none. Nothing
+///    reached through this function will observe that `HOME`.
+/// 2. Several callers locate *another tool's* configuration — `~/.codex`,
+///    `~/.hermes`, `~/.claude`. A Node-based tool's `os.homedir()` reads
+///    `USERPROFILE`, so if a profile is redirected the two disagree and edda
+///    writes its hook config where that tool will not look for it.
+///
+/// This function deliberately preserves the existing behaviour rather than
+/// switching to an environment-first lookup: changing it would move paths for
+/// users who already have state on disk. It exists so that the behaviour is
+/// stated once, in one place, instead of being rediscovered at ten call sites.
+pub fn home_dir() -> Option<PathBuf> {
+    // The one sanctioned call. See the module docs and clippy.toml.
+    #[allow(clippy::disallowed_methods)]
+    dirs::home_dir()
+}
--- a/crates/edda-ledger/src/paths.rs
+++ b/crates/edda-ledger/src/paths.rs
@@ -189,7 +189,7 @@ impl EddaPaths {
     }
 
     fn find_root_walk(start: &Path, ceiling: Option<&Path>) -> Option<PathBuf> {
-        let home = dirs::home_dir();
+        let home = edda_core::paths::home_dir();
 
         // Normalisation belongs to `find_root_bounded`, not here: the
         // unbounded `find_root` is production API and this PR does not
--- a/crates/edda-store/src/lib.rs
+++ b/crates/edda-store/src/lib.rs
@@ -56,7 +56,7 @@ pub fn store_root() -> PathBuf {
     }
     if let Some(data_dir) = dirs::data_dir() {
         data_dir.join("edda")
-    } else if let Some(home) = dirs::home_dir() {
+    } else if let Some(home) = edda_core::paths::home_dir() {
         home.join(".edda")
     } else {
         PathBuf::from(".edda-store")
--- a/crates/edda-transcript/Cargo.toml
+++ b/crates/edda-transcript/Cargo.toml
@@ -10,6 +10,7 @@ categories.workspace = true
 keywords.workspace = true
 
 [dependencies]
+edda-core = { path = "../edda-core", version = "0.4.0" }
 edda-store = { path = "../edda-store", version = "0.4.0" }
 anyhow.workspace = true
 thiserror.workspace = true
--- a/crates/edda-transcript/src/pi.rs
+++ b/crates/edda-transcript/src/pi.rs
@@ -53,7 +53,7 @@ pub fn pi_session_dir_for_cwd(cwd: &Path) -> Option<PathBuf> {
         }
     }
     safe.push_str("--");
-    let home = dirs::home_dir()?;
+    let home = edda_core::paths::home_dir()?;
     Some(home.join(".pi").join("agent").join("sessions").join(safe))
 }
 


fagemx:
## Code Review: Round 1 — PR #812 @ 895d00b6a0441b80e8056260dcdd5379d90eb687

- model_requested: (not specified in brief)
- model_observed: unverified — interactive Claude Code session, no `claude -p --output-format json` `modelUsage` key to read from the system; harness context reports `claude-opus-5`, which is self-labelling and is not offered as evidence (§6)
- spec: review-spec-v1
- escalations: none — `D5` and `R2` are the only `[判斷]` rules in the spec; neither is routed (no `docs`/`skills` class; the diff adds no shell lines). This is not a zero-findings review: 3 P2 and 2 FOLLOW-UP ISSUEs are enumerated below, and every routed rule carries a row.
- cost: ~30 min; ~20 tool calls; 3 × `cargo clippy -p edda-core` (12s cold, ~2s warm) + 1 × focused `cargo test -p` over 5 crates, all in the **existing** `verifier` lane (`%LOCALAPPDATA%\fleet-workstation\lanes\verifier`) with `CARGO_INCREMENTAL=0`. The brief named no lane; an existing lane was reused rather than a new target directory created, per `.claude/CLAUDE.md` § Build lanes. Review worktree was detached, disposable, and removed.

### IN SCOPE

Changed behaviour and paths: 15 files, +66/-10. One new `pub mod paths` in `edda-core` holding `home_dir()`; two `disallowed-methods` entries added to the pre-existing `clippy.toml`; ten call sites re-pointed at the wrapper; `edda-bridge-cursor` and `edda-transcript` gain a direct `edda-core` dependency.

Direct callers/consumers: the ten routed sites, plus every crate that compiles under `cargo clippy` (the new ban changes what "green" means, which is why this was reviewed against §5.4 as well as §5.3 even though `clippy.toml` is not on the §3.1 enumerated risk list — the diff already carries `code-risk` via `Cargo.lock` and `crates/edda-core/**`).

Acceptance ceiling: issue #808 `doneWhen`, **as amended by the author's own narrowing comment on that issue** ("doneWhen items in the body that refer to spawn or to '17 sites' are void; the count is ten"). That comment is a durable, dated amendment on the issue itself, so the ceiling this PR is judged against is the narrowed one, not the original body. The narrowing was independently re-checked against the code — see Findings P2-2.

Security / data-loss regressions introduced or exposed: none. `home_dir()` takes no input, returns the same `Option<PathBuf>` on the same platform code path, and no caller-supplied value is interpolated anywhere in the diff.

Current-base integration: `MERGEABLE`; the base's `clippy.toml` content is fully contained in the head's — verified byte-for-byte below.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 15 files, all inside the issue's suspected surface (`clippy.toml`; the banned call sites; the sanctioned wrapper the issue explicitly asks the PR to "propose, do not assume"). No file outside it. |
| U2 | any | P1 | PASS | body grep and `git log --format=%B origin/main..$SHA` grep for closing keywords both empty (exit 1) |
| U3 | any | P1 | PASS | `Issue: #808` present |
| U4 | any | P1 | PASS | 1 commit, `chore(repo): ban dirs::home_dir, resolve home through edda_core::paths`; offender filter empty |
| U6 | any | P0 | PASS | `gh pr checks 812` → 9/9 pass (Format, Clippy ×3 OS, Test ×3 OS, Detect code changes, CI Gate). L1 receipt READ from the PR body: fmt / clippy `-D warnings` / `cargo test --workspace` 2680 passed 0 failed 56 suites, clean tree, `CARGO_INCREMENTAL=0`, rustc 1.93.1. One RAN beyond it, reason stated at C5. |
| U7 | any | P1 | PASS | whole scoped audit finished before posting; single round, all findings batched |
| C1 | code-plain | P0 | PASS | doneWhen matched to code, not to the body: (1) `clippy.toml` carries `dirs::home_dir` + `std::env::home_dir` with `reason` strings — spawn entries void per the issue amendment; (2) fail-first list quotes exactly 10 sites whose `file:line` match the pre-change lines in the diff hunks 1:1, and the ban was independently proven to fire (RAN below); (3) the single `#[allow(clippy::disallowed_methods)]` carries a comment — `crates/edda-core/src/paths.rs:35-36`; (4) behaviour unchanged — all ten rewrites verified one by one, see below. |
| C2 | code-plain | P0 | N.A.(commit type is `chore:`, not `feat:`/`fix:`; and doneWhen explicitly waives new tests — "No test asserts new runtime behavior; the regression protection is the lint itself, and the fail-first output is the evidence it works". The lint's effectiveness was verified by RAN instead of taken on the body's word.) |
| C3 | code-plain | P1 | PASS | `git diff origin/main...$SHA -- 'crates/**' \| grep -nE '^\+.*\.(unwrap\|expect)\('` → empty (exit 1) |
| C4 | code-plain | P1 | PASS | `unsafe {` / `#[allow(clippy::all)]` / `#![allow` grep → empty (exit 1). The one allow is `#[allow(clippy::disallowed_methods)]` on a single statement — targeted, permitted by `.claude/CLAUDE.md` §3.1 |
| C5 | code-plain | P1 | PASS (RAN) | Selector printed 5 crates outside the CI Windows subset: `edda-bridge-codex`, `edda-bridge-cursor`, `edda-bridge-hermes`, `edda-bridge-openclaw`, `edda-core`. Focused check RAN on Windows 11 / rustc 1.93.1: `cargo test -p edda-core -p edda-bridge-codex -p edda-bridge-cursor -p edda-bridge-hermes -p edda-bridge-openclaw` → **exit 0, 10 suites, 284 passed, 0 failed**. Rerun reason (U6): these five crates' test targets are run by no CI job on Windows, and the only Windows evidence for them is the author's own receipt on a PR the same session authored the issue for. |
| R1 | code-risk | P0 | PASS | destructive-command grep over added lines → empty (exit 1) |
| R2 | code-risk | `[判斷]` | N.A.(no added line mixes `\|\|` and `&&`; the diff adds no shell) |
| R3 | code-risk | P0 | N.A.(no `*.sh` / `*.bash` changed) |
| R4 | code-risk | P1 | N.A.(no temp directory, file handle or lock is created in the diff) |
| R5 | code-risk | P1 | PASS | property held: no caller-supplied value reaches a shell, a filename encoder, or a SHA match in this diff. `home_dir()` is nullary; its return type and `None` domain are identical to the call it replaces. |
| §5.5 | any | required slot | PASS | filled below |

### Wiring

|---|---|---|---|---|
| `pub mod paths` — `crates/edda-core/src/lib.rs:9` | module declaration in the workspace base crate | the 10 call sites; `scripts/wiring-scan.sh` reports 815 outside references for the module path | n/a (declaration) | compiled into every crate already depending on `edda-core` |
| `pub fn home_dir() -> Option<PathBuf>` — `crates/edda-core/src/paths.rs:33` | wraps `dirs::home_dir()` verbatim at `crates/edda-core/src/paths.rs:36`; no new state, no I/O beyond the wrapped call | 10 readers: `edda-bridge-claude/src/dispatch/helpers.rs:23`, `edda-bridge-codex/src/admin.rs:38`, `edda-bridge-cursor/src/admin.rs:18` and `:188`, `edda-bridge-hermes/src/admin.rs:38` and `:234`, `edda-bridge-openclaw/src/admin.rs:150`, `edda-ledger/src/paths.rs:192`, `edda-store/src/lib.rs:59`, `edda-transcript/src/pi.rs:56` (`wiring-scan.sh`: 10 outside references) | **nothing newly swallowed.** `None` is produced in exactly the states `dirs::home_dir()` produced it, and each caller's handling arm is unchanged in kind: `?` at helpers.rs:23 and pi.rs:56, `.map(..)` at codex:38 / hermes:38,234 / openclaw:150, `.map(..).ok_or_else(anyhow!)` at cursor:18, `let … else { return false }` at cursor:188, bare binding at ledger:192, `else if let Some` arm at store:59 | enforcement layer is `clippy.toml`, and it was measured to reach the code — see RAN |
| `disallowed-methods` keys — `clippy.toml:10-13` | 2 hand-written TOML entries, each with a `reason` | `cargo clippy` in 3 CI OS jobs + local L0/L1 | compile-time error under `-D warnings`. Measured limits: an entry whose **item name** is wrong warns loudly (`does not refer to a reachable function`, i.e. red under `-D warnings`); an entry whose **crate root** is absent from the crate's dependency graph is dropped **silently**. Both shipped entries were proven to fire. | every crate that links `dirs` (9 of 23); the `std::env` entry reaches all 23 |

### Findings

- [P2] Seven crates now carry an unused `dirs` dependency — this PR removed their last `dirs::` reference and left the manifest line. Evidence: at `895d00b`, `git grep -c '\bdirs::' -- crates/<c>` returns **0 matching files** for `edda-bridge-claude`, `edda-bridge-codex`, `edda-bridge-cursor`, `edda-bridge-hermes`, `edda-bridge-openclaw`, `edda-ledger`, `edda-transcript`, while each `crates/<c>/Cargo.toml` still declares `dirs.workspace = true`. (`edda-store` legitimately keeps it for `dirs::data_dir()` at `src/lib.rs:57`; `edda-core` for the sanctioned call.) Harmless to the build — nothing enables `unused_crate_dependencies` — but it is dead manifest surface created by this change, and dropping it would make "which crates can even reach `dirs`" match "which crates are covered by the ban".
- [P2] The spawn-drop census in the PR body does not add up, though its conclusion survives independent checking. The body says *"four already use `kill_on_drop(true)`, two are deliberately detached with `CREATE_NO_WINDOW`, two are tests"* — that is 8, against "the nine call sites". Measured at `895d00b`: 9 `.spawn()` sites. `crates/edda-conductor/src/agent/codex_app_server.rs` opens `#[cfg(test)] mod tests` at line **649**, so the sites at `:998` and `:1023` are **both test code**, not one production "cancellation child" plus one test; and `crates/edda-conductor/src/agent/stream.rs` has spawns at **both** `:690` and `:697` (the two arms of one `cfg!(windows)` branch inside a `#[tokio::test]`, `mod tests` opening at line 350) — the second is not enumerated anywhere. Correct inventory: 5 production sites (`codex_app_server.rs:36` and `pi_rpc.rs:327` with `kill_on_drop(true)`; `launcher.rs:293` managed and awaited, **not** `kill_on_drop`; `cmd_reconcile.rs:2314` and `:2549` deliberately detached, `creation_flags(0x0800_0000)` with null stdio) and 4 test sites. I re-verified the substantive claim and **it holds**: no site is an unmanaged long-lived child, and `edda ask fleet.lane-launch` confirms the cited incident is the opposite failure ("job membership is inherited from the CREATING process… nohup and Start-Process detach the console but NOT job membership"), fixed at the PowerShell/Task-Scheduler layer with no Rust `spawn` implicated. So the narrowing is right and the numbers backing it are not.
- [P2] The `std::env::home_dir` entry's `reason` names the sanctioned alternative but no failure — `clippy.toml:12`, *"resolve through edda_core::paths::home_dir so every home-anchored path in the workspace agrees"* — while doneWhen item 1 asks each `reason` to name "the edda incident and the sanctioned alternative". There is no edda incident to name (zero call sites; the ban is preemptive), so this is not a gap that can be closed as written — but the entry is also silently the *interesting* one: `std::env::home_dir` on Windows does consult `USERPROFILE`, i.e. it behaves **differently** from the `dirs` call the wrapper keeps. One clause saying that is what makes the ban legible to the next contributor who reaches for it as the obvious fix.

### FOLLOW-UP ISSUE

- **The disagreement the PR title names is centralised, not resolved — and nothing tracks resolving it.** `edda_core::paths::home_dir()` still returns `SHGetKnownFolderPath(FOLDERID_Profile)`, so the lane wrapper's explicit `HOME` (required by `fleet.lane-launcher`, confirmed in the ledger: *"CARGO_TARGET_DIR and HOME are EMPTY there and must be set explicitly by the lane wrapper"*) and a Node tool's `os.homedir()`/`USERPROFILE` are still unobserved by every one of the ten sites. The PR is explicit about this being deliberate and doneWhen says "Behavior is unchanged", so it is correctly out of scope here. But the author's own issue comment calls it *"the one real design question"* and names no tracking issue, and the four bridge-admin sites — which write **another tool's** config — are where it bites. Worth an issue that decides between environment-first resolution (with the migration question attached) and reporting the disagreement, now that there is exactly one place to change.
- **`.claude/CLAUDE.md` §3.1 still says nothing about the workspace method bans.** The issue's suspected surface asked for "one line noting that the workspace bans specific methods and where the reasons live"; §3.1 at `895d00b` still covers only `-D warnings` and targeted allows. Not a doneWhen item, and the issue's own mechanism (the reason is printed in the clippy diagnostic) means nobody has to read it — so this is genuinely optional, not a gap.

### RAN vs READ

**RAN** (in a detached worktree at `895d00b` under the session scratchpad; the reviewed checkout was never modified, nothing was committed, pushed, merged, or labelled):

1. **`clippy.toml` integrity — the restoration is complete and correct.** `git show origin/main:clippy.toml` is exactly one line, `too-many-lines-threshold = 150`. `git show 895d00b:clippy.toml` contains that line **byte-identical**, and the diff is `+12/-0` with no deletion. Nothing else existed in the base file, so nothing else could have been lost. The branch carries a single commit, so there is no intermediate tree in which the ratchet was absent. The file parses: clippy read `disallowed-methods` out of it and reported a diagnostic anchored at `clippy.toml:12:5`, which it could not do if the TOML were rejected.
2. **The ban fires, and does not have the bypasses one would look for.** Six deliberately-different reintroduction shapes were added to `crates/edda-core/src/paths.rs` in the throwaway worktree and linted with `cargo clippy -p edda-core`:

   ```
   paths.rs:45  warning: use of a disallowed method `dirs::home_dir`      <- plain dirs::home_dir()
   paths.rs:52  warning: use of a disallowed method `dirs::home_dir`      <- use dirs::home_dir as aliased_home; aliased_home()
   paths.rs:57  warning: use of a disallowed method `std::env::home_dir`  <- std::env::home_dir()
   paths.rs:66  warning: use of a disallowed method `dirs::home_dir`      <- through a local `pub use dirs::home_dir` re-export
   paths.rs:71  warning: use of a disallowed method `dirs::home_dir`      <- ::dirs::home_dir()
   paths.rs:76  warning: use of a disallowed method `dirs::home_dir`      <- the bare fn item as a value, never called
   ```

   All six fire. The lint matches on `DefId`, so renaming the dependency, aliasing the import, re-exporting it, or taking a function pointer does not evade it. Both shipped entries resolve to real items — including `std::env::home_dir`, the one that could have been silently dead.
3. **What a future edit to this file would and would not catch.** Two bogus entries were added to the worktree's `clippy.toml`: `dirs::hoem_dir` produced `warning: `dirs::hoem_dir` does not refer to a reachable function` (red under `-D warnings`), while `totally::not::a::crate::home_dir` produced **nothing at all**. So a typo inside a known crate fails loudly; a path whose crate root is not in that crate's graph is dropped in silence. That is consistent with CI being green while 15 of 23 crates do not depend on `dirs`, and it is why item 2 above was measured rather than assumed. `allow-invalid` is not set on either shipped entry, which is the correct setting.
4. **Behaviour preservation, all ten sites.** Each rewrite was read against its base hunk. The wrapper's signature `fn home_dir() -> Option<PathBuf>` is identical to `dirs::home_dir`, and every call site keeps its original arm verbatim — `?` (helpers.rs:23, pi.rs:56), `.map` (codex:38, hermes:38, hermes:234, openclaw:150), `.map(..).ok_or_else(..)` (cursor:18), `let … else` (cursor:188), plain binding (ledger:192), `else if let Some` (store:59). No `Option` is collapsed, no error path is added or removed, no `?` changes what it propagates. Zero control-flow deltas.
5. **Source citations are exact.** `dirs 6.0.0 src/win.rs:5` is `pub fn home_dir() -> Option<PathBuf> { dirs_sys::known_folder_profile() }` — line 5, verbatim. `dirs-sys 0.5.0 src/lib.rs:172` is `pub fn known_folder_profile() -> Option<PathBuf> { known_folder(Shell::FOLDERID_Profile) }` — line 172, verbatim. Both read from the vendored registry sources this workspace compiles. The doc comment's platform claim is correct.
6. **Layering.** `crates/edda-core/Cargo.toml` at `895d00b` still has **zero** in-workspace dependencies, so no cycle is possible; `cargo` accepted the graph. The two new edges cost nothing in dependency weight: `edda-bridge-cursor` and `edda-transcript` both already depend on `edda-store`, which has depended on `edda-core` all along, so `edda-core` was already in their transitive graph — the new lines make an existing edge direct. `dirs = "6"` is an external crates.io crate already in the workspace, so `edda-core` gains no new publish-order constraint for the crates.io work in #648.
7. `cargo test -p` over the five crates outside the CI Windows subset — exit 0, 284 tests, 0 failures (see C5).
8. `sh scripts/lint-markdown-content.sh` → exit 0. `sh scripts/wiring-scan.sh origin/main 895d00b` → `paths (mod)` and `home_dir (fn)` with 10 outside references, no "no consumer" row for this diff.
9. `gh pr checks 812`, `gh issue view 808 --comments`, `edda ask fleet.lane-launcher`, `edda ask fleet.lane-launch`, and the §3 classifier block from `REVIEW.md` at the base SHA (`classes=code-risk code-plain`, `canonical_class=code-risk`).

**READ** (not re-run):

- The L1 gate receipt in the PR body — full SHA `895d00b`, clean tree, `CARGO_INCREMENTAL=0`, rustc 1.93.1, worktree default `target/`: `cargo fmt --all --check` pass, `cargo clippy --workspace --all-targets -- -D warnings` pass, `cargo test --workspace` 2680 passed / 0 failed / 56 suites. Covers everything CI covers plus the Windows test targets of all 16 crates CI skips on Windows. Not covered by it: nothing in this diff — the C5 rerun above was spent on the five touched crates in that gap, for the reason recorded in the U6 row, and no other recorded gate was repeated.
- Exact-head CI, 9/9 green: workspace clippy on Linux, macOS **and** Windows (so the new ban is enforced on all three), `cargo test --workspace` on Linux and macOS, and the 7-crate Windows test subset. What CI does **not** cover is the Windows test targets of the other 16 crates — the gap C5 exists for.
- Issue #808 body and its two author comments (the narrowing, and the `clippy.toml`-already-exists correction). Treated as the amended acceptance ceiling, and each factual claim in them was checked against code rather than accepted.

### Verdict

LGTM (P0=0, P1=0) — the highest-risk item, the `clippy.toml` restoration, is complete and correct: the base file's single line survives byte-identical, the diff deletes nothing, and the single-commit branch has no tree in which the GH-779 ratchet was missing. The ban is real, not decorative — both entries resolve and fire, against six different reintroduction shapes including aliased imports, re-exports and bare function items. All ten rewrites are control-flow-identical to their base hunks. The dropped spawn ban was checked against the actual nine sites and against the ledger decision it cited, and dropping it is the right call even though the census backing it miscounts. Three P2s and two follow-ups are recorded above; none blocks. Merge remains the operator's action.

Issue #884, opened 2026-09-04
design(fleet): requalify glm-5.3-flash on canary v0 under brief v2 — R2 shell parse tree is zero-discretion, severity by table; the recorded code-risk disqualification is a brief artifact (design #618 §3 learning 1, §7 item 4)


fagemx:
Operator direction 2026-09-04 (Tim): flash-tier engines are not as weak as the current routing assumes; they need guidance. The record agrees, and the recorded disqualification of glm-5.3-flash for code-risk is a brief artifact that the design document itself already diagnosed and scheduled for a fix that was never made.

## What happened

Basis `origin/main` `6486228c728dff5d488b5756e918af0bccde0eb5`.

**The calibration record.** `docs/superpowers/specs/2026-09-02-substitutable-reviewer-design.md`:

```
:91   | c1-shell-precedence | P0 | caught … | escalated——解析樹與觸發條件全對（含「fast 成功路徑也會刪」…），但按 [判斷] 規則只標需升級、未列 finding | …
:43   | glm-5.3-flash | … | docs-skills（provisional）；code-risk 不合格 | … | code-risk 不合格的原因與 brief v1 的 [判斷] 標籤有關（§3 學習 1），brief v2 修正後重校，不是引擎本身判死 |
:190  1. brief v1 把 shell 解析樹標成 [判斷] 是誤標：解析樹與觸發條件是機械可驗的（glm 全對卻因 [判斷] 規則不能列 finding，P0 閘因此 1/2…）。v2 應把「寫解析樹＋觸發條件」移為零裁量項，只留嚴重度裁量。
:292  4. brief v2——把「shell 解析樹＋觸發條件」移出 [判斷]（校準學習 1）。
```

The brief rule that produced the miss, `docs/superpowers/specs/2026-09-02-reviewer-brief-template-v1.md`:

```
:113  1. shell／腳本：`||` 與 `&&` 混用時逐行寫出解析樹；任何刪除/覆寫類指令 … 標出確切觸發條件。`[判斷]`
      規則 2. 標 [判斷] 的項目：若你是清單型引擎，只能標「需升級」，不得自行裁定。
```

**What happened after.** Nothing in §7 item 4 was done, and the canon carried the mislabel forward as if it were a capability fact:

```
REVIEW.md:540   **R2 — shell operator precedence. `[判斷]`.** … it is the item the cheap engines escalate rather than adjudicate — see §6.
[...]

PR #616, opened 2026-09-02
docs(plan): unified roadmap — one numbering, two tracks (loop objects ∥ trust objects), stage exit tests

fagemx:
## 這是什麼

`docs/plan/roadmap.md` — **唯一的階段編號正本**。2026-09-02 的討論產生了三套編號（策略改進 A–D、終點 0–3、全自動化「閉環段」），操作者要求重排；本頁合成一個編號、兩條軌：**0 自舉可靠 → 1a 迴圈物件（#560）∥ 1b 信任物件（#612）→ 2 閉環 → 3a 開源／3b 企業 → 4 策略選擇可學**，每階段的出口測試＝下一階段門票（「蓋與考同梯」），三條排序規則（1a 設計餵 1b 規格；2 需要 1a 身分與 1b 規格但不需要簽章；簽章是對外階段的門票），七項第 0 段出口測試與 2026-09-02 的量測（0.5/7），舊編號對照表。destinations 設計文件 §4 加一行指向本頁。

## 驗證（docs-only 階梯）

- 無 code / product blob / Cargo.lock / toolchain 變更 → 本地不跑 Cargo 閘門，exact-head CI 為閘。
- `sh scripts/lint-markdown-content.sh` 本地 exit 0。
- 本頁每個 issue 號都在 2026-09-02 開立於本 repo（#560、#574–#612 區間）；不關閉任何 issue。

## 作者身分

規劃 session（控制者角色）起草。**作者已在 #600 兩輪被抓到「文件裡沒實跑的指令宣稱」；本頁不含任何指令宣稱，只有階段、出口與 issue 號。** 審查請以 issue 號存在性、與 #612／#560 body 的一致性、與 `product.*` 決策的一致性為清單。

PR #915, opened 2026-09-05
feat(fleet): L0 rule runner runs every marked REVIEW.md check block as one pass (GH-882)

fagemx:
Closes #882.

Issue: #882

## What

`scripts/review-l0.sh` extracts every marked check block from `REVIEW.md` and runs them against a diff as one pre-push pass. `REVIEW.md` gains the block markers; `scripts/test-review-l0.sh` covers extraction and execution; the fleet-worker skill and the brief template point lanes at it.

## Why now

Measured on the all-flash window's day 0: a review round costs **11–19 minutes of wall clock** and \$0.03–0.06. Three of the eight rounds run that day resolved to a single mechanical P1 — a shape a script catches in seconds. The flash tier is cheap in dollars and expensive in *rounds*, so the highest-value optimisation is eliminating rounds that teach nothing, not shaving cost inside one.

Running this before push means the reviewer only ever pays for findings that need judgement.

## Gates (L0, touched surface is shell + markdown — no crate touched)

- `sh -n scripts/review-l0.sh` — exit 0
- `sh -n scripts/test-review-l0.sh` — exit 0
- `sh scripts/test-review-l0.sh` — all checks pass
- `sh scripts/lint-file-length.sh --tree` — exit 0
- pre-commit hooks (doc citations, markdown content, file length) passed on commit

Exact-head CI is the L1 gate.

## Note on review routing

This PR changes `REVIEW.md`. Under the amended window rule on #888, a glm-only round must not self-certify the rules it is judged by (`review.brief-source`) — so this one waits for an Opus round, not a flash SHADOW round.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #749, opened 2026-09-03
design(fleet): layered review - script for zero-discretion checks, flash-tier dossier and verification, Opus only for hypothesis and verdict; validated on the #618 calibration set


fagemx:
## What happened

Operator direction 2026-09-03: split review the way development is split — the parts that need judgment go to a high-tier model, the parts that are retrieval or mechanical go to a script or a flash-tier model. Evidence that the split is real and where the cost sits:

- **Calibration v0** (#618 / #638, `fleet.review-calibration`): target 5 files / 57 lines. Opus $1.4869, gpt-5.6-sol $0.0798, glm-5.3-flash $0.00092. glm qualified for docs/skills only — it missed c1, the finding that needs escalation judgment — and is **not** qualified for code-risk; sol and Opus caught c1–c5 but both under-rated c4/c5 severity.
- **Cost is procedure, not judgment.** 57 lines cannot cost $1.49 in judgment. The Opus lane reads a ~36 KB brief plus `REVIEW.md`, runs every backticked `gh` / `git` / `edda` command (contract item 5), collects callers, maps doneWhen items to hunks — zero-discretion and retrieval work billed at Opus rates.
- **What only the judgment layer produces:** #713 P1-3 (→ #741): the reviewer noticed on its own that a discarded `Err` arm turns a reconciled save into a silent overwrite. Nobody asked that question.
- **What a cheap verification layer would have caught:** #684 R1 false P0 from a stale installed binary (#693, #746) — a SHA comparison, not a judgment.
- **Operator's characterization of flash-tier models (2026-09-03):** execution is increasingly reliable; initiative and judgment are not. Given exhaustive instructions they complete them; **what is not said will not be done.** Design consequence: every flash-layer task is enumerated, schema-fixed and exhaustive by construction; no step says "as needed" or "if relevant"; every discretionary item routes to the judgment layer.

## Why it matters

Independent review is the only layer catching real defects (`fleet.agent-model-split` rationale), and #742 makes it mandatory on every PR. At $1.5–3 per round × rounds × PRs/day the gate becomes the throughput ceiling unless the Opus share shrinks. The split also improves quality: a dossier narrows Opus's context to the relevant surface, and a verification layer catches Opus false positives and severity drift.

## Proposed design — four layers, deterministic orchestration

| Layer | Engine | Input → output | Rules |
|---|---|---|---|
[...]

PR #916, opened 2026-09-05
feat(fleet): canary calibration runner scripts the README recipe (GH-881)

### LINKED ISSUE #881: feat(fleet): canary calibration runner — script the tests/canaries README recipe: per-engine read-only runs, model_observed from the system, machine-scored table for the ledger (design #618 §7 item 7)
Operator direction 2026-09-04 (Tim): build the mechanism that lets flash-tier engines take over review work by measurement, class by class. The canary set is that mechanism's instrument; today it has no runner.

## What happened

Basis `origin/main` `6486228c728dff5d488b5756e918af0bccde0eb5`.

`tests/canaries/` holds calibration set v0 — five known-answer diffs (c1–c5), merged 2026-09-02 in #638 — and `tests/canaries/README.md` §「如何跑一次校準」 is a hand-typed recipe: throwaway clone in `$TMPDIR`, one fixture commit, one canary commit, `git diff HEAD~1..HEAD`, then one read-only run per engine and hand-scoring against each `expected.md`. Nothing executes it:

```
git ls-tree -r --name-only origin/main | grep -iE 'canar|calibrat'   → only tests/canaries/** and the two design docs
grep -rn canaries scripts/ .github/                                   → no runner (the only hits are unrelated test-fixture "canary" strings in test-review-capabilities.sh and the lane-privilege spike)
```

Consequences already on record:

- It ran once (2026-09-02, `fleet.review-calibration` v0). The Opus remeasurement that `fleet.review-engine-model` (2026-09-02) itself requires has not happened; the ledger has no calibration note after 09-02.
- `docs/superpowers/specs/2026-09-02-substitutable-reviewer-design.md` §7 item 7 lists this exact gap: 「金絲雀重校的自動化——把 §1.2 的跑法變成腳本／lane…目前是手動程序＋README」.
- The 「線只升不降」 rule (a finding one engine catches and another misses becomes a new canary) is also manual; the set is still five.
- `expected.md` is prose (class, severity, one-line finding, scoring hint), so "caught" is scored by a person; #749's doneWhen needs the same table rerun through a layered pipeline and has nothing to run.

## Why it matters

Engine substitution is decided by qualification, not brand (`fleet.review-engine=replaceable-by-qualification-not-brand`), and qualification comes only from this table. Without a runner, every promotion of a cheap engine is an argument instead of a measurement, and the Opus-only cost structure (calibration v0: $1.49 vs $0.0009 per canary run) stays by default. A run costs fractions of a cent for glm; the blocker is the manual procedure, not money.

## Suspected surface

- `scripts/calibrate-canaries.sh` (new) — the README recipe as a script: build the throwaway clone outside the worktree, apply fixtures then patches, produce the target diff, launch each requested engine once per run read-only, collect `model_observed` from the system (pi session file `"model"` field; claude `--output-format json` `modelUsage` key), score, print a Markdown table plus a `for-ledger` block, delete the clone.
- `tests/canaries/*/expected.md` — add a fixed YAML front matter (`id`, `class`, `severity`, `file`, `match` regex) so the script can score `caught` mechanically; prose stays for humans.
- `tests/canaries/README.md` — point at the script; keep the recipe as the description of what the script does.
- `scripts/test-calibrate-canaries.sh` (new) — offline test that stubs the engine with fixture transcripts.

## Predicted surface

`scripts/calibrate-canaries.sh`, `scripts/test-calibrate-canaries.sh`, `tests/canaries/README.md`, `tests/canaries/code-risk/*/expected.md`, `tests/canaries/docs-skills/*/expected.md`. No crate.

## doneWhen

- `sh scripts/calibrate-canaries.sh --engine pi:openrouter/z-ai/glm-5.3-flash --brief <path> --runs 3 --dry-run` prints the clone path it would create under `$TMPDIR`, the two commits it would make, the target diff stat, and the exact launch line per run (pi: `--tools read,grep,find,ls --session-id calib-<engine>-<run>`; claude: the review allowlist), and launches nothing.
- Engine selector grammar is `<backend>:<catalogue id>` with the id copied verbatim (design §3 learning 3: fuzzy resolution is a silent substitution); `claude:` with any non-Anthropic id and `pi:` with any Anthropic id exit 2 before launch (`fleet.claude-subscription-transport`).
- A real run writes one row per (canary × run): `caught | missed | false-positive`, `severity_match yes|no`, `model_requested`, `model_observed`, `cost_usd`, plus a `union` row per canary across runs; `model_observed` comes from the session file / JSON, never from the transcript's own text (#616).
- Every `expected.md` carries the front matter and `grep -L '^id:' tests/canaries/*/*/expected.md` is empty.
- `sh scripts/test-calibrate-canaries.sh` stubs `pi` and `claude` with fixture transcripts and asserts: caught, missed, false positive, severity mismatch, and `model_observed` mismatch → row marked `void` (not silently scored); it writes nothing outside its temp dir and exits 0.
- The script never posts to GitHub, never writes to the ledger (it prints the `for-ledger` block; the controller records), and removes its clone on every exit path (`trap`).
- `sh -n` on both scripts exits 0.

## Wiring audit

| Component | 1. Writer & shape | 2. Reader | 3. Failure signal | 4. Layer reach |
|---|---|---|---|---|
| `expected.md` front matter | canary author; fixed keys `id class severity file match` | `calibrate-canaries.sh` scorer | missing key → script exits 2 naming the file (no silent "missed") | file → scorer → table |
| results table + `for-ledger` block | script stdout | controller → `edda decide fleet.review-calibration=...`; #749 acceptance | engine exit ≠ 0 or `model_observed` mismatch → row `void`, run continues, exit code 1 at the end | script → PR body → ledger |
| throwaway clone | script, `$TMPDIR/edda-calib-$$` | engine cwd | `trap` removes it on exit/INT/TERM; a leftover directory is a finding | script only |

- #618 / #638 — the set and the design; this is design §7 item 7, unfiled until now.
- #749 — its calibration rerun and negative control run through this script; not a duplicate (that issue designs layers, this one runs canaries).
- #880 — the pi arm for production review; this script launches engines directly and does not depend on it.
- #884 (glm requalification) consumes this runner but can fall back to the README recipe; no `blocked-by`.
- #746 / #693 — binary identity for "RAN on this workstation" claims; out of scope here.
- Decisions: `fleet.review-calibration`, `fleet.review-engine-pool`, `review.execution-policy`, `fleet.claude-subscription-transport`.
- Dedupe ran: `gh issue list --search calibration` → #749 #652 #766 #742 (no runner); `--search canary` → #879 #749 #652 #613 (none); `edda ask "review-calibration"` → v0 record only.

fagemx:
Issue: #881
Closes #881.

## What

`scripts/calibrate-canaries.sh` turns the `tests/canaries/README.md` calibration recipe into one command. Per engine it builds a throwaway clone, commits the fixture pre-state and then the canary set (the README's two commits exactly), launches the engine **read-only** once, reads `model_observed` from the system, scores each canary mechanically against its `expected.md` front matter, and prints the results table plus a for-ledger block.

`expected.md` gains machine-readable front matter, so scoring stops being a human reading of the prose rubric.

## Two design points worth flagging

**`model_requested != model_observed` voids the run and exits 1.** A calibration scored against the wrong model is worse than no calibration — it produces a number that looks like evidence. `model_observed` is read from the system (pi session file `model` field; `claude --output-format json` `modelUsage`), never from the transcript's own text (#616).

**It never writes.** No GitHub post, no ledger write, no `edda decide` — it prints the for-ledger block and the controller records it (design #618 §3).

## Why this matters for the window

The all-flash window's whole premise is a comparison, and right now the comparison rests on hand-run calibrations. #884 records that the previously logged glm code-risk disqualification was a **brief artifact**, not a model failure — a scoring result that was wrong because the brief was written wrong, and that nobody could re-run cheaply enough to catch. This is the script that makes re-running cheap.

## Gates (L0, touched surface is shell + markdown — no crate touched)

- `sh -n scripts/calibrate-canaries.sh` — exit 0
- `sh -n scripts/test-calibrate-canaries.sh` — exit 0
- `sh scripts/test-calibrate-canaries.sh` — all checks pass, including the void-on-mismatch path
- `sh scripts/lint-file-length.sh --tree` — exit 0
- pre-commit hooks (doc citations, markdown content, file length) passed on commit

Exact-head CI is the L1 gate.

## Surface

`scripts/**` plus `tests/canaries/**`. No `.github/**` job and no `lefthook.yml` entry invokes either script, and nothing here ships to a user.

### DIFF
--- a/scripts/calibrate-canaries.sh
+++ b/scripts/calibrate-canaries.sh
@@ -1,0 +1,563 @@
+#!/bin/sh
+# calibrate-canaries.sh — the tests/canaries/README.md §「如何跑一次校準」
+# recipe as a script (issue #881; design #618 §1.2 and §7 item 7).
+#
+# For each requested engine (<backend>:<catalogue id>, id copied verbatim) it
+# builds a throwaway clone under ${TMPDIR:-/tmp}/edda-calib-$$, commits the
+# canary fixture pre-state, then commits the canary set (two commits, exactly
+# the README recipe), produces the target diff HEAD~1..HEAD, launches the
+# engine READ-ONLY once per run with cwd = the clone, collects model_observed
+# from the system (pi session file "model" field; claude --output-format json
+# "modelUsage" key — never from the transcript's own text, #616), scores each
+# canary mechanically against its expected.md front matter, and prints a
+# Markdown results table plus a for-ledger block. The clone is removed on
+# every exit path (trap on EXIT/INT/TERM).
+#
+# The script is calibration plumbing only: it never posts to GitHub, never
+# writes to the ledger and never runs `edda decide` — it prints the for-ledger
+# block; the controller records it (design §3: 本 lane 不執行 edda decide).
+#
+# Launch lines follow the review transport conventions (README recipe and
+# scripts/review-pr.sh): pi is read-only via the --tools read,grep,find,ls
+# allowlist; claude via the review allowlist
+# --allowedTools "Read,Grep,Glob,Bash(git *),Bash(sh *)".
+# Opus / Anthropic ids are never sent through pi
+# (fleet.claude-subscription-transport=claude-code-only): the check is data —
+# the id set is parsed out of the pi catalogue (`pi --list-models`), not
+# hard-coded (design §3 learning 3: ids are matched, never fuzzy-resolved).
+#
+# Scoring (mechanical, per tests/canaries/<class>/<name>/expected.md front
+# matter keys id/class/severity/file/match):
+#   caught          a FINDING line on the expected file whose text matches
+#                   the `match` regex (ERE);
+#   false-positive  no such finding, but a FINDING line on the canary's
+#                   surface (the expected file or anything under its
+#                   directory) that is not the expected finding;
+#   missed          otherwise.
+# severity_match compares the caught finding's reported severity with the
+# front-matter severity. A run whose engine exit ≠ 0 or whose model_observed
+# differs from model_requested marks every row of that run `void` — never
+# silently scored (model_requested ≠ model_observed is a P0 incident).
+# The mechanical score is conservative: human scoring per the expected.md
+# 評分提示 remains the authority for qualification decisions.
+#
+# usage: calibrate-canaries.sh --engine <backend>:<id> [--engine ...]
+#                              --brief <path> [--runs N] [--dry-run]
+#
+# exit: 0 all runs scored clean; 1 any run void/failed (all runs still
+#       complete); 2 usage / selector / expected.md front-matter errors —
+#       always before any engine launch.
+set -u
+
+usage() {
+  cat >&2 <<'EOF'
+usage: calibrate-canaries.sh --engine <backend>:<id> [--engine ...]
+                             --brief <path> [--runs N] [--dry-run]
+  --engine   repeatable; <backend> is `pi` or `claude`; <id> is a catalogue
+             id copied verbatim (e.g. pi:openrouter/z-ai/glm-5.3-flash)
+  --brief    reviewer brief path (reviewer-brief-template-v1 filled); the
+             script appends the mechanical-scoring output protocol to it
+  --runs     runs per engine (default 1)
+  --dry-run  print the clone path, the two commits, the target diff stat and
+             the exact launch line per run; launch nothing
+EOF
+}
+
+die() { # die <code> <message...>
+  code=$1; shift
+  printf 'calibrate-canaries.sh: %s\n' "$*" >&2
+  exit "$code"
+}
+
+ENGINES=''
+BRIEF=''
+RUNS=1
+DRY_RUN=0
+
+while [ $# -gt 0 ]; do
+  case $1 in
+    --engine)
+      [ $# -ge 2 ] || { usage; exit 2; }
+      ENGINES="$ENGINES
+$2"
+      shift 2 ;;
+    --engine=*)
+      ENGINES="$ENGINES
+${1#*=}"
+      shift ;;
+    --brief)
+      [ $# -ge 2 ] || { usage; exit 2; }
+      BRIEF=$2
+      shift 2 ;;
+    --brief=*)
+      BRIEF=${1#*=}
+      shift ;;
+    --runs)
+      [ $# -ge 2 ] || { usage; exit 2; }
+      RUNS=$2
+      shift 2 ;;
+    --runs=*)
+      RUNS=${1#*=}
+      shift ;;
+    --dry-run)
+      DRY_RUN=1
+      shift ;;
+    -h|--help)
+      usage
+      exit 0 ;;
+    *)
+      usage
+      exit 2 ;;
+  esac
+done
+
+[ -n "$BRIEF" ] || { usage; exit 2; }
+[ -f "$BRIEF" ] && [ -r "$BRIEF" ] || die 2 "brief not readable: $BRIEF"
+case $RUNS in
+  ''|*[!0-9]*) die 2 "--runs must be a positive integer: $RUNS" ;;
+esac
+[ "$RUNS" -ge 1 ] || die 2 "--runs must be a positive integer: $RUNS"
+ENGINES=${ENGINES#?
+}
+[ -n "$ENGINES" ] || { usage; exit 2; }
+
+SELF_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd) || exit 2
+WT=$(CDPATH= cd -- "$SELF_DIR/.." && pwd) || exit 2
+CANARIES_DIR="$WT/tests/canaries"
+[ -d "$CANARIES_DIR" ] || die 2 "canary set not found: $CANARIES_DIR"
+
+# ---------------------------------------------------------------------------
+# Discover canaries and validate the expected.md front matter (wiring audit
+# row 1: a missing key is a hard exit 2 naming the file — never a silent
+# "missed").
+# ---------------------------------------------------------------------------
+front_matter_field() { # <expected.md> <key> -> value
+  # Single awk process reading the file directly: a `sed | ... | head -n 1`
+  # pipeline has an early-exit reader and is prone to MSYS/Git Bash pipe
+  # deadlocks (observed as a hang in the validation loop on Windows).
+  awk -v key="$2" '
+    NR == 1 { next }
+    /^---[[:space:]]*$/ { exit }
+    index($0, key ":") == 1 {
+      v = $0
+      sub("^" key ":[[:space:]]*", "", v)
+      sub(/\r$/, "", v)
+      if (v ~ /^\047/) { sub(/^\047/, "", v); sub(/\047$/, "", v) }
+      else if (v ~ /^\042/) { sub(/^\042/, "", v); sub(/\042$/, "", v) }
+      print v
+      exit
+    }' "$1"
+}
+
+CANARIES=$(cd "$CANARIES_DIR" && ls -d */*/ 2>/dev/null | sed 's/\/$//;s/\r$//' | sort) || CANARIES=''
+[ -n "$CANARIES" ] || die 2 "no canary directories under $CANARIES_DIR"
+
+for cdir in $CANARIES; do
+  exp="$CANARIES_DIR/$cdir/expected.md"
+  [ -f "$exp" ] || die 2 "missing expected.md: $exp"
+  [ "$(sed -n '1{s/[[:space:]]*$//;p}' "$exp" | tr -d '\r')" = '---' ] \
+    || die 2 "expected.md without front matter (line 1 must be ---): $exp"
+  for key in id class severity file match; do
+    val=$(front_matter_field "$exp" "$key")
+    [ -n "$val" ] || die 2 "expected.md front matter missing key '$key': $exp"
+  done
+  sev=$(front_matter_field "$exp" severity)
+  case $sev in
+    P[0-3]) ;;
+    *) die 2 "expected.md severity must be P0..P3, got '$sev': $exp" ;;
+  esac
+done
+
+# ---------------------------------------------------------------------------
+# Engine selector validation — data, not vibes: the Anthropic-id set is
+# parsed from the pi catalogue. `claude:` + non-Anthropic id and `pi:` +
+# Anthropic id exit 2 before any launch
+# (fleet.claude-subscription-transport=claude-code-only).
+# ---------------------------------------------------------------------------
+load_catalogue() {
+  if command -v pi >/dev/null 2>&1; then
+    pi --list-models 2>/dev/null
+  elif command -v edda >/dev/null 2>&1; then
+    edda dispatch --agent pi --list-models 2>/dev/null
+  else
+    return 1
+  fi
+}
+
+CATALOGUE=$(load_catalogue) \
+  || die 2 "cannot validate engine selectors: pi catalogue unavailable (install pi or edda)"
+
+# rows: "<A|N> <pi selector id>" — A = Anthropic (pi transport forbidden),
+# N = non-Anthropic (claude transport forbidden).
+ID_CLASSES=$(printf '%s\n' "$CATALOGUE" | awk '
+  NF >= 2 && $1 != "provider" {
+    if ($1 == "anthropic") { print "A " $2; print "A anthropic/" $2 }
+    else if ($2 ~ /^~?anthropic\//) { print "A " $1 "/" $2 }
+    else print "N " $1 "/" $2
+  }')
+
+id_class() { # <id> -> A | N | ''
+  printf '%s\n' "$ID_CLASSES" | awk -v id="$1" '$2 == id { print $1; exit }'
+}
+
+sanitize() { # engine id -> session-id-safe token
+  printf '%s' "$1" | sed 's/[^A-Za-z0-9._-]/-/g'
+}
+
+VALIDATED=''
+for sel in $ENGINES; do
+  case $sel in
+    *:*) backend=${sel%%:*}; id=${sel#*:} ;;
+    *) die 2 "engine selector grammar is <backend>:<catalogue id>, got: $sel" ;;
+  esac
+  case $backend in
+    pi|claude) ;;
+    *) die 2 "unknown backend '$backend' in selector $sel (pi|claude)" ;;
+  esac
+  [ -n "$id" ] || die 2 "empty catalogue id in selector $sel"
+  case $id in
+    *:*) die 2 "catalogue id must not contain ':' (copy the id verbatim from pi --list-models): $id" ;;
+  esac
+  cls=$(id_class "$id")
+  if [ "$backend" = pi ] && [ "$cls" = A ]; then
+    die 2 "refusing pi transport for Anthropic id '$id' (fleet.claude-subscription-transport=claude-code-only)"
+  fi
+  if [ "$backend" = claude ] && [ "$cls" = N ]; then
+    die 2 "refusing claude transport for non-Anthropic id '$id' (fleet.claude-subscription-transport=claude-code-only)"
+  fi
+  VALIDATED="$VALIDATED
+$backend|$id"
+done
+VALIDATED=${VALIDATED#?
+}
+
+# ---------------------------------------------------------------------------
+# Launch-line builders (single source of truth for --dry-run and real runs).
+# ---------------------------------------------------------------------------
+pi_launch_line() { # <id> <san> <run> <clone>
+  printf '(cd "%s" && pi -p --model %s --tools read,grep,find,ls --session-dir "%s/sessions/r%s" --session-id calib-%s-%s "$(cat calib-brief.md)")' \
+    "$4" "$1" "$4" "$3" "$2" "$3"
+}
+
+claude_launch_line() { # <id> <san> <run> <clone>
+  printf '(cd "%s" && claude -p --model %s --allowedTools "Read,Grep,Glob,Bash(git *),Bash(sh *)" --output-format json < calib-brief.md)' \
+    "$4" "$1"
+}
+
+FIXTURE_COMMIT_MSG='calibration: canary fixture pre-state'
+CANARY_COMMIT_MSG='calibration: canary set v0'
+PATCHES=''
+for cdir in $CANARIES; do
+  PATCHES="$PATCHES
+$CANARIES_DIR/$cdir/diff.patch"
+done
+PATCHES=${PATCHES#?
+}
+
+DRY_TARGET_STAT=''
+for p in $PATCHES; do
+  [ -f "$p" ] || die 2 "missing canary patch: $p"
+  DRY_TARGET_STAT="$DRY_TARGET_STAT$(cat "$p")
+"
+done
+DRY_TARGET_STAT=$(printf '%s' "$DRY_TARGET_STAT" | git apply --stat - 2>/dev/null)
+
+# ---------------------------------------------------------------------------
+# Dry run: describe everything, launch nothing (not even a review run).
+# ---------------------------------------------------------------------------
+if [ "$DRY_RUN" -eq 1 ]; then
+  echo "clone: ${TMPDIR:-/tmp}/edda-calib-$$  (branch calib-canary-v0, from origin/main)"
+  echo
+  echo "commit 1: \"$FIXTURE_COMMIT_MSG\""
+  for cdir in $CANARIES; do
+    if [ -d "$CANARIES_DIR/$cdir/fixture" ]; then
+      echo "  cp -r tests/canaries/$cdir/fixture -> canaries-fixture/$(basename "$cdir")"
+    fi
+  done
+  echo "commit 2: \"$CANARY_COMMIT_MSG\""
+  for cdir in $CANARIES; do
+    echo "  git apply tests/canaries/$cdir/diff.patch"
+  done
+  echo
+  echo "target diff (HEAD~1..HEAD) stat:"
+  printf '%s\n' "$DRY_TARGET_STAT"
+  echo
+  echo "brief: $BRIEF (a fixed output-protocol addendum for mechanical scoring is appended)"
+  echo
+  for pair in $VALIDATED; do
+    backend=${pair%%|*}
+    id=${pair#*|}
+    san=$(sanitize "$id")
+    run=1
+    while [ "$run" -le "$RUNS" ]; do
+      if [ "$backend" = pi ]; then
+        pi_launch_line "$id" "$san" "$run" '${TMPDIR:-/tmp}/edda-calib-'"$$"
+      else
+        claude_launch_line "$id" "$san" "$run" '${TMPDIR:-/tmp}/edda-calib-'"$$"
+      fi
+      echo "  ^ run $run/$RUNS for $backend:$id"
+      run=$((run + 1))
+    done
+  done
+  exit 0
+fi
+
+# ---------------------------------------------------------------------------
+# Real run: throwaway clone, two commits, target diff, per-run launches.
+# ---------------------------------------------------------------------------
+CLONE="${TMPDIR:-/tmp}/edda-calib-$$"
+cleanup() { rm -rf -- "$CLONE"; }
+trap cleanup EXIT
+trap 'cleanup; trap - EXIT; exit 130' INT
+trap 'cleanup; trap - EXIT; exit 143' TERM
+
+git clone --quiet "$WT" "$CLONE" \
+  || die 2 "git clone $WT -> $CLONE failed"
+cd "$CLONE" || die 2 "cd $CLONE failed"
+git checkout -q -B calib-canary-v0 origin/main 2>/dev/null \
+  || git checkout -q -B calib-canary-v0 \
+  || die 2 "checkout calib-canary-v0 failed"
+
+# Commit 1 — fixture pre-state (fact sources land before the review target).
+mkdir -p canaries-fixture
+for cdir in $CANARIES; do
+  if [ -d "$CANARIES_DIR/$cdir/fixture" ]; then
+    cp -r "$CANARIES_DIR/$cdir/fixture" "canaries-fixture/$(basename "$cdir")"
+  fi
+done
+if [ -n "$(git status --porcelain -- canaries-fixture)" ]; then
+  git add canaries-fixture
+  git commit -q -m "$FIXTURE_COMMIT_MSG" || die 2 "fixture pre-state commit failed"
+fi
+
+# Commit 2 — the canary set (the review target).
+for cdir in $CANARIES; do
+  git apply "$CANARIES_DIR/$cdir/diff.patch" \
+    || die 2 "git apply failed for $cdir"
+done
+git add -A
+git commit -q -m "$CANARY_COMMIT_MSG" || die 2 "canary commit failed"
+
+echo "clone: $CLONE"
+echo "target diff (HEAD~1..HEAD) stat:"
+git diff --stat HEAD~1..HEAD
+echo
+
+# Brief = the supplied brief verbatim + the output-protocol addendum that the
+# mechanical scorer parses (FINDING lines; everything else is narrative).
+{
+  cat "$BRIEF"
+  cat <<'EOF'
+
+## 輸出協定（校準評分——機械解析，務必遵守）
+
+每個 finding 用一行固定格式輸出（其餘敘述不會被計分）：
+
+FINDING P<n> <repo 相對路徑> — <一行 finding 描述>
+
+- P<n> 是你對該 finding 的 severity 判定（P0/P1/P2）。
+- <repo 相對路徑> 指向證據所在檔案（可帶 :行號）。
+- 沒有 finding 時，輸出一行：NO FINDINGS
+EOF
+} > calib-brief.md
+
+
+any_void=0
+
+score_run() { # <transcript> -> ROW lines (canary result severity_match) on stdout
+  transcript=$1
+  findings_file=$1.findings
+  sed -n -E 's/^FINDING[[:space:]]+(P[0-3])[[:space:]]+([^[:space:]]+)[[:space:]]+(-|—)[[:space:]]+(.*)$/\1\t\2\t\4/p' \
+    "$transcript" > "$findings_file" 2>/dev/null
+
+  for cdir in $CANARIES; do
+    exp="$CANARIES_DIR/$cdir/expected.md"
+    cid=$(front_matter_field "$exp" id)
+    sev=$(front_matter_field "$exp" severity)
+    file=$(front_matter_field "$exp" file)
+    regex=$(front_matter_field "$exp" match)
+    exp_dir=$(dirname "$file")
+
+    caught=0
+    sev_ok=0
+    on_surface=0
+    while IFS="$(printf '\t')" read -r fsev fpath ftext; do
+      [ -n "$fpath" ] || continue
+      ffile=${fpath%%:*}
+      case $ffile in
+        "$exp_dir"/*) on_surface=1 ;;
+      esac
+      case $ffile in
+        "$file")
+          if printf '%s\n' "$ftext" | grep -Eq -- "$regex" 2>/dev/null; then
+            caught=1
+            [ "$fsev" = "$sev" ] && sev_ok=1
+          else
+            on_surface=1
+          fi
+          ;;
+      esac
+    done < "$findings_file"
+
+    if [ "$caught" -eq 1 ]; then
+      result=caught
+      [ "$sev_ok" -eq 1 ] && smatch=yes || smatch=no
+    elif [ "$on_surface" -eq 1 ]; then
+      result=false-positive
+      smatch=-
+    else
+      result=missed
+      smatch=-
+    fi
+    printf 'ROW %s %s %s\n' "$cid" "$result" "$smatch"
+  done
+  rm -f "$findings_file"
+}
+
+for pair in $VALIDATED; do
+  backend=${pair%%|*}
+  id=${pair#*|}
+  san=$(sanitize "$id")
+  echo "## engine $backend:$id"
+  echo
+  echo '| canary | run | result | severity_match | model_requested | model_observed | cost_usd |'
+  echo '|---|---|---|---|---|---|---|'
+
+  union_file="$CLONE/union-$san"
+  : > "$union_file"
+  costs=''
+  last_observed=unknown
+  last_cost='-'
+
+  run=1
+  while [ "$run" -le "$RUNS" ]; do
+    out="$CLONE/out-$san-$run.txt"
+    sess_dir="$CLONE/sessions/r$run"
+    rm -rf "$sess_dir"
+
+    if [ "$backend" = pi ]; then
+      eval "$(pi_launch_line "$id" "$san" "$run" "$CLONE")" > "$out" 2>"$out.err"
+    else
+      eval "$(claude_launch_line "$id" "$san" "$run" "$CLONE")" > "$out.json" 2>"$out.err"
+    fi
+    exit_code=$?
+
+    # model_observed comes from the system — pi session file "model" field /
+    # claude JSON envelope modelUsage key — never from the transcript's own
+    # text (#616). A mismatch voids the run instead of being silently scored.
+    observed=unknown
+    cost=''
+    if [ "$backend" = pi ]; then
+      sess_file=$(ls -t "$sess_dir"/*.jsonl 2>/dev/null | head -n 1)
+      if [ -n "${sess_file:-}" ] && [ -f "$sess_file" ]; then
+        observed=$(grep -o '"model":"[^"]*"' "$sess_file" | tail -n 1 | sed 's/.*:"//;s/"$//')
+        [ -n "$observed" ] \
+          || observed=$(grep -o '"modelId":"[^"]*"' "$sess_file" | tail -n 1 | sed 's/.*:"//;s/"$//')
+        cost=$(grep -o '"cost":{[^}]*}' "$sess_file" | tail -n 1 \
+          | sed -n 's/.*"total":\([0-9.e+-]*\).*/\1/p')
+      fi
+    else
+      if command -v jq >/dev/null 2>&1; then
+        observed=$(jq -r '.modelUsage | keys[0] // "unknown"' "$out.json" 2>/dev/null)
+        cost=$(jq -r '.total_cost_usd // empty' "$out.json" 2>/dev/null)
+        jq -r '.result // empty' "$out.json" 2>/dev/null > "$out"
+      else
+        echo "calibrate-canaries.sh: jq not found on PATH — cannot read claude modelUsage/cost; run voided" >&2
+      fi
+    fi
+    [ -n "$observed" ] || observed=unknown
+
+    if [ "$exit_code" -ne 0 ]; then
+      any_void=1
+      reason="engine exit $exit_code"
+    elif [ "$observed" != "$id" ]; then
+      any_void=1
+      reason='model_observed mismatch (model_requested ≠ model_observed is a P0 incident)'
+    else
+      reason=''
+    fi
+
+    rows_file="$CLONE/rows-$san-$run"
+    if [ -n "$reason" ]; then
+      for cdir in $CANARIES; do
+        cid=$(front_matter_field "$CANARIES_DIR/$cdir/expected.md" id)
+        printf 'ROW %s void -\n' "$cid"
+      done > "$rows_file"
+      printf 'calibrate-canaries.sh: run void — %s:%s run %s (%s; requested=%s observed=%s)\n' \
+        "$backend" "$id" "$run" "$reason" "$id" "$observed" >&2
+    else
+      score_run "$out" > "$rows_file"
+    fi
+
+    if [ -n "$cost" ]; then costcell=$cost; else costcell='-'; fi
+    while IFS="$(printf ' 	')" read -r _tag cid result smatch; do
+      [ -n "$cid" ] || continue
+      printf '| %s | %s | %s | %s | %s | %s | %s |\n' \
+        "$cid" "$run" "$result" "$smatch" "$id" "$observed" "$costcell"
+      printf '%s %s %s %s\n' "$cid" "$result" "$smatch" "$costcell" >> "$union_file"
+    done < "$rows_file"
+    rm -f "$rows_file"
+
+    case $cost in
+      ''|*[!0-9.]*) ;;  # not a plain number — excluded from the sum
+      *) costs="$costs
+$cost" ;;
+    esac
+    last_observed=$observed
+    last_cost=$costcell
+    run=$((run + 1))
+  done
+
+  # Union row per canary across this engine's runs: caught beats
+  # false-positive beats missed; severity_match yes iff every caught run
+  # reported the expected severity.
+  for cdir in $CANARIES; do
+    exp="$CANARIES_DIR/$cdir/expected.md"
+    cid=$(front_matter_field "$exp" id)
+    ucaught=0; ufp=0; uallmatch=1
+    while IFS="$(printf ' \t')" read -r _cid uresult usmatch _ucost; do
+      [ -n "$uresult" ] || continue
+      case $uresult in
+        caught) ucaught=1; [ "$usmatch" = yes ] || uallmatch=0 ;;
+        false-positive) ufp=1 ;;
+      esac
+    done <<UNIONEOF
+$(grep "^$cid " "$union_file")
+UNIONEOF
+    if [ "$ucaught" -eq 1 ]; then
+      uresult=caught
+      [ "$uallmatch" -eq 1 ] && usmatch=yes || usmatch=no
+    elif [ "$ufp" -eq 1 ]; then
+      uresult=false-positive
+      usmatch=-
+    else
+      uresult=missed
+      usmatch=-
+    fi
+    usum=$(grep "^$cid " "$union_file" | awk '$4 ~ /^[0-9.]+$/ {s+=$4} END {if (s != "") printf "%.6f", s}')
+    [ -n "$usum" ] || usum='-'
+    printf '| %s | union | %s | %s | %s | %s | %s |\n' \
+      "$cid" "$uresult" "$usmatch" "$id" "$last_observed" "$usum"
+  done
+  rm -f "$union_file"
+
+  # for-ledger block — the controller records it; this script never runs
+  # `edda decide`, never writes the ledger, never posts to GitHub.
+  usum=$(printf '%s\n' "$costs" | awk 'NF && $1 ~ /^[0-9.]+$/ {s+=$1} END {if (s != "") printf "%.6f", s; else print "-"}')
+  echo
+  echo '## for-ledger — fleet.review-calibration'
+  echo
+  echo '本腳本不執行 edda decide、不寫帳本、不發 GitHub 請求；以下區塊由操作者逐字記帳本。'
+  echo
+  echo "engine: $backend:$id"
+  echo "requested: $id"
+  echo "observed: $last_observed"
+  echo "cost_usd: $usum"
+  echo "runs: $RUNS (void runs are marked void in the table)"
+  echo
+done
+
+if [ "$any_void" -eq 1 ]; then
+  exit 1
+fi
+exit 0
--- a/scripts/calibrate-canaries.sh
+++ b/scripts/calibrate-canaries.sh
@@ -266,7 +266,7 @@ DRY_TARGET_STAT=$(printf '%s' "$DRY_TARGET_STAT" | git apply --stat - 2>/dev/nul
 # Dry run: describe everything, launch nothing (not even a review run).
 # ---------------------------------------------------------------------------
 if [ "$DRY_RUN" -eq 1 ]; then
-  echo "clone: ${TMPDIR:-/tmp}/edda-calib.XXXXXX/repo  (branch calib-canary-v0, from origin/main)"
+  echo 'clone: ${TMPDIR:-/tmp}/edda-calib.XXXXXX/repo  (branch calib-canary-v0, from origin/main)'
   echo
   echo "commit 1: \"$FIXTURE_COMMIT_MSG\""
   for cdir in $CANARIES; do
--- a/scripts/calibrate-canaries.sh
+++ b/scripts/calibrate-canaries.sh
@@ -3,7 +3,7 @@
 # recipe as a script (issue #881; design #618 §1.2 and §7 item 7).
 #
 # For each requested engine (<backend>:<catalogue id>, id copied verbatim) it
-# builds a throwaway clone under ${TMPDIR:-/tmp}/edda-calib-$$, commits the
+# builds a throwaway clone under a fresh mktemp -d directory, commits the
 # canary fixture pre-state, then commits the canary set (two commits, exactly
 # the README recipe), produces the target diff HEAD~1..HEAD, launches the
 # engine READ-ONLY once per run with cwd = the clone, collects model_observed
@@ -266,7 +266,7 @@ DRY_TARGET_STAT=$(printf '%s' "$DRY_TARGET_STAT" | git apply --stat - 2>/dev/nul
 # Dry run: describe everything, launch nothing (not even a review run).
 # ---------------------------------------------------------------------------
 if [ "$DRY_RUN" -eq 1 ]; then
-  echo "clone: ${TMPDIR:-/tmp}/edda-calib-$$  (branch calib-canary-v0, from origin/main)"
+  echo "clone: ${TMPDIR:-/tmp}/edda-calib.XXXXXX/repo  (branch calib-canary-v0, from origin/main)"
   echo
   echo "commit 1: \"$FIXTURE_COMMIT_MSG\""
   for cdir in $CANARIES; do
@@ -291,9 +291,9 @@ if [ "$DRY_RUN" -eq 1 ]; then
     run=1
     while [ "$run" -le "$RUNS" ]; do
       if [ "$backend" = pi ]; then
-        pi_launch_line "$id" "$san" "$run" '${TMPDIR:-/tmp}/edda-calib-'"$$"
+        pi_launch_line "$id" "$san" "$run" '${TMPDIR:-/tmp}/edda-calib.XXXXXX/repo'
       else
-        claude_launch_line "$id" "$san" "$run" '${TMPDIR:-/tmp}/edda-calib-'"$$"
+        claude_launch_line "$id" "$san" "$run" '${TMPDIR:-/tmp}/edda-calib.XXXXXX/repo'
       fi
       echo "  ^ run $run/$RUNS for $backend:$id"
       run=$((run + 1))
@@ -305,8 +305,14 @@ fi
 # ---------------------------------------------------------------------------
 # Real run: throwaway clone, two commits, target diff, per-run launches.
 # ---------------------------------------------------------------------------
-CLONE="${TMPDIR:-/tmp}/edda-calib-$$"
-cleanup() { rm -rf -- "$CLONE"; }
+# A fresh mktemp -d, never "$$": PIDs are reused, the traps below fire on
+# every exit path including the failure path, and several lanes run at once on
+# this workstation — a PID collision would have one run delete another run's
+# directory. The clone goes in a subdirectory so the path is both unique and
+# ours, and cleanup removes the whole workspace rather than the clone alone.
+CALIB_TMP=$(mktemp -d "${TMPDIR:-/tmp}/edda-calib.XXXXXX") || die 2 "mktemp -d failed"
+CLONE="$CALIB_TMP/repo"
+cleanup() { rm -rf -- "$CALIB_TMP"; }
 trap cleanup EXIT
 trap 'cleanup; trap - EXIT; exit 130' INT
 trap 'cleanup; trap - EXIT; exit 143' TERM
--- a/scripts/test-calibrate-canaries.sh
+++ b/scripts/test-calibrate-canaries.sh
@@ -1,0 +1,334 @@
+#!/bin/sh
+# test-calibrate-canaries.sh — offline test for scripts/calibrate-canaries.sh
+# (issue #881). Stubs `pi` and `claude` with fixture transcripts on PATH and
+# asserts the mechanical scorer end to end:
+#
+#   1. sh -n passes on both scripts;
+#   2. every expected.md carries front matter (grep -L '^id:' is empty);
+#   3. --dry-run prints the clone path, both commits, the target diff stat and
+#      the exact launch line per run, and launches no engine run;
+#   4. selector grammar <backend>:<id>: pi + Anthropic id and claude +
+#      non-Anthropic id exit 2 before launch (the id set is data, parsed from
+#      the stubbed pi catalogue);
+#   5. scoring: caught, missed, false positive, severity mismatch
+#      (caught + severity_match no);
+#   6. model_observed from the session file / JSON — a mismatch marks the row
+#      void (never silently scored) and the script exits 1;
+#   7. the for-ledger block and the per-canary union row are printed;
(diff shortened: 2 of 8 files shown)


fagemx:
## Code Review: Round 1 — PR #916 @ c2dc3c3e2cf4db722770274775c08b5b6204bf69

- elapsed: unmeasured
- model_observed: claude-opus-5 (Claude Code operator session)
- reviewer_session: operator session on the docs machine
- spec: REVIEW.md @ base `6849b90b8b34b235592235f9b68686bf84defdb4`
- class: code-risk (`scripts/**`, `*.sh`); surface: 審判面 under `docs/fleet/rules.md` R22 — `tests/canaries/**` and a script that measures reviewers are review machinery, so this round is authoritative and a glm round here is SHADOW
- escalations: none
- cost: one operator session, no lane spend

**Reviewer conflict, disclosed.** I wrote #881, so its doneWhen is my own acceptance list. As on #915 and #921, this round does the mechanical check and then re-reads the deliverable as someone with no access to the issue — both findings below come from the second pass, not from the checklist.

**Cross-machine note.** #916 was built on the 4090 and its author cannot review it (#926). I am a different session on the docs machine and did not write this code, so #926's blocker does not apply to this PR.

### IN SCOPE

`scripts/calibrate-canaries.sh` (new, 563 lines), its offline test (334 lines), 40 lines of `tests/canaries/README.md`, and front matter on five `expected.md`. Consumers: #884's glm requalification, #749's calibration rerun, and the controller who copies the `for-ledger` block into `fleet.review-calibration`. Acceptance ceiling: #881's seven doneWhen items.

### What is right

This is a measuring instrument, and it behaves like one. The `--dry-run` prints the clone path, both commits with their exact file operations, the target diff stat (5 files, 57 insertions) and one verbatim launch line per run, and launches nothing — you can audit the whole experiment before spending a cent. `model_observed` is read from the **system**, never from the transcript: pi's session JSONL `"model"` (falling back to `"modelId"`), claude's `.modelUsage | keys[0]`, and a mismatch **voids** the run rather than scoring it (#616's rule, enforced rather than restated). Transport refusals are fail-closed before any launch — all four selector cases exit 2 with a reason. The clone is removed on EXIT, INT **and** TERM, and the offline suite asserts removal after every scenario. No `edda decide`, no `gh` write: the four ledger mentions are comments or printed text, and the only executed `edda` is a read-only `dispatch --list-models`.

### Rules

|---|---|---|---|---|
| U1 surface | any | P0 | PASS | eight changed paths == #881's `Predicted surface`, exactly |
| U2 closing keyword ⟺ doneWhen | any | P1 | PASS | `Closes #881`; all seven items delivered (below) |
| U3 `Issue: #N` | any | P1 | **FAIL** | see P1-1 |
| U4 conventional subject | any | P1 | PASS | one commit, `feat(fleet): …` |
| U5 markdown lint | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → exit 0 (RAN) |
| U6 gates READ before RAN | any | P0 | PASS | run `33958527510`: all ten jobs green including Test ×3 and Clippy ×3 |
| U7 whole scoped surface | any | P1 | PASS | both scripts read in full |
| D1 factual claims | docs | P1 | PASS | the README's commands resolve; `edda dispatch --list-models` is the only executed probe |
| D3 paths and links | docs | P1 | PASS | every path the README adds exists in this tree (see the note on the L0 run below — its `MISSING` is an artifact of where I ran it) |
| D4 authority boundary | docs | P0 | PASS | the script prints the `for-ledger` block and states in two places that the **controller** records it; it never ratifies |
| C1 doneWhen vs real code | code | P0 | PASS | seven items, six RAN below |
| C2 `feat:` with an integration test | code | P0 | PASS | `test-calibrate-canaries.sh` → exit 0, covering caught / missed / false-positive / severity mismatch / `model_observed` mismatch → void |
| R1 destructive commands | code-risk | P0 | **FAIL** | see P1-2 — the trigger is statable for four of the five hits, not for the fifth |
| R3 shell scripts parse | code-risk | P0 | PASS | `sh -n` on both → exit 0 (RAN) |
| R4 resource release | code-risk | P1 | PASS | `trap cleanup EXIT`, plus INT → 130 and TERM → 143, each removing the clone; the test uses `mktemp -d` and asserts removal per scenario |
| R5 boundary and injection | code-risk | P1 | PASS | the engine runs with `--tools read,grep,find,ls` in a throwaway clone; the selector is split on the first `:` and the catalogue id is passed verbatim, never fuzzy-resolved (#618 §3 learning 3) |

**doneWhen, item by item** (RAN at this head). (1) The dry-run printed all four required parts and launched nothing. (2) `claude:` with a glm id, `pi:` with `anthropic/claude-opus-5`, `pi:` with `claude-opus-5`, and a selector with no colon → **exit 2** each, with a message naming the rule. (3) Row shape verified through the offline suite's stubs — `| c1-shell-precedence | 1 | caught | yes |`, `| c2-stale-ratify-claim | 1 | missed | - |`, false-positive, and the void row carrying **both** models; the union row and its cost sum are assembled per canary. A real paid run is calibration, not review, and is #884's job. (4) `grep -L '^id:' tests/canaries/*/*/expected.md` → empty, across all five. (5) `sh scripts/test-calibrate-canaries.sh` → **exit 0**, `all checks passed`, writing only under its `mktemp -d`. (6) No ledger or GitHub write; `trap` on all three signals. (7) `sh -n` on both → exit 0.

### Findings

**P1-1 — U3: the PR body has no `Issue: #881` line.** It opens with `Closes #881.` and never states the conventional line; `gh pr view 916 --json body --jq .body | awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → empty. Same miss as #915 had, and the same one-line fix.

**P1-2 — R1: the clone path is predictable and the EXIT trap deletes it.**

```
scripts/calibrate-canaries.sh:308  CLONE="${TMPDIR:-/tmp}/edda-calib-$$"
scripts/calibrate-canaries.sh:309  cleanup() { rm -rf -- "$CLONE"; }
scripts/calibrate-canaries.sh:310  trap cleanup EXIT
```

Four of the five `rm -rf` hits in this diff have a trigger I can state — they remove a `mktemp -d` directory the script created. This one does not: `$$` is a PID, PIDs are reused, and the trap fires on **every** exit path including the failure path. So:

1. a calibration run leaves `/tmp/edda-calib-4711` behind (killed lane, full disk, `Ctrl-C` outside the trap's window);
2. a later run — or a concurrent lane on this workstation, where several run at once — draws PID 4711;
3. `git clone` into the non-empty directory fails, the script `die 2`s, and the EXIT trap deletes **the other directory's contents**.

The blast radius is a throwaway clone rather than user data, which is why this is P1 and not P0. The fix is one line and this PR already contains it: `scripts/test-calibrate-canaries.sh:31` uses `WORK=$(mktemp -d "${TMPDIR:-/tmp}/test-calibrate-canaries.XXXXXX") || exit 1`. Use the same shape for `CLONE`, then `git clone` into a subdirectory of it, so the path is guaranteed unique and guaranteed yours.

### The L0 runner's own reading of this PR

`scripts/review-l0.sh` merged today (#915), so this is its first use on a PR it did not ship with. Run as `sh scripts/review-l0.sh origin/main c2dc3c3e 916` **from a `main` checkout**:

| row | runner | truth |
|---|---|---|
| U3 | FAIL | **correct** — it found P1-1 mechanically, which I found by hand |
| U2 | `N.A.(needs reviewer input)`, evidence `1:Closes #881.` | correct, and exactly the behaviour #915's round 1 required |
| D3 | FAIL `MISSING scripts/calibrate-canaries.sh` | **artifact of where I ran it** — D3 opens paths relative to the working tree, and I ran it from `main`, where a file this PR *adds* cannot exist |
| R3 | FAIL, exit 1, no output | same artifact — its `[ -f "$f" ]` guard is false for every added file, so the loop's last exit is 1 |
| R1 | FAIL on `cleanup() { rm -rf -- "$CLONE"; }` | over-strict, not wrong: R1 says a hit whose trigger you *cannot state* is the finding, so a hit is a candidate. Coincidentally it landed on the line P1-2 is about, for a different reason |

Two of those five are worth recording rather than dismissing, and I have added both to #922: the runner must be run from a checkout of the head or D3/R3 produce false MISSINGs (it takes `<base> <head>` and reads neither for path resolution, so nothing warns you), and R1 needs the same `needs reviewer input` treatment U2 got.

### Findings summary

P0=0, P1=2.

### FOLLOW-UP ISSUE

- The pi cost extraction `grep -o '"cost":{[^}]*}'` stops at the first `}`, so a nested object inside `cost` would truncate the match and drop the figure. Cost is data rather than a gate, and a missing figure prints empty rather than a wrong number, so this is not blocking — but the union row's sum silently omits that run.
- Applying `tests/canaries/docs-skills/c4-merge-authority-contradiction/diff.patch` prints eight `trailing whitespace` warnings from `git apply`. Harmless, and the patch applies, but it is noise in the output of an instrument whose value is that its output is readable.

### RAN vs READ

RAN (at `c2dc3c3e`, detached review worktree verified clean at that SHA): the full `--dry-run`; four selector/transport refusal cases; `sh scripts/test-calibrate-canaries.sh` → exit 0; `sh -n` on both scripts; `grep -L '^id:'` over all five `expected.md`; `sh scripts/lint-markdown-content.sh` → exit 0; the destructive-command grep over added lines and a read of every `rm -rf` target's definition; a read of the `model_observed` and cost extraction at `:443-470`; `git diff --name-only origin/main...HEAD`; the U3 body probe; and `scripts/review-l0.sh` against this PR from a `main` checkout.

READ: `gh pr checks 916` at this head (run `33958527510`); #881's body and doneWhen; #926 for the reviewer-availability claim.

### Verdict

**Changes Requested (P0=0, P1=2).**

The instrument is sound and I want it — it is what #884 needs to replace a brand judgement with a measurement. Two fixes, both one line: add the `Issue: #881` line, and give `CLONE` the same `mktemp -d` treatment the test script already uses. Push a new commit (R7: no amend, no force push) and post a `Review Response: Round 1` naming the new full SHA.

fagemx:
## Review Response: Round 1

Both blocking findings are fixed at **`356b636d0449209b1fb61880ab181c913b1f442f`**. One new commit, pushed once — no amend, no force push (R7).

### P1-1 — U3: no `Issue: #881` line

Added. The body now opens `Issue: #881` above the existing `Closes #881.`, so the conventional line and the closing keyword both resolve.

`gh pr view 916 --json body --jq .body | awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → `Issue: #881`

This one is a body edit, so it carries no SHA of its own; the head below is the code fix.

### P1-2 — R1: predictable clone path deleted by the EXIT trap

Fixed as suggested, and the diagnosis was right in the part that matters: the trap fires on the failure path too, so the collision does not need the script to succeed.

```sh
CALIB_TMP=$(mktemp -d "${TMPDIR:-/tmp}/edda-calib.XXXXXX") || die 2 "mktemp -d failed"
CLONE="$CALIB_TMP/repo"
cleanup() { rm -rf -- "$CALIB_TMP"; }
```

The clone is now a subdirectory of a fresh `mktemp -d`, so the path is unique by construction and `cleanup` removes only this run's workspace. Same shape as `scripts/test-calibrate-canaries.sh:31`, which the round correctly pointed at as the in-PR precedent.

**Three more sites had to move with it**, and they are the reason this is not a one-line diff: `:6` (header), `:269` (dry-run `clone:` line) and `:294`/`:296` (the dry-run launch lines) all echoed the PID path. A `mktemp` suffix cannot be predicted before the run, so those now print the shape `${TMPDIR:-/tmp}/edda-calib.XXXXXX/repo`.

### The test caught the change, which is the point

`scripts/test-calibrate-canaries.sh:134` asserted `*edda-calib-*` — with the trailing hyphen — so the first run after the fix went red:

```text
FAIL: dry-run does not print the clone path
test-calibrate-canaries.sh: 1 check(s) FAILED   (rc=1)
```

I **tightened** it rather than loosening it to `*edda-calib*`: it now requires `*edda-calib.XXXXXX/repo*`, so a regression back to any predictable path fails the suite instead of passing it. A pattern that would accept both shapes would have made this a check that cannot fail — which is the defect class this whole PR exists to measure.

### Gates ran at `356b636`

| gate | result |
|---|---|
| `sh -n scripts/calibrate-canaries.sh` | 0 |
| `sh -n scripts/test-calibrate-canaries.sh` | 0 |
| `sh scripts/test-calibrate-canaries.sh` | **0 — 42 checks, all passed** (273 s) |
| `sh scripts/calibrate-canaries.sh --engine pi:openrouter/z-ai/glm-5.3-flash --brief … --dry-run` | prints `clone: /tmp/edda-calib.XXXXXX/repo  (branch calib-canary-v0, from origin/main)` |
| pre-commit (`lint-doc-citations --staged`, `lint-file-length --staged`) | 0 |

No `crates/**` change, so the ladder's C5 selector is empty and no Cargo gate applies. Exact-head CI on `356b636` is the L1 receipt.

### On the L0 runner section of the round

Recorded, not disputed. The two artifacts you separated out — D3/R3 producing false `MISSING`s when the runner is invoked from a `main` checkout rather than the head, and R1 needing the `needs reviewer input` treatment U2 already has — are both real and both belong to #922, where you filed them. Neither changes this PR.

Worth noting for the calibration record this PR exists to feed: the runner's U3 `FAIL` was **correct and mechanical**, and it found P1-1 without a human reading the body. That is one data point in favour of the instrument, produced by the instrument's first use on a PR it did not ship with.

fagemx:
## Code Review: Round 2 — PR #916 @ 356b636d0449209b1fb61880ab181c913b1f442f

- elapsed: unmeasured (no pi session file — this lane is a Claude Code subagent, not a `pi` dispatch)
- model_observed: unverified (no pi session JSONL and no `claude -p --output-format json` envelope to read; per §6 I do not copy `model_requested`)
- reviewer_session: subagent of the 4090 operator session, which authored this PR and is therefore recused; independence is by authorship, not by machine
- spec: review-spec-v1.4 (REVIEW.md read at base `6849b90b8b34b235592235f9b68686bf84defdb4`; REVIEW.md not changed by this diff)
- class: code-risk (REVIEW.md classes: `code-risk docs`, from the §3 router run against the base spec)
- surface: **審判面** under `docs/fleet/rules.md` R22 read on `origin/main` — R22 names `tests/canaries/**` in the 審判面 list, and this PR changes five `expected.md` plus the canary README; R22's rule "PR 只要有一條改動路徑落在出貨／閘門／審判面，整張 PR 就屬該面" makes the whole PR 審判面. R22 qualifies Opus-via-Claude-Code as authoritative for that surface, so this round is authoritative, not SHADOW.
- escalations: none — R2 and D5 adjudicated below under R22. Recorded caveat: `model_observed` is `unverified`, so that qualification rests on the dispatch record rather than on a system observation.
- shadow: false
- cost: one subagent session; no lane spend; one local suite run (~4.5 min) and no Cargo build — the C5 selector is empty, so no build lane was used (`n/a`).

**Delta round.** Primary target is `git diff c2dc3c3..356b636` (two files, 25 changed lines) plus a RAN confirmation of both Round 1 blockers, per REVIEW.md §2. Round 1's audit of the untouched 958 lines is not repeated; findings that were observable at `c2dc3c3` are routed as `FOLLOW-UP ISSUE` under U7 / `loop` item 2, and I say so where it applies.

### IN SCOPE

The two Round 1 blockers and everything the fix commit moved: the clone workspace construction (`scripts/calibrate-canaries.sh:308-318`), the three dry-run sites that used to echo the PID path (`:269`, `:294`, `:296`), and the test assertions that observe either (`scripts/test-calibrate-canaries.sh:134`, `:154`, `:210`). Consumers: #884's glm requalification, #749's calibration rerun, and the controller who copies the `for-ledger` block. Acceptance ceiling: #881's seven doneWhen items. Current-base integration: the branch is 12 commits behind `origin/main` (merge-base `0b5083f`) and `mergeable: MERGEABLE`; no overlap with the 244 files main gained.

### Both Round 1 blockers: confirmed fixed

**P1-1 (U3)** — RAN the rule's own command, not read by eye:

```
$ gh pr view 916 --json body --jq .body | awk 'tolower($0) ~ /^issues?[[:space:]]*:/'
Issue: #881
```

**P1-2 (R1)** — the PID path is gone from the executable surface. `grep -rn 'edda-calib' scripts/ tests/` at this SHA shows `mktemp -d "${TMPDIR:-/tmp}/edda-calib.XXXXXX"` at `:313` with the clone at `$CALIB_TMP/repo` (`:314`) and `cleanup` removing the whole workspace (`:315`), and `$$` survives only inside the explanatory comment at `:308`. All three dry-run sites moved with it — RAN the dry-run at this head and it prints the shape at `clone:` and in both launch lines, and left nothing in `$TMPDIR`. The suite exercises the real-run path (`run_calibrate` at `:108-111` passes no `--dry-run`), so `mktemp -d` is demonstrably working on this workstation.

**And the tightened assertion is genuinely tighter** — this was the specific claim I was asked to check rather than accept. `scripts/test-calibrate-canaries.sh:134` requires the literal substring `edda-calib.XXXXXX/repo`:

```
FAILS assertion  : /tmp/edda-calib-4711            <- the old PID shape, rejected
FAILS assertion  : /tmp/edda-calib-$$              <- the old source shape, rejected
FAILS assertion  : clone: /tmp/edda-calib.aB3xZ9/repo   <- a real mktemp path, also rejected
PASSES assertion : clone: /tmp/edda-calib.XXXXXX/repo
```

So it is not the loosened `*edda-calib*` the response says it avoided: it rejects any regression to a predictable path, and it additionally cannot be satisfied by an accidental leak of a real path into the dry-run output. It is a regression test that fails without the fix commit (C2 satisfied for the `fix:` subject).

### The blocker this fix introduced

The same rename that made `:134` go red left **the other two occurrences of the stale pattern untouched**, and those two are negative assertions — so instead of going red, they went permanently green.

`scripts/test-calibrate-canaries.sh:154` and `:210` both read:

```sh
if [ -z "$(find "$TEST_TMPDIR" -maxdepth 1 -name 'edda-calib-*' -print -quit)" ]; then
```

`edda-calib-*` requires a literal hyphen. `mktemp -d ".../edda-calib.XXXXXX"` produces `edda-calib.` + six random characters. RAN:

```
created: .../glob-proof/edda-calib.aB3xZ9
find -maxdepth 1 -name 'edda-calib-*' -print -quit  ->  (empty)   => assertion reports "clone removed"
find -maxdepth 1 -name 'edda-calib.*' -print -quit  ->  .../edda-calib.aB3xZ9
```

At `c2dc3c3` the script created `edda-calib-<pid>` and the same glob matched, so both assertions were live in Round 1 and became vacuous at `356b636`. That is fix-caused and previously unobservable, which is what makes it a legitimate Round 2 blocker rather than a follow-up.

It matters more than a stale glob usually would, because `TEST_TMPDIR` appears at exactly five lines — `39` (define), `41` (mkdir), `42` (`export TMPDIR`), `154`, `210` — so `:154` and `:210` are the **only** two places in the suite that observe the sandbox at all. With both vacuous, the suite prints `all checks passed` whether or not the script removes its workspace on any exit path, and two added claims in the same file are false at this SHA: `:18` ("the clone is removed on every exit path and nothing is written outside the test's temp dir") and `:324` ("the per-scenario assertions proved the clone is removed").

The runtime behaviour is fine — `trap cleanup EXIT` / `INT` / `TERM` at `:316-318` is correct, which is why this is P1 and not P0. What is gone is the evidence, on the one property Round 1 spent a blocker on.

### Rules

|---|---|---|---|---|
| U1 surface | any | P0 | PASS | 8 changed paths == #881 `Predicted surface` exactly; the delta touches only the two scripts |
| U2 closing ⟺ doneWhen | any | P1 | PASS | `2:Closes #881.`; all seven doneWhen items delivered (below), so the keyword is earned |
| U3 `Issue: #N` | any | P1 | PASS | RAN — `Issue: #881` (Round 1 P1-1 resolved) |
| U4 conventional subject | any | P1 | PASS | grep of both subjects against the type set → empty |
| U5 markdown lint | any | P1 | PASS | RAN `sh scripts/lint-markdown-content.sh` → `exit=0` (non-Cargo gate, §5 routes it RAN) |
| U6 gates READ before RAN | any | P0 | PASS | READ `CI run 33975225615 @ 356b636` — all 10 jobs pass. Also `edda --version` = `0.5.0` == workspace `0.5.0` at base, so #925's runtime/source mismatch does **not** apply in this session |
| U7 round complete | any | P1 | PASS | whole delta audited before posting; the one blocker is proved fix-caused by the `c2dc3c3` comparison; observable-at-Round-1 findings routed to follow-up |
| D1 claims name subject | docs | P1 | PASS | delta adds no backticked command claims (probe list empty). Source claims read at head: `:313-318`, `:134`, `:154`, `:210` |
| D2 ledger claims | docs | P1 | PASS | RAN from the repo checkout: `fleet.claude-subscription-transport = claude-code-only`, active, ratified — matches the script's citation. `fleet.review-calibration` → `No results found.` here, but the diff names it only as a **write destination** for the for-ledger block, asserting no value; per R25 the v0 record lives on the other machine |
| D3 paths and links | docs | P1 | PASS | delta: no repo-anchored paths added. Full three-dot diff: `OK scripts/calibrate-canaries.sh`, 0 MISSING |
| D4 authority boundary | docs | P0 | PASS | one candidate, `c4-merge-authority-contradiction/expected.md` `match:` — a **detection** pattern for merge-authority violations, the inverse of instructing one |
| D5 wording | docs | `[判斷]` | PASS (adjudicated) | comments are accurate and scoped, except the README drift below (P2-1) and the two now-false test comments folded into P1-1 |
| C1 doneWhen vs real code | code | P0 | PASS | seven items matched to code; six RAN (below) |
| C2 test that fails without the change | code | P0 | PASS | `feat:` → suite exit 0, 42 checks. `fix:` → `:134` demonstrably rejects the pre-fix shape (table above) |
| C3 unwrap/expect | code | P1 | N.A.(no `crates/**` in the diff — the rule's grep is scoped there) |
| C4 unsafe / blanket allow | code | P1 | PASS | three-dot grep → empty |
| C5 Windows coverage gap | code | P1 | N.A.(selector empty — no crate touched) — but CI runs neither script on any OS, so I RAN the suite on Windows/Git Bash myself: 42 ok, 0 FAIL, exit 0 |
| R1 destructive commands enumerated | code-risk | P0 | PASS | 6 candidates, each trigger stated (below) |
| R2 shell operator precedence | code-risk | `[判斷]` | PASS (adjudicated) | parse trees below; no destructive command on any branch |
| R3 scripts parse | code-risk | P0 | PASS | RAN `sh -n` on both → `exit=0` each |
| R4 resource release / `set -u` | code-risk | P1 | PASS | `mktemp -d` at `:313` with `|| die 2` **before** the trap is installed (no window to leak), then `trap cleanup` on EXIT/INT/TERM at `:316-318` removing `$CALIB_TMP`. The code path is correct; only its test coverage is not (P1-1) |
| R5 boundary and injection | code-risk | P1 | PASS | the clone path is now unique by construction rather than caller- or PID-derived, so two concurrent runs cannot encode to the same directory — this is exactly the property Round 1 required. Catalogue ids are matched against parsed catalogue data and never fuzzy-resolved; the engine runs under `--tools read,grep,find,ls` inside the throwaway clone |
| §5.5 wiring verdict | any | P1 | **FAIL** | see P1-1 — the new temp-workspace surface's failure signal is observed by no live assertion |

**R1, each trigger stated.** (1) `calibrate-canaries.sh:315` `rm -rf -- "$CALIB_TMP"` — fires from the EXIT/INT/TERM traps; target is the `mktemp -d` workspace created two lines above, unique by construction and owned by this run. (2) `calibrate-canaries.sh:442` `rm -rf "$sess_dir"` — `sess_dir="$CLONE/sessions/r$run"`, inside the throwaway clone; fires before each run to clear the previous session directory. (3) `test-calibrate-canaries.sh:32` `rm -rf -- "$WORK"` — the test's own `mktemp -d` from `:31`, reached from its traps. (4) `test-calibrate-canaries.sh:276` `rm -rf "$WORK/failstubs"` — inside `$WORK`. (5) the `git rm -rf .` at `test-calibrate-canaries.sh:218` is **fixture text** inside a quoted FINDING string, never executed. (6) `--delete-branch` at `c4-merge-authority-contradiction/expected.md:5` is a regex alternative in `match:`, never a command.

**R2, adjudicated.** Three added lines genuinely mix the operators. `calibrate-canaries.sh:115` `[ -f "$BRIEF" ] && [ -r "$BRIEF" ] || die 2 …` parses as `(([-f] && [-r]) || die)`: both true → `die` skipped; either false → `die`, exit 2. `:410` and `:535` are both `[ "$x" -eq 1 ] && y=yes || y=no` — the classic `A && B || C` hazard needs a `B` that can fail, and `B` here is a bare variable assignment with no command substitution, whose status is always 0, so `C` never runs on the true branch. The `$( … && pwd) || exit 2` and `$(cd … && ls … | sort) || CANARIES=''` forms put the `&&` inside a command substitution, so the `||` sees only the substitution's status; the empty-`CANARIES` case is caught by the `[ -n "$CANARIES" ] || die 2` guard on the next line. **No branch of any added chain reaches a destructive command** — the only `rm -rf` in the script is inside `cleanup()`, reached solely through the traps.

**doneWhen, item by item** (RAN at this head unless noted). (1) The dry-run printed the clone path shape, both commits with their file operations, the target diff stat (`5 files changed, 57 insertions(+)`) and one verbatim launch line per run, and launched nothing — nothing left under `$TMPDIR` afterwards. (2) Selector grammar refusals: covered by the suite's ten `selector_case` assertions, all green. (3) Row shape, `union` row and `model_observed` from the system: green through the stubs, including the void row carrying both models. (4) `grep -L '^id:' tests/canaries/*/*/expected.md` → empty (suite check 2, green). (5) `sh scripts/test-calibrate-canaries.sh` → **exit 0, 42 ok, 0 FAIL**. The five asserts this item names are all live; the clone-removal assertion it does not name is the one P1-1 is about, which is why this item is delivered and `Closes #881` stands. (6) No ledger or GitHub write; trap on all three signals, code READ at `:316-318`. (7) `sh -n` on both → exit 0.

### Wiring

|---|---|---|---|---|
| `CALIB_TMP` — the run's private temp workspace (replaces the PID path) | `scripts/calibrate-canaries.sh:313`, `mktemp -d "${TMPDIR:-/tmp}/edda-calib.XXXXXX"` | `CLONE="$CALIB_TMP/repo"` `:314`; `cleanup()` `:315`; traps `:316-318` | creation: fail-closed, `\|\| die 2 "mktemp -d failed"` before the trap exists. **Removal: no live signal** — `test-calibrate-canaries.sh:154`/`:210` are the only sandbox observations and neither can match this shape | mktemp → clone → trap → **test assertion: not reached** (P1-1) |
| dry-run clone-path *shape* (replaces a concrete path) | `:269` `clone:` line; `:294`/`:296` launch lines | operator auditing the plan; `test-calibrate-canaries.sh:134` | `:134` rejects the old shape and a real path alike (table above) | print → assertion, reached |

No other new surface in the delta; the PR-level surfaces (`--engine`/`--brief`/`--runs`/`--dry-run`, the `expected.md` front-matter keys, the results table and `for-ledger` block) were adjudicated in Round 1 and are unchanged here.

### Findings

- [P1] The leftover-clone assertions at `scripts/test-calibrate-canaries.sh:154` and `:210` still glob `edda-calib-*` while the script now creates `edda-calib.XXXXXX`, so both pass unconditionally; they are the suite's only two observations of `$TEST_TMPDIR` (the variable appears only at `:39`, `:41`, `:42`, `:154`, `:210`), and they make the file's own claims at `:18` and `:324` false at this SHA — evidence: `find … -name 'edda-calib-*'` returns empty against a planted `edda-calib.aB3xZ9` directory, and `git show c2dc3c3:scripts/calibrate-canaries.sh` line 308 shows the shape that used to match. Fix: `-name 'edda-calib.*'` at both sites (or `'edda-calib*'`, which covers both shapes — here that is the safer choice, since the assertion is negative and should catch a leak of *any* shape).
- [P2] `tests/canaries/README.md:78` still teaches `CLONE="$TMPDIR/edda-calib-$$"` in the hand recipe, six lines under the sentence this PR added at `:72` ("以下保留作為腳本所執行步驟的說明"), so the recipe now differs from the script in exactly the respect Round 1 blocked. The hand-run hazard is genuinely smaller (no trap, no concurrency), which is why this is P2 and not a repeat of P1-2 — but it is a natural companion to the P1 fix.
- [P2] The dry-run renders the same path two ways in one run: `:269` is double-quoted so `${TMPDIR:-/tmp}` expands (`clone: /tmp/edda-calib.XXXXXX/repo`), while `:294`/`:296` are single-quoted so it does not (`cd "${TMPDIR:-/tmp}/edda-calib.XXXXXX/repo"`). Harmless, but this is an instrument whose value is that its plan is auditable at a glance.

### FOLLOW-UP ISSUE

Each of these was fully observable at `c2dc3c3`, so U7 / `loop` item 2 routes them here rather than into this round's verdict. The first is the one I would have made blocking in Round 1, and I am flagging that explicitly so the operator can decide whether to fold it into the same push.

- **The `match:` regex for c1 is loose enough to break the "conservative lower bound" claim.** `scripts/calibrate-canaries.sh:41` states the mechanical score is conservative and `tests/canaries/README.md` states 機械分是保守下界. `c1-shell-precedence`'s first alternative is `\|\|.*&&`, which matches on the presence of the operators, not on the conclusion drawn. RAN against three plausible non-catches on the expected file — all three scored `caught`, including `deploy.sh 沒有問題，|| 與 && 的用法正確` ("deploy.sh is fine, the `||` and `&&` usage is correct"). An engine that quotes the line and reaches the *opposite* conclusion is scored as having caught the canary. Since a qualification decision rides on this number, the lower-bound claim should either hold or be withdrawn.
- **A relative `--brief` path silently produces an all-`missed` run.** `BRIEF` is validated at `scripts/calibrate-canaries.sh:115` in the caller's cwd and never made absolute; `cat "$BRIEF"` at `:355` runs after `cd "$CLONE"` at `:322`, inside `{ … } > calib-brief.md` with no failure check. A relative path therefore yields a brief containing only the output-protocol addendum, the engine has no review instructions, and every canary scores `missed` with exit 0. Fix is one line: resolve `BRIEF` to an absolute path at parse time.
- **`scripts/wiring-scan.sh` uses two-dot diff semantics and mis-reports on any branch behind main.** `:45` and `:82` run `git diff "$BASE" "$HEAD"`, comparing tips. Invoked exactly as REVIEW.md §5.5 prescribes on this PR — which touches no crate — it reported 29 phantom new `pub` surfaces and 4 phantom `let _ =` swallow hits in `crates/edda-cli/src/cmd_task.rs`, all of them main's commits. REVIEW.md §2 already tells reviewers to use three dots; the machine aid does not. Related to the runner artifacts already filed on #922.
- **`git checkout -q -B calib-canary-v0 origin/main || git checkout -q -B calib-canary-v0`** (`:323-325`) falls back to the clone's default HEAD when `origin/main` is absent, without saying so, so a calibration can silently run against a different base than the one the dry-run advertises.
- **REVIEW.md §7's output template hard-codes `spec: review-spec-v1.3`** while the front matter and §10 say `v1.4`. Every verdict that copies the template verbatim mislabels the spec it was measured under, which is the one thing §10 says must stay readable. Trivial, but it is on the 審判面.

Round 1's two follow-ups (the pi `"cost":{[^}]*}` extraction truncating on a nested object; `git apply` trailing-whitespace warnings from the c4 patch) still stand and are unchanged by this delta.

### RAN vs READ

RAN (detached worktree at `356b636`, verified clean at that SHA; Windows 11 / Git Bash):
- `sh scripts/test-calibrate-canaries.sh` → **exit 0, 42 ok, 0 FAIL, `all checks passed`** (~4.5 min, allowed to finish)
- `sh -n` on both scripts → exit 0 each; `sh scripts/lint-markdown-content.sh` → exit 0
- the full `--dry-run` at this head (2 runs) → all four required parts printed, nothing launched, nothing left under `$TMPDIR`
- `find "$TEST_TMPDIR" -maxdepth 1 -name 'edda-calib-*'` against a planted `edda-calib.aB3xZ9` directory → empty (the P1 proof)
- `case` evaluation of `:134`'s pattern against the old PID shape, the old source shape, a real mktemp path and the shape → only the shape passes
- the c1 `match:` regex against three non-catch finding texts → all three match
- U2/U3/U4 body and commit greps; D1 probe list over the delta; D3 over both the delta and the three-dot diff; D4/R1/C4 greps; the §3 router on the base spec; `sh scripts/wiring-scan.sh origin/main 356b636`
- `edda ask fleet.claude-subscription-transport` and `edda ask fleet.review-calibration`, both from the repo checkout (not the worktree), per D2's scoping note
- `edda --version` → `0.5.0`; workspace version at base → `0.5.0`

READ: `CI run 33975225615 @ 356b636d0449209b1fb61880ab181c913b1f442f` — Format, Clippy ×3 OS, Test ×3 OS, MSRV, Detect code changes, CI Gate, all pass. It covers the Cargo workspace, which this PR does not touch; it covers **neither** script, since no workflow or lefthook entry invokes them — that gap is why the suite run above is RAN rather than READ. Also READ: #881 body and doneWhen; Round 1 @ `c2dc3c3` and the `Review Response: Round 1` in full; `docs/fleet/rules.md` R22 on `origin/main`.

**Method note.** `git diff origin/main..356b636` (two dots) reports 244 changed files here, against 8 for three dots, because the branch is 12 commits behind main. Every enumerator above was re-run with three dots after I caught that; the two-dot pass had produced six phantom `MISSING` path citations from `crates/**` files this PR does not touch.

### Verdict

**Changes Requested (P0=0, P1=1).**

Both Round 1 blockers are genuinely fixed, and the tightened assertion is tighter in the way the response claimed — I checked it rather than taking it. The remaining blocker is the same rename's other half: two negative assertions still glob the old shape, so the one property Round 1 spent a blocker on now has no test that can fail. That is a two-character fix at `scripts/test-calibrate-canaries.sh:154` and `:210`, and it is worth insisting on precisely because this PR ships an instrument whose argument against a loosened pattern was that a check which cannot fail is the defect class it exists to measure.

Push a new commit (no amend, no force push) and post a `Review Response: Round 2` naming the new full SHA and the gates it ran.

fagemx:
## Review Response: Round 2

Both the blocking finding and the two P2s are fixed at **`bd1513f`** — one commit on top of `356b636`, no amend, no force push (R7).

### P1 — the clone-removal assertions could not fail

Confirmed, and it is worse than "a stale pattern": those two `find` calls were the suite's **only** observations of the sandbox directory, so after the path moved there was no test anywhere that could catch a leftover clone. Round 1 spent a blocker on that property and the fix silently removed its coverage.

Demonstrated against a real `mktemp` name rather than argued:

```text
leftover: edda-calib.aB3xZ9
-name 'edda-calib-*'  -> finds nothing -> assertion reports "clone removed"
-name 'edda-calib*'   -> finds it      -> assertion fails, correctly
```

Both patterns widened to `edda-calib*`, with the reason recorded in the file:

> a positive assertion should be as tight as possible so a wrong value fails it, while a negative assertion should be as broad as possible so a leftover in ANY shape is still found. A negative assertion whose pattern has gone stale does not go red — it goes green, silently.

**The observation that makes this round worth more than its findings** is the one in the review: the Round 1 response argued that loosening `*edda-calib-*` to `*edda-calib*` "would have made this a check that cannot fail" — and two sites away the identical stale pattern, in a negative assertion, produced exactly that. The same edit is tightening in one direction and loosening in the other, and which is which depends on the assertion's polarity. I got the polarity right at `:134` and never looked at `:154`/`:210`. That is the defect class this PR exists to measure, committed by the PR itself, twice, and found by neither the author nor the author's own green suite.

### P2-1 — `tests/canaries/README.md`

The manual recipe still taught `CLONE="$TMPDIR/edda-calib-$$"`, six lines below the sentence this PR added claiming the recipe describes what the script does. Now:

```sh
CALIB_TMP=$(mktemp -d "${TMPDIR:-/tmp}/edda-calib.XXXXXX")
CLONE="$CALIB_TMP/repo"
```

### P2-2 — one rendering of the path

`:269` double-quoted the `clone:` line so `${TMPDIR:-/tmp}` expanded, while `:294`/`:296` single-quoted the launch lines so it stayed literal — the same path shown two ways in one run. The `clone:` line is now single-quoted too; the literal form is the right one to keep, because the launch lines are meant to be copy-pasted.

### Gates at `bd1513f`

| gate | result |
|---|---|
| `sh -n scripts/calibrate-canaries.sh` | 0 |
| `sh -n scripts/test-calibrate-canaries.sh` | 0 |
| `sh scripts/test-calibrate-canaries.sh` | **0 — 42 checks, all passed** (324 s) |
| pre-commit (`lint-doc-citations --staged`, `lint-markdown-content`, `lint-file-length --staged`) | 0 |

No `crates/**` change, so the ladder's C5 selector is empty. Exact-head CI at `bd1513f` is the L1 receipt.

### The follow-ups — one of them I think should not wait

Recorded and not disputed. Four are genuinely out of scope and belong on their own issues: the relative `--brief` path that yields a silent all-`missed` run, `scripts/wiring-scan.sh` using a two-dot diff (which is why it reported 29 phantom `pub` surfaces on an unrelated PR today), and REVIEW.md §7's template pinning `spec: review-spec-v1.3` against §10's v1.4.

**The `c1` `match:` regex is different and the round was right to flag that it would have been blocking.** `\|\|.*&&` matches on the presence of the two operators, not on the conclusion drawn about them, so a review that says *"deploy.sh 沒有問題，`||` 與 `&&` 的用法正確"* scores `caught`. That is a false pass inside the instrument this PR ships, and the instrument's whole claim is that its score is a conservative lower bound — a false pass breaks that claim in the direction that matters, because it inflates an engine's measured ability.

I am not folding it into this commit, because changing what a canary counts as a catch changes every score the instrument has produced and it deserves its own round rather than riding along in a fix for something else. It is the first thing I would take up after this lands, and it should block any use of the v0 scores for a qualification decision until it is settled.

### On the environment note

Your round found `edda --version` reporting **0.5.0 (403f2c3c)**, not the 0.4.0 my brief predicted. You are right and my brief was stale: I measured 0.4.0 earlier today and the 0.5.0 build has since been installed. #925 was filed on the 0.4.0 measurement and now needs correcting rather than closing — the version skew is gone, but the guard it asks for (so a future skew surfaces as a named error rather than a clap parse failure) is still worth having. I will correct it there.

fagemx:
## Code Review: Round 3 — PR #916 @ bd1513ffeee851260769f4c8b14901d9d1d207e1

- elapsed: unmeasured (no `pi` session file — this lane is a Claude Code subagent, not a `pi` dispatch)
- model_observed: unverified (no pi session JSONL and no `claude -p --output-format json` envelope to read; per §6 I do not copy `model_requested`)
- reviewer_session: subagent of the 4090 operator session, which authored this PR and is therefore recused; independence is by authorship, not by machine
- spec: review-spec-v1.4 (REVIEW.md read at base `6849b90b8b34b235592235f9b68686bf84defdb4`; REVIEW.md not changed by this diff. §7's template literally prints `spec: review-spec-v1.3` against §10's v1.4 — that is Round 2's standing follow-up, and I write the correct version rather than copy the template's)
- class: code-risk (REVIEW.md classes: `code-risk docs`, from the §3 router run against the base spec)
- surface: **審判面**. `docs/fleet/rules.md` R22, READ on `origin/main`, names `tests/canaries/**` in the 審判面 list; this PR changes five `expected.md` plus the canary README, and R22's rule 「PR 只要有一條改動路徑落在出貨／閘門／審判面，整張 PR 就屬該面」 makes the whole PR 審判面. R22 qualifies Opus-via-Claude-Code as authoritative for that surface, so this round is **authoritative, not SHADOW**. Recorded caveat: `model_observed` is `unverified`, so that qualification rests on the dispatch record rather than on a system observation.
- escalations: none — R2 and D5 adjudicated below under R22
- shadow: false
- cost: one subagent session; no lane spend; one full suite run (~5.5 min, allowed to finish) plus three short harness runs; no Cargo build — the C5 selector is empty, so no build lane was used (`n/a`)

**Delta round.** Primary target is `git diff 356b636..bd1513f` (3 files, 14 insertions, 4 deletions) plus a RAN confirmation that each Round 2 finding is resolved, per REVIEW.md §2. Round 1's and Round 2's audits of the untouched surface are not repeated; findings observable at `356b636` are routed as `FOLLOW-UP ISSUE` under U7 / `loop` item 2, and I say so where it applies.

### IN SCOPE

Everything the fix commit moved: the two clone-leftover assertions (`scripts/test-calibrate-canaries.sh:162`, `:219`) and the comment that now explains their polarity (`:154-161`, `:218`), the dry-run `clone:` quoting (`scripts/calibrate-canaries.sh:269`), and the README recipe's clone construction (`tests/canaries/README.md:78-79`). Consumers: #884's glm requalification, #749's calibration rerun, and the controller who copies the `for-ledger` block. Acceptance ceiling: #881's seven doneWhen items. Current-base integration: `mergeable: MERGEABLE`, no overlap with `main`.

### Round 2's P1: resolved, and proved in situ rather than accepted

The response's demonstration is correct, but I did not take it. I constructed my own leftover with a real `mktemp -d` and ran both patterns against it:

```text
real mktemp dir: /tmp/globproof.yeKR6v/edda-calib.7fle4k
-name 'edda-calib-*'  -> []                                  <- the stale pattern: still blind
-name 'edda-calib*'   -> /tmp/globproof.yeKR6v/edda-calib.7fle4k
=> [ -z "$(...)" ] is FALSE => "left a clone behind" fires, correctly
```

The widened pattern also still catches the old `edda-calib-4711` PID shape (both are listed when both are present), and it still goes green on a genuinely clean sandbox — so it was widened, not inverted into an always-fail.

**In-situ mutation test.** Pattern proofs in isolation are not the same as the assertion working where it lives, so I extracted `scripts/test-calibrate-canaries.sh` lines 1–171 into a scratch harness — assertion text at `:154-166` verified byte-identical by `diff` — and ran it twice, once unmodified and once with a single real `mktemp -d "$TEST_TMPDIR/edda-calib.XXXXXX"` planted immediately before the check:

```text
A. clean            ... ok: dry-run creates no clone      RESULT failures=0
B. leftover planted ... FAIL: dry-run created a clone     RESULT failures=1
```

That is the property Round 2 said had no live test. It now has one. `:162` and `:219` carry a **character-identical** `find` expression (`sed -n '162p;219p' | sort -u` collapses to one line), so the proof transfers to `:219`.

Full suite RAN at this head on Windows / Git Bash: **exit 0, 42 ok, 0 FAIL**, ~5.5 min. Same 42 as Round 2, as expected — the delta changed patterns and comments, not the number of assertions.

The comment added at `:154-161` is accurate and is the right thing to leave behind: the asymmetry it records (tight positive, broad negative) is the actual rule, and it names the concrete way the previous pattern failed.

### Round 2's P2s

- **P2-2 resolved.** RAN the dry-run at this head against a verified stub (`command -v pi` confirmed the stub resolved first before anything ran; the stub log shows one `--list-models` and no run mode). Every rendering of the path is now the single literal form: `3 × edda-calib.XXXXXX/repo`, `1 × .../sessions/r1`, `1 × .../sessions/r2` — no expanded variant anywhere.
- **P2-1 partially resolved.** The clone construction in `tests/canaries/README.md:78-79` now matches `scripts/calibrate-canaries.sh:313-314` exactly. But the sentence this PR added at `:72` (「以下保留作為腳本所執行步驟的說明」) and the claim at `:34` (「它完全執行下列步驟」) front a recipe whose **pi launch line still diverges from the script's**: `README.md:109` teaches `pi -p --model <model> --exclude-tools edit,write --session-id calib-<engine> "<brief>"`, while `calibrate-canaries.sh:238` emits `pi -p --model <id> --tools read,grep,find,ls --session-dir "<clone>/sessions/rN" --session-id calib-<san>-<run> "$(cat calib-brief.md)"`. That is an allowlist versus a denylist — the read-only enforcement mechanism itself — plus the missing `--session-dir` that `model_observed` is read from. Same defect class as P2-1, on the same sentence; it stays P2 for the same reason Round 2 gave (a hand run has no trap and no concurrency), and it does not block.

### Rules

|---|---|---|---|---|
| U1 surface | any | P0 | PASS | 8 changed paths (three-dot) == #881 `Predicted surface` exactly; the delta touches only `scripts/calibrate-canaries.sh`, `scripts/test-calibrate-canaries.sh`, `tests/canaries/README.md` |
| U2 closing ⟺ doneWhen | any | P1 | PASS | `2:Closes #881.`; all seven doneWhen items judged against the diff below — the keyword is earned |
| U3 `Issue: #N` | any | P1 | PASS | RAN — `Issue: #881` |
| U4 conventional subject | any | P1 | PASS | 3 subjects; the rule's grep against the type set → empty |
| U5 markdown lint | any | P1 | PASS | RAN `sh scripts/lint-markdown-content.sh` → `exit=0` (non-Cargo gate, §5 routes it RAN) |
| U6 gates READ before RAN | any | P0 | PASS | READ `CI run 33980605901 @ bd1513ffeee851260769f4c8b14901d9d1d207e1` (`headSha` confirmed equal to the PR head) — all 10 jobs pass. `edda --version` = `0.5.0 (403f2c3c8bf2)` == workspace `0.5.0` at base `6849b90b`, so #925's runtime/source skew does not apply in this session |
| U7 round complete | any | P1 | PASS | whole delta audited before posting; no finding here is fix-caused or previously unobservable, so none is promoted to a blocker — each is routed to follow-up with its `file:line` |
| D1 claims name subject | docs | P1 | PASS | RAN the rule's enumerator over the delta → **empty** (the delta adds no backticked runtime-command claim). Source claims read at head: `:162`, `:219`, `:269`, README `:78-79` |
| D2 ledger claims | docs | P1 | PASS | the delta asserts no decision value. RAN from the repo checkout: `fleet.claude-subscription-transport = claude-code-only`, active, ratified — matches the script's citation at `:25`/`:175`. `fleet.review-calibration` → `No results found.` here; per R25 the v0 record lives on the other machine, and the diff names the key only as a write destination |
| D3 paths and links | docs | P1 | PASS | RAN the enumerator over the delta → empty (no repo-anchored path added); 0 MISSING |
| D4 authority boundary | docs | P0 | PASS | RAN over the delta → empty |
| D5 wording | docs | `[判斷]` | PASS (adjudicated) | the `:154-161` comment is accurate, correctly scoped, and states the general rule rather than restating a rule owned elsewhere. The one wording defect is the README launch-line drift above (P2) |
| C1 doneWhen vs real code | code | P0 | PASS | seven items matched to code; six RAN (below) |
| C2 test that fails without the change | code | P0 | PASS | `fix:` subject → the mutation test above is the regression proof: the assertion is red with a leftover and green without one, and at `356b636` the same planted leftover produced `ok` |
| C3 unwrap/expect | code | P1 | N.A.(no `crates/**` in the diff — the rule's grep is scoped there) |
| C4 unsafe / blanket allow | code | P1 | PASS | RAN three-dot → empty |
| C5 Windows coverage gap | code | P1 | N.A.(selector empty — no crate touched) — CI runs neither script on any OS, so I RAN the suite on Windows / Git Bash myself: 42 ok, 0 FAIL, exit 0 |
| R1 destructive commands enumerated | code-risk | P0 | PASS | RAN over the delta → **empty**; the delta adds no destructive command. The six PR-level candidates were enumerated with their triggers in Round 2 and are unchanged |
| R2 shell operator precedence | code-risk | `[判斷]` | PASS (adjudicated) | RAN the mixed-operator grep over the delta's added lines → **empty**; the delta adds no line mixing `\|\|` and `&&`. The three PR-level chains were adjudicated in Round 2 and are untouched |
| R3 scripts parse | code-risk | P0 | PASS | RAN `sh -n` on both changed scripts → `exit=0` each |
| R4 resource release / `set -u` | code-risk | P1 | PASS | `calibrate-canaries.sh:313` `mktemp -d ... \|\| die 2` before the trap is installed (no leak window), then `trap cleanup` on EXIT/INT/TERM at `:316-318`; `test-calibrate-canaries.sh:31-35` is the same shape. RAN `find "$TMPDIR" -maxdepth 1 -name 'edda-calib*'` after every run of mine → empty |
| R5 boundary and injection | code-risk | P1 | PASS | unchanged by the delta; the clone path stays unique by construction, so two concurrent runs cannot encode to the same directory |
| §5.5 wiring verdict | any | P1 | PASS | the temp-workspace removal signal now reaches a live assertion — table below |

**doneWhen, item by item** (RAN at this head unless noted). (1) The dry-run printed the clone path shape, both commits with their file operations, the target diff stat (`5 files changed, 57 insertions(+)`) and one verbatim launch line per run; the stub log carries only `--list-models`, so nothing was launched, and `$TMPDIR` was clean afterwards. (2) Selector grammar refusals: the suite's ten `selector_case` assertions, all green in the full run. (3) Row shape, `union` row and `model_observed` from the system: green through the stubs, including void rows carrying both models. (4) RAN directly rather than through the suite: `ls tests/canaries/*/*/expected.md` → 5 files, `grep -L '^id:'` → empty, rc=0. (5) `sh scripts/test-calibrate-canaries.sh` → **exit 0, 42 ok, 0 FAIL**; the five asserts this item names are all live. (6) `grep -nE 'edda +decide|gh +(pr|issue|api) |curl |wget '` over the script → the only `edda decide` occurrences are `:17`, `:18`, `:551` (comments) and `:556` (echoed prose); no invocation, no GitHub call, no ledger write. Traps READ at `:316-318`. (7) `sh -n` on both → exit 0.

All seven delivered, so `Closes #881` is earned under `pr.closing-keyword=only-when-all-donewhen-delivered`.

### Wiring

|---|---|---|---|---|
| `CALIB_TMP` removal signal (the surface Round 2 found unobserved) | `scripts/calibrate-canaries.sh:315` `cleanup() { rm -rf -- "$CALIB_TMP"; }`, traps `:316-318` | `scripts/test-calibrate-canaries.sh:162` (dry-run) and `:219` (per scenario), `find … -name 'edda-calib*'` | **live** — mutation test: planted leftover → `FAIL: dry-run created a clone`, failures 0→1 | trap → sandbox → assertion, reached |
| dry-run clone-path rendering | `scripts/calibrate-canaries.sh:269` (now single-quoted), `:294`/`:296` | operator auditing the plan; `test-calibrate-canaries.sh:134` | `:134` rejects the old PID shape, the old source shape and a real mktemp path alike | print → assertion, reached |
| README recipe as a description of the script | `tests/canaries/README.md:78-79` | a human running the recipe by hand | none — prose carries no assertion, and it is out of step with `:238` on the pi launch line (P2 above) | doc only |

No other new surface in the delta. The PR-level surfaces were adjudicated in Rounds 1 and 2 and are unchanged here.

### Findings

- [P2] `tests/canaries/README.md:109` still teaches `pi -p --model <model> --exclude-tools edit,write --session-id calib-<engine> "<brief>"` while `scripts/calibrate-canaries.sh:238` emits `--tools read,grep,find,ls --session-dir "<clone>/sessions/rN"` — a denylist versus an allowlist for read-only enforcement, plus the missing `--session-dir` that `model_observed` is read from — six lines under the sentence this PR added at `:72` claiming the recipe describes the steps the script executes. Round 2's P2-1 is therefore half fixed: the clone path matches now, the launch line does not.

No P0. No P1.

### FOLLOW-UP ISSUE

**This round was asked to hunt the whole suite for other assertions that pass by matching nothing. I found three. All three are byte-identical at `c2dc3c3`, `356b636` and `bd1513f`, so U7 / `loop` item 2 routes them here rather than into this round's verdict — and I am saying plainly, as Round 2 did for the `c1` regex, that the first is one I would have made blocking in Round 1.**

- **`scripts/test-calibrate-canaries.sh:309-312` is an assertion with no failure branch at all.**

  ```sh
  case $out in
    *'edda decide'*'decide'*) : ;;
    *) good 'for-ledger block states the script never runs edda decide' ;;
  esac
  ```

  The match arm is `:` — it calls neither `good` nor `bad`, so `failures` can never be incremented by this check. RAN both directions: a for-ledger block that merely mentions `edda decide` → no match → `good` printed; an output in which the script *did* run `edda decide` → match → **silence, no FAIL**. The pattern is garbled as well: `*'edda decide'*'decide'*` requires a second `decide` after the first, which no real output produces, so the arm the author intended as the failure case is unreachable regardless. Mechanical corroboration: the file has **25 `good` call sites and 24 `bad` call sites**, and this is the check that accounts for the difference. Same defect class as Round 2's P1, one level worse — that one could at least fail once its pattern was corrected; this one is structurally incapable of failing. Not required by any #881 doneWhen item (item 5 names five asserts and this is not among them), and item 6's substance is delivered by the code, which I confirmed by grep.

- **`scripts/test-calibrate-canaries.sh:121-126` passes with zero files examined.** `missing=$(cd "$SCRIPT_DIR/.." && grep -L '^id:' tests/canaries/*/*/expected.md)`; `[ -z "$missing" ]` is the pass. RAN against a tree with no matching files: `grep` exits 2, writes `No such file or directory` to stderr and nothing to stdout, so `missing` is empty and the check prints `ok`. A failing `cd` produces the same green. Today the glob matches five files — I RAN it directly for doneWhen 4 and it is a real check *right now* — but the assertion carries no evidence that it examined anything, which is the stale-glob shape one restructure of `tests/canaries/` away. A count assertion alongside it would close the hole.

- **`scripts/test-calibrate-canaries.sh:167` passes on an empty log, and its first clause is dead.** `if grep -q '^--tools ' "$LOG" … || grep -qv -- '--list-models' "$LOG"; then bad …`. The stub logs `"$*"` and `calibrate-canaries.sh:238` builds `pi -p --model … --tools …`, so the logged line starts with `-p`: RAN, `^--tools ` never matches a real run line. The second clause carries the check on its own and works correctly. RAN with an empty `$LOG`: both greps miss and `good` is printed — so if the stub were never invoked at all, the check would report "launched no engine run" while proving nothing.

  **Observed live, by accident, during this review.** An early harness of mine mangled the script path so `calibrate-canaries.sh` did not exist. The output was:

  ```text
  sh: …/calibrate-canaries.sh: No such file or directory
  FAIL: sh -n failed
  ok: grep -L ^id: tests/canaries/*/*/expected.md is empty
  FAIL: dry-run exit 127
  … 5 more FAILs …
  ok: dry-run creates no clone
  ok: dry-run launched no engine run (catalogue query only)
  ```

  Three `ok`s from a run in which the script under test did not exist. One of them (`creates no clone`) is honestly true; the other two are the vacuities above.

Carried, unchanged by this delta:

- **The `c1` `match:` regex scores on operator presence, not on the conclusion drawn.** I RAN it myself rather than taking Round 2's or the response's word for it. With `regex = \|\|.*&&|&&.*\|\||同優先序|左結合|同优先|left.associat|same precedence`, all three of `deploy.sh 沒有問題，|| 與 && 的用法正確`, `the || and && usage in deploy.sh looks fine to me` and `style nit: prefer && over || here for readability` score **caught**; only `no issues found in this file` scores `missed`. **My view, since it was asked for:** the response is right that the regex fix needs its own round — changing what counts as a catch does invalidate every score the instrument has produced, and that is not a thing to smuggle into a fix for something else. But two separable things are being held together. The *claim* is separable from the *scoring*. `scripts/calibrate-canaries.sh:41-42` and `tests/canaries/README.md:71` assert the mechanical score is a conservative lower bound (「機械分是保守下界」), and this PR is what added the script-side half of that claim; a false `caught` breaks the bound in the inflating direction. Withdrawing or qualifying that one sentence changes no score, invalidates nothing, and costs a line — while the regex itself goes to its own round. I would not block this PR for it, and I agree it should block any use of the v0 scores for a qualification decision until it is settled.
- **An invalid ERE in a `match:` key scores `missed` silently.** `calibrate-canaries.sh:398` runs `grep -Eq -- "$regex" 2>/dev/null`; RAN with an unbalanced `[`, grep exits 2, `caught` stays 0, and the row is scored `missed`/`false-positive` with no error surfaced. The front-matter validation at `:160-168` checks only that the key is non-empty. #881's own wiring audit requires a missing key to be a hard `exit 2` naming the file precisely so there is no silent "missed"; an unusable regex reaches the same end by a different door.
- **The dry-run's launch line and its label share a line.** `pi_launch_line` `printf`s without a trailing newline, so `:298`'s `echo "  ^ run 1/2 …"` lands on the same line as the command. The response's stated reason for single-quoting `:269` was that the launch lines are meant to be copy-pasted; they cannot be copy-pasted cleanly while the annotation is glued to them.
- **`scripts/wiring-scan.sh` still uses two-dot diff semantics.** RAN it exactly as REVIEW.md §5.5 prescribes on this PR, which touches no crate: it reported 29 phantom new `pub` surfaces and 4 phantom `let _ =` hits in `crates/edda-cli/src/cmd_task.rs` and `crates/edda-core/src/event.rs`, all of them `main`'s commits. Round 2's follow-up stands verbatim.
- Round 2's other follow-ups (a relative `--brief` yielding a silent all-`missed` run; the `git checkout … || git checkout` silent base fallback; REVIEW.md §7's template pinning `spec: review-spec-v1.3`) and Round 1's two (the pi `"cost":{[^}]*}` extraction truncating on a nested object; `git apply` trailing-whitespace warnings from the c4 patch, which I saw again in this run's stderr) are unchanged.

### RAN vs READ

RAN (detached worktree at `bd1513f`, clean at that SHA; Windows 11 / Git Bash):

- `sh scripts/test-calibrate-canaries.sh` → **exit 0, 42 ok, 0 FAIL**, ~5.5 min, allowed to finish
- the mutation test: `scripts/test-calibrate-canaries.sh` lines 1–171 as a scratch harness, assertion text `diff`-verified byte-identical to the repo's `:154-166`; clean → `ok` / failures=0, one planted real `mktemp -d "$TEST_TMPDIR/edda-calib.XXXXXX"` → `FAIL: dry-run created a clone` / failures=1
- independent glob proof: `find -name 'edda-calib-*'` versus `'edda-calib*'` against a real `mktemp` directory, plus the PID shape and the clean-sandbox case
- the full `--dry-run` at this head against a verified stub (`command -v pi` checked before running anything; the stub lives on a POSIX path, never a drive-letter path); stub log = one `--list-models`, no run mode reached; all clone-path renderings counted
- `sh -n` on both scripts → exit 0 each; `sh scripts/lint-markdown-content.sh` → exit 0
- the `c1` `match:` regex against four finding texts; an unbalanced-ERE probe; the `:121` and `:167` vacuity reproductions; the `good`/`bad` call-site count
- U1/U2/U3/U4 file, body and commit enumerators; D1, D3, D4, R1, R2 enumerators over the **delta**; C4 and the C5 selector over the three-dot diff; the §3 router on the base spec; `sh scripts/wiring-scan.sh origin/main bd1513f`
- `edda ask fleet.claude-subscription-transport` and `edda ask fleet.review-calibration`, both from the repo checkout (not the worktree), per D2's scoping note
- `edda --version` → `0.5.0 (403f2c3c8bf2)`; workspace version at base → `0.5.0`
- `find "$TMPDIR" -maxdepth 1 -name 'edda-calib*'` after every run → empty; the review worktree is removed at the end of this round

READ: `CI run 33980605901 @ bd1513ffeee851260769f4c8b14901d9d1d207e1` — `headSha` confirmed against the PR head; Format, Clippy ×3 OS, Test ×3 OS, MSRV (1.91.0), Detect code changes, CI Gate, all pass. It covers the Cargo workspace, which this PR does not touch; it covers **neither** script, since no workflow and no lefthook entry invokes them — that gap is why the suite run above is RAN rather than READ. Also READ: #881 body and doneWhen; Round 2 @ `356b636` and the `Review Response: Round 2` in full; `docs/fleet/rules.md` R22 on `origin/main`; REVIEW.md at base `6849b90b`.

**Method note.** Every enumerator above used three dots (`origin/main...bd1513f`) or the explicit delta range. Two dots reports 244 changed files here because the branch is behind `main`, and Round 2 recorded that it produced phantom findings; I did not repeat that.

### Verdict

**LGTM (P0=0, P1=0).**

Round 2's blocker is genuinely fixed, and I proved it the way it deserved: with the assertion text verified identical to the repo's, `:162` is red with a planted leftover and green without one. Both P2s were addressed; one of them, the README recipe, only half — its pi launch line still teaches `--exclude-tools` where the script uses a `--tools` allowlist. That stays P2 and does not block.

The hunt this round was asked to run found three more assertions in the same suite that pass by matching nothing, and one of them — `:309-312` — has no failure branch at all and cannot fail under any input. Every one is byte-identical at `c2dc3c3`, so `loop` item 2 and U7 route them to follow-up rather than into this verdict, and `loop` item 6 points the same way: this is the third round on a harness-only PR. I record the routing explicitly so it is auditable, not to soften it — `:309-312` is worth a commit of its own, and the instrument built to measure whether reviewers catch checks-that-cannot-fail should not ship carrying one.

Merge is the operator's action, never the reviewer's (`loop`; GATE-01). No label and no commit status are set by this round.

Issue #927, opened 2026-09-05
test(fleet): eight scripts/test-*.sh and three scripts/fleet/test-*.ps1 are still machine-run by nothing after #896


fagemx:
## What happened

#896 / PR #910 wires `scripts/fleet/test-*.sh` into CI. Its scope stops at that directory, and the review of #910 turned up the rest of the same defect one directory over.

After #910 lands, these test files are still executed by no machine gate:

**`scripts/test-*.sh` — eight files, none globbed by the new runner** (it matches `scripts/fleet/test-*.sh` only):

- `scripts/test-doc-citations.sh`
- `scripts/test-fleet-claim-issue.sh`
- `scripts/test-git-config-guard.sh`
- `scripts/test-lint-markdown-content.sh`
- `scripts/test-pr-review-watch.sh`
- `scripts/test-review-capabilities.sh`
- `scripts/test-review-ownership.sh`
- `scripts/test-review-posix-snapshot.sh`
- `scripts/test-review-pr.sh`

**`scripts/fleet/test-*.ps1` — three files, outside the `.sh` glob:**

- `scripts/fleet/test-detached-dispatch.ps1`
- `scripts/fleet/test-lane-reap.ps1`
- `scripts/fleet/test-lane-terminal-receipts.ps1`

Plus `scripts/fleet/test-lane-helpers.sh`, which #910's runner skips by name because it needs Windows Scheduled Tasks, `pwsh.exe` and `taskkill` — it is raised as a blocking finding on #910 itself and is listed here only so the full picture is in one place.

## Why it matters

The subjects of the ungated tests are not incidental. `scripts/test-pr-review-watch.sh` covers the script that writes the `Independent Review` merge-gate status; `scripts/test-review-pr.sh` and `scripts/test-review-capabilities.sh` cover the review launcher and its read-only capability probes; `scripts/test-fleet-claim-issue.sh` covers the claim gate R21 requires before any lane starts. These are the exact files where a silent regression escapes the review gate rather than merely annoying someone — and they are on the 審判面 and 閘門面, where the rules assume the machine is checking.

It is the same shape #896 was filed for: a test that nothing runs is a test in name only.

## Done when

- Every `scripts/test-*.sh` on `main` runs in CI on a push that touches it, and a red test blocks. Extending #910's `detect` path filter and its glob is the obvious route; a second glob term plus `scripts/*` in the filter is likely the whole change.
[...]

PR #910, opened 2026-09-05
test(fleet): run scripts/fleet/test-*.sh in a CI gate (GH-896)

fagemx:
Issue: #896

## Summary

No machine gate ran any fleet shell test — grep over `.github/` found nothing and `lefthook.yml` runs cargo only — so their doneWhen rested on hand execution. This wires them into CI.

Running them for the first time found that several were already broken. That is the gate working, and it is why this PR carries **no closing keyword**: one of the tests #896 names by hand runs nowhere after this lands, so #896 stays open for it.

## What runs where, at this head

| where | what |
|---|---|
| `fleet-tests` (ubuntu) | `sh scripts/fleet/run-fleet-tests.sh` — glob over `scripts/fleet/test-*.sh`, minus the three quarantines below |
| `fleet-tests-windows` | `sh scripts/test-review-capabilities.sh` — one test, invoked directly |
| quarantined, runs nowhere | `scripts/fleet/test-lane-helpers.sh` (**#963**), `scripts/fleet/test-next-loop.sh` (**#964**), `scripts/fleet/test-collision-scan.sh` (**#971**) |

The runner's match list is **one glob**, `scripts/fleet/test-*.sh`. `scripts/test-review-capabilities.sh` is **not** matched by it — it lives one directory up and the windows job invokes it directly. Widening to `scripts/test-*.sh` is deliberately not done here (#927).

Both jobs check out with `fetch-depth: 0`: tests that resolve `origin/main` exited on `fatal: ambiguous argument` under the default depth-1 checkout.

`ci-gate` takes both jobs as `needs` and evaluates them **before** its early success exits, so a red fleet job cannot be swallowed; a skip is accepted only when `detect.outputs.fleet == 'false'`.

## The five defects the gate found, none introduced here

| defect | where | disposition |
|---|---|---|
| depth-1 checkout has no `origin/main` | both jobs | fixed here (`fetch-depth: 0`) |
| generated helper carries `set -o pipefail`; ubuntu's `sh` is dash (the `set -o pipefail` it writes into that runner — no line number: this citation has already gone stale once) | `test-review-capabilities.sh` on ubuntu | test moved to the windows job; fixing `review-pr.sh` belongs to the PRs that own it |
| red on `windows-latest` (case 0) **and** on a Windows workstation against `origin/main` (case 25, Task Scheduler) | `test-lane-helpers.sh` | **quarantined — #963** |
[...]

PR #867, opened 2026-09-04
fix(fleet): finish review lane terminal lifecycle

fagemx:
Issue: #772

Closes #772

Scheduled fleet tasks now record terminal outcomes and unregister on completion or failure. Reaping checks both the current task state and its registration identity before removal, so a stale snapshot cannot remove a replacement task.

Writer lanes require canonical repository-relative ownership scopes and forward every supplied scope to the dispatch claim/heartbeat runtime delivered in #860. Optional build lanes retain the shared `lane-warm -PrintEnv` environment contract.

Review tracks share atomic round reservations, derive the published round floor from trusted authors, and require successful dispatch and final terminal receipts before publishing a verdict. The merge helper accepts only a trusted current-head qualified verdict and uses a matching-head merge guard.

Validation covers task replacement and state races, launch/stop failure receipts, real Scheduler dry-run completion with multiple ownership scopes, generated Windows review startup failures, trusted round allocation, and failed-review admission. Exact-head CI and the independent SOL review are recorded in the numbered review loop below. No local Cargo run is needed for this scripts/docs delta.

PR #890, opened 2026-09-04
fix(fleet): review-pr.sh opt-in pi arm — EDDA_REVIEW_AGENT=pi with the read-only allowlist, Anthropic models refused on pi, pi-dispatch capability canaries (GH-880)

fagemx:
## Problem and change

`scripts/review-pr.sh` could launch a reviewer through exactly two arms, both Claude: `edda dispatch --agent claude` and the oversized-brief `claude -p` stdin fallback, so the pool's pi-hosted cheap engine (e.g. `openrouter/z-ai/glm-5.3-flash`) had no production transport (GH-708 removed the pi arm for the Opus transport). This adds an opt-in `EDDA_REVIEW_AGENT=pi` arm that launches `edda dispatch --agent pi` with the read-only allowlist `read,grep,find,ls` (review.execution-policy), passes the per-PR `--session-id` (no `--resume`), and lets the `Model requested:` / `Model observed:` receipt come from `edda dispatch` itself. With `EDDA_REVIEW_AGENT` unset, the generated lane/runner and the `--dry-run` output are byte-identical to the base SHA (verified by cmp below). Per `fleet.claude-subscription-transport`, `EDDA_REVIEW_AGENT=pi` with any `anthropic/*` or `claude-*` model — and any unknown `EDDA_REVIEW_AGENT` value — exits 2 with a one-line reason before any directory is created or file written, `--dry-run` included. The capability guard grows a third arm `pi-dispatch` in both `scripts/reviewer-capabilities.sh` and `scripts/fleet/reviewer-capabilities.ps1`, the watcher's provider probe follows the agent, and `scripts/test-review-capabilities.sh` gains pi canaries plus the two refusal cases.

### RAN

- `sh scripts/test-review-capabilities.sh` exit 0; last four lines (the two git CRLF warnings are fixture noise):
  - `review capability canaries passed (unguarded baseline; old/modern dispatch and fallback; all source scopes; pi-modern launch and pi-old/pi-old-edda refusals; anthropic/codex refusals)`
  - `Windows generated lane canaries passed (old/modern transport and source-snapshot cases; pi-modern launch and pi-old/pi-old-edda refusals)`
[...]

PR #929, opened 2026-09-05
fix(fleet): capability canaries read DISPATCH_EXIT and report every case

### LINKED ISSUE #928: fix(fleet): test-review-capabilities.sh exits at the unguarded-baseline case since #867 — the exit-code clause folds in teardown, hiding every pi and refusal case
## Evidence

`scripts/test-review-capabilities.sh` fails on `origin/main` and has since 2026-09-05 02:37Z.

Bisected to one commit:

| SHA | tree | result |
|---|---|---|
| `1bc2e57` (`f675991^`) | before PR #867 | `review capability canaries passed (…)` and `Windows generated lane canaries passed (…)` |
| `f675991` (PR #867) onward, incl. `1418557` (current main) | after | `FAIL unguarded baseline: old dispatch shape did not write owned canary`, exit 1 |

The failing case is the suite's **negative control**: it takes the generated POSIX runner, `sed`s away the capability guard and the restrictive flags, runs it, and asserts the canary file reads `TAMPERED` — proving that the guard, and not something incidental, is what stops an unrestricted dispatch. With that control dead, every "refusal" assertion after it is unproven, and the suite exits before reaching them: on this tree it never runs the modern-dispatch, fallback, source-scope, or Windows generated-lane cases at all.

## Likely cause, stated as such

#867 wrapped the runner's dispatch in a source-snapshot precondition:

```
scripts/review-pr.sh:676  if [ \$? -eq 0 ]; then source_ready=1; else source_ready=0; echo 'review-pr runner: source snapshot unavailable before dispatch' >> '$LOG'; fi
scripts/review-pr.sh:692  if [ "\$source_ready" = 1 ]; then
```

The fixture strips the capability guard but sets nothing up for the snapshot, so the dispatch block is very likely skipped whole. Supporting observation rather than proof: the fixture cats the runner's stdout on failure and it is **empty**, which fits "the block never ran" and not "the dispatch ran and failed" — and the runner's own explanation goes to `$LOG`, which the fixture does not read. Confirming this is the first step, not an assumption to build on.

## Why it merged

Nothing runs this test. It is not in `.github/workflows/ci.yml` and not in `lefthook.yml`; it is invoked by hand or by a lane brief. #896 and #910 describe the same gap for `scripts/fleet/test-*.sh`; this is that gap on the file that guards the review machinery's own capability boundary.

Two things are true about how it got in, and both matter:

- The **gate** gap is the fixable one: an offline test that no gate runs will go red and stay green-looking.
- The **round** also missed it: #867 changes `scripts/**`, and REVIEW.md routes that to code-risk, where the reviewer is expected to run the change's own offline test. #867 was one of six PRs merged on 2026-09-05 on a glm authoritative verdict with the gate's auto-merge (#888). This is the first measured quality datum on those six and it is negative — but the honest reading is that the engine was operating without the mechanical backstop this issue asks for, not that the engine is the root cause. The 09-07 compare should weigh it that way.

## Consequence right now

PR #890 is blocked on it. Its merge-from-`main` round produces a tree whose failure set is **identical** to main's (measured: the same single `FAIL unguarded baseline` line, nothing new), but because the suite exits at the first failure, the pi-arm canaries the PR exists to prove — `pi-modern` launch, `pi-old`/`pi-old-edda` refusals, the `TOOL_FLAGS` receipt equality — never execute. #890 cannot demonstrate its own acceptance until this is green.

## doneWhen

1. The cause is identified and stated in the PR body: which precondition the fixture fails to satisfy, and whether the fixture or the runner owns the fix.
2. `sh scripts/test-review-capabilities.sh` → exit 0 on the PR head, with both summary lines printed.
3. The unguarded-baseline control genuinely still controls: with the capability guard restored, the same fixture must **not** write the canary. State both runs in the PR body — a control that passes in both directions or neither is not a control.
4. The suite reports **every** failing case rather than exiting at the first: a failure increments a counter and the run continues, so one dead control can no longer hide the rest of the suite. (This is what let a single break mask eleven downstream assertions.)
5. `sh -n` on `scripts/review-pr.sh`, `scripts/reviewer-capabilities.sh` and `scripts/test-review-capabilities.sh` → exit 0.

## Predicted surface

`scripts/test-review-capabilities.sh`, and `scripts/review-pr.sh` only if doneWhen 1 finds the runner at fault

## Class and review

審判面 (`docs/fleet/rules.md` R22): both files are review machinery, so the authoritative round is Opus and a glm round is SHADOW.

## Sequencing

Before PR #890. Getting this test into a gate is #910's job, not this issue's — this one makes it green so a gate is worth adding.

fagemx:
## Problem and change

`scripts/test-review-capabilities.sh` has exited 1 at its first case since #867 (bisected: `1bc2e57` green, `f675991` onward red), so neither summary line has been reached and every case after the first has gone unrun. Nothing noticed, because no CI job or hook runs this test.

The unguarded-baseline control was **not** broken. It works: the canary reads `TAMPERED`, a call is recorded, and the receipt says `DISPATCH_EXIT=0`. What broke is that five assertions read the runner's **process** exit to mean "did the dispatch run". Since #867 that exit is `FINAL_EXIT`, which also folds in worktree and task teardown — and this fixture's `wt-review-pr9998` is a standalone `git init` repo rather than a worktree registered against `$ROOT`, so teardown cannot succeed there and `FINAL_EXIT` is 2 on a perfectly good dispatch.

Two changes, both in the test:

1. Every "did the dispatch run" assertion reads `DISPATCH_EXIT` from `review-pr9998-r1.done` — the field #867 added to the receipt for exactly this distinction. Five sites: the unguarded control, the POSIX capable-backend case, and the Windows capable-backend case (the last two covering `modern` and `fallback` each). Their failure messages now print `DISPATCH_EXIT` and the process exit side by side, so the next reader does not have to re-derive which clause fired.
2. Every case records a failure and continues; the single `exit 1` gate moved to the end of the file.

**Change 2 is what found the other four.** Before it, the suite exited at the unguarded case and the four `modern` / `fallback` failures — POSIX and Windows — were invisible. They are the same defect, and they had been there since #867 too.

### Audit of the other exit-code assertions

The `mutate` cases at both blocks test `[ "$rc" -eq 0 ]` expecting a **non-zero** exit, which is correct under #867's semantics: a rejected source change *should* make the runner fail. The `old` / `old-claude` refusal cases likewise assert on the process exit of a run that never dispatched. Both are unchanged.

### RAN at 872f7cec428549596c5093e35ea253f17a9e3fd3

- before round 1: exit 1, one `FAIL unguarded baseline`, **zero** `canaries passed` lines
- after round 1 (early exits removed, control fixed): exit 1 with **four** newly visible `FAIL` lines
- after round 2: **exit 0**, zero `FAIL` lines, both summary lines present
- `sh -n scripts/test-review-capabilities.sh` → exit 0
- `grep -c 'exit 1'` → 1 (the final gate only)
- `git diff --cached --name-only` → this file only

### READ

- the bisect and the preserved-fixture diagnosis in #928
- #867's `DISPATCH_EXIT` / `FINAL_EXIT` receipt fields in `scripts/review-pr.sh`

## Not in this PR

Putting this test behind a gate is #910 / #927. Until one of those lands it is still only run by hand — which is how five broken assertions survived a day on `main`.

Issue: #928
Closes #928

### DIFF
--- a/scripts/test-review-capabilities.sh
+++ b/scripts/test-review-capabilities.sh
@@ -88,15 +88,22 @@ BACKEND=unguarded; export BACKEND
 printf ORIGINAL > "$CANARY"
 : > "$CALLS"
 printf brief > "$tmp/scratch/review-pr9998-r1-brief.md"
+failures=0
 sh "$tmp/unguarded-old-run.sh" > "$tmp/old-out" 2>&1 || old_rc=$?
 old_rc=${old_rc:-0}
-if [ "$(cat "$CANARY")" != TAMPERED ] || [ ! -s "$CALLS" ] || [ "$old_rc" -ne 0 ]; then
+# This control asks one thing: does an UNGUARDED dispatch reach the backend?
+# So it reads DISPATCH_EXIT from the receipt, not the runner's process exit.
+# Since GH-867 the process exit also folds in worktree and task teardown, and
+# this fixture's wt-review-pr9998 is a standalone `git init` repo rather than a
+# worktree registered against $ROOT, so `git worktree remove` fails here by
+# construction and FINAL_EXIT is always 2 — which made the control look dead
+# when it was working (GH-928).
+unguarded_dispatch_exit=$(sed -n 's/^DISPATCH_EXIT=//p' "$tmp/scratch/review-pr9998-r1.done" | head -1)
+if [ "$(cat "$CANARY")" != TAMPERED ] || [ ! -s "$CALLS" ] || [ "$unguarded_dispatch_exit" != 0 ]; then
   cat "$tmp/old-out" >&2 || true
-  echo 'FAIL unguarded baseline: old dispatch shape did not write owned canary' >&2
-  exit 1
+  echo "FAIL unguarded baseline: unguarded dispatch did not reach the backend (canary=$(cat "$CANARY") DISPATCH_EXIT=${unguarded_dispatch_exit:-absent} process_exit=$old_rc)" >&2
+  failures=$((failures + 1))
 fi
-
-failures=0
 for BACKEND in old modern old-claude fallback mutate; do
  export BACKEND
  printf ORIGINAL > "$CANARY"
@@ -128,16 +135,24 @@ for BACKEND in old modern old-claude fallback mutate; do
      echo 'FAIL mutate: tracked, untracked, or ignored source change was accepted' >&2; failures=$((failures + 1))
    fi ;;
  *)
-   if [ ! -s "$CALLS" ] || [ "$rc" -ne 0 ]; then
-     echo "FAIL $BACKEND: capable backend never launched"; failures=$((failures + 1))
+   # Same reading as the unguarded control above: "did the dispatch run" is
+   # DISPATCH_EXIT, not the process exit, which since GH-867 also carries
+   # teardown the fixture cannot satisfy (GH-928).
+   dispatch_exit=$(sed -n 's/^DISPATCH_EXIT=//p' "$tmp/scratch/review-pr9998-r1.done" | head -1)
+   if [ ! -s "$CALLS" ] || [ "$dispatch_exit" != 0 ]; then
+     echo "FAIL $BACKEND: capable backend never launched (DISPATCH_EXIT=${dispatch_exit:-absent} process_exit=$rc)"; failures=$((failures + 1))
    fi
    if ! grep -q '^TOOL_FLAGS=.*--permission-mode.*mcp__' "$tmp/scratch/review-pr9998-r1.done"; then
      echo "FAIL $BACKEND: no actual flags receipt"; failures=$((failures + 1))
    fi ;;
  esac
 done
-[ "$failures" -eq 0 ] || exit 1
-echo 'review capability canaries passed (unguarded baseline; old/modern dispatch and fallback; all source scopes)'
+# No early exit: the Windows block below must run even when a POSIX case
+# failed, so the operator sees every failing case in one run (GH-928). The
+# single gate is at the end of the file.
+if [ "$failures" -eq 0 ]; then
+  echo 'review capability canaries passed (unguarded baseline; old/modern dispatch and fallback; all source scopes)'
+fi
 
 # On Windows execute the generated PowerShell lane too. Functions stand in
 # for the two binaries; any non-help call with missing flags writes canary.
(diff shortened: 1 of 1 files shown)


fagemx:
## Code Review: Round 1 — PR #929 @ 872f7cec428549596c5093e35ea253f17a9e3fd3

- elapsed: unmeasured
- model_observed: claude-opus-5 (Claude Code operator session, docs machine)
- spec: REVIEW.md @ base `efa40679094e6a368e089771ed8b0fac34475c3d`
- class: code-risk (`scripts/**`, `*.sh`); surface: 審判面 (`docs/fleet/rules.md` R22 — this file is how the review machinery proves its own capability boundary), so this round is authoritative and a glm round here is SHADOW
- escalations: none
- cost: one operator session; the change took two lane rounds, $0.02 total

**The strongest reviewer conflict so far, disclosed.** I filed #928, diagnosed its cause, and wrote every `oldText`/`newText` in the brief that produced this diff. The lane was a typist. So "did the lane do what was asked" is worth almost nothing here, and I have not leaned on it: this round is built on two adversarial mutations that try to show **my own change is wrong**, plus an independent re-derivation of whether the fix could have loosened the thing it touches.

### IN SCOPE

`scripts/test-review-capabilities.sh` only: +36/−15. Five assertions move from the runner's process exit to `DISPATCH_EXIT`, and six early `exit 1`s become a failure counter with one gate at the end. No production script changes. Acceptance ceiling: #928's corrected doneWhen (the list in its newest comment, not the one in the body).

### The risk this change carries

Reading a *different, weaker* field to make a red test go green is indistinguishable from fixing it, unless you show the test still fails when it should. That is the whole review. Two mutations, both run at this head in a detached worktree, both restored afterwards.

**Mutation A — neuter the capability guard.** Inserted `return 0` at the top of `review_capabilities()` in `scripts/reviewer-capabilities.sh`, so every backend is admitted regardless of its flags.

```
MUTATED_EXIT=1
FAIL old: capability refusal did not precede launch
FAIL old-claude: capability refusal did not precede launch
```

The suite goes red, on exactly the two cases that should notice, and reports **both** rather than stopping at the first — which is the counter change doing its job. So the capability boundary is still protected: this PR did not loosen what the suite is for.

**Mutation B — make the receipt lie.** Changed the generated PowerShell receipt in `scripts/review-pr.sh` to always write `DISPATCH_EXIT=9` while everything else stayed correct, isolating the field this PR now reads.

```
review capability canaries passed (unguarded baseline; old/modern dispatch and fallback; all source scopes)
FAIL Windows modern: no successful launch (DISPATCH_EXIT=9 process_exit=2)
```

The POSIX section passes — correct, I only mutated the Windows receipt — and the Windows case fires. So the new assertion is **not vacuous**: it detects a dishonest `DISPATCH_EXIT`, and the rewritten message prints both numbers so the next reader sees which clause fired without re-deriving it.

Together: the fix reads a field that still discriminates, and the boundary it guards still holds.

### Was any coverage lost?

Moving off the process exit means these five assertions no longer notice a **teardown** failure. That sounds like a real loss and is not one: in this fixture `wt-review-pr9998` is a standalone `git init` repo, so `git worktree remove` can never succeed and `FINAL_EXIT` was 2 on every run. The clause was constant-false — it caught nothing before and catches nothing now. Teardown *detection* remains covered by the `mutate` cases, which assert `WORKTREE_CHECK=failed` and deliberately still read the process exit. Teardown *cleanup* is untested, as it was before this PR; noted as a follow-up rather than charged to this change.

### Rules

|---|---|---|---|---|
| U1 surface | any | P0 | PASS | one changed path == #928's `Predicted surface` |
| U2 closing keyword ⟺ doneWhen | any | P1 | PASS | `Closes #928`; the six corrected doneWhen items are checked below |
| U3 `Issue: #N` | any | P1 | PASS | `Issue: #928` present — the miss I found on #915 and #916 did not repeat |
| U4 conventional subject | any | P1 | PASS | `fix(fleet): capability canaries read DISPATCH_EXIT and report every case` |
| U5 markdown lint | any | P1 | N.A. | no markdown changed |
| U6 gates READ before RAN | any | P0 | PASS | run at this head: all ten jobs green, `CI Gate` pass (`ci.yml:78` whitelists `scripts/*`, so the full cargo matrix runs despite no crate) |
| U7 whole scoped surface | any | P1 | PASS | 36 added lines, all read |
| C1 doneWhen vs real code | code | P0 | PASS | six items, five RAN below |
| C2 `feat:`/`fix:` with a test | code | P0 | PASS | the change *is* to the test; its fail-without-the-change property is mutations A and B |
| R1 destructive commands | code-risk | P0 | PASS | none added |
| R3 shell scripts parse | code-risk | P0 | PASS | `sh -n` → exit 0 (RAN) |
| R4 resource release | code-risk | P1 | PASS | no resource added; the `sed` reads a receipt the fixture already writes |
| R5 boundary and injection | code-risk | P1 | PASS | `sed -n 's/^DISPATCH_EXIT=//p' … | head -1` over a file the harness itself produced; no interpolation of external text |

**doneWhen (the corrected list), item by item.** (1) The control reads `DISPATCH_EXIT` — present, and mutation B proves it discriminates. (2) The audit is in the PR body and is right: the `mutate` cases expect a **non-zero** exit and correctly still read it; the `old`/`old-claude` refusals assert on a run that never dispatched. (3) Both directions hold in one run — the unguarded case asserts the canary **is** `TAMPERED` while every guarded case asserts it **is not**, and mutation A shows the guarded direction fails when it should. (4) Every case records and continues: mutation A printed two failures where the old code would have printed one; `grep -c 'exit 1'` → 1, at line 234 of 234. (5) `sh scripts/test-review-capabilities.sh` → **exit 0**, both summary lines. (6) `sh -n` → exit 0.

### Findings

none (P0=0, P1=0).

### What this change bought, stated because it is the reason to merge it

The early exit hid four failures for a day. Removing it surfaced them in the same run that fixed the first — `FAIL modern`, `FAIL fallback`, and both Windows equivalents, all the same defect. Five broken assertions in the file that proves reviewers cannot launch an unrestricted backend, none of them visible, because one dead clause stopped the suite at case one and no gate runs it.

### FOLLOW-UP ISSUE

- Worktree **cleanup** failure is asserted nowhere, before or after this PR. The fixture cannot exercise it (its `wt-review-pr9998` is not a registered worktree), so covering it needs a fixture that registers a real linked worktree — a different piece of work from this one.
- The `old`/`old-claude` refusal cases test `[ "$rc" -eq 0 ] || [ -s "$CALLS" ]`. In this fixture `rc` is never 0, so that clause is dead and the assertion rests entirely on the calls file. It is doing the right thing for half a reason; mutation A confirms it fires, but the dead clause invites the same confusion this PR just resolved.
- Getting this suite behind a gate is #910 / #927. Until one lands, it is still only run by hand — which is how five broken assertions survived a day on `main`.

### RAN vs READ

RAN (at `872f7cec`, detached review worktree, verified clean before and restored after each mutation): mutation A (guard neutered → exit 1, two expected failures) and mutation B (receipt forced to `DISPATCH_EXIT=9` → Windows case fires, POSIX unaffected); `sh scripts/test-review-capabilities.sh` → exit 0 with both summary lines; `sh -n` → exit 0; `grep -c 'exit 1'` → 1 and its line number vs the file length; the full diff read line by line; `git diff --name-only origin/main...HEAD`; the U3 body probe.

READ: `gh pr checks 929` at this head; #928's corrected doneWhen; #867's receipt fields in `scripts/review-pr.sh`.

### Verdict

**LGTM (P0=0, P1=0).**

`Independent Review` will be set to success on this SHA from this round, per R22. Merging it also unblocks PR #890: that PR's merge-from-`main` currently inherits a suite that exits before reaching the pi canaries it exists to prove — measured on its resolved tree, which produces the same single `unguarded baseline` failure as `main` and no new one.

PR #755, opened 2026-09-03
fix(scripts): pr-review-launch resolves pwsh.exe to an absolute path and DryRun fails on non-zero LastTaskResult (GH-694)

fagemx:
Issue: #694

Residue of doneWhen item 5: `scripts/pr-review-launch.ps1:69` registered the
watcher task with the bare `-Execute "pwsh.exe"`. Task Scheduler cannot
resolve that name when PowerShell is a Store/WindowsApps install, so the
watcher died before ever running its wrapper — one reason this repo's PRs
sat unreviewed.

## Change (674b4c4, scripts-only)

- `scripts/pr-review-launch.ps1:72-73` — resolve `pwsh.exe` via
  `Get-Command` to an absolute path; `throw` with a clear message when not
  found, instead of registering a task that cannot start. Same pattern as
  the already-fixed `scripts/fleet/lane-launch.ps1:108-111`.
- `scripts/pr-review-launch.ps1:75` — `New-ScheduledTaskAction -Execute $PwshExe`.
- `scripts/pr-review-launch.ps1:102-108` — `-DryRun` now **exits 1** when
  `LastTaskResult` is non-zero and prints the hex code, so a dead
  registration can no longer silently pass (before: it printed the
  non-zero result and exited 0).

### Only one registration site

`pr-review-launch.ps1` has exactly one `New-ScheduledTaskAction` (line 75,
post-fix); both the real path and `-DryRun` share it — `Register-ScheduledTask`
(line 84) runs before the `$DryRun` branch, which differs only in the wrapper
name and `--once --dry-run` watcher args. The GH-694 "real + DryRun both"
concern for this file is therefore covered by this single fix.

## DryRun receipts (this machine, pwsh = Store/WindowsApps install)

Before (base `df3bd4d`):

```
> pwsh -NoProfile -ExecutionPolicy Bypass -File scripts/pr-review-launch.ps1 -DryRun
dry-run task=edda-pr-review-watcher state=Ready
LastRunTime=09/03/2026 11:20:54 LastTaskResult=2147942402
task=edda-pr-review-watcher unregistered (dry-run)
EXIT=0
```

`2147942402` = `0x80070002` ERROR_FILE_NOT_FOUND — and note exit 0: the
defect passed silently.

After (this PR, 674b4c4):

```
> pwsh -NoProfile -ExecutionPolicy Bypass -File scripts/pr-review-launch.ps1 -DryRun
dry-run task=edda-pr-review-watcher state=Ready
[...]

Issue #896, opened 2026-09-04
test(fleet): no scripts/fleet/test-*.sh is machine-run anywhere — wire the fleet shell tests into CI or a hook


fagemx:
Found by the glm shadow review of PR #895 (Round 1, @ 6ba187d, FOLLOW-UP slot).

## What happened

`scripts/fleet/test-brief-from-issue.sh` (GH-885) and its siblings are never executed by any machine gate: grep over `.github/` returns no matches and `lefthook.yml` runs cargo only. The doneWhen guarantees (`sh -n` exit 0, test exit 0) currently rest on hand execution.

## Why it matters

A fleet shell test that nothing runs is a test in name only; a regression in any `scripts/fleet/*.sh` helper ships to lanes green.

## Done when

- The fleet shell tests (`scripts/fleet/test-*.sh`, at least `test-brief-from-issue.sh`, `test-review-capabilities.sh`, `test-lane-helpers.sh`, `test-ready-queue-lint.sh`, `test-daily-digest.sh`) run in CI or a pre-push hook on every push that touches them, and a red test blocks.
- Evidence in the PR: the CI/hook config diff plus a run showing the gate firing on a seeded failure.

fagemx:
結案・已交付：#910（scripts/fleet/test-*.sh 進 CI gate）已合併。其餘未被機器跑的 shell 測試依 #1010 裁定 A 屬過渡層，不另開單。

Issue #650, opened 2026-09-02
docs(cli): docs/reference/cli.md stops at 0.2 — task, dispatch, verdict, ratify, checkpoint, peers, phase and 25 more verbs are undocumented


fagemx:
Stage 1 門票項目（裁定 `roadmap.stage1-order`，Tim 2026-09-02）。Lane B（docs-only，無 build lane）。

## What happened

在 main `275b0b39dfe0a13064cd7898c39e2d59068e0f07`：`edda --help` 列出 60 個 top-level 動詞（不含 `help`）；`docs/reference/cli.md` 有 30 個 `### edda <verb>` 節。以下 0.3/0.4 的核心動詞在 cli.md **零提及**：

`task` `dispatch` `verdict` `ratify` `checkpoint` `peers` `phase`

其餘同樣未出現的：`actor` `group` `sync` `unclaim` `coord` `setup` `recap` `export` `hook` `intake` `prs` `pipeline` `bundle` `brief` `policy` `notify` `pair` `serve` `user` `rules` `scan` `propose-issue` `propose-patch` `skill` `tool-tier`。

也就是 cli.md 停在 0.2 的形狀；2026-09-02 外部審計原話：「0.3 和 0.4 的核心功能在參考文件裡不存在」。

## Why it matters

參考文件是陌生人判斷「這東西能做什麼」的第一個地方。README 賣 Layer 2（dispatch / verdict / task），cli.md 卻一個字沒有——讀者會判定 README 在吹。

## Suspected surface

- `docs/reference/cli.md` — 補齊；分組沿用既有節（Getting started / Memory & querying / Recording / Coordination / Branches & drafts / Integration / Maintenance / Orchestration），新增節放對組
- `scripts/check-cli-docs.sh`（新）— 比對 `edda --help` 的動詞集合與 cli.md 的 `### edda <verb>` 標題集合；差集非空即 exit 1（規則能被機器檢查就自動化，不靠人記得）
- `.github/workflows/ci.yml` — 在 **Linux Test job** 的 `cargo test` 之後加一個 step 跑該腳本（那裡 edda binary 已建好；Format job 不建 binary，不放那裡）；**不新增 required check**（CI Gate 聚合不動）
[...]

PR #972, opened 2026-09-06
fix(fleet): a killed lane leaves a readable log and an honest status; -Agent claude gets a session id the backend accepts (GH-748, GH-694)

### LINKED ISSUE #694: fix(fleet): the merged lane launcher cannot start a lane on this machine — bare 'pwsh.exe' is unresolvable to Task Scheduler, and the default session id is rejected by the claude backend
Found 2026-09-02 by the workstation controller while dispatching the #669/#678/#672/#690 wave. Basis `origin/main` `030b36b2048431a443407d8c0ba3d035351c8a0f` (PR #668 / GH-606 merged 2026-09-02T14:42:14Z).

## 缺陷

**`scripts/fleet/lane-launch.ps1` 在這台機器上啟動的每一條 lane 都會立刻死。** 兩個獨立缺陷，兩個都在 merged 的程式碼裡，`-DryRun` 兩個都測不到。

### 缺陷一：`-Execute "pwsh.exe"` 排程器解析不到

`lane-launch.ps1:~/Register-ScheduledTask` 用裸檔名註冊：

```powershell
$action = New-ScheduledTaskAction -Execute 'pwsh.exe' ...
```

本機的 pwsh 是 WindowsApps/Store 安裝：

```
> (Get-Command pwsh.exe).Source
C:\Program Files\WindowsApps\Microsoft.PowerShell_7.6.5.0_x64__8wekyb3d8bbwe\pwsh.exe
```

WindowsApps 的 alias 目錄不在排程任務的解析路徑上，任務因此秒死：

```
> (Get-ScheduledTask -TaskName 'edda-lane-gh669').Actions | % { $_.Execute }
pwsh.exe
> (Get-ScheduledTaskInfo -TaskName 'edda-lane-gh669').LastTaskResult
2147942402          # 0x80070002 ERROR_FILE_NOT_FOUND
```

launcher 自己的 post-check 抓到了狀態不對，但錯誤訊息指向錯的方向：

```
lane-launch: task edda-lane-gh669 did not reach Running after Start-ScheduledTask (state: Ready)
```

log 檔零位元組、done-file 不存在——看起來像 lane 沒被啟動，實際是 lane 啟動了、找不到直譯器。

**被取代的舊機制沒有這個問題**：`C:\ai_agent\fleet-workstation\lane-gh649.ps1` 對應的任務用的是絕對路徑 `C:\Program Files\WindowsApps\Microsoft.PowerShell_7.6.5.0_x64__8wekyb3d8bbwe\pwsh.exe`。**這是 #668 的回歸。**

`scripts/pr-review-launch.ps1` 與 `scripts/review-pr.sh:211` 用同一個裸檔名寫法，**PR 審查 watcher 有同一個缺陷**。

### 缺陷二：預設 session id 不是 UUID，claude backend 直接拒絕

```powershell
if (-not $SessionId) { $SessionId = "lane-$Name" }     # lane-launch.ps1
```

claude backend 要求 session id 是合法 UUID。逐字重現：

```
> claude -p "reply OK" --verbose --output-format stream-json --session-id probe-1 --permission-mode bypassPermissions
Error: Invalid session ID. Must be a valid UUID.
rc=1

> claude -p "reply OK" --verbose --output-format stream-json --session-id e8cb7e0e-6044-4964-9bcf-bedf5b425de8 --permission-mode bypassPermissions
{"type":"system","subtype":"hook_started",...}
rc=0
```

所以修好缺陷一之後，任何 `-Agent claude` 的 lane 仍然秒死。lane log 逐字：

```
Cost: n/a (no usage data reported)
Session: lane-gh669
Error: agent exited with code 1 without result
```

`--agent pi` 與 `--agent codex` 不受影響（session id 語意各自不同），所以這個缺陷**只在 claude lane 上出現**——而 claude 正是設計／審查 lane 用的 backend。

## 為什麼重要

1. **兩個缺陷合起來讓 merged 的 launcher 在本機完全不能用。** 今天這一波四條 lane（#669 #678 #672 #690）沒有一條靠 launcher 自己起得來，全部要外部修補任務動作＋自帶 UUID 才啟動。
2. **`-DryRun` 通過但真實啟動失敗。** DryRun 換掉 `__RUN__`（跑 `Start-Sleep 20` 而非 agent），所以缺陷二測不到；而 DryRun 走的是同一個裸 `pwsh.exe`——如果 DryRun 在開發機上曾經通過，代表該機的 pwsh 不是 Store 安裝，**驗收環境與執行環境不同**。
3. **失敗訊號指錯方向**（與 #669／#672 同一家族）：「did not reach Running」讀起來像排程器沒啟動，實際是啟動了但找不到 pwsh；log 零位元組讓人以為 lane 沒跑。
4. **審查 watcher 同樣中招**（`review-pr.sh:211`），所以 PR 審查也起不來——這正是本 repo 現在八張 PR 全部未審查的原因之一。

## Suspected surface

- `scripts/fleet/lane-launch.ps1` — `New-ScheduledTaskAction -Execute 'pwsh.exe'`（真實與 DryRun 兩處）、`$SessionId = "lane-$Name"` 預設值
- `scripts/pr-review-launch.ps1` — 同樣的 `-Execute "pwsh.exe"`
- `scripts/review-pr.sh:211` — 生成的 launch 腳本裡同樣的裸檔名

## doneWhen

- lane-launch 註冊任務時用**解析後的絕對路徑**（`(Get-Command pwsh.exe).Source`），解析不到就 `Fail` 並說明，而不是註冊一個必死的任務。
- 預設 session id 是合法 UUID；或 launcher 對 `-Agent claude` 驗證 `-SessionId` 形狀，不合法就 `Fail` 並說明。
- **先驗證 FAIL**：在修改前用今天的 launcher 起一條 `-Agent claude` 的 lane，記下 `LastTaskResult=2147942402`（修缺陷一前）與 `Invalid session ID`（修缺陷一後）兩份收據。
- `-DryRun` 能抓到這兩個缺陷：它必須用真實的 Execute 路徑註冊，且必須以真實 backend 能接受的 session id 形狀驗證（或至少對形狀做斷言）。
- `scripts/pr-review-launch.ps1` 與 `scripts/review-pr.sh` 的同一寫法一併修，或在本單說明為何不受影響（附證據）。

## Wiring audit

| 面 | writer & shape | reader | failure signal | layer reach |
|---|---|---|---|---|
| 排程任務的 `Execute` | `lane-launch.ps1` / `pr-review-launch.ps1` / `review-pr.sh:211`，裸 `pwsh.exe` | Windows 排程器 | **今天指錯方向**：`LastTaskResult=0x80070002` + 零位元組 log + 「did not reach Running」讀起來像沒啟動 | fleet harness：本機每一條 lane 與每一次 PR 審查 |
| lane 的 session id | `lane-launch.ps1` 預設 `lane-<name>` | `edda dispatch` → claude backend | `agent exited with code 1 without result`，不說是 session id | 所有 `-Agent claude` 的 lane（設計與審查） |

- #672（`Stop-ScheduledTask` 不殺 process tree）是同一支腳本的另一半：那張讓「停」真的會停，本張讓「起」真的會起。
- #669（dispatch 把失敗說成成功）：本單的缺陷二被 dispatch 如實回報成 exit 1，但人類可讀輸出仍先印了 `Cost:` / `Session:` 摘要才印錯誤——正是 #669 doneWhen 第三條要修的形狀。
- #683（review-pr.sh 從 Git Bash 啟動不了，result 64）：同一支腳本的鄰近缺陷，可一併處理。

fagemx:
Issue: #748
Closes #694.

**#748 is not closed here.** Round 1 found two of its four doneWhen items
undelivered: item 1 (a lane killed mid-L1 still carries a `Review Response`,
with a verified-FAIL/PASS pair) is untouched by this diff, and item 4 (measure
and record local L1 wall-clock) has no record — `edda ask` has none and the
only measurement in the tree is GH-810's debug-symbol one. The `範圍外` note
in #748 excludes the four-lane pool, not the measurement, so it is not cover
for either. #694's five items all hold, so that keyword stays.

Both issues sit on one surface — the lane lifecycle — and both make a lane that did all the work indistinguishable from a lane that did none. #748's residual (log preservation, TIMEOUT translation) and #694's residual (`-Agent claude` session-id shape; #755 already fixed the `pwsh.exe` path half for `pr-review-launch.ps1`) are delivered together as one hygiene change.

## The suspected cause of `logBytes=0` was wrong

#748 named the wrapper's `Tee-Object` buffer. `scripts/fleet/test-lane-delivery.ps1` disproves it: with an agent that emits as it runs, the emitted line survives a mid-run kill on the **unmodified** wrapper.

The real cause is upstream. `crates/edda-cli/src/cmd_dispatch.rs` hard-coded `verbose: false` into `LauncherOptions`, and `print_live` (`crates/edda-conductor/src/agent/stream.rs:207`) is the only thing that puts agent activity on stdout while a turn runs. Nothing reached stdout until the turn ended, so a lane killed at its execution limit had printed nothing for the wrapper to write — which is exactly why all three lanes on 2026-09-03 reported `logBytes=0`.

## What changed

| Surface | Change |
|---|---|
| `crates/edda-cli/src/cmd_dispatch.rs` | new `--verbose`; refused with `--json` (which promises exactly one object on stdout), mirroring the existing `--list-models` refusal |
| `scripts/fleet/lane-launch.ps1` | every dispatch line asks for `--verbose`; default session id is a name-based UUID over `edda-lane-<name>`; an explicit non-UUID id for `-Agent claude` is refused before anything is registered |
| `scripts/fleet/lane-status.ps1` | scheduler result codes are named, and a new `delivery=` field says `complete` / `pending` / `not-started` / `UNDELIVERED(<name>)` |
| `scripts/fleet/test-lane-delivery.ps1` | new offline fixture, 13 assertions |

The session id is **checked, never rewritten** — silently replacing a caller's id would break the resume it was passed for. The derived UUID follows RFC 4122 §4.3, the same shape `review_session_uuid` produces (`scripts/review-pr.sh:456`), so it is stable across relaunches of a lane and distinct between lanes.

## Before / after

```
before: edda-lane-gh748-timeout state=Ready lastTaskResult=267014 logBytes=0 done=False head=- ...
after:  edda-lane-gh748-timeout state=Ready lastTaskResult=267014(SCHED_S_TASK_TERMINATED) logBytes=0 done=False delivery=UNDELIVERED(SCHED_S_TASK_TERMINATED) head=- ...
```

## Verification (L0, this worktree, Windows)

Verified-FAIL first, as both issues require. Baseline on `b96a801` with the fixture in place: 9 of 13 assertions failed, including the verbatim `state=Ready lastTaskResult=267014 logBytes=0 done=False` line #748 reports.

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo clippy -p edda --all-targets -- -D warnings` | clean |
| `cargo test -p edda` | 548 + 24 + 11 + 11 + 10 + … all `ok`, 0 failed |
| `sh scripts/lint-file-length.sh --tree` | rc=0 |
| `pwsh -File scripts/fleet/test-lane-delivery.ps1` | 13/13 PASS |
| `pwsh -File scripts/fleet/test-lane-terminal-receipts.ps1` | 6 PASS, 0 FAIL (no regression) |

The fixture registers, starts and unregisters no real Scheduled Task, and dispatches no agent.

## Out of scope

`verification.cost-discipline`'s four-lane pool and the timeout calibration stay untouched (#626 and the Build lanes footprint come first, per #748's own scope note).


fagemx:
## Code Review: Round 1 — PR #972 @ a07e8816e5e15aacd6798fc3cd3e7f10d2069aad

- elapsed: unmeasured (no pi session file; this round did not run through `pi`)
- model_observed: unverified (no session file / `modelUsage` key available to this round)
- reviewer_session: n/a (round not launched through `lane-launch.ps1` / `review-pr.sh`)
- spec: review-spec-v1.5
- escalations: D5 (adjudicated below, marked `provisional`)
- shadow: false
- cost: 1 fixture run (~40 s), 1 offline `lane-status` repro, ~40 read-only `gh`/`git`/`grep` calls. No Cargo gate run (C5 selector empty).

### IN SCOPE

Changed paths: `crates/edda-cli/src/cmd_dispatch.rs`, `cmd_dispatch_acp.rs`, new `cmd_dispatch_verbose_tests.rs`, `docs/reference/cli.md`, `scripts/file-length-ceilings.txt`, `scripts/fleet/lane-launch.ps1`, `lane-status.ps1`, new `scripts/fleet/test-lane-delivery.ps1`.

Changed behaviour: a new `dispatch --verbose` flag reaching `LauncherOptions.verbose`; the lane launcher asking for it on all three dispatch command lines; a derived name-based session UUID plus a `-Agent claude` UUID refusal; a named scheduler result code and a new `delivery=` field in lane status.

Direct callers/consumers audited: `main.rs:1331` → `cmd_dispatch::run` → `run_inner` (the single entry); `detached_dispatch::launch` / `foreground_argv`; `cmd_dispatch_acp::run` / `validate_args`; `agent_kind::build_launcher` → `ClaudeCodeLauncher`/`PiRpcLauncher`/`CodexLauncher::with_verbose`; `stream.rs:152` and `pi_rpc.rs:522` as the two live-output gates; `lane-reap.ps1` / `lane-stop.ps1` as neighbours of the status field; `scripts/lint-file-length.sh` for the ceiling bump; `run-fleet-tests.sh` / `ci.yml` for the new fixture.

Issue acceptance: #748 doneWhen (4 items) and #694 doneWhen (5 items), both offered under closing keywords.

Security / data-loss: no new destructive command outside the fixture's own `%TEMP%` scratch; no new interpolation into a shell command that bypasses `PsQuote`; no ledger or store write added.

Current-base integration: the branch is behind `main` (its `.github/workflows/ci.yml` predates the GH-896 fleet-test jobs). CI ran on the merge ref and is fully green, so integration itself is clean; the staleness only matters for the follow-up below.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 8 changed files, all inside #748's *Suspected/Predicted surface* (`lane-launch.ps1`, `lane-status.ps1`, log preservation) and #694's (`lane-launch.ps1` session id); `cmd_dispatch.rs` is the upstream cause the PR argues for, `cli.md` + `file-length-ceilings.txt` are its consequences |
| U2 | any | P1 | **FAIL** | `Closes #748. Closes #694.` (body line 1; commit body lines 59–60). #694 is fully delivered; #748 is not — see F4 |
| U3 | any | P1 | **FAIL** | `gh pr view 972 --json body --jq .body \| awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → empty. See F5 |
| U4 | any | P1 | PASS | 3 subjects, all conventional (`docs(edda-cli):`, `fix(edda-cli):`, `fix(fleet):`); enumerator printed nothing |
| U6 | any | P0 | PASS | READ `gh pr checks 972` + `gh run view 34016410311`: all 12 jobs `success` at this head, `Test (windows-latest)` included (pending at dispatch, since finished green). The three reds at `b78bb6c` — file-length ratchet, undocumented `dispatch --verbose`, stale-base `sdk-contract` — are all gone at this head. `edda --version` → `edda 0.5.0 (403f2c3c8bf2 2026-09-05)`; base `Cargo.toml` `version = "0.5.0"` |
| U7 | any | P1 | PASS | Whole scoped audit finished before this verdict; all 5 blockers batched here |
| D1 | docs | P1 | PASS | Enumerator over added lines yields one command, `edda dispatch`; `edda dispatch --help` → `exit=0` on the installed runtime. Source claim for the documented flag: `crates/edda-cli/src/cmd_dispatch.rs:133-136` at this head carries `#[arg(long)] pub verbose: bool`. CI's `Check CLI docs coverage (GH-650)` is green, so the `cli.md` row and the clap flag agree |
| D2 | docs | P1 | PASS | Run from the repo checkout (ledger reachable): `edda ask fleet.lane-finish-ordering` → `push-and-respond-before-L1`, active; `verification.cost-discipline` → `fixed-lane-pool-one-verify-per-sha`, active; `fleet.lane-launch` → `task-scheduler`, active. The diff asserts no decision status |
| D3 | docs | P1 | PASS | Path/link enumerator over added lines printed no rows (no `MISSING`) |
| D4 | docs | P0 | PASS | Enumerator empty — no added line instructs merge, force-push, branch deletion, `--admin` or `--no-verify` |
| D5 | docs | `[判斷]` | **需升級 (provisional)** | Adjudicated below; its one substantive item is already counted in F1 |
| C1 | code-plain | P0 | PASS | Every doneWhen item this diff actually delivers is matched to code + test: #694 items 1–5 (see F4), #748 item 2 (`lane-status.ps1:166-176` + fixture assertions 11–13) and item 3 (`lane-launch.ps1:244,369,463` + the fixture's killed-lane log assertion). The two #748 items the closing keyword over-claims are routed once, through U2/F4, rather than counted twice here |
| C2 | code-plain | P0 | PASS | Two `fix:` commits; 2 new `#[test]`s (`cmd_dispatch_verbose_tests.rs`) plus a 13-assertion PowerShell fixture. Verified-FAIL reported at `b96a801` (9/13 failing, including the verbatim `state=Ready lastTaskResult=267014 logBytes=0 done=False` line). RAN at this head: 13/13 PASS |
| C3 | code-plain | P1 | PASS | `git diff … -- 'crates/**' \| grep -nE '^\+.*\.(unwrap\|expect)\('` → empty |
| C4 | code-plain | P1 | PASS | Enumerator empty. The only `#[allow]` near the change, `#[allow(clippy::too_many_lines)]` at `cmd_dispatch.rs:446`, is pre-existing and targeted |
| C5 | code-plain | P1 | N.A.(selector empty) | The selector printed nothing: the only touched crate dir is `crates/edda-cli`, which is the `edda` entry of CI's Windows subset, and `Test (windows-latest)` is green at this head. No focused local Cargo run is owed and none was spent |
| R1 | code-risk | P0 | PASS | One hit: `test-lane-delivery.ps1` `finally` → `Remove-Item -LiteralPath $scratch -Recurse -Force -ErrorAction SilentlyContinue`. Trigger, stated: unconditionally on fixture exit, where `$scratch = Join-Path $env:TEMP "gh748-delivery-$PID"` — a PID-scoped directory the fixture itself created. `Join-Path` rejects an empty `$env:TEMP` before anything is created, so the target can never widen to a parent |
| R2 | code-risk | P0/P1 | PASS | `git diff --unified=0 -- '*.sh' '*.bash' '*.ps1' '.github' \| grep -E '\|\|.*&&\|&&.*\|\|'` → empty; re-checked the non-shell added lines by eye, none mixes the operators. No parse tree to draw |
| R3 | code-risk | P0 | N.A.(no `*.sh`/`*.bash` in the diff) | changed-file list contains only `.rs`, `.md`, `.txt`, `.ps1` |
| R4 | code-risk | P1 | PASS | `test-lane-delivery.ps1:14` sets `$ErrorActionPreference='Stop'` and wraps the whole body in `try { … } finally { Remove-Item … }`; the `%PATH%` mutation around the kill test saves `$oldPath` and restores it in its own inner `finally`. No temp directory, file or env mutation lacks a release path on the error branch |
| R5 | code-risk | P1 | PASS | Properties confirmed: (a) two distinct lanes cannot encode to one session id — `-Name` is constrained to `^[A-Za-z0-9][A-Za-z0-9._-]*$` (`lane-launch.ps1:135`) and the id is a SHA-1 over the distinct literal `edda-lane-<name>`; (b) every value interpolated into the generated wrapper goes through `PsQuote` (`:109-111`), `$SessionId` included; (c) the session-id shape check is anchored `^…$` (`:214`) |
| §5.5 | any | — | **FAIL** | see Wiring — two of the new flag's three consumer paths read nothing |

### Wiring

|---|---|---|---|---|
| `--verbose` / `DispatchArgs::verbose` (CLI flag + pub field) | `crates/edda-cli/src/cmd_dispatch.rs:133-136` — `#[arg(long)] pub verbose: bool` | legacy path: `cmd_dispatch.rs:599` → `LauncherOptions.verbose` → `agent_kind.rs:248,255,263` `with_verbose` → `stream.rs:152` `print_live` (claude), `pi_rpc.rs:522` (pi). **ACP path: no reader** (`cmd_dispatch_acp.rs` — only the test fixture at `:282`). **Detached path: no reader** (`detached_dispatch.rs:97-138`) | Refused with `--json` at `cmd_dispatch.rs:475-477` (`bail!` → stderr + non-zero; it dominates all four JSON emitters — `:545`, `:572`, `:650`, and the ACP `println!` reached via `:563`). **Silently dropped** on the ACP and `--detach` paths — no refusal, no warning → **F2, F3** | flag→builder→launcher holds for pi/claude/codex by construction; the spawn-line end is asserted by `test-lane-delivery.ps1` (`$wrapperText -match 'dispatch --verbose'`). No test asserts the `args.verbose → LauncherOptions.verbose` hop despite the test being named for it (P2 below) |
| `LaneSessionUuid` (new function) + the `-Agent claude` UUID gate | `scripts/fleet/lane-launch.ps1:117-123` (writer), `:211` (default), `:214-216` (gate) | the `--session-id` token on all three `argLine` builds (`:244`, `:369`, `:463`) → `edda dispatch` → claude backend | `Fail` → stderr + `exit 1`. Fires before the wrapper is written and before `Register-ScheduledTask`; asserted by fixture assertion 3 | launcher→wrapper→dispatch argv, proven by the fixture (UUID shape, stability across relaunch, distinctness between lanes) |
| `delivery=` field + `TaskResultName` (new status surface) | `scripts/fleet/lane-status.ps1:35-44`, `:166-176` | the controller reading `lane-status` output; no script parses it (`lane-reap.ps1` / `lane-stop.ps1` only reference the tool by name) | **Broken for a whole input class** — `delivery=` empty and two `InvalidArgument` errors on stderr while the script still exits 0 → **F1** | one layer (script → stdout); the fixture pins only the `267014` case, which is the one value its mock happens to type as `Int32` |
| `scripts/fleet/test-lane-delivery.ps1` (new file) | the fixture itself, 13 assertions | **no machine gate** — `run-fleet-tests.sh` globs `scripts/fleet/test-*.sh`; `fleet-tests-windows` runs only `scripts/test-review-capabilities.sh` | none: it can rot green-by-absence | routed as FOLLOW-UP, not blocking — three `test-*.ps1` files already share this gap |
| `cli.md` `--verbose` row; `file-length-ceilings.txt` `2106 → 2116` | `docs/reference/cli.md:705`; `scripts/file-length-ceilings.txt:17` | CI `Check CLI docs coverage (GH-650)`; `scripts/lint-file-length.sh` | both green at this head | the ceiling raise is explicitly "an ordinary reviewable diff to the data file" (`lint-file-length.sh:10-11`); +10 lines is the minimum the flag needs, and the `include!` test split matches 4 pre-existing uses in this repo |

### Findings

- [P1] **F1 — `lane-status.ps1` throws, and prints an empty `delivery=`, for every scheduler result above `Int32.MaxValue` — including `2147942402`, the value #694 itself records and the new switch names.** `Get-ScheduledTaskInfo`'s `LastTaskResult` is `System.UInt32` (measured read-only against a real task), so `[int]$info.LastTaskResult` at `scripts/fleet/lane-status.ps1:166` and `:171` cannot convert it. The `ERROR_FILE_NOT_FOUND` arm at `:41` is therefore unreachable, the four-state machine documented at `:21-25` emits a fifth undocumented state (empty), and the script still exits 0 so a caller sees success. This is introduced here — the base formatted the raw value with no cast. Evidence, RAN offline with the scheduler cmdlets mocked exactly as the fixture mocks them, the only change being that the mock returns `[uint32]2147942402` instead of an untyped `267014`:

  ```text
  InvalidArgument: ...\lane-status.ps1:166
   166 |   $resultName = TaskResultName ([int]$info.LastTaskResult)
       | Cannot convert value "2147942402" to type "System.Int32". Error: "Value was either too large or too small for an Int32."
  InvalidArgument: ...\lane-status.ps1:171
   171 |     elseif ([int]$info.LastTaskResult -eq 267011) { 'not-started' }
  edda-lane-gh694-notfound state=Ready lastTaskResult=2147942402 logBytes=0 done=False delivery= head=- ...
  lane-status exit=0
  ```

  The fixture does not catch it because its mock writes `LastTaskResult = 267014`, an untyped literal PowerShell types as `Int32` (`test-lane-delivery.ps1:171`). `0xC000013A` (3221225786, a task stopped by the user) has the same shape. Fix it and add a case whose mock returns `[uint32]2147942402`.

- [P1] **F2 — `--verbose` is accepted and silently does nothing on the ACP path.** `cmd_dispatch.rs:563-564` routes ACP agents to `cmd_dispatch_acp::run` before any launcher is built; `cmd_dispatch_acp.rs:116-134` `validate_args` exists precisely to refuse "legacy backend options that it cannot enforce" and names eight of them (`--budget-usd`, `--permission-mode`, `--model`, `--thinking`, `--tools`, `--exclude-tools`, `--session-dir`, `--list-models`) — `verbose` is not among them, and the identifier appears nowhere else in that file. This contradicts the honesty gate stated in the very function the diff edits, `cmd_dispatch.rs:452-458`: *"refuse unsupported backend/option combinations instead of accepting them and silently doing nothing."* `cmd_dispatch_acp.rs` is in this diff, so the omission is not adjacent.

- [P1] **F3 — `--verbose` is not forwarded to a detached worker, so `--detach` still leaves the zero-byte log this PR exists to remove.** `detached_dispatch.rs:97-138` `foreground_argv` builds the child argv explicitly, forwarding `--resume` (`:131`) and `--json` (`:134`) among others; `--verbose` is absent, so the child dispatch runs with `verbose:false`. `--detach` is the other unattended-lane shape and the one `--build-lane` requires (`docs/reference/cli.md`), and the parent accepts the flag without a word. Either forward it or refuse the combination.

- [P1] **F4 — U2: `Closes #748` over-claims; two of its four doneWhen items are not delivered.** Delivered: item 2 (`lane-status` shows a recognisable non-delivery state) and item 3 (a killed lane leaves a non-empty log). Not delivered:
  - *Item 1* — "一條在 L1 中途被殺的 lane，其 PR 上仍有 `Review Response` 與正確的 head SHA", with an explicit verified-FAIL / verified-PASS pair. No file in this diff changes lane ordering or a brief template; the mechanism #748 predicts for it is `fleet.lane-finish-ordering = push-and-respond-before-L1`, which already sits in the ledger and in `.claude/CLAUDE.md`'s ladder from earlier work, and no such kill experiment is reported in this PR.
  - *Item 4* — "量測並記錄本機 L1 的實際 wall-clock". `edda ask` finds no such record and `grep -rn 'wall-clock' docs/ .claude/CLAUDE.md` finds only the GH-810 debug-symbol measurement. The PR body cites #748's `範圍外` as cover, but that note excludes exactly one thing — "不改 `verification.cost-discipline` 的四格池本身" — not the measurement.

  Merging as written auto-closes #748 with those two items open. Remedy: `Issue: #748` in place of `Closes #748`. **`Closes #694` is correct** — item 1 (resolved `pwsh` path) holds at `lane-launch.ps1:240-241`, item 2 by `:211` and `:214-216`, item 4 by the gate running before the `-DryRun` branch (the fixture proves `-DryRun` now catches defect two), item 5 by `review-pr.sh:261-262,662-665` already resolving `pwsh` and `:482-509` already deriving a UUID.

- [P1] **F5 — U3: the PR body carries no `Issue: #N` line.** `gh pr view 972 --json body --jq .body | awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` prints nothing. The closing keywords do supply the brief with both doneWhen sets, so nothing downstream is starved — this is the convention miss REVIEW.md U3 rates P1, and F4's remedy line resolves it at the same time.

- [P2] The PR body cites `scripts/review-pr.sh:456` for `review_session_uuid`; at this head that line is the REVIEW.md classifier invocation and the function is at `scripts/review-pr.sh:482-509`. The substance holds — both derive a name-based UUID with version nibble `5` and variant `10xx`, so the in-code comment at `lane-launch.ps1:113-116` ("the same shape") is accurate.

- [P2] The UUID gate at `lane-launch.ps1:214` is lowercase-only (`[0-9a-f]`). A caller resuming with an upper- or mixed-case UUID — valid per RFC 4122 and accepted by the backend — is refused. It fails safe and the message names the remedy, but `(?i)` would remove a false refusal on the one path the gate exists to keep working.

- [P2] "Refused **before anything is registered**" is literally true, and fixture assertion 3 pins it. Two side effects do precede it, though: `git-config-guard.ps1 -Backup` at `lane-launch.ps1:195` (writes and rotates `config.guard.bak` / `.prev` in the git common dir) and `New-Item -Path $LogDir` at `:204`. Both are non-destructive, but the gate would sit more naturally with the other guard rails at `:133-169`, which is also where a reader looks for it.

- [P2] `cmd_dispatch_verbose_tests.rs:2-18` is named `verbose_streams_live_activity_into_the_launcher` but asserts only that clap parses the flag and that the default is false; it never touches `LauncherOptions` or the launcher. The name claims the hop the test does not cover.

- [P2] The PR body says `print_live` (`stream.rs:207`) "is the only thing that puts agent activity on stdout while a turn runs". It is the claude path only. The pi path — `lane-launch.ps1`'s **default** `-Agent` — has its own gate at `pi_rpc.rs:522-527` (`Some("tool_execution_start") if verbose => println!(...)`). The conclusion is unaffected (see below); the attribution is incomplete.

### Root-cause claim — independent finding

**The re-diagnosis holds in substance.** `LauncherOptions.verbose` is the only route to live output, and the base hard-coded it false at `cmd_dispatch.rs:599`; from there it is the sole gate on both live streams — `stream.rs:152` (`if self.verbose { print_live(&msg) }`, claude) and `pi_rpc.rs:522` (`Some("tool_execution_start") if verbose`, pi) — reached through `agent_kind.rs:248/255/263`. With it false, nothing reaches stdout until the turn's final `render_text()`, so a lane killed at its execution limit had genuinely printed nothing for the wrapper to tee. `print_live` uses `println!`, and Rust's stdout is a `LineWriter`, so each line does leave the process at the newline rather than sitting in a block buffer.

**The `Tee-Object` theory is narrowed, not disproven.** The fixture's kill test emits one short line six seconds before the kill and finds it on disk, which is enough to rule `Tee-Object` out as *the* cause of the 2026-09-03 `logBytes=0`. It says nothing about a large buffered volume, and the PR body states the stronger claim. Not blocking; worth not carrying forward as settled.

### FOLLOW-UP ISSUE

- `scripts/fleet/test-lane-delivery.ps1` runs in no machine gate. `run-fleet-tests.sh` is glob-driven over `scripts/fleet/test-*.sh`, and `fleet-tests-windows` runs exactly one test, `scripts/test-review-capabilities.sh` (`.github/workflows/ci.yml` on `origin/main`, jobs `fleet-tests` / `fleet-tests-windows`). Three `test-*.ps1` files already share the gap — `test-detached-dispatch.ps1`, `test-lane-reap.ps1`, `test-lane-terminal-receipts.ps1` — so this is pre-existing and routed rather than blocking. It is the same hole GH-896 closed for `.sh`: a `test-*.ps1` runner on the existing `windows-latest` job, behind the same `detect.fleet` path filter, would close it for all four at once.

### D5 adjudication (`[判斷]`, provisional)

Wording and structure of the added prose are clear, correctly scoped and non-duplicative: the `cli.md` row, the three PowerShell comment blocks and the fixture header each state why the change exists and cite the issue, without restating a rule owned by `.claude/CLAUDE.md` or REVIEW.md. One substantive item: the `lane-status.ps1:21-25` header enumerates four `delivery` states as exhaustive, and the code emits a fifth. That is the documentation half of F1 and is counted there, not separately. Marked `provisional` per §6.1 — this round is a checklist-type engine for `code-risk` and the brief does not name it qualified.

### RAN vs READ

RAN (all read-only or offline; no Scheduled Task registered, started, modified or unregistered, no agent dispatched, no label, status, push or merge):

- `gh pr view 972 --json …`, `gh issue view 748` / `694`, `gh pr diff 972 --name-only`, `gh run view 34016410311 --json jobs`, `gh api …/jobs`
- REVIEW.md §3 classifier on the changed-file list → `classes=code-risk code-plain docs`, `canonical_class=code-risk`
- U2/U3/U4 enumerators; `sh scripts/lint-markdown-content.sh` → exit 0; the D1/D3/D4, C2/C3/C4/C5 and R1/R2/R3 enumerators; `sh scripts/wiring-scan.sh origin/main a07e881`
- `edda --version`; `edda dispatch --help` → exit 0; `edda ask fleet.lane-finish-ordering` / `verification.cost-discipline` / `fleet.lane-launch`, run from the repo checkout so the ledger is reachable
- `pwsh -File scripts/fleet/test-lane-delivery.ps1` → **13/13 PASS**, exit 0. Its offline claim was verified before running: the driver scripts shadow `Register-` / `Start-` / `Unregister-ScheduledTask` and `New-ScheduledTaskAction` / `SettingsSet` with local functions, the agent is a stub `edda.cmd` prepended to `PATH`, and the repository is a throwaway `git init` under `%TEMP%`. Confirmed after the run, read-only: no `edda-lane-gh748-*` or `edda-lane-gh694-*` task exists (the one pre-existing `edda-lane-gh694` is Disabled, last run 2026-09-03).
- an offline `lane-status.ps1` reproduction with `[uint32]2147942402` (F1 evidence), same mocking shape as the fixture
- `Get-ScheduledTaskInfo` on one arbitrary existing task, read-only, to establish that `LastTaskResult` is `System.UInt32`

READ (not re-run):

- L1 receipt: **CI run 34016410311 @ a07e8816e5e15aacd6798fc3cd3e7f10d2069aad** — Format; Clippy on ubuntu/macOS/windows; Test on ubuntu + macOS (full workspace) and windows (7-crate subset); MSRV 1.91.0; Detect code changes; Fleet shell tests (×2); CI Gate. All 12 `success`. Plus `Workspace publish dry-run` (run 34016410211) and `sdk-contract` (run 34016410177), both `success`.
- That set covers `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -D warnings` on all three OSes, and `cargo test --workspace` on Linux and macOS. What it does **not** cover is the runtime behaviour of the 16 crates outside the Windows subset — irrelevant here, since the C5 selector is empty (`crates/edda-cli` is the subset's `edda` entry) — and it runs **no** PowerShell fixture, which is why `test-lane-delivery.ps1` was run locally rather than READ.
- No workspace Cargo gate was run locally, and none was owed.

### Verdict

Changes Requested, P0=0, P1=5 — `lane-status` throws and reports an empty `delivery=` for every scheduler result above `Int32.MaxValue` (F1); `--verbose` is accepted and silently dropped on both the ACP and `--detach` paths (F2, F3); and `Closes #748` would auto-close two undelivered doneWhen items while the body carries no `Issue: #N` line (F4, F5).

fagemx:
## Review Response: Round 1

New head: `7efbd6f843ed2bd2a2e6239392db3c15e46c06db`

All five P1s addressed. The three code findings each carry a regression case proved to fail without its fix.

### F1 — `lane-status.ps1` `[int]` cast on a `System.UInt32` — fixed

The finding was exact, and the reason it matters is worse than a crash. Measured before touching anything:

```text
THROWS: Cannot convert value "2147942402" to type "System.Int32".
        Error: "Value was either too large or too small for an Int32."
COMPARE THROWS: (same, on the 267011 comparison)
```

Both the `TaskResultName` call and the `-eq 267011` comparison threw. So `delivery=` printed empty — a fifth, undocumented state — the `ERROR_FILE_NOT_FOUND` arm was unreachable, and the script still exited 0.

**`2147942402` is 0x80070002.** It is the code GH-694 records, and it is what a relative `-File` argument produces — the exact failure `scripts/fleet/digest-schedule.ps1` was carrying until an hour ago. The one result code that most needs reporting was the one that broke the reporter.

`[long]` at all three sites. After:

```text
2147942402  -> name=[ERROR_FILE_NOT_FOUND] delivery=UNDELIVERED(ERROR_FILE_NOT_FOUND)
267014      -> name=[SCHED_S_TASK_TERMINATED] delivery=UNDELIVERED(SCHED_S_TASK_TERMINATED)
```

Your diagnosis of why the fixture missed it was right and is now the fix: the existing mock writes an untyped `267014`, which PowerShell types as `Int32`. A second driver returns a real `[uint32]2147942402`. Verified FAIL with `[int]` restored (2 failures), PASS with `[long]` — 15/15 overall.

### F3 — `--verbose` not forwarded through the detached supervisor — fixed

This is the finding that matters most for the issue. GH-748's `logBytes=0` was reported on **lanes**, and lanes dispatch through `--detach`. `foreground_argv` forwarded `--resume` and `--json` but not `--verbose`, so the PR fixed every path *except the one where the bug was observed*. `detached_worker_inherits_verbose` asserts both directions; verified FAIL without the forward.

### F2 — `--verbose` accepted and dropped on the ACP path — fixed

Refused rather than implemented, per the GH-574 honesty gate you cited: ACP has no live activity stream, and a caller who passes `--verbose` and gets an empty log would read it as "the agent produced nothing". Added to the existing `acp_dispatch_refuses_substitutes_and_unenforceable_options` case; verified FAIL without the guard.

### F4 — `Closes #748` — you were right, removed

I re-read #748's `## Done when` against the tree rather than defending the keyword. Item 1 (a lane killed mid-L1 still carries a `Review Response`, with a verified-FAIL/PASS pair) is untouched by this diff, and item 4 (measure and record local L1 wall-clock) has no record — `edda ask` has none and the only measurement in the tree is GH-810's debug-symbol one. Your reading of the `範圍外` note is also right: it excludes the four-lane pool, not the measurement.

Body now carries `Issue: #748` and `Closes #694`, with the two undelivered items named. #694's five items hold, so that keyword stays.

### F5 — U3 — fixed

`Issue: #748` is the first line.

### On the root-cause claim

Your correction is accepted and is a real narrowing: `print_live` is claude-specific, and pi — the launcher's default agent — has its own `if verbose` gate at `pi_rpc.rs:522-527`. Equally fair that the `Tee-Object` theory is *narrowed* rather than disproven: one short line six seconds before the kill rules it out as the operative cause here, not as a buffering hazard at volume. The body overstated "the only thing"; the mechanism it names is still the one that produced the observed zero bytes.

### Gates (L0, this worktree, Windows)

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | 0 |
| `cargo clippy -p edda --all-targets -- -D warnings` | 0 |
| `cargo test -p edda` | 547 passed, **2 failed** — see below |
| `pwsh -File scripts/fleet/test-lane-delivery.ps1` | 15/15 PASS |
| `Parser::ParseFile` on both changed `.ps1` | parses |

The two failures are `cmd_review::evidence::tests::deadline_stops_descendant_and_does_not_start_next_gate` and `cmd_reconcile::tests::runner::simultaneous_reconciles_create_one_attempt_and_release_the_lock`. Neither touches this diff's surface; both pass on rerun (twice, `--test-threads=1`), and both are green in CI at `a07e881`. This workstation was carrying eight concurrent review agents and a full workspace compile at the time. I am reporting them rather than filtering them out, and I am not claiming a cause beyond the load.

C5 selector empty — only `crates/edda-cli` is touched, which is the subset's `edda` entry.

fagemx:
## Code Review: Round 2 — PR #972 @ 7efbd6f843ed2bd2a2e6239392db3c15e46c06db

- elapsed: unmeasured (no pi session file; this round did not run through `pi`)
- model_observed: unverified (no session file / `modelUsage` key available to this round)
- reviewer_session: n/a (round not launched through `lane-launch.ps1` / `review-pr.sh`)
- spec: review-spec-v1.5
- class: code-risk  (REVIEW.md classes, full PR: code-risk code-plain docs; delta list: code-risk code-plain)
- escalations: D5 (adjudicated below, marked `provisional`)
- shadow: false
- cost: 1 fixture run (~35 s), 1 offline verified-FAIL/PASS pair for the uint32 fix, 1 PowerShell shadowing probe, ~35 read-only `gh`/`git`/`grep` calls. No Cargo gate run (C5 selector empty). One over-verification to declare: I started `sh scripts/lint-file-length.sh --tree` locally before confirming CI's `Format` job already runs it (`ci.yml:119`) — redundant, and READ from CI instead.

### IN SCOPE

Delta round over Round 1's `a07e8816e5e15aacd6798fc3cd3e7f10d2069aad`. Round 1's audit of the unchanged surface is cited, not repeated (`loop` items 2 and 6).

Delta surface (`git diff a07e881..7efbd6f --name-only`, 4 files, +106/−3): `crates/edda-cli/src/cmd_dispatch_acp.rs`, `crates/edda-cli/src/detached_dispatch.rs`, `scripts/fleet/lane-status.ps1`, `scripts/fleet/test-lane-delivery.ps1`. Plus the PR body edit (`Closes #748` → `Issue: #748`).

Changed behaviour: `TaskResultName` and its two call sites widened `[int]` → `[long]`; `--verbose` forwarded onto the detached worker argv; `--verbose` refused on the ACP path; one new fixture driver returning a real `[uint32]`.

Direct callers/consumers audited: `foreground_argv` ← `detached_dispatch::launch:54` ← `cmd_dispatch.rs:571`; `validate_args` ← `cmd_dispatch_acp::run:45` and `preflight:25` ← `cmd_dispatch.rs:537` (before the claim-guard write at `:544`); `TaskResultName` / `$delivery` ← `lane-status.ps1:171-178` → its stdout line; the fixture's `$env:GH748_STATUS` indirection.

Issue acceptance: #748 doneWhen (4 items) and #694 doneWhen (5 items); ruling on both keywords below.

Security / data-loss: no new destructive command; the delta's only `Remove-Item` is the pre-existing fixture `finally`. No new interpolation, no ledger or store write.

Current-base integration: CI at this exact head is fully green on the merge ref, including `Test (windows-latest)`.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 4 delta files (9 for the PR), all inside #748's *Suspected/Predicted surface* and the upstream cause the PR argues for. Note: §5's `git diff "origin/$BASE..$SHA"` enumerators are **two-dot** and inflate this branch's surface to 49 files; three-dot gives the true 9. Spec defect, tracked as **#1003** — used three-dot throughout for `git diff`, two-dot for `git log` as §5 has it |
| U2 | any | P1 | **FAIL** | Body arm now clean (`Closes #694.` only). Commit arm is **not**: `git log --format=%B origin/main..7efbd6f \| grep -Ein '(close[sd]?\|fix(e[sd])?\|resolve[sd]?)[[:space:]]+#[0-9]+'` → `91:Closes #748`, `92:Closes #694`, both in `b78bb6c`'s message (lines 33–34). See G1 |
| U3 | any | P1 | PASS | `gh pr view 972 --json body --jq .body \| awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → `Issue: #748`. Round 1's F5 resolved |
| U4 | any | P1 | PASS | 4 subjects over `origin/main..7efbd6f`, all conventional; enumerator printed nothing |
| U5 | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → `exit=0` (RAN) |
| U6 | any | P0 | PASS | READ `gh pr checks 972` + `gh run view 34019692072 --json headSha,jobs`: `headSha=7efbd6f843ed2bd2a2e6239392db3c15e46c06db`, all 12 jobs `success` — `Test (windows-latest)` was `pending` when this round opened and finished green (7m49s). `edda --version` → `edda 0.5.0 (403f2c3c8bf2 2026-09-05)`; base `586cb07e`'s `Cargo.toml` → `version = "0.5.0"` |
| U7 | any | P1 | PASS | Whole delta audit finished before this verdict; both blockers batched here |
| D1 | docs | P1 | PASS | Enumerator over the delta's added lines yields no backticked command. **But** the delta adds two prose claims about repository shape that are false — routed to G2 under the §5 P1 definition ("a false claim"), since D1's enumerator only sees backticked commands |
| D2 | docs | P1 | PASS | Delta asserts no decision key (enumerator over added lines → none). Cross-checked from the repo checkout so the ledger is reachable: `edda ask fleet.lane-launch` → `task-scheduler`, active — which is the evidence under G2 |
| D3 | docs | P1 | PASS | Path/link enumerator over the delta printed no rows |
| D4 | docs | P0 | PASS | Enumerator over the delta empty |
| D5 | docs | `[判斷]` | **需升級 (provisional)** | Adjudicated below; its substantive item is counted in G2, not separately |
| C1 | code-plain | P0 | PASS | #694's five items hold at head (see the keyword ruling); #748's two delivered items unchanged from Round 1. The two undelivered #748 items are routed once, through U2/G1 |
| C2 | code-plain | P0 | PASS | `fix:` commit with a regression case per finding: `detached_worker_inherits_verbose` (`detached_dispatch.rs:96-114`), the `--verbose` leg of `acp_dispatch_refuses_substitutes_and_unenforceable_options` (`cmd_dispatch_acp.rs:304-310`), and fixture assertions 14–15 (`test-lane-delivery.ps1:207-208`). All three shown below to fail without their fix |
| C3 | code-plain | P1 | PASS | `git diff a07e881..7efbd6f -- 'crates/**' \| grep -nE '^\+.*\.(unwrap\|expect)\('` → empty |
| C4 | code-plain | P1 | PASS | Enumerator over the delta empty |
| C5 | code-plain | P1 | N.A.(selector empty) | `git diff --name-only origin/main...7efbd6f \| grep '^crates/' \| cut -d/ -f2 \| sort -u \| grep -Ev '^(edda-store\|…\|edda-cli)$'` → nothing. `crates/edda-cli` is CI's Windows-subset `edda` entry and `Test (windows-latest)` is green at this head, so no focused local Cargo run is owed and a workspace gate would be over-verification. None was spent |
| R1 | code-risk | P0 | PASS | Enumerator over the delta empty. The pre-existing `finally` → `Remove-Item -LiteralPath $scratch -Recurse -Force` is unchanged and its trigger is Round 1's (PID-scoped `%TEMP%` dir the fixture itself created); confirmed released — no `gh748-delivery-*` directory survived my run |
| R2 | code-risk | P0/P1 | PASS | `git diff --unified=0 -- '*.sh' '*.bash' '*.ps1' '.github' \| grep -E '\|\|.*&&\|&&.*\|\|'` → empty. Re-read the added non-shell lines by eye; none mixes the operators. No parse tree to draw |
| R3 | code-risk | P0 | N.A.(no `*.sh`/`*.bash` in the delta) | Instead, RAN the PowerShell equivalent on both changed `.ps1`: `[Parser]::ParseFile` → 0 errors each |
| R4 | code-risk | P1 | PASS | The delta adds no temp directory, file, lock or env mutation. The new `$notFoundDriver` and `$notFoundLog` live under the existing `$scratch` that the `finally` releases; `$env:GH748_NOTFOUNDLOG` follows the fixture's existing env convention and dies with the process |
| R5 | code-risk | P1 | PASS | Properties re-confirmed for the delta: the widened cast changes no filename encoding; nothing new is interpolated into a shell command; the new driver's here-string is single-quoted (`@'…'@`), so `$env:GH748_STATUS` is resolved by the child, not spliced by the parent |
| §5.5 | any | — | PASS | See Wiring — all four delta surfaces have a reader and a failure signal |

### Wiring

|---|---|---|---|---|
| `--verbose` token on the detached worker argv | `crates/edda-cli/src/detached_dispatch.rs:141-143` — `if args.verbose { out.push("--verbose") }` | the spawned child `edda dispatch` → `cmd_dispatch.rs:599` `LauncherOptions.verbose` → `agent_kind.rs:248,255,263` → `stream.rs:152` / `pi_rpc.rs:522` | absence of the token is the silent-drop class the fix removes; `detached_worker_inherits_verbose` (`:96-114`) asserts **both** directions, so the negative half cannot pass vacuously. `--verbose`+`--json` cannot reach here at all: refused at `cmd_dispatch.rs:475` before the detach branch at `:570` | flag→supervisor argv→child dispatch→launcher. Round 1's F3 gap closed |
| ACP `--verbose` refusal | `crates/edda-cli/src/cmd_dispatch_acp.rs:128-132` — `bail!` | `validate_args` ← `run:45` (first statement) and `preflight:25` (first statement) | `bail!` → stderr + non-zero. Ordering verified: `run_inner:537` runs the ACP preflight **before** the claim-guard GitHub write at `:544`, so no `taking:` claim is left behind; test at `:304-310` | option→validate→refusal. The list is now complete: every other `DispatchArgs` field is either honored on the ACP path (`owns`, `cwd`, `task_id`, `timeout_sec`→`prompt_timeout:79`, `json:106`, `issue`/`machine` via `cmd_dispatch.rs:544`) or structurally unreachable (`build_lane`/`detach_log_dir` carry clap `requires = "detach"`, and `detach` bails at `:118`) |
| `TaskResultName([long]$Code)` + two `[long]` call-site casts | `scripts/fleet/lane-status.ps1:40`, `:171`, `:176` | the `lastTaskResult=…` / `delivery=…` line the controller reads (`:180-181`) | before: two `InvalidArgument` throws, empty `delivery=`, exit 0. After: named result + `UNDELIVERED(ERROR_FILE_NOT_FOUND)`. Pinned by fixture assertions 14–15 | one layer (script → stdout). Exactly three sites; a mechanical diff of head-vs-reverted shows the three and nothing else |
| `status-notfound-driver.ps1` (new fixture driver, real `[uint32]2147942402`) | `scripts/fleet/test-lane-delivery.ps1:193-208` | the fixture itself (15 assertions now, up from 13) | **no machine gate** — unchanged pre-existing gap, already tracked by **#927**; routed as FOLLOW-UP, not blocking | routed as FOLLOW-UP per §5.5's fixed adjudication ("with a follow-up issue number → pass") |

### Findings

- [P1] **G1 — `Closes #748` still stands in the branch's commit message, and every merge strategy this repo allows carries it to `main`, so merging still auto-closes an issue the response itself says is not delivered.** The PR body was fixed; the commit was not. `git log -1 --format=%B b78bb6c561c822ff2b8d7f692b03c70f063b5a90` carries `Closes #748` / `Closes #694` at lines 33–34, which is why REVIEW.md U2's enumerator is two-armed (body **and** `git log --format=%B`) and why Round 1's F4 cited "body line 1; **commit body lines 59–60**". Only the body half was answered. This is not theoretical:

  ```text
  $ gh api repos/fagemx/edda --jq '{squash_commit_message:.squash_merge_commit_message,
                                    merge:.allow_merge_commit, rebase:.allow_rebase_merge}'
  {"squash_commit_message":"COMMIT_MESSAGES","merge":true,"rebase":true}
  ```

  With `squash_merge_commit_message: COMMIT_MESSAGES`, a squash merge composes its body from the commit messages, not the PR body; merge-commit and rebase preserve them verbatim. So under all three permitted strategies `Closes #748` reaches the default branch and GitHub closes #748 — with, by the response's own account, item 1 (a lane killed mid-L1 still carries a `Review Response`, with a verified-FAIL/PASS pair) and item 4 (measure and record local L1 wall-clock) undelivered. #748 is `OPEN`. The keyword has to leave the commit message, not just the body.

- [P1] **G2 — the reason recorded beside the `--verbose` forward is false: no lane in this repository dispatches through `--detach`.** The forward itself is correct and I am not asking for it back. Two added comments assert the opposite of the tree:

  - `crates/edda-cli/src/detached_dispatch.rs:139-140` — *"the detached path is the lane path, which is where logBytes=0 was reported"*
  - `crates/edda-cli/src/detached_dispatch.rs:97-99` — *"the zero-byte log was reported on LANES, which dispatch through the detached supervisor"*

  The lane path is the Task Scheduler wrapper, and its dispatch line has no `--detach` token. `scripts/fleet/lane-launch.ps1:244`, `:369` and `:463` each build `dispatch --verbose --agent … --prompt-file … --session-id … --cwd … --timeout-sec …`; the fixture prints one verbatim:

  ```text
  edda dispatch --verbose --agent claude --prompt-file '…gh694-default.dryrun-brief.md'
       --session-id 'b0d249c4-0f3e-5663-a10c-4cd658eeaf52' --cwd '…\repo'
       --timeout-sec 1800 --owns 'scripts/fleet/lane-launch.ps1'
  ```

  `scripts/review-pr.sh:766,772,902,909` has the same foreground shape. `grep -rn -- "--detach" scripts/ docs/ .claude/ skills/` returns only `git worktree add --detach`, `docs/reference/cli.md:701-703` documenting the flag, and the fixture `scripts/fleet/test-detached-dispatch.ps1:100`; `grep -rn '"--detach"' crates/ --include=*.rs` adds only `cmd_run.rs:157` and `cmd_review/git.rs:97`, both `git worktree add`. The decision agrees: `edda ask fleet.lane-launch` → `task-scheduler`, active. And #748's own three lanes (`edda-lane-fix717`, `edda-lane-fix736`, `edda-lane-gh651r2`) are `lane-launch.ps1` Scheduled Tasks, per the table in its body.

  So `--detach` is not "the lane path" and is not where `logBytes=0` was observed; it is a second unattended shape that had the same silent-drop defect. Severity is the §5 table lookup — P1, "a false claim" — and it is fix-caused (added in `7efbd6f`), so it is admissible in this round under `loop` item 2. It matters more than a typo would: this PR exists because a wrong causal story about the lane's plumbing sent #748's first diagnosis at `Tee-Object`, and this comment writes a second wrong one into the source. Remedy is two comment lines: the forward closes the *other* unattended path, and the observed `logBytes=0` came from the foreground lane dispatch that `cmd_dispatch.rs:599` now streams.

- [P2] Four `[int]` casts of CIM `uint32` fields survive in `lane-status.ps1` — `:75`, `:76` (`Win32_Process.ParentProcessId` / `.ProcessId`), `:143`, `:158` — with the same shape in `lane-stop.ps1:156,157,180,197,268` and `test-lane-delivery.ps1:152`. All are pre-existing: the full three-dot diff of `lane-status.ps1` adds no `[int]` line. Unlike an HRESULT, a Windows PID above `Int32.MaxValue` is not a realistic input class, so this is a note rather than a repeat of G1's defect — but the fix comment at `:35-39` now states the rule for `LastTaskResult` only, and a reader may take the file as audited.

- [P2] The two `cargo test -p edda` failures the response reports are load-sensitive, not diff-caused — I checked rather than took it. `cmd_review/evidence/tests.rs:236` asserts `started.elapsed() < Duration::from_secs(4)` for a 1-second deadline over a `sleep 2` gate and then sleeps 1500 ms before asserting two files are absent; `cmd_reconcile/tests/runner.rs:185-196` races two threads through a `Barrier` onto a workspace lock and asserts exactly one wins. Both are wall-clock/contention bound by construction. Neither reaches `foreground_argv`, `validate_args`, `DispatchArgs`, or any file in this diff, and the delta's two new tests add no global state (no env mutation, no filesystem, no process spawn) that could perturb a neighbour — unlike the pre-existing `MANIFEST_ENV_LOCK` cases in the same module. `Test (windows-latest)` is green on `-p edda` at this exact head. Claim accepted.

### Are the four fixes wired to a case that fails without them?

- **`[int]` → `[long]`, three sites — yes, RAN.** I copied `lane-status.ps1` to scratch, reverted exactly the three casts (mechanical diff shows three hunks and nothing else), and drove both copies with the fixture's own not-found driver:

  ```text
  ########## REVERTED ([int] restored) ##########
  InvalidArgument: …\lane-status-reverted.ps1:171
   171 |   $resultName = TaskResultName ([int]$info.LastTaskResult)
       | Cannot convert value "2147942402" to type "System.Int32".
       | Error: "Value was either too large or too small for an Int32."

  ########## FIXED (head, [long]) ##########
  edda-lane-gh694-notfound state=Ready lastTaskResult=2147942402(ERROR_FILE_NOT_FOUND) …
    delivery=UNDELIVERED(ERROR_FILE_NOT_FOUND) …
  lane-status exit=0
  ```

  Both new assertions fail without the fix. Three sites changed, and no `[int]` cast of `LastTaskResult` remains.

- **`--verbose` forwarded on the detached path — yes, by construction.** The forward is unconditional on the flag (`if args.verbose`), and `foreground_argv`'s whole body is `:97-145`: the fixed-flag loop at `:109-127` enumerates ten flags, none of them `--verbose`, so with `:141-143` removed the first assertion of `detached_worker_inherits_verbose` cannot pass. No other dispatch-shaped option is dropped on that path — see the Wiring row; `build_lane` and `detach_log_dir` are consumed by the supervisor itself at `:19-36`.

- **ACP refuses `--verbose` — yes, by construction.** `verbose` is the only non-default field `acp_args()` gets in that leg, so with `:128-132` removed `validate_args` returns `Ok(())` and `.unwrap_err()` at `:309` panics. The refusal fires before any side effect (`run:45`, `preflight:25`, and `run_inner:537` before the claim write at `:544`), and the refusal list is complete.

- **The fourth "fix" — the PR-body keyword — is only half done.** See G1.

### Keyword ruling

- **`Issue: #748` — correct, and required.** I re-read #748's `## Done when` myself. Item 2 (`lane-status` shows a recognisable non-delivery state) and item 3 (a killed lane leaves a non-empty log) are delivered and tested. Item 1 asks for a lane killed mid-L1 that still carries a `Review Response`, *with an explicit verified-FAIL/verified-PASS pair* — no file in this diff changes lane ordering or a brief template, and no such kill experiment is reported. Item 4 asks for a measured, recorded local L1 wall-clock — `edda ask` has none, and the only measurement in the tree is GH-810's debug-symbol one. The `範圍外` note excludes exactly one thing, `verification.cost-discipline` 的四格池, not the measurement. Dropping the keyword was right. **It just did not actually leave the branch — G1.**
- **`Closes #694` — still earned.** Item 1: `lane-launch.ps1:240-241` resolves `pwsh.exe` absolutely and `Fail`s if it cannot. Item 2: `:211` derives the UUID, `:214-216` refuses a non-UUID for `-Agent claude` without rewriting a caller's id. Item 3 (verified-FAIL receipts): both are recorded verbatim in #694's own body — `LastTaskResult=2147942402` and `Error: Invalid session ID. Must be a valid UUID.` Item 4: both defects are now `-DryRun`-visible — the shape gate at `:214-216` runs before the `if ($DryRun)` branch at `:362`, and the dry-run action uses the same resolved `$PwshExe` at `:395`; fixture assertion 3 pins it and I saw it fire. Item 5: `scripts/pr-review-launch.ps1:75-76` and `scripts/review-pr.sh:261-262,662-665` both resolve `pwsh` absolutely, and `review_session_uuid` at `:482` derives the UUID. Round 1's ruling holds and the delta does not disturb it. Note that `Closes #694` in the commit message is the *correct* half of G1's hit and should stay.

### FOLLOW-UP ISSUE

- **#927** already tracks it: `scripts/fleet/test-lane-delivery.ps1` — now 15 assertions and the only proof of three of this PR's four fixes — runs in no machine gate. On `origin/main`, `scripts/fleet/run-fleet-tests.sh:37` globs `scripts/fleet/test-*.sh` only, and `ci.yml`'s `fleet-tests-windows` runs exactly `sh scripts/test-review-capabilities.sh`. Unchanged from Round 1's routing; not blocking.
- The two load-sensitive tests above (`cmd_review/evidence/tests.rs:236`, `cmd_reconcile/tests/runner.rs:185-196`) are pre-existing and outside this PR's surface. A wall-clock budget of 4 s and a two-thread lock race are both machine-dependent; worth an issue of their own rather than a rerun ritual in every `Review Response`.
- **#996** intersects this PR's fixture rather than its product. `lane-launch.ps1 -DryRun` reaches `Register-`/`Start-ScheduledTask` at `:406-407`, and this fixture is safe only because the driver's function shadowing intercepts them. I verified that mechanism before running anything (below); #996's doneWhen already asks for it to be documented as working or not. No change asked of this PR.

### D5 adjudication (`[判斷]`, provisional)

The delta's added prose is otherwise clear, correctly scoped and non-duplicative: `lane-status.ps1:35-39`, `test-lane-delivery.ps1:186-192` and `cmd_dispatch_acp.rs:124-127` each state why the change exists, cite the issue, and restate no rule owned by `.claude/CLAUDE.md` or REVIEW.md. The one substantive item is the false attribution in `detached_dispatch.rs:97-99,137-140`, counted in G2 rather than here. Also worth noting the delta *closes* Round 1's documentation half: the header's four `delivery` states at `:21-25` are once again exhaustive, since the empty fifth state is gone. Marked `provisional` per §6.1 — this round is a checklist-type engine for `code-risk` and the brief does not name it qualified.

### RAN vs READ

RAN (all read-only or offline; **no Scheduled Task registered, started, modified or unregistered**, no agent dispatched, no label, status, push, force-push, rebase, amend, branch/worktree/ref deletion or merge):

- `gh pr view 972 …`, `gh pr checks 972`, `gh run view 34019692072 --json headSha,jobs`, `gh issue view 748 / 694 / 927 / 996`, `gh api repos/fagemx/edda` (merge settings)
- REVIEW.md §3 classifier: full PR list → `classes=code-risk code-plain docs`, `canonical_class=code-risk`; delta list → `classes=code-risk code-plain`
- U2/U3/U4 enumerators; `sh scripts/lint-markdown-content.sh` → exit 0; the D1/D3/D4, C2/C3/C4/C5 and R1/R2 enumerators over `a07e881..7efbd6f`; `sh scripts/wiring-scan.sh a07e881 7efbd6f` → *no new pub surfaces*, one swallow-pattern hit that is a comment line
- `[System.Management.Automation.Language.Parser]::ParseFile` on both changed `.ps1` → 0 errors each (R3's substitute; there is no `*.sh` in the delta)
- `edda --version`; `edda ask fleet.lane-launch` / `fleet.lane-dispatch`, run from the repo checkout so the ledger is reachable
- **Offline claim verified before running the fixture, given #996.** `lane-launch.ps1 -DryRun` really does reach `Register-ScheduledTask` / `Start-ScheduledTask` (`:406-407`), so the fixture is safe only if the drivers' function shadowing intercepts them. I probed the mechanism in isolation, touching nothing schedulable: a parent script defining the fixture's own mock shapes, calling a child with `& <child>.ps1` — the child got `SHADOWED-GET-DATE` and `System.Management.Automation.PSCustomObject` from `New-ScheduledTaskAction` / `New-ScheduledTaskSettingsSet`, i.e. the mocks, not the module cmdlets. Same discovery order (Alias→Function→Cmdlet) governs the three write cmdlets, and no `Import-Module ScheduledTasks` or module-qualified call exists in `lane-launch.ps1`.
- `pwsh -File scripts/fleet/test-lane-delivery.ps1` → **15/15 PASS**, exit 0. Task inventory taken read-only before and after: **172** `edda-*` tasks both times, and the only `gh748|gh694` match both times is the pre-existing `edda-lane-gh694` (Disabled). `%TEMP%\gh748-delivery-*` was released by the `finally`. The header's offline claim holds.
- The F1 verified-FAIL/PASS pair above, against scratch copies outside the worktree — the reviewed tree was never edited.
- `git worktree list` / `git status` confirm the worktree is clean at `7efbd6f`.

READ (not re-run):

- L1 receipt: **CI run 34019692072 @ 7efbd6f843ed2bd2a2e6239392db3c15e46c06db** — `headSha` matches exactly. Detect code changes; Format; Clippy on ubuntu/macOS/Windows; Test on ubuntu + macOS (full workspace) and Windows (7-crate subset, 7m49s); MSRV 1.91.0; Fleet shell tests (×2); CI Gate. All 12 `success`. Plus `Workspace publish dry-run` (run 34019692063) and `sdk-contract` (run 34019692061), both `success`.
- That set covers `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -D warnings` on all three OSes, `cargo test --workspace` on Linux and macOS, and — via the `Format` job's `bash scripts/lint-file-length.sh --tree` (`ci.yml:119`) and `sh scripts/lint-markdown-content.sh` (`:121`) — both non-Cargo tree gates. What it does **not** cover is any `.ps1`, which is why the fixture was RAN locally, and the runtime behaviour of the 16 crates outside the Windows subset — irrelevant here, C5 selector empty.
- No workspace Cargo gate was run locally, and none was owed.

### Verdict

Changes Requested, P0=0, P1=2 — `Closes #748` still stands in `b78bb6c`'s commit message and this repo's squash/merge/rebase settings all carry it to `main`, so merging auto-closes two undelivered doneWhen items (G1); and the comments justifying the `--verbose` forward assert that lanes dispatch through `--detach`, which no lane in this repository does (G2). Both fixes themselves — and all three code fixes from Round 1 — are correct and each is wired to a case that fails without it.

fagemx:
## Review Response: Round 2

New head: `1b51981b84763483be5004dfa7a41c01f1f0ab19`

Both P1s are yours and both were right.

### G2 — the `--verbose` rationale asserted something false — fixed

I verified it rather than taking it: `--detach` appears nowhere in `scripts/fleet/*.ps1` as a dispatch argument. The only hits repo-wide are `git checkout --detach` and `git worktree add --detach`, which are unrelated. `lane-launch.ps1` dispatches plainly (`:222`, `:347`, `:441`) and detaches via a **Scheduled Task wrapper**, and it already asks for `--verbose` on its own dispatch line.

So my claim — "the detached path is the lane path, which is where `logBytes=0` was reported" — was wrong in both directions: lanes do not take that path, and the lane path was already covered by this PR's own `lane-launch.ps1` change.

The forward itself stays, and the reason is narrower than I wrote: `--detach` is the *second* way to get a background worker and it had the same gap for the same cause. Both the code comment and the test comment now say that. **A false comment is worse than no comment**, because the next reader trusts it — which is exactly what your round demonstrates.

### G1 — `Closes #748` survives in `b78bb6c`'s commit message — acknowledged, and it is a merge-time action

You are right about the mechanism and I cannot fix it by pushing. I removed `Closes #748` from the PR **body**, but `b78bb6c`'s commit message still carries it, and this repo's squash setting composes the squash body from the commit messages — so a default squash would auto-close #748 with two doneWhen items undelivered (item 1: a lane killed mid-L1 still carries a `Review Response`; item 4: local L1 wall-clock measured and recorded).

R7 forbids amending a pushed commit, so the remedy belongs to whoever merges:

> Merge this PR with an explicit squash body that does **not** contain `Closes #748`, e.g. `gh pr merge 972 --squash --body-file <file>`. `Closes #694` is earned and should stay.

Recording it here so it is on the PR rather than in anyone's memory, and flagging that this is a general hazard: **every PR that drops a closing keyword from its body after the commit is pushed has the same exposure**, and nothing checks for it.

### What is unchanged

Round 1's three code fixes and their regression cases are untouched — you confirmed each is wired to a case that fails without it, and this delta is comment text plus this response. No behaviour changed.

### Gates

| Gate | Result |
|---|---|
| `cargo fmt --all --check` | 0 |
| `cargo test -p edda --bin edda detached_worker_inherits_verbose` | ok, 1 passed |

Comment-only delta on top of a head whose full gates you already read; I did not rerun the workspace suite for a comment change. C5 selector empty — only `crates/edda-cli`, which is inside CI's Windows subset as `edda`.

fagemx:
報廢處置（closed-unmerged）— 依 #1010 裁定 A/B 與操作者 2026-09-07 核准的 Phase 0 方案。lane 生命週期（lane-launch/stop/reap 的 log 可讀性與 session id 誠實回報）屬機器層，按既有裁定歸 fleet-workstation 層，edda 明確不吸收（Phase 4 maintain-only）。載體 #748、#694 維持凍結，性質留在原單；本 shell 補丁不再佔用審查引擎輪。分支保留不刪。

PR #943, opened 2026-09-05
fix(fleet): no pre-push guard — the window's force-push prohibition is prose only, and a force push is the one operation that can destroy the tree a SHA-pinned verdict describes

fagemx:
## Problem and change
The window's prohibition on force pushes existed only as prose (window decision, CLAUDE.md review loop, lane brief finish step); nothing enforced it, and the 2026-09-05 window controller actually ran `git commit --amend` + `git push --force-with-lease` on PR #906 before self-disclosing. A force push over a reviewed head is the one git operation that can silently destroy the tree a SHA-pinned verdict describes (`review.verdict-carrier` holds every-push-voids "by construction" only while pushes are append-only).

This adds the missing gate:

- `scripts/fleet/guard-push.sh` (new) — reads lefthook's pre-push stdin (`<local-ref> <local-sha> <remote-ref> <remote-sha>`). Branch deletion, new branch, and fast-forward records accept with no network call. A non-fast-forward record triggers the only `gh` call (`gh pr list --head <branch> --state open`); an open PR means refusal with the PR number, the remote SHA and the local SHA; a `gh` failure refuses too (fail closed). Every refusal names `FLEET_ALLOW_FORCE_PUSH=1`, the single documented escape. The script never runs `git push` and writes nothing.
- `lefthook.yml` — new `pre-push` `guard-push` command ahead of the existing `clippy` command (clippy untouched).
- `scripts/fleet/test-guard-push.sh` (new) — installs the guard as the pre-push hook of a throwaway clone and drives real `git push` commands against a local bare origin with a stub `gh`; six cases, each asserting the exit code and the remote ref state.
- `docs/guides/brief-template.md` — the finish step now names the guard as the enforcement mechanism instead of the bare prohibition.

### RAN
- base_sha `660014331f3fc82ffad9bd4c5112f901c2e76749` (origin/main at lane start), delivery_sha `fd4d5116988b28d3335cb95e54a6cf74143baef1`.
- `sh scripts/fleet/test-guard-push.sh` → exit 0, six PASS lines: fast-forward accepted (remote moved), amend + non-ff with open PR refused (stderr names `#906`, both SHAs and the escape; `git ls-remote` byte-identical before/after), same push with `FLEET_ALLOW_FORCE_PUSH=1` accepted, new-branch push accepted, branch deletion accepted, non-ff with failing `gh` refused (fail closed).
- `sh -n scripts/fleet/guard-push.sh scripts/fleet/test-guard-push.sh` → exit 0.
[...]

Issue #913, opened 2026-09-05
fix(fleet): no pre-push guard — the window's force-push prohibition is prose only, and a force push is the one operation that can destroy the tree a SHA-pinned verdict describes


fagemx:
Evidence: on 2026-09-05 the window controller ran `git commit --amend` + `git push --force-with-lease` on PR #906, a step the window policy explicitly forbids (`fleet.all-flash-window-2026-09-05`: never force push). It self-disclosed within minutes and stopped; no SHA-pinned verdict was invalidated. This issue is the missing gate, not the incident.

## What happened

Basis `origin/main` `6486228c728dff5d488b5756e918af0bccde0eb5`.

The prohibition on force pushing exists in three places as prose — the window decision, `.claude/CLAUDE.md`'s review loop ("every push invalidates the prior verdict"), and every lane brief's finish step ("This is a normal push; no force option is authorized"). Nothing enforces it:

```
lefthook.yml            → no pre-push hook (cargo gates only)
.git/hooks/pre-push     → absent
```

A force push over a reviewed head is the one git operation that can silently destroy the evidence a SHA-pinned verdict refers to. In this instance the overwritten commit `a815a57` had lived under two minutes and carried no verdict, so the loss was zero — but that was luck about timing, not a property of the system. The three verdicts on #906 happened to be pinned to `ab76bd4`, `b838980` and `2e177d8`, all still reachable.

Related prior scar: `commit-tree` reuse silently reverting base changes, which is why branch rewriting is supposed to go through cherry-pick.

## Why it matters

`review.verdict-carrier` makes the `Independent Review` commit status the merge gate, and its every-push-voids property holds "by construction" only if pushes are append-only. A force push can move a branch back onto a SHA that already carries a green status, or destroy the tree a standing verdict described. The rule is load-bearing and unguarded.

## Suspected surface

- `lefthook.yml` — a `pre-push` hook running a new `scripts/fleet/guard-push.sh`.
- `scripts/fleet/guard-push.sh` (new) — reads lefthook's pre-push stdin (`<local-ref> <local-sha> <remote-ref> <remote-sha>`); refuses when the push is non-fast-forward on a branch that has an open PR, or when `remote-sha` is reachable-but-not-an-ancestor of `local-sha`; prints the PR number and the SHAs. Escape hatch is a single explicit env var (`FLEET_ALLOW_FORCE_PUSH=1`) that the message names, so an authorized rewrite is a deliberate one-line act and shows up in the shell history.
[...]

PR #906, opened 2026-09-05
feat(fleet): shadow review shape — shadow field in REVIEW.md, verdict-label, watcher exclusion, review-compare.sh

fagemx:
## Problem and change

GH-887 executes the operator's ruling (`review.gh880-shadow`: 「#880 這張用 glm 影子審，Opus 補審等額度」) by giving a non-verdict review round a machine-recognizable shape and a compare tool. Stacked on `feat/gh886-controller-loop` (PR #899), which posts such rounds.

- `REVIEW.md` — §7 gains the `- shadow: true` header field, §8 one sentence (a shadow round sets no `review:*` label, no `Independent Review` status, and the union ignores it), §9 one provenance row; `grep -n shadow REVIEW.md` resolves to exactly those three places.
- `scripts/review-pr.sh` — `verdict-label` prints `shadow` for a body carrying the field (both with and without the header dash) and otherwise behaves byte-identically; `EDDA_REVIEW_SHADOW=1` writes the field and the heading suffix into the brief's output contract. The verdict-label stdin is slurped once — a stdin grep would consume the body before the Verdict section is read.
- `scripts/pr-review-watch.sh` — the verdict extractor recognizes the shadow suffix and excludes the round from the union entirely; `finish_verdict` drops a shadow round's pending entry with a log line and applies no label and no status.
- `scripts/review-compare.sh` (new) — `<pr> <sha>`: picks the shadow round and the authoritative round pinned to that SHA, matches findings by file:line tokens (rule ids land with #883), and prints missed / unconfirmed / matched / severity-drift rows plus one `for-ledger` line for `fleet.review-calibration`. No SHADOW round exits 2; a pending authoritative round prints the shadow findings and exits 0. It reads GitHub only.
- The heading pin in both scripts avoids a regex spanning the em dash: gawk in the C locale counts bytes and the dash is three of them, so a character-count regex silently misaligns — the pin splits on ` @ ` and compares the tail string.

### RAN

- `sh scripts/test-review-compare.sh` → PASS (3 cases: both-rounds counts and for-ledger line, no-shadow exits 2, pending-authoritative exits 0); stub `gh` resolves `--jq` like the real one
[...]

PR #976, opened 2026-09-06
fix(fleet): pre-push guard reads the remote ref, refuses tag moves, and ships in the git-native installer

### LINKED ISSUE #957: fix(fleet): pre-push guard follow-ups — tag refs bypass the branch-scoped lookup; git-native install path wires no pre-push hook; no automated lefthook-layer test
Follow-ups from the #943 review (GH-913), none extends that PR:

1. **Tag refs bypass the guard's lookup.** `guard-push.sh` strips `refs/heads/` and passes the rest to `gh pr list --head <branch>`; a forced tag update (`refs/tags/…`) yields a branch name that matches nothing, so the push passes. Out of #913's branch scope — decide explicitly whether tags get their own refusal or an explicit allow.
2. **The git-native installer wires no pre-push hook.** `CONTRIBUTING.md` points at `scripts/githooks/install.sh` (core.hooksPath, pre-commit + commit-msg only), so clones installed that way are unguarded. #913's wiring audit delegated this to the collision-scan/companion sweep.
3. **The lefthook stdin contract is READ-verified only.** lefthook is absent from lane hosts, so the `use_stdin: true` → records-delivered behavior has no automated proof (the offline suite drives a native git hook). Worth a skippable/marked integration test where a lefthook binary exists.

Also related: none of `scripts/fleet/test-guard-push.sh`, `scripts/fleet/test-verdict-drift.sh`, `scripts/test-review-l0.sh` has a CI consumer — tracked on #927.

fagemx:
Issue: #957
Closes #957

Three follow-ups from the #943 review (GH-913). The issue carries no
`doneWhen` block — its three numbered items are the acceptance list, and each
row below is one of them.

## 1. Tag refs bypassed the guard's lookup — decided as a **refusal**

`guard-push.sh` stripped `refs/heads/` off the **local** ref and handed the
remainder to `gh pr list --head`. For `refs/tags/v1.2.3` that is a head name
matching nothing, so a forced tag move passed.

Refusal, not an allow, because outside `refs/heads/` the fast-forward test
decides nothing: re-pointing `v1.2.3` at a later commit is still a rewrite of
what an already-shipped version resolves to — `install.sh` resolves release
tags — and no tag can have an open PR for the lookup to consult, which is the
same fail-closed rule the guard already applies when `gh` is missing. The
namespace check therefore runs **before** the fast-forward test.

Deliberately unchanged: **creating** a tag still passes (its zero
`remote_sha` leaves the loop first), and **deleting** a ref stays the separate
operation this guard has never covered, branch or tag.

While fixing that line I found it wrong in a second way, same defect class:
a PR's head **is the remote branch**, so `git push origin scratch:main`
rewrote PR #906's head while the local ref said `scratch` and the lookup asked
about the wrong branch. `git push origin HEAD:main` had the same shape. The
lookup now reads `remote_ref`.

## 2. The git-native installer wired no pre-push hook

`CONTRIBUTING.md` points clones at `scripts/githooks/install.sh`, which set
`core.hooksPath` for `pre-commit` and `commit-msg` only — so a clone installed
the documented zero-dependency way pushed with **no force-push guard at all**,
while a lefthook clone had one.

New `scripts/githooks/pre-push` hands git's own stdin records — byte-identical
to what lefthook forwards under `use_stdin: true` — to the single
`scripts/fleet/guard-push.sh` both layers share, so there is no second
implementation to drift. `install.sh` chmods and announces it;
`CONTRIBUTING.md` documents what it refuses and its escape.

## 3. The lefthook stdin contract now has automated proof (two layers)

Cases 1–10 drive a **native** git hook: they prove the guard, never that
lefthook *delivers* the records to it. Drop `use_stdin: true` and the guard
reads an empty stdin, runs zero iterations, and accepts every push — with every
other case still green.

- **(a) runs everywhere:** the repo's own `lefthook.yml` is parsed and asserted
  to still route `pre-push` through the guard **with** `use_stdin: true`.
- **(b) runs where a lefthook binary exists:** the same refusal is driven
  through real `lefthook install`; a refusal naming the PR is only reachable if
  the records arrived. lefthook is absent from the lane hosts, so this leg
  prints a loud `SKIP` here rather than passing silently.

## Evidence — red before, green after

```
PASS 1 fast-forward push accepted
PASS 2 non-fast-forward push with open PR refused
PASS 3 authorized force push accepted
PASS 4 new branch push accepted
PASS 5 branch deletion accepted
PASS 6 failed open-PR check refused (fail closed)
PASS 7 new tag push accepted
PASS 8 moving an existing tag refused
PASS 9 open-PR lookup reads the remote ref
PASS 10 git-native install.sh wires the pre-push guard
SKIP 11 lefthook stdin contract: no lefthook binary on PATH (lefthook.yml contract asserted above; run this on a host that has it)
ALL PASS
```

Each new case fails on this branch's base, isolated one at a time:

| case | with the product change reverted |
|---|---|
| 8 (tag move) | `FAIL 8: moving an existing tag was accepted` |
| 9 (remote-ref lookup) | `FAIL 9: non-fast-forward push of a differently named local ref over an open PR head was accepted` |
| 10 (installer wiring) | `FAIL 10: scripts/githooks has no pre-push hook to install` |

Case 9 needed the `gh` stub to become **branch-aware** (`GH_PR_BRANCH`): the old
stub printed the PR number whatever `--head` it was given, so a guard looking up
the *wrong* branch still got an answer and no case could fail.

## Gates ran (docs + scripts only — no crate, `Cargo.*` or toolchain change, so no Cargo gate applies)

| gate | result |
|---|---|
| `sh scripts/fleet/test-guard-push.sh` | ALL PASS, 10 pass + 1 skip (was 6 cases) |
| `sh scripts/githooks/test.sh` | ALL SCENARIOS PASSED |
| `bash scripts/lint-file-length.sh --tree` | pass |
| `sh scripts/lint-markdown-content.sh` | pass |
| `bash scripts/lint-doc-citations.sh --tree` | pass |
| `sh -n install.sh scripts/lint-markdown-content.sh` | pass |
| `sh -n` on `guard-push.sh`, `githooks/pre-push`, `githooks/install.sh`, `test-guard-push.sh` | pass |
| `git diff --check` | clean |

`scripts/githooks/pre-push` is committed mode `100755` with an **LF** blob
(`.gitattributes` `scripts/githooks/* text eol=lf`) — a CRLF shebang breaks
`#!/bin/sh` on Git Bash.

SHA: `51969d09ca3d6c8af2ea4ba6f0f1733336c788f2`

## Not extended here

- **No CI consumer for `test-guard-push.sh`** (nor `test-verdict-drift.sh`,
  `test-review-l0.sh`). Named in this issue's closing note and tracked on #927 —
  this PR adds cases to a suite nothing runs automatically, which is #927's
  problem to fix, not this one's.
- **Deleting a tag or branch** still passes the guard. That is the pre-existing
  decision, not the defect this issue names; changing it is a separate call.

## Review surface

R22: `lefthook.yml` is 閘門面, so this PR needs an authoritative engine (Opus or
sol), not a SHADOW round.


fagemx:
## Gate receipt

`CI run 34003701199 @ 51969d09ca3d` — success (Format; Clippy ×3 OS; Test on ubuntu/macos/windows; MSRV 1.91.0; CI Gate).

Local, same SHA, on Windows 11 / Git Bash:
- `sh scripts/fleet/test-guard-push.sh` — ALL PASS, 10 pass + 1 skip (lefthook binary absent on this host)
- `sh scripts/githooks/test.sh` — ALL SCENARIOS PASSED
- `bash scripts/lint-file-length.sh --tree`, `sh scripts/lint-markdown-content.sh`, `bash scripts/lint-doc-citations.sh --tree`, `sh -n install.sh scripts/lint-markdown-content.sh`, `git diff --check` — clean

Build lane: n/a — no crate, `Cargo.*`, or toolchain change, so no Cargo gate applies and nothing was compiled.

fagemx:
## Review Response: Round 1

New head: `2817ddb0518e81daedc5eea4331f60db6ed4aadc`

### Blocking finding

**[P1] f1 — CONTRIBUTING.md:35 claimed a single escape while this PR documented a second.** Confirmed and fixed. `install.sh` (added by this PR) advertised `git push --no-verify`, which makes git skip the hook entirely, so the guard never runs at all. The paragraph now names both, and states that CI compensates for neither — CI gates what a head *contains* and cannot restore a reviewed head a force-push already overwrote.

### Non-blocking findings, also fixed

- **[P2] f3** — "any update to an existing tag" understated the code. The namespace check refuses an update to any ref that already exists outside `refs/heads/`. Corrected in both CONTRIBUTING.md and the installer summary.
- **[P2] f4** — the installer carried the commit hook's CI rationale over to `git push --no-verify`. The two bypasses are now stated separately, with the push side explicitly *not* covered by CI.
- **[P2] f2** — `guard-push.sh` still called itself the "lefthook pre-push gate". The header now names both callers on the same stdin contract.

### Gates

`CI run 34015412171 @ 2817ddb0518e81daedc5eea4331f60db6ed4aadc` — all 12 checks SUCCESS.

L0 (this head, Windows, worktree default target):
- `sh -n` on `guard-push.sh`, `githooks/install.sh`, `githooks/pre-push` → exit 0 each
- `sh scripts/fleet/test-guard-push.sh` → 10 PASS, 1 SKIP (no lefthook binary; the `lefthook.yml` contract is asserted regardless), `ALL PASS`
- `bash scripts/lint-file-length.sh --tree` → exit 0

No product code changed — the guard's behaviour at this head is byte-identical to `51969d09`. The diff is the four corrected statements about it.

fagemx:
## Code Review: Round 2 — PR #976 @ 2f53a29299a44b1b9823295d979c5a327b84da6f

- elapsed: unmeasured (no pi session file — this round ran outside the pi lane transport)
- model_requested: not recorded in this dispatch
- model_observed: unverified (§6 — self-labelling is not evidence; no session `model` field or `modelUsage` key available here)
- reviewer_session: none assigned by this dispatch
- spec: review-spec-v1.5
- class: code-risk  (REVIEW.md classes: code-risk docs)
- escalations: D5 (adjudicated below, marked `provisional`)
- shadow: false
- cost: ~45 tool calls; no Cargo build (no crate, `Cargo.*` or toolchain change — C5 selector empty)

### Record note (observed fact, not a finding against this diff)

This round is numbered 2 because a `## Review Response: Round 1` stands on this PR
(2026-09-06T06:19:02Z), but **no `## Code Review: Round 1` comment exists in the PR
record**. `gh pr view 976 --json comments,reviews` returns exactly two comments —
`## Gate receipt` and `## Review Response: Round 1` — zero reviews, and grepping both
bodies for `^## Code Review` returns nothing. Per `loop` ("Internal verifier reports,
task receipts and CI do not replace PR comments") that round is unrecorded, so its
findings f1–f4 were treated here as unverified assertions and re-checked, and findings
a first round would have caught are in scope for this one rather than being deferred as
late-round blockers.

### Head SHA re-pin

Dispatch pinned `2817ddb0518e81daedc5eea4331f60db6ed4aadc`. The head has since moved to
`2f53a29299a44b1b9823295d979c5a327b84da6f` (`docs(fleet): "bypass everything" is false
once a pre-push hook exists`, 2026-09-06 14:28:52 +0800). REVIEW.md §1 says to restart
at the new head rather than mix two trees, so this round is pinned to `2f53a292` and
covers the whole PR including the `2817ddb0..2f53a292` delta (three comment lines in
`scripts/githooks/install.sh:12-16` — it still said `Bypass everything: git commit
--no-verify`, which the Round 1 response's f1 fix had corrected in `CONTRIBUTING.md`
but not in the installer header). That delta is correct and completes f1.

### IN SCOPE

Changed behaviour and paths (5 files, +232/−19): `scripts/fleet/guard-push.sh` gains a
ref-namespace refusal and reads the PR-lookup branch from the **remote** ref;
`scripts/githooks/pre-push` is a new git-native hook; `scripts/githooks/install.sh`
wires and announces it; `scripts/fleet/test-guard-push.sh` adds cases 8/9/10/11;
`CONTRIBUTING.md` documents it. Direct consumers: `lefthook.yml:10-14` (unchanged,
still `use_stdin: true`) and the new hook — both reach the one guard implementation.
Acceptance ceiling: issue #957 carries no `## doneWhen` block; its three numbered items
are the list, and all three are delivered — (1) tags decided as a refusal on update
(`guard-push.sh:55-59`), (2) the git-native installer wires `pre-push`
(`install.sh:30,34`), (3) the lefthook stdin contract gains a two-layer test
(`test-guard-push.sh:248-273`). Current-base integration: `main` has moved from this
PR's base `586cb07e` to `39ba0429`; the merge-ref CI at this head is green on that base.

Because this is a guard that decides whether a push is allowed, both failure directions
were probed adversarially against a throwaway repo under `mktemp -d` (never this repo's
remote, never this repo's `.git`):

**Fail-open: none found, and none introduced.** Every path this diff changes is strictly
more closed than the base. Verified refusals: multi-ref push with the offending record
first *and* second (git rejects all refs atomically; a new branch in the same invocation
was not created); detached-HEAD `git push --force origin HEAD:main`; `scratch:main`;
lightweight *and* annotated tag moves; a `remote_sha` whose object is absent locally
(`merge-base` errors → falls through to the PR lookup → refuse); `gh` missing or
failing; the last stdin record lacking a trailing newline. `refuse()` is not in a
subshell, so its `exit 1` aborts the script and the whole push. The only accept-all path
is an **empty** stdin (`printf '' | sh guard-push.sh` → exit 0) — that is the pre-existing
`use_stdin: true` regression shape, and case 11(a) is precisely the assertion added here
to catch it.

**Fails-closed-wrongly: one behaviour change, documented.** New branch, new tag
(lightweight and annotated), branch deletion, tag deletion, fast-forward, and
`FLEET_ALLOW_FORCE_PUSH=1` all pass correctly. But a **fast-forward** update of
`refs/notes/*` is now refused where the base guard accepted it (measured both ways) —
intended per the namespace rule and disclosed in `CONTRIBUTING.md`, noted as P2 only for
its message wording.

**Installer half: verified end-to-end.** `install.sh` is idempotent (byte-identical
output and exit 0 on a second run), and a tag move is refused through the installed
`core.hooksPath` route with the hand-written `.git/hooks/pre-push` removed.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 5 files, all inside #957's three items: `guard-push.sh`, `githooks/pre-push`, `githooks/install.sh`, `test-guard-push.sh`, `CONTRIBUTING.md` |
| U2 | any | P1 | PASS | body+commits carry `Closes #957`; #957 has no `## doneWhen`, its 3 numbered items are the acceptance list and all 3 are delivered (see IN SCOPE), so the closing keyword is correct under `pr.closing-keyword=only-when-all-donewhen-delivered` |
| U3 | any | P1 | **FAIL** | `gh pr view 976 --json body --jq .body \| awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → empty. See [P1] below |
| U4 | any | P1 | PASS | `git log --format=%s origin/main..2f53a292 \| grep -Ev '<type>(<scope>): …'` → empty over all 3 commits |
| U5 | any | P1 | PASS | RAN `sh scripts/lint-markdown-content.sh` → `exit=0` |
| U6 | any | P0 | PASS | READ `gh pr checks 976` — CI run **34016606979 @ 2f53a292**, conclusion `success`, 12/12 including `CI Gate` |
| U7 | any | P1 | PASS | whole scoped audit completed and every blocker batched before posting; no product cycle spent |
| D1 | docs | P1 | PASS | `gh pr --help` → 0 OK; `git commit -h` → 129 OK; `git push -h` → 129 OK (git probed with `-h` per #691) |
| D2 | docs | P1 | N.A.(no ledger decision value or status is asserted anywhere in the diff — the citations are issue numbers GH-913/GH-957/#927, not decision keys) | — |
| D3 | docs | P1 | PASS | all repo-anchored paths resolve at the head tree: `lefthook.yml`, `scripts/fleet/guard-push.sh`, `scripts/githooks`, `scripts/githooks/install.sh`, `scripts/githooks/pre-push` (`git ls-tree` → `100755 blob 2a2a859`; the enumerator's one `MISSING` was an artifact of running it from a base checkout) |
| D4 | docs | P0 | PASS | 14 candidates. Each `--no-verify` / `push --force` hit is either the guard's own disclosed escape carrying the caveat in the same paragraph (`CONTRIBUTING.md`: "CI compensates for neither — it gates what a head contains, and cannot restore a reviewed head that a force-push has already overwritten"; `install.sh:49-51` same), or a sandbox `git push --force` inside `test-guard-push.sh`. Nothing instructs a reader to merge, delete a branch, or skip review |
| D5 | docs | `[判斷]` | 需升級 (provisional) | Adjudicated: prose is clear, correctly scoped and matches the code. One duplication concern — the same refusal semantics are now restated in four places (`CONTRIBUTING.md:30-41`, `install.sh:41-43`, `guard-push.sh:2-25`, `githooks/pre-push:5-18`) and will drift; filed as P2 below, not a blocker |
| C1 | code-plain | P0 | PASS | all three #957 items matched to code **and** to a test that fails without it (cases 8/9/10); independently reproduced red-before, see RAN |
| C2 | code-plain | P0 | PASS | `fix:` subject with regression tests: cases 8, 9, 10 added; base guard exits 0 and head guard exits 1 on the tag-move and `scratch:main` records (RAN) |
| C3 | code-plain | P1 | N.A.(no `crates/**` in the diff) | — |
| C4 | code-plain | P1 | PASS | enumerator empty — no `unsafe {`, no `clippy::all`, no `#![allow(` |
| C5 | code-plain | P1 | N.A.(selector empty — no crate touched, so there is no Windows CI coverage gap to close) | — |
| R1 | code-risk | P0 | PASS | 4 hits, all `git push --force` in `test-guard-push.sh` (:183, :202, :230, :264). Trigger for every one: running `sh scripts/fleet/test-guard-push.sh`; the target is `$work/origin.git`, the bare repo created by `git init --bare` at :55 inside `mktemp -d` (:24) and removed by the EXIT trap (:25). Never a real remote |
| R2 | code-risk | P0/P1 | PASS | 2 candidates, both `( A && B ) >file 2>&1 \|\| fail N` (`test-guard-push.sh:221`, `:261`). The `&&` is inside a subshell, so this is not the `A \|\| B && C` shape — parse tree is `subshell` then `\|\|`. Truth table: `cd` fails → subshell non-zero → `fail` runs; installer/lefthook fails → subshell non-zero → `fail` runs; both succeed → `fail` does not run. No row runs an R1 command, on either branch → not P0, not P1 |
| R3 | code-risk | P0 | PASS | RAN `sh -n` on all four changed scripts → exit 0 each |
| R4 | code-risk | P1 | PASS | `test-guard-push.sh:24-25` pairs `work=$(mktemp -d)` with `trap 'rm -rf "$work"' EXIT`, so the error path releases too. `guard-push.sh` and `githooks/pre-push` create no temp resource |
| R5 | code-risk | P1 | PASS (one note) | Properties confirmed: every ref/SHA interpolation is quoted — `gh pr list --head "$branch"`, `git merge-base --is-ancestor "$remote_sha" "$local_sha"` — and untrusted values otherwise reach only quoted message strings. Note: a stdin record with fewer than 4 fields is silently **accepted** rather than refused (P2 below) |
| §5.5 | any | P1 | PASS | table below; no "no consumer" row, no swallowed error on a ledger/coordination/cost path |

### Wiring

|---|---|---|---|---|
| `scripts/githooks/pre-push` (new file, mode `100755`, LF blob via `.gitattributes:4`) | `scripts/githooks/install.sh:34` `git config core.hooksPath scripts/githooks`; `:30` adds it to the `chmod +x` list | git's pre-push protocol; the file itself `exec`s the guard at `githooks/pre-push:23` | `exec` makes the guard's exit code the hook's; `set -eu` + a failing `git rev-parse` exits non-zero. **Verified fail-closed**: with `scripts/fleet/guard-push.sh` deleted the push exits 1 (`sh: …/guard-push.sh: No such file or directory`) and the remote ref does not move | install.sh → `core.hooksPath` → `git push` → `pre-push` → `guard-push.sh` → refusal. Proven by case 10 in the CI log **and** independently by driving a tag move through the installed hook with `.git/hooks/pre-push` removed |
| Ref-namespace refusal branch, `guard-push.sh:55-59` | the guard, on `remote_ref` not matching `refs/heads/*` | `git push` exit code + stderr | `refuse()` → exit 1; message names the ref, both SHAs, and the escape | reaches the push verdict. Proven for lightweight (CI case 8) and annotated (RAN) tags, and through the installed hooksPath route (RAN) |
| `branch=${remote_ref#refs/heads/}`, `guard-push.sh:65` (was `local_ref`) | the guard | `gh pr list --head "$branch"` at `:70` | non-zero `gh` → `refuse` (fail closed); empty/`null` → `continue` | flag → lookup → refusal. Proven by CI case 9 and by `git push --force origin HEAD:main` from a detached HEAD (RAN) |
| `GH_PR_BRANCH` stub key, `test-guard-push.sh:45-47` | the test's own `gh` stub | consumed by cases 9 and 10 (`:200`, and still exported at `:230`) | makes the stub branch-specific so a guard asking the wrong branch gets no answer | test-only; not dead on arrival |
| 3 new operator-facing `echo` lines, `install.sh:41-43` | `install.sh` | the operator's terminal | **none — printed unconditionally, never verified**. This is the P2 below | install.sh → stdout only |

### Findings

- **[P1] The PR body carries no `Issue: #N` line** (REVIEW.md U3 — "Empty output is the failure"). The body opens with `Closes #957` only. U3 is explicit that this stays P1 even though the closing keyword already supplies the brief with the issue's acceptance list, because the repo wants the issue named in that exact conventional form — sibling PRs #943 and #967 both carry `Issue: #913` / `Issue: #912`. Evidence: `gh pr view 976 --json body --jq .body | awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → empty. Fix is one line in the body; no code change and no re-run of any gate.
- **[P2] `install.sh` announces the pre-push guard, and exits 0, even when `scripts/githooks/pre-push` does not exist.** `install.sh:30` is `chmod +x "$dir/pre-commit" "$dir/commit-msg" "$dir/pre-push" 2>/dev/null || true` — the missing-file error is discarded twice — and nothing else checks that the hook is present. Measured in a scratch clone with `scripts/githooks/pre-push` removed: install exits 0, prints `pre-push : scripts/fleet/guard-push.sh — refuses a non-fast-forward push…`, prints no warning, and a subsequent tag move is **accepted** (remote tag re-pointed). Reachability is low — the hook blob and the new `install.sh` land in the same commit, so only a hand-damaged or partial tree can reach it — which is why this is P2 and not P1. It is worth closing anyway because the same swallow covers the exec bit, and Linux silently skips a non-executable hook, so the one thing this installer cannot currently tell an operator is whether the guard it just announced will actually run. A `[ -f … ] || exit 1` before the announcement would make the claim true by construction.
- **[P2] The new tag-move refusal is defeated by delete-then-recreate, using two ordinary pushes and no escape.** Measured: `git push origin :refs/tags/atag` → accepted (exit 0, tag gone), then `git push origin atag` pointing at a different commit → accepted (exit 0), net effect a re-pointed published tag with no refusal and without `FLEET_ALLOW_FORCE_PUSH=1`. The deletion half is the pre-existing, explicitly-retained decision (PR body "Not extended here"), so this is not a regression — but it does mean the stated threat ("re-pointing a published tag rewrites what an already-shipped version resolves to", `guard-push.sh:19-21`) is only defended against the one-push form, while `CONTRIBUTING.md:36-38` enumerates "Two things get past it". Worth one sentence in `CONTRIBUTING.md` naming the third, or the follow-up below.
- **[P2] A fast-forward `refs/notes/*` update is now refused where the base guard accepted it.** Measured both ways against the same record: head guard exit 1, base guard exit 0. This follows from the namespace rule and is disclosed in `CONTRIBUTING.md:34-36` (`refs/notes/*`, `refs/replace/*`), so the behaviour is intended. The wording is what jars: the refusal text says "re-pointing a published tag rewrites what an already-shipped version resolves to" and the escape prints "this rewrite is on your shell history" — both describe a rewrite, for what is an append-only notes fast-forward. If notes pushes are expected here, allowing a genuine fast-forward outside `refs/heads/` would keep the tag protection intact while removing the false positive.
- **[P2] A stdin record with fewer than four fields is silently accepted.** `printf 'refs/heads/main <sha>\n' | sh guard-push.sh` → exit 0; likewise 1-, 2- and 3-field records. `read` leaves the missing variables empty, and an empty `local_sha`/`remote_sha` matches the deletion / new-ref arms at `:40-47`. Not reachable from git, which always writes four fields, so this is hardening rather than a defect — but it is the one place the script's own stated principle ("a guard that cannot check must not wave the push through", `:9-10`) is not applied.
- **[P2] The PR body's "No CI consumer for `test-guard-push.sh`" is stale against the current base.** `scripts/fleet/run-fleet-tests.sh` globs `scripts/fleet/test-*.sh` and does not quarantine this one; it landed on `main` after this branch's base (`586cb07e` has no such file, `39ba0429` does), and the merge-ref CI at this head ran it on both runners. The `Fleet shell tests` job log for run 34016606979 shows `RUN scripts/fleet/test-guard-push.sh`, `PASS 1`…`PASS 10`, `SKIP 11`, `ALL PASS`. This is good news — the suite is gated, and the evidence behind this PR is stronger than its own body claims — but the "Not extended here" paragraph now says the opposite of what CI does.
- **[P2] The same refusal semantics are restated in four places** (`CONTRIBUTING.md:30-41`, `install.sh:41-43`, `guard-push.sh:2-25`, `githooks/pre-push:5-18`). Round 1's f3 already had to correct two of them in lockstep when one drifted from "any update to an existing tag" to the broader namespace rule; the fourth copy is the next one to drift. (This is the D5 adjudication, marked `provisional` per §6.1.)

### FOLLOW-UP ISSUE

- **`install.sh` silently disables a pre-existing `.git/hooks/pre-push`.** `core.hooksPath` makes git ignore `.git/hooks/` wholesale. Measured: with an executable `.git/hooks/pre-push` in place that echoes and refuses, after `sh scripts/githooks/install.sh` that hook never runs and never prints, while the `scripts/githooks` guard does; `install.sh` says nothing about it (0 mentions of "existing"/"override"/"disabled"/".git/hooks"). Pre-existing — the mechanism and the silence predate this PR, which only adds a third hook to it — so it does not extend this PR. It is worth an issue because adding a *security* hook to that mechanism raises what a silent override costs.
- **The guard ignores which remote is being pushed to.** `githooks/pre-push:23` passes `"$@"` (remote name and URL) through explicitly unread, and `guard-push.sh:70` calls `gh pr list` with no `--repo`, so the open-PR question is always asked of `gh`'s default repo. Pushing to a second remote therefore consults the wrong repository's PRs. Pre-existing (the `gh pr list` shape is unchanged by this diff); noted because this PR is the first to document the pass-through as deliberate.
- **`FLEET_ALLOW_FORCE_PUSH=1 git push …` is POSIX-shell-only.** The refusal message tells the operator to "set `FLEET_ALLOW_FORCE_PUSH=1` for this one push"; that inline form is a parser error in PowerShell, which is a primary shell on this project's Windows workstations (`.claude/CLAUDE.md` makes the same point for `CARGO_INCREMENTAL`). A `$env:FLEET_ALLOW_FORCE_PUSH = "1"` hint in `CONTRIBUTING.md` would close it. Pre-existing message text.

### RAN vs READ

RAN (all against a throwaway repo under `mktemp -d`, with a `gh` stub verified in effect via `command -v gh` on a POSIX path; this repo's `.git` was never touched and no ref of `fagemx/edda` was written):
- `sh scripts/lint-markdown-content.sh` → `exit=0` (U5).
- `sh -n` on `guard-push.sh`, `test-guard-push.sh`, `githooks/install.sh`, `githooks/pre-push` (head blobs) → exit 0 each (R3).
- U2/U3/U4/C4/C5/R1/R2/D1/D3/D4 enumerators and the §3 classifier block from `REVIEW.md` → `classes=code-risk docs`, `canonical_class=code-risk`.
- Adversarial matrix the CI suite does not cover: multi-ref push with the bad record first and second (refused, no ref moved, sibling branch not created); annotated tag create → accepted / move → refused; tag delete → accepted then recreate elsewhere → accepted (the P2 above); detached-HEAD `HEAD:main` non-FF over an open PR → refused; `refs/notes/*` FF update → refused at head, accepted at base; empty stdin → exit 0; 1-, 2-, 3-field records → exit 0; final record with no trailing newline → refused, no hang; `remote_sha` object absent locally → refused; `FLEET_ALLOW_FORCE_PUSH=1` on a tag move → accepted; `install.sh` run twice → identical output, exit 0 both; `install.sh` with the hook file absent → exit 0 and guard inert; `guard-push.sh` absent under the installed hook → exit 1, fail-closed; `git push --no-verify` → bypasses, as documented.
- Red-before, independently of the PR body's table: base `guard-push.sh` exits **0** on both the tag-move record and the `refs/heads/scratch → refs/heads/main` non-FF record; head `guard-push.sh` exits **1** on both.

READ (not re-run):
- L1 receipt **`CI run 34016606979 @ 2f53a29299a44b1b9823295d979c5a327b84da6f`** — conclusion `success`, 12/12: Format, Detect code changes, Clippy ×3 OS, Test ×3 OS, MSRV (1.91.0), Fleet shell tests, Fleet shell tests (windows), CI Gate. Covers the four workspace Cargo `gates:` entries, which are therefore not re-run here (`ladder` L1), **and** — unlike what the PR body assumes — covers `scripts/fleet/test-guard-push.sh` itself on ubuntu and Windows. Does not cover: the adversarial inputs above, which is why they were RAN.
- `CI run 34015412171 @ 2817ddb0` — success, verifying the Round 1 response's gate claim at the then-head.
- Verified the Round 1 response's substantive claim by reading the trees, not by trusting it: `guard-push.sh` with comment lines stripped is **byte-identical** across `51969d09`, `2817ddb0` and `2f53a292`, so "no product code changed since 51969d09" holds, and the three commits after the fix are documentation only.

### Verdict

Changes Requested, P0=0, P1=1 — the guard's product change is correct in both failure directions and I found no fail-open path; the single blocker is REVIEW.md U3, the missing `Issue: #957` line in the PR body. `escalations:` is non-empty (D5, `provisional`), so this verdict is provisional per §6.4.

fagemx:
## Code Review: Round 3 — PR #976 @ 2f53a29299a44b1b9823295d979c5a327b84da6f

- elapsed: 116 s (edda review session)
- model_observed: claude-opus-5 (in-band)
- reviewer_session: d38251a9-72c0-5011-bcce-cf4478bc9dce
- spec: review-spec-v1.5
- class: code-risk  (REVIEW.md classes: docs-skills, code-risk)
- escalations: D5; §6.5 zero findings on a non-trivial diff — both adjudicated below
- shadow: false
- cost: $2.27 / 116 s / tool_policy=hard / independence=session (same-model)

### IN SCOPE
Delta round over `2817ddb0..2f53a292` (5 files, 251 lines). Only `scripts/githooks/install.sh` moved (`da5949b`→`3d0e7a5`), usage header lines 12-18 only.

### Rules
|---|---|---|---|---|
| Round-2 f1 resolved | D5 | P2 | PASS | `install.sh:12-16` now separates "Bypass the commit-time hooks" from "Bypass the push guard (a separate hook, and `--no-verify` does not carry from one to the other)" |
| Round-1 f1-f4 remain resolved | — | P1/P2 | PASS | the blobs carrying those fixes did not move in this delta |
| U1 surface | U1 | P0 | PASS | diff touches only the GH-957 guard/hook/doc surface |
| U2 / U3 | U2,U3 | P1 | N.A.(PR body not in EVIDENCE) | — |
| U4 conventional commit subject | U4 | P1 | PASS | head subject conforms |
| U5 markdown content lint | U5 | P1 | PASS (READ) | `ci.yml:227` requires fmt success unconditionally; `ci.yml:105` runs `lint-markdown-content.sh` |
| U6 gates READ before RAN | U6 | P0 | PASS | `ci.yml:78` matches `scripts/*` so `code=true`; the docs-only skip branch at `:232` requires `CODE=false` |
| U7 round completeness / two-cycle stop | U7 | P1 | PASS | zero findings this round; rounds 2 and 3 were both non-product prose cycles, so per `loop` item 6 no further wording round is opened |
| D1 `git push --no-verify` | D1 | P1 | PASS (RAN) | probe `git --no-pager push -h` |
| D1 `git commit --no-verify` | D1 | P1 | PASS (RAN) | probe `git --no-pager commit -h` |
| D1 `gh pr list --head` | D1 | P1 | PASS (RAN) | probe `gh pr --help` |
| D1 source claim — install.sh resolves release tags | D1 | P1 | PASS | `install.sh:130-137,164-176`; `release.yml:3-5` triggers on tag push — but see the adjudication's P2 on the citation |
| D3 added paths resolve | D3 | P1 | PASS | all five named paths present |
| D4 authority boundary | D4 | P0 | PASS | `install.sh:12-16` discloses `--no-verify` as hook mechanics; `:49-51` states the compensating-control gap |
| R1 destructive commands | R1 | P0 | PASS | unchanged from Round 1: `rm -rf "$work"` under `trap EXIT` (`test-guard-push.sh:25`), `rm -f` on the clone hook (`:217`), `git push --force` confined to the sandbox |
| R2 shell operator precedence | R2 | P0/P1 | PASS | this delta adds only comment lines; Round 1's two mixed lines are `( A && B ) || fail` |
| R3 changed scripts parse | R3 | P0 | PASS | see adjudication — `sh -n` ×4, exit 0 each |
| R4 resource release | R4 | P1 | PASS | `trap 'rm -rf "$work"' EXIT` covers `fail()` and `set -e` paths |
| R5 boundary and injection | R5 | P1 | PASS | `guard-push.sh:65,70` quote `$branch` into gh argv with no shell re-parse; `refuse()` interpolates refs into stderr only |
| C5 Windows-gap crates | C5 | P1 | N.A.(selector empty) | no `crates/**` path |
| §5.5 Wiring | 5.5 | P1 | PASS | comment-only delta adds no surface; wiring-scan over `586cb07..2f53a29` reports no new pub surfaces |
| D5 wording and structure | D5 | 判斷 | 需升級 | adjudicated below |

### Wiring
|---|---|---|---|---|
| no new surfaces in this delta | — | — | — | Round 1 table unchanged |

### Findings
none this round (P0=0, P1=0, P2=0)

### FOLLOW-UP ISSUE
See the adjudication below — four evidenced items, none extending this PR.

### RAN vs READ
RAN:  `git --no-pager push -h`, `git --no-pager commit -h`, `gh pr --help` (D1 probes).
READ: `CI Gate` green. Implementer receipt: `CI run 34016606979 @ 2f53a29299a44b1b9823295d979c5a327b84da6f`, all 12 checks SUCCESS.

### Verdict
LGTM (P0=0, P1=0) — provisional under §6.4 pending the adjudication below.

fagemx:
## Escalation Adjudication — PR #976 @ 2f53a29299a44b1b9823295d979c5a327b84da6f

Adjudicated by `claude-opus-5` via Claude Code — R22 authoritative on all four surfaces, including the 閘門面 the git hooks belong to. Separate session from the author and from the round reviewer.

**(A) D5 — PASS. (B) §6.5 zero-findings — PASS. P0 = 0, P1 = 0.** Three P2s and four follow-ups below.

### (A) D5

Round 3's decision that `guard-push.sh:25` ("Single escape") is not a finding is **confirmed as to severity, partially overturned as to "not a finding"** — it is a P2, not nothing, and not a P1:

- The line is **pre-existing**: `586cb07e:scripts/fleet/guard-push.sh` carries it verbatim at base line 9, and it appears as a context line in `gh pr diff 976`. Under `loop` item 1 that makes it a follow-up, not a blocker, unless newly false or newly exposed.
- As a statement about the script's own control flow it is **true**: `:28` is the only early `exit 0`, and `git push --no-verify` is not an escape *from* the guard — git never invokes it (`git push -h`: `--no-verify  bypass pre-push hook`).
- The round-1 P1 was a contradiction between two *added* contributor-facing enumerations. Every contributor-facing surface this PR adds now states both escapes — `CONTRIBUTING.md:36-38`, `install.sh:49-51`, `pre-push:18`. No safety gap.
- Still a P2 because this PR rewrote that header block to frame the file at the hook layer ("Two callers reach it…"), leaving `:25` as the one place in the changed set without the qualifier the PR added everywhere else — in the file whoever next edits the guard reads first.

Everything else in the added prose verified true, including "Before GH-957 this hook did not exist" (`scripts/githooks/pre-push` is absent at base; base `install.sh:25` chmod'd only `pre-commit commit-msg`), the `refs/heads/` refusal description, the POSIX-`sh` claim, and `"$@"` being passed through unread.

Two further P2s:

- **[P2]** `guard-push.sh:18-24` and `:49-54` are near-verbatim duplicates — same argument, same `v1.2.3` example, same "a brand-new tag is unaffected" sentence — with the rationale stated a third time in the refusal string at `:58`. Three statements in a 79-line file; D5 covers "non-duplicative".
- **[P2]** `guard-push.sh:20-21` / `:51-52` — the parenthetical "(install.sh resolves release tags)" does not carry the harm it is offered to prove. Root `install.sh:130-137` reads `tag_name` from the **Releases API** (`releases/latest`) and `:162-165` downloads from `releases/download/${TAG}`; moving the underlying git tag does not alter a stored release's assets. The underlying claim is sound via git/source resolution (`git checkout v1.2.3`, `archive/refs/tags/…`) — the citation is the weak part, not the claim, hence P2.

### (B) §6.5 — independent fail-open audit

Accept paths enumerated exhaustively from `guard-push.sh:28-79` and each classified. **No reachable fail-open beyond the ones the PR documents and asserts against.**

| # | Property | Result |
|---|---|---|
| 1 | Final record without a trailing newline is processed | refuses either way — the `[ -n "${local_ref:-}" ]` continuation at `:38` is correct |
| 2 | That idiom cannot loop forever | terminated immediately under `timeout 20` |
| 3 | Deletion (`local_sha` all-zero) | accept — documented at `:23-24, :42`; see follow-up 3 |
| 4 | New ref (`remote_sha` all-zero) | accept — nothing to overwrite; matches tests 4 and 7 |
| 5 | Malformed short record (2-3 fields) | accept, but **not reachable**: git always emits 4 fields and ref names cannot contain spaces |
| 6 | Empty / newline-only stdin | accept — the `use_stdin` fail-open the PR **asserts against** at `test-guard-push.sh:249-252` |
| 7 | A refusal aborts the whole push | rc=1 — `refuse` exits the script; the loop is not in a subshell |
| 8 | Every record examined until one refuses | rc=1 — only `refuse` exits |
| 9 | Namespace test not evadable by a near-miss prefix (`refs/headsfoo`) | refuse — the `case` at `:55` is fail-closed |
| 10 | Unknown remote object treated as non-FF, not FF | `is-ancestor` rc=128, falls through to the PR lookup. Fail-closed |
| 11-12 | `gh` absent / `gh` fails | refuse, "failing closed" in stderr |
| 13 | `gh` returns `null`/empty | accept by design (no open PR = no protected head); any other non-empty value refuses |
| 14 | `set -e` cannot silently accept | a failure exits non-zero, so git aborts the push |
| 15 | Missing guard file / not a repo | `exec sh` → 127, `rev-parse` → 128. Fail-closed |
| 16 | Exit code reaches git through the native hook | test 10, installed path with `.git/hooks/pre-push` removed first |
| 17 | `FLEET_ALLOW_FORCE_PUSH` scoped to one push | only exact `"1"` bypasses, matching the `SKIP_CLIPPY` convention; an `export` would persist for the shell but every bypassed push prints the loud notice at `:29`, so it cannot be silent. Not a finding |
| 18 | R5 quoting of interpolated values | all quoted; no `eval`; `printf` with a literal format; `gh` is not a shell |
| 19 | R2 mixed operators | none added |
| 20 | R3 all changed scripts parse | `sh -n` ×4, exit 0 each |
| 21 | Suite green, its skip honest | `PASS 1`-`PASS 10`, `SKIP 11`, `ALL PASS`, exit 0 |

GH-957's three items are all delivered: tag refusal (`:55-59`, test 8), git-native install wiring (`install.sh:30,41-43`, test 10 through the installed path), skippable lefthook-layer test (11a static + 11b conditional).

### FOLLOW-UP ISSUE (evidenced, does not extend this PR)

1. `docs/guides/brief-template.md:228-232` — "runs as the lefthook pre-push hook" is now incomplete, and its "`FLEET_ALLOW_FORCE_PUSH=1` as its single documented escape" is contradicted by `CONTRIBUTING.md:36-38`. Outside the diff.
2. `scripts/githooks/pre-commit:37` and `docs/guides/operator-runbook.md:166` carry the same "bypass everything" drift this PR's own commit subject names, in files outside the diff.
3. **Deletion asymmetry.** `git push --delete origin v1.2.3` passes silently while re-pointing that same tag is refused — under the guard's own rationale it refuses the lesser harm and permits the greater. Honestly documented at `:23-24`, and GH-957 item 1 asked only for an explicit decision, which this PR made.
4. **Wrong-remote lookup.** `pre-push:23` passes `"$@"` unread, so `gh pr list` resolves the repo from gh's own default rather than the pushed remote. Fail-closed when gh cannot resolve; can pass when gh resolves to a different repo with no open PR. A stated design choice, not reachable in this repo's push-to-origin workflow.

### RAN

`gh pr diff 976`; `sh -n` on all four changed shell scripts (exit 0 each); `sh scripts/fleet/test-guard-push.sh` (`PASS 1`-`PASS 10`, `SKIP 11`, `ALL PASS`); 12 crafted-stdin probes plus 3 `FLEET_ALLOW_FORCE_PUSH` value probes under `timeout 20`; `git merge-base --is-ancestor` on an unknown object → rc=128; `git push -h`; `git show 586cb07e:…` base comparisons; `grep -rn` sweeps. The suite is sandboxed in its own `mktemp -d` clone and bare origin; nothing outside it was written.

**`escalations:` is now empty. The round-3 LGTM is no longer provisional.**

PR #946, opened 2026-09-05
fix(fleet): mergeStateStatus is not a readiness signal — a stacked PR with zero verdicts reports CLEAN, and the digest's blocked-PR filter skips it, so #899 was reported complete on a stale SHA

fagemx:
## Problem and change
`mergeStateStatus` is not a readiness signal. The ruleset protects `main` only, so a stacked PR whose base is a feature branch reports `CLEAN` while carrying zero verdicts — and two consumers read it as ready: the human/agent controller (the day-0 report called #899 complete while its head had moved 78 minutes past the adjudication) and `daily-digest.sh`, whose 擋住什麼 filter (`mergeStateStatus == "BLOCKED" or "DIRTY"`) is silent on exactly those PRs.

Changes:

- `scripts/fleet/verdict-drift.sh` (new, read-only) — the readiness check as a standalone command. One line per open PR: `#<n> <head12> <base> <state>` with state `no verdict on head` / `stale from <sha12>` / `SHADOW only` / `LGTM` / `Changes Requested`, decided by the newest first-line R23 verdict comment (`## Code Review: Round <N> — PR #<n> @ <sha>`); a ` base=<branch> (status contexts not enforced)` annotation is appended when the base is not `main`. Exit 1 while any PR is not ready, exit 2 on a failed read (never a silent clean bill), exit 0 otherwise. Non-`main` PRs annotated; nothing posts, labels, merges, or writes the ledger.
- `scripts/fleet/daily-digest.sh` — 擋住什麼 gains the drift states: every open PR that is `no verdict on head`, `stale from …`, or `SHADOW only` now appears, with its drift reason appended when the PR is also BLOCKED/DIRTY today. No row the digest prints today is lost; `mergeStateStatus` reasons print exactly as before. The drift check is invoked, not re-implemented: its exit 1 is the normal not-ready signal and does not abort the digest, its exit 2 takes the digest's existing gh-failure path.
- `scripts/fleet/test-verdict-drift.sh` (new) — stub `gh` routes to raw JSON fixtures and applies the caller's `--jq` filter with the real jq, so the R23 heading regex itself is exercised; seven cases assert the exact line and exit code.
- `docs/guides/pi-controller-runbook.md` — new `## 回報` section with the three reporting rules (generated digest, drift check before any completion claim, `mergeStateStatus` caveat).

[...]

Issue #914, opened 2026-09-05
fix(fleet): mergeStateStatus is not a readiness signal — a stacked PR with zero verdicts reports CLEAN, and the digest's blocked-PR filter skips it, so #899 was reported complete on a stale SHA


fagemx:
Evidence: the day-0 window report called PR #899 complete. It was not — its head had moved past the verdict, it carried no `Independent Review` status, and GitHub still showed it `CLEAN`. Every readiness signal a controller can cheaply read agreed with the wrong answer.

## What happened

Basis `origin/main` `6486228c728dff5d488b5756e918af0bccde0eb5`, measured 2026-09-05.

Ruleset `18852689 Protect main` requires two status contexts, `CI Gate` and `Independent Review`. It protects **`main` only**. The day-0 lane stacked its PRs onto each other:

| PR | base | head combined status | `mergeStateStatus` |
|---|---|---|---|
| #895 | `main` | `pending` (no Independent Review) | **BLOCKED** ✓ |
| #907 | `main` | `pending` | **BLOCKED** ✓ |
| #899 | `feat/gh885-brief-from-issue` | `pending` | **CLEAN** ✗ |
| #906 | `feat/gh886-controller-loop` | `pending` | **CLEAN** ✗ |

All four heads carry the *same* absent verdict. The two that target `main` are correctly blocked; the two that target a feature branch report `CLEAN`, because no ruleset covers those refs.

**The gate at the `main` boundary still holds** — merging a stacked PR into its parent branch moves the parent's head, which voids the parent's status, so nothing reaches `main` unreviewed. This is not a hole in the merge gate, and it must not be described as one. What it is: **`mergeStateStatus` is not a readiness signal, and on a stacked PR it actively reports the opposite of the truth.**

Two consumers already read it that way:

1. The human/agent controller. The day-0 report listed #899 as `✅ 完成 @ a55d64f` while the head was `091bd10e`, pushed 78 minutes after the adjudication. Nothing contradicted it.
2. `scripts/fleet/daily-digest.sh:141-143` — the 擋住什麼 section filters `select(.mergeStateStatus == "BLOCKED" or .mergeStateStatus == "DIRTY")`, so a stacked PR sitting at `CLEAN` with zero verdicts never appears in the digest either. The one machine-generated report the operator gets is silent on exactly the PRs whose readiness is least trustworthy.

## Why it matters

[...]

PR #899, opened 2026-09-05
feat(fleet): skill-less controller loop — next-issue.sh, next-review.sh, and the pi controller runbook

fagemx:
## Problem and change

GH-886: during the all-flash window the controller is pi + glm with no skills, so the control loop must exist as scripts and an enumerated runbook. This PR stacks on `feat/gh885-brief-from-issue` (PR #895) because `next-issue.sh` consumes the GH-885 renderer.

- `scripts/fleet/next-issue.sh <issue> <machine>/<role> [--dry-run]` — queue lint, claimed/non-ready refusal naming the claimant from the `taking:` comment, brief rendered by `brief-from-issue.sh` into `~/.edda/fleet/brief-gh<N>.md`, refusal to launch while `<<AUTHORED STEPS>>` is unfilled (an authored, marker-free brief is kept, never overwritten), then `edda task new` (scope paths taken from the rendered facts line), the claim guard, the worktree (`<parent>/edda-wt-gh<N>`, parallel to the main checkout, correct even when run from inside another worktree), and `lane-launch.ps1 -TimeoutSec 5400 -BudgetUsd 3`; relaunches pick the next free `-r<n>` suffix. `--dry-run` prints every command in the doneWhen order and creates, claims, and launches nothing.
- `scripts/fleet/next-review.sh <pr> [--shadow] [--operator-granted] [--dry-run]` — pins the full head SHA (40-hex validated), counts `## Code Review: Round` comments, exits 2 with the `needs-operator` label at the three-round cap without `--operator-granted`, builds the same brief `review-pr.sh --dry-run` writes, copies the base-SHA spec into a detached review worktree, runs the read-only glm engine (`edda dispatch --agent pi --tools read,grep,find,ls --thinking max --json`), refuses to post when the head moved between brief and post, and with `--shadow` posts the SHADOW shape (heading suffix, `shadow: true`, corrected cost line, no labels, no status). The engine output is stored next to the worktree for #887's compare.
- `docs/guides/pi-controller-runbook.md` — every step with its exact command and expected output, plus the STOP conditions that route to `needs-operator`.
- `scripts/fleet/test-next-loop.sh` — offline stub test (gh stub resolves `--jq` like the real gh), six cases.
- `docs/README.md` — one pointer line.

### RAN

[...]

PR #979, opened 2026-09-06
docs(fleet): pi-controller-runbook — a flash controller must route STOP to needs-operator, not fix the brief

fagemx:
## Problem and change

`docs/guides/pi-controller-runbook.md` (#899) instructed the controller, on `STOP step=<n>`, to fix the brief and relaunch — with no engine qualification. In the all-flash window the runbook's audience is glm-5.3-flash, so the current text instructed a flash controller to do the one function the 09-05 window measured as flash-unsafe: brief authoring (five STOPs, all brief-authoring defects, zero engine misjudgments; a flash re-author meets the original author's failure modes and has no validate-in-a-throwaway-worktree habit).

Changed behavior, per #933's doneWhen:

- **Watch the lane** — the `STOP step=<n>` bullet now routes by controller engine: an authoritative engine (Opus/sol) reads the five lines, fixes the brief, and relaunches (the previous text, preserved for the qualified case); a flash-level controller never modifies the brief — it labels `needs-operator` and stops, with the 09-05 measurement cited as the reason. The #930-validator relaxation stays out of scope as the issue states.
- **STOP conditions** — the routing line is added: a flash-level controller's `STOP step=<n>` on a lane routes to `needs-operator`, not a brief edit.

### RAN

- Authored-span dry-run: `sh scripts/fleet/brief-validate.sh <brief>` → **VALID** at the pinned base `586cb07e762170bd0768be8d66b4451b764dca45` (both edits matched exactly once in a throwaway worktree; all output expectations matched; the validator's measured stat corrected the author's line arithmetic once).
- At the delivery head: `grep -c "needs-operator"` = 7 (was 5; +1 per edit), `grep -c "flash-level"` = 2 (was 0), `sh scripts/lint-doc-citations.sh` exit 0, `git diff --check` clean.

### READ

- doneWhen item 3: `git diff --name-only` shows exactly `docs/guides/pi-controller-runbook.md` — no other file touched.
- doneWhen item 1 is the text itself (engine-routed STOP handling; flash → `needs-operator`, no brief modification); item 2 is the new STOP-conditions line.

Issue: #933

PR #980, opened 2026-09-06
fix(fleet): readiness follow-ups — shared PR limit, R24's third field, a jq-applying digest stub, a declared L0 gate marker

### LINKED ISSUE #958: fix(fleet): readiness-check follow-ups — drift/digest limit mismatch, mergeable field, digest-side fixture for the select-drop
Follow-ups from the #946 review (GH-914), none extends that PR:

1. **Limit mismatch:** `verdict-drift.sh` enumerates `--limit 200` while `daily-digest.sh` walks `--limit 100` — PR beyond 100 get drift lines but can never reach a digest row.
2. **R24 field (3):** rules.md R24 names three readiness fields; the drift check covers (1) newest-verdict-SHA == head and (2) the mergeStateStatus caveat, not (3) `mergeable != CONFLICTING` (left to the digest's DIRTY row today). Relevant if the drift check becomes the sole standalone gate.
3. **Digest-side regression fixture:** the existing digest stub bypasses `--jq`, so no fixture fails if a `mergeStateStatus` select were restored; a stub mode applying the caller's `--jq` (as `test-verdict-drift.sh` already does) or a case asserting a CLEAN drift-not-ready row would close it.
4. **Structured gate marker** for `review-l0.sh`'s $N-gate (today a textual `REVIEW_FILES` grep — correct for every current block, loud on failure, but a structured marker would remove the coupling); and the `oneline()` back-off comment could describe the boundary over-strip case it also handles.

fagemx:
Closes #958

Four follow-ups from the #946 review (GH-914). The issue carries no `doneWhen`
block — its four numbered items are the acceptance list, one section each.

## 1. Limit mismatch

`verdict-drift.sh` enumerated `--limit 200`, `daily-digest.sh` walked
`--limit 100`. A PR past the digest's 100 got a drift line that could never
become a digest row — invisible in the one artefact R24 calls *the report*.

Both now read `EDDA_OPEN_PR_LIMIT` (default 200), and the digest **exports the
value it used** to the drift subprocess, so a run is governed by one number
rather than two literals that agree today. A saturated enumeration is announced
on stderr instead of silently dropping the tail.

## 2. R24's third readiness field

R24 names three; the drift check covered (1) newest-verdict-SHA == head and
(2) the `mergeStateStatus` caveat. `mergeable != CONFLICTING` was left to the
digest's DIRTY row, so it was simply absent whenever the check ran standalone —
which is exactly the case the issue flags.

`CONFLICTING` now annotates the line and sets the not-ready exit. `UNKNOWN` is
annotated but does **not** hold the PR: GitHub has not computed the merge yet,
which is a not-yet answer, not a verdict. `MERGEABLE` and an absent field (an
older `gh`) add nothing to the line, so every existing output shape is
unchanged.

## 3. Digest-side regression fixture

The digest's `gh` stub `cat`ed TSV fixtures and never ran the script's own
`--jq`. Restoring a `select(.mergeStateStatus …)` to the open-PR query would
therefore have dropped PRs from the digest **with every fixture still green** —
the #914 defect exactly (the blocked-PR filter skipped a CLEAN PR, and #899 was
reported complete on a stale SHA).

The stub now applies the caller's `--jq` with real `jq` when the fixture is raw
JSON — the pattern `test-verdict-drift.sh` already used — and passes TSV
through untouched, so every older case reads unchanged. New **case 3c** walks a
CLEAN, drift-not-ready PR end to end and also asserts both enumerations request
the same `--limit`.

One thing worth flagging for review: the stub captures `--jq` from the real
`"$@"` at script top, not inside the emit function. A filter like
`.[] | [.number, .title] | @tsv` carries spaces, and re-splitting a flattened
`"$*"` keeps only its first word — which is how the first version of this stub
failed.

## 4. Structured gate marker, and the oneline() comment

`review-l0.sh` decided whether a block could run without a PR number by
`grep`ping its command text for the string `REVIEW_FILES`. Correct for every
block that exists, loud on failure — but it coupled the runner to one variable
name inside someone else's command, and any block mentioning it for another
reason would have been waved through.

The capability is now **declared on the block's own marker**:

```
# review-spec:check U1 no-pr-needed
```

Attributes are parsed by the same extractor pass, carried through
`@@BLOCK@@ <RULE> <n> <attrs>` and the block map, and passed to `run_block`.
They are **validated**: an unrecognised attribute exits 2, because a typo in
`no-pr-needed` would otherwise silently send U1/C5/R3 back to
`N.A.(needs PR number)` on every pre-push pass — a check quietly not running is
the failure mode this whole runner exists to prevent.

`oneline()`'s back-off comment now states the case it was silent about: the
pattern matches **any** trailing lead byte plus continuations, so when the byte
cap lands exactly on a character boundary it strips one **complete** character
rather than nothing. Accepted, with the reason recorded — the cell is truncated
evidence either way, one glyph changes no verdict, and decoding to count
characters is more machinery than a table cell is worth. What it must never do
is emit invalid UTF-8, and over-stripping cannot.

## Evidence — red before, green after

```
verdict-drift fixtures passed          (7 cases -> 11)
daily-digest fixtures passed           (case 3c added)
test-review-l0.sh: all fixture assertions held   (case 6 added, two assertions)
```

Each new assertion fails on this branch's base, isolated one at a time:

| probe | result on base |
|---|---|
| revert `daily-digest.sh` | `FAIL: case 3c: digest and verdict-drift enumerate different --limit values: 100 200` |
| add `select(.mergeStateStatus != "CLEAN")` to the digest's open query | `FAIL: case 3c: a CLEAN drift-not-ready PR is missing from 擋住什麼` — the point of item 3: with the old cat-the-TSV stub this injection changed nothing |
| drop `no-pr-needed` from U1's marker (body still says `REVIEW_FILES`) | U1 becomes `N.A.(needs PR number)`; under the old textual gate it still ran |
| misspell the attribute `no-pr-neded` | runner exits 2, `unknown review-spec:check attribute(s): U1:no-pr-neded` |
| drift cases 8/9 | `mergeable` was not requested at all before this change |

Also updated: `test-review-l0.sh` case 3 anchored `/^# review-spec:check U1$/`,
which the attribute broke; it now matches `U1( |$)`.

## Gates ran (docs + scripts only — no crate, `Cargo.*` or toolchain change, so no Cargo gate applies)

| gate | result |
|---|---|
| `sh scripts/fleet/test-verdict-drift.sh` | passed, 11/11 |
| `sh scripts/fleet/test-daily-digest.sh` | passed |
| `sh scripts/test-review-l0.sh` | all fixture assertions held |
| `bash scripts/lint-file-length.sh --tree` | pass |
| `sh scripts/lint-markdown-content.sh` | pass |
| `bash scripts/lint-doc-citations.sh --tree` | pass |
| `sh -n` on all five changed shell scripts | pass |
| `git diff --check` | clean |

SHA: `89d49dd1f4906c7ffe321f3bf1b65baa5330e9bb`

## Not extended here

- **`scripts/fleet/test-verdict-drift.sh` and `scripts/test-review-l0.sh` still
  have no CI consumer** — tracked on #927. This PR adds cases to suites nothing
  runs automatically.
- **Beyond `EDDA_OPEN_PR_LIMIT` the enumeration still truncates.** The stderr
  warning reaches a direct caller; it does not reach the operator's digest,
  which would want a 例外 row. Follow-up, not this issue's item 1 (the
  *mismatch*), and the digest surfacing is its own decision.

## Review surface

R22: `REVIEW.md`, `scripts/review-l0.sh` and `scripts/fleet/daily-digest.sh` are
all 審判面, so this PR needs an authoritative engine (Opus or sol), not a SHADOW
round.

### DIFF
--- a/REVIEW.md
+++ b/REVIEW.md
@@ -275,7 +275,7 @@ must hold and input shapes to confirm — never as an attack plan (decision
 **U1 — surface. P0.** Every changed file is inside the surface the issue or the
 lane brief allows. A file outside it is a lane-boundary violation.
 
-# review-spec:check U1
+# review-spec:check U1 no-pr-needed
 ```sh
 if [ -n "${REVIEW_FILES:-}" ]; then cat "$REVIEW_FILES"; else gh pr diff "$N" --name-only; fi
 ```
@@ -564,7 +564,7 @@ the ladder's own L2 command for that crate — `cargo test -p <crate>` on
 Windows — not a new gate invented here (`ladder`). One focused check per
 uncovered crate; never the whole workspace to reach one.
 
-# review-spec:check C5
+# review-spec:check C5 no-pr-needed
 ```sh
 { if [ -n "${REVIEW_FILES:-}" ]; then cat "$REVIEW_FILES"; else gh pr diff "$N" --name-only; fi; } \
   | grep '^crates/' | cut -d/ -f2 | sort -u \
@@ -609,7 +609,7 @@ git diff "origin/$BASE..$SHA" --unified=0 -- '*.sh' '*.bash' '*.ps1' '.github' \
 
 **R3 — every changed shell script parses. P0.** Exit code is the signal.
 
-# review-spec:check R3
+# review-spec:check R3 no-pr-needed
 ```sh
 for f in $(if [ -n "${REVIEW_FILES:-}" ]; then cat "$REVIEW_FILES"; else gh pr diff "$N" --name-only; fi | grep -E '\.(sh|bash)$'); do
   [ -f "$f" ] && { sh -n "$f"; echo "$f -> sh -n exit=$?"; }
--- a/scripts/fleet/daily-digest.sh
+++ b/scripts/fleet/daily-digest.sh
@@ -144,7 +144,13 @@ blocked_body="$tmp/blocked-body.md"
 # CLEAN with zero verdicts and must still appear. BLOCKED/DIRTY PRs keep the
 # status-reasons treatment below; every other state is reachable only through
 # its drift reason.
-open_rows=$(gh pr list --repo "$EDDA_REPO" --state open --limit 100 \
+# GH-958: the same enumeration limit verdict-drift.sh uses, from the same
+# variable. They were 100 here and 200 there, so PRs 101-200 got a drift
+# line that could never reach a digest row — invisible in the one artefact
+# R24 calls the report. It is exported below so the drift subprocess reads
+# the value this run actually used, not its own default.
+open_pr_limit=${EDDA_OPEN_PR_LIMIT:-200}
+open_rows=$(gh pr list --repo "$EDDA_REPO" --state open --limit "$open_pr_limit" \
     --json number,title,mergeStateStatus,headRefOid \
     --jq '.[] | [.number, .title, .mergeStateStatus, .headRefOid] | @tsv' \
 ) || { printf '%s: gh pr list (open) failed\n' "$prog" >&2; exit 1; }
@@ -159,7 +165,7 @@ open_rows=$(gh pr list --repo "$EDDA_REPO" --state open --limit 100 \
 # post-jq `gh pr list` output is not re-parsed as raw JSON by the drift check.
 drift_out="$tmp/drift.txt"
 drift_rc=0
-GH_OPEN_JSON= sh "$self_dir/verdict-drift.sh" >"$drift_out" 2>"$tmp/drift.err" || drift_rc=$?
+GH_OPEN_JSON= EDDA_OPEN_PR_LIMIT="$open_pr_limit" sh "$self_dir/verdict-drift.sh" >"$drift_out" 2>"$tmp/drift.err" || drift_rc=$?
 if [ "$drift_rc" -ge 2 ]; then
     printf '%s: verdict-drift.sh failed (exit %s)\n' "$prog" "$drift_rc" >&2
     cat "$tmp/drift.err" >&2
--- a/scripts/fleet/test-daily-digest.sh
+++ b/scripts/fleet/test-daily-digest.sh
@@ -33,18 +33,49 @@ cat >"$STUBBIN/gh" <<'EOF'
 #!/bin/sh
 echo "gh $*" >>"$GH_STUB_LOG"
 echo "gh $*" >>"$ORDER_LOG"
+# GH-958: apply the caller's --jq when the fixture is raw JSON, the way
+# scripts/fleet/test-verdict-drift.sh already does. With the old cat-the-TSV
+# stub the script's own --jq never ran, so restoring a
+# `select(.mergeStateStatus …)` to the open-PR query would have dropped PRs
+# from the digest with every fixture still green — the #914 defect exactly.
+# TSV fixtures are passed through untouched, so the older cases still read.
+# The --jq argument is captured HERE, from the real argv: a filter like
+# '.[] | [.number, .title] | @tsv' carries spaces, and re-splitting a
+# flattened "$*" inside the function would keep only its first word.
+gh_jq=
+gh_prev=
+for a in "$@"; do
+    [ "$gh_prev" = "--jq" ] && gh_jq=$a
+    gh_prev=$a
+done
+emit() { # <fixture file>
+    [ -n "${1:-}" ] && [ -f "$1" ] || return 0
+    case "$(head -c 1 "$1")" in
+        '['|'{')
+            if [ -n "$gh_jq" ]; then jq -r "$gh_jq" <"$1"; else cat "$1"; fi ;;
+        *) cat "$1" ;;
+    esac
+}
 case "$1" in
   pr)
     case "$2" in
       list)
         case "$*" in
-          *"--state merged"*) [ -n "${GH_MERGED_JSON:-}" ] && cat "$GH_MERGED_JSON" ; exit 0 ;;
-          *"--state open"*)   [ -n "${GH_OPEN_JSON:-}" ] && cat "$GH_OPEN_JSON" ; exit 0 ;;
+          *"--state merged"*) emit "${GH_MERGED_JSON:-}" ; exit 0 ;;
+          # The digest clears GH_OPEN_JSON for the verdict-drift subprocess,
+          # so an empty one means this call came from the drift check.
+          *"--state open"*)
+            if [ -n "${GH_OPEN_JSON:-}" ]; then emit "$GH_OPEN_JSON"
+            else emit "${GH_DRIFT_OPEN_JSON:-}"
+            fi
+            exit 0 ;;
         esac
         exit 0
         ;;
       view)
-        [ -n "${GH_COMMENTS_FILE:-}" ] && cat "$GH_COMMENTS_FILE"
+        if [ -n "${GH_OPEN_JSON:-}" ]; then emit "${GH_COMMENTS_FILE:-}"
+        else emit "${GH_DRIFT_COMMENTS_JSON:-${GH_COMMENTS_FILE:-}}"
+        fi
         exit 0
         ;;
     esac
@@ -100,7 +131,8 @@ export PATH="$STUBBIN:$PATH"
 reset_stubs() {
     : >"$GH_STUB_LOG"; : >"$EDDA_STUB_LOG"; : >"$ORDER_LOG"
     unset GH_MERGED_JSON GH_OPEN_JSON GH_STATUS_JSON GH_CHECK_RUNS_JSON GH_COMMENTS_FILE \
-          GH_BOARD_FILE GH_READY_JSON EDDA_RECAP_FILE 2>/dev/null || true
+          GH_BOARD_FILE GH_READY_JSON EDDA_RECAP_FILE \
+          GH_DRIFT_OPEN_JSON GH_DRIFT_COMMENTS_JSON 2>/dev/null || true
 }
 
 # canned recap digest (the three ledger blocks, exactly as edda recap --digest prints)
@@ -184,6 +216,8 @@ awk '/^## 擋住什麼/{f=1;next} /^## /{f=0} f' "$tmp/case2.out" | grep -qxF '�
 printf '%s\n' "$out" | grep -qF -- '- 審查（gh verdict 留言 Cost:）：n/a' \
     || fail 'case 2: review cost line should be n/a'
 
+SHA79=7979797979797979797979797979797979797979
+
 # --- case 3: one BLOCKED PR with a failing Independent Review ------------------
 reset_stubs
 printf '77\tBlocked thing\tBLOCKED\tabc123\n' >"$tmp/open.tsv"
@@ -210,6 +244,33 @@ export EDDA_RECAP_FILE="$RECAP_CANNED"
 out=$(run_digest --dry-run 2>&1) || fail 'case 3b: daily-digest.sh exited non-zero'
 printf '%s\n' "$out" | grep -qF -- '- #78 Check run thing — Independent Review=failure' \
     || fail "case 3b: check-run CI Gate should be accepted, got: $(printf '%s' "$out" | grep '#78' || true)"
+
+# --- case 3c: a CLEAN PR that the drift check calls not-ready still gets a row -
+# GH-958. Two things are proven only with the jq-applying stub above:
+#   * the open-PR query's own --jq runs, so restoring a
+#     `select(.mergeStateStatus …)` would drop #79 and fail here (the #914
+#     defect: the blocked-PR filter skipped a CLEAN PR and #899 was reported
+#     complete on a stale SHA);
+#   * both enumerations use the same --limit, so no PR gets a drift line it
+#     can never turn into a digest row.
+reset_stubs
+printf '[{"number":79,"title":"Clean but unreviewed","mergeStateStatus":"CLEAN","headRefOid":"%s"}]\n' "$SHA79" >"$tmp/open-clean.json"
+printf '[{"number":79,"headRefOid":"%s","baseRefName":"main","mergeable":"MERGEABLE"}]\n' "$SHA79" >"$tmp/drift-open.json"
+printf '{"comments":[]}\n' >"$tmp/drift-comments.json"
+export GH_OPEN_JSON="$tmp/open-clean.json"
+export GH_DRIFT_OPEN_JSON="$tmp/drift-open.json"
+export GH_DRIFT_COMMENTS_JSON="$tmp/drift-comments.json"
+export EDDA_RECAP_FILE="$RECAP_CANNED"
+out=$(run_digest --dry-run 2>&1) || fail 'case 3c: daily-digest.sh exited non-zero'
+printf '%s\n' "$out" | grep -qF -- '- #79 Clean but unreviewed — no verdict on head' \
+    || fail "case 3c: a CLEAN drift-not-ready PR is missing from 擋住什麼, got: $(printf '%s' "$out" | grep '#79' || true)"
+printf '%s\n' "$out" | awk '/^## 擋住什麼/{f=1;next} /^## /{f=0} f' | grep -qF '#79' \
+    || fail 'case 3c: the #79 row is not inside the 擋住什麼 section'
+# both enumerations request the same page size — the mismatch GH-958 names
+limits=$(grep -- '--state open' "$GH_STUB_LOG" | sed -n 's/.*--limit \([0-9]*\).*/\1/p' | sort -u)
+[ "$(printf '%s\n' "$limits" | grep -c .)" = 1 ] \
+    || fail "case 3c: digest and verdict-drift enumerate different --limit values: $(printf '%s' "$limits" | tr '\n' ' ')"
+unset GH_DRIFT_OPEN_JSON GH_DRIFT_COMMENTS_JSON
 
 # --- case 4: board comment with needs-operator lands under 例外 ----------------
 reset_stubs
--- a/scripts/fleet/test-verdict-drift.sh
+++ b/scripts/fleet/test-verdict-drift.sh
@@ -26,6 +26,7 @@ trap 'rm -rf "$tmp"' 0 HUP INT TERM
 mkdir -p "$tmp/bin"
 cat >"$tmp/bin/gh" <<'EOF'
 #!/bin/sh
+[ -n "${GH_ARGV_LOG:-}" ] && echo "gh $*" >>"$GH_ARGV_LOG"
 if [ -n "${GH_FAIL:-}" ]; then
     echo "gh: stub failure" >&2
     exit 1
@@ -137,5 +138,50 @@ unset GH_FAIL
 grep -q 'could not read' "$tmp/err" ||
     fail 7 "stderr must say the read failed, got: $(cat "$tmp/err")"
 echo "PASS 7"
+
+# --- case 8: mergeable CONFLICTING holds the PR (R24 field 3, GH-958) ----------
+# The head carries an authoritative LGTM, so every field the check covered
+# before says ready. R24's third field does not, and used to be left to the
+# digest's DIRTY row — invisible whenever this check ran standalone.
+printf '[{"number":8,"headRefOid":"%s","baseRefName":"main","mergeable":"CONFLICTING"}]\n' "$SHA1" >"$tmp/prs.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #8 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA1" >"$tmp/comments-8.json"
+run_drift
+expect 8 1 "#8 111111111111 main LGTM mergeable=CONFLICTING"
+
+# --- case 9: mergeable UNKNOWN is surfaced but does not hold the PR ------------
+# GitHub has not computed the merge yet; that is a transient answer, not a
+# verdict, so it is annotated and the exit code stays 0.
+printf '[{"number":9,"headRefOid":"%s","baseRefName":"main","mergeable":"UNKNOWN"}]\n' "$SHA2" >"$tmp/prs.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #9 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA2" >"$tmp/comments-9.json"
+run_drift
+expect 9 0 "#9 222222222222 main LGTM mergeable=UNKNOWN"
+
+# --- case 10: the enumeration limit is EDDA_OPEN_PR_LIMIT, shared with the -----
+# digest (GH-958). The two scripts hardcoded 200 and 100, so a PR past the
+# digest's 100 got a line here that could never reach a digest row.
+printf '[{"number":10,"headRefOid":"%s","baseRefName":"main","mergeable":"MERGEABLE"}]\n' "$SHA3" >"$tmp/prs.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #10 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA3" >"$tmp/comments-10.json"
+: >"$tmp/gh-argv.log"
+GH_ARGV_LOG="$tmp/gh-argv.log"
+export GH_ARGV_LOG
+EDDA_OPEN_PR_LIMIT=7
+export EDDA_OPEN_PR_LIMIT
+run_drift
+unset EDDA_OPEN_PR_LIMIT
+grep -q -- '--limit 7' "$tmp/gh-argv.log" ||
+    fail 10 "the pr list call ignored EDDA_OPEN_PR_LIMIT: $(cat "$tmp/gh-argv.log")"
+[ "$rc" = 0 ] || fail 10 "exit $rc, expected 0 (stderr: $(cat "$tmp/err"))"
+echo "PASS 10"
+
+# --- case 11: a saturated enumeration says so on stderr instead of dropping ----
+# the tail in silence.
+EDDA_OPEN_PR_LIMIT=1
+export EDDA_OPEN_PR_LIMIT
+run_drift
+unset EDDA_OPEN_PR_LIMIT
+grep -q 'hit its limit' "$tmp/err" ||
+    fail 11 "a saturated enumeration printed no warning: $(cat "$tmp/err")"
+echo "PASS 11"
+unset GH_ARGV_LOG
 
 echo "verdict-drift fixtures passed"
--- a/scripts/fleet/verdict-drift.sh
+++ b/scripts/fleet/verdict-drift.sh
@@ -7,19 +7,36 @@
 # reports CLEAN with zero verdicts.
 #
 # Output, one line per open PR:
-#   #<n> <head12> <base> <state> [ base=<branch> (status contexts not enforced) ]
+#   #<n> <head12> <base> <state> [ mergeable=<v> ] [ base=<branch> (...) ]
 # with <state> one of:
 #   no verdict on head | stale from <sha12> | SHADOW only | LGTM | Changes Requested
 # The base annotation is appended when the PR's base is not `main`, whose
 # status contexts are the only ones any ruleset enforces.
 #
-# Exit 0 every PR carries a verdict on its head; exit 1 when any PR has
-# `no verdict on head` or is `stale from ...`; exit 2 when a gh read fails —
-# a check that could not read must never print nothing and exit 0.
+# GH-958: R24 names THREE readiness fields, and this check now covers all
+# three — (1) the newest verdict's SHA equals the head, (2) the
+# mergeStateStatus caveat above, and (3) `mergeable` is not CONFLICTING,
+# which used to be left to the digest's DIRTY row and so went missing
+# whenever this check ran standalone. CONFLICTING is a not-ready state;
+# UNKNOWN (GitHub has not computed the merge yet) is annotated but does not
+# hold the PR, because it is a transient answer, not a verdict.
+#
+# The open-PR enumeration limit is shared with daily-digest.sh through
+# EDDA_OPEN_PR_LIMIT (GH-958): the two used 200 and 100, so a PR past the
+# digest's 100 got a drift line here that could never reach a digest row —
+# invisible in the one artefact R24 calls the report.
+#
+# Exit 0 every PR carries a verdict on its head and is not CONFLICTING;
+# exit 1 when any PR has `no verdict on head`, is `stale from ...`, or is
+# CONFLICTING; exit 2 when a gh read fails — a check that could not read
+# must never print nothing and exit 0.
 # Read-only: no posting, no labels, no merges, no ledger writes.
 set -eu
 
 repo=${EDDA_REPO:-fagemx/edda}
+# Shared with daily-digest.sh; see the note above. A saturated enumeration
+# is announced on stderr rather than silently dropping the tail.
+open_limit=${EDDA_OPEN_PR_LIMIT:-200}
 
 fail_read() {
     echo "verdict-drift: could not read PR state from gh ($1) — refusing to print a clean bill" >&2
@@ -29,13 +46,16 @@ fail_read() {
 # The R23 verdict heading, canonical REVIEW.md §7 shape: the ` (SHADOW)`
 # suffix follows `Round <N>`. The trailing-suffix variant is also accepted so
 # the issue #914 spelling of the pattern matches the same comments.
-open_rows=$(gh pr list --repo "$repo" --state open --limit 200 \
-    --json number,headRefOid,baseRefName \
-    --jq '.[] | [.number, .headRefOid, .baseRefName] | @tsv' \
+open_rows=$(gh pr list --repo "$repo" --state open --limit "$open_limit" \
+    --json number,headRefOid,baseRefName,mergeable \
+    --jq '.[] | [.number, .headRefOid, .baseRefName, .mergeable] | @tsv' \
 ) || fail_read "pr list"
+if [ "$(printf %s "$open_rows" | grep -c .)" -ge "$open_limit" ]; then
+    echo "verdict-drift: the open-PR enumeration hit its limit of $open_limit; PRs past it were not examined (raise EDDA_OPEN_PR_LIMIT)" >&2
+fi
 
 not_ready=0
-while IFS="$(printf '\t')" read -r num head base; do
+while IFS="$(printf '\t')" read -r num head base mergeable; do
     [ -n "$num" ] || continue
     verdicts=$(gh pr view "$num" --repo "$repo" --json comments --jq '
         .comments[]
@@ -86,7 +106,16 @@ EOF
         state="Changes Requested"
     fi
 
+    # R24 field (3). CONFLICTING holds the PR; UNKNOWN is surfaced without
+    # holding it; MERGEABLE and an empty value (an older gh without the
+    # field) add nothing to the line.
     line="#$num $head12 $base $state"
+    if [ "$mergeable" = "CONFLICTING" ]; then
+        line="$line mergeable=CONFLICTING"
+        not_ready=1
+    elif [ "$mergeable" = "UNKNOWN" ]; then
+        line="$line mergeable=UNKNOWN"
+    fi
     if [ "$base" != "main" ]; then
         line="$line base=$base (status contexts not enforced)"
     fi
--- a/scripts/fleet/verdict-drift.sh
+++ b/scripts/fleet/verdict-drift.sh
@@ -107,8 +107,10 @@ EOF
     fi
 
     # R24 field (3). CONFLICTING holds the PR; UNKNOWN is surfaced without
-    # holding it; MERGEABLE and an empty value (an older gh without the
-    # field) add nothing to the line.
+    # holding it; MERGEABLE and an empty value add nothing to the line. An
+    # empty value means the object carried no mergeable key at all, as the
+    # older fixtures do — NOT an older gh, which rejects an unknown --json
+    # field before the request and so lands in fail_read instead.
     line="#$num $head12 $base $state"
     if [ "$mergeable" = "CONFLICTING" ]; then
         line="$line mergeable=CONFLICTING"
--- a/scripts/review-l0.sh
+++ b/scripts/review-l0.sh
@@ -29,9 +29,11 @@
 #            committed range — the pre-push shape of Example B step 20.
 #   [PR-number]  blocks referencing "$N" (U2, U3, U6) run only when a PR
 #            number is passed; without one they report N.A.(needs PR number)
-#            instead of failing. U1, C5 and R3 enumerate the changed files
-#            from REVIEW_FILES (GH-922 d-004 option A), so the pre-push pass
-#            — no PR number — runs them too. Offline callers can still fake
+#            instead of failing. A block whose marker declares
+#            `no-pr-needed` (U1, C5, R3) runs anyway: it enumerates the
+#            changed files from REVIEW_FILES (GH-922 d-004 option A) and
+#            reaches the "$N" only in a fallback branch that stays
+#            unreached, so the pre-push pass runs it too. Offline callers can still fake
 #            the PR surface by putting a stub gh earlier in PATH —
 #            scripts/test-review-l0.sh does exactly that.
 #   REVIEW_L0_SPEC  spec file to extract from (default REVIEW.md in the cwd —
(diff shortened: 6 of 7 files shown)


fagemx:
## Gate receipt

`CI run 34004736871 @ 89d49dd1f490` — success (Format; Clippy ×3 OS; Test on ubuntu/macos/windows; MSRV 1.91.0; CI Gate).

Local, same SHA, on Windows 11 / Git Bash:
- `sh scripts/fleet/test-verdict-drift.sh` — passed, 11/11
- `sh scripts/fleet/test-daily-digest.sh` — passed
- `sh scripts/test-review-l0.sh` — all fixture assertions held
- `bash scripts/lint-file-length.sh --tree`, `sh scripts/lint-markdown-content.sh`, `bash scripts/lint-doc-citations.sh --tree`, `git diff --check` — clean

Build lane: n/a — no crate, `Cargo.*`, or toolchain change, so no Cargo gate applies and nothing was compiled.

Follow-up opened from this PR's findings: #986.

fagemx:
## Review Response: Round 1

New head: `5252f12008a691b35b1c052d84fe121380db4a69`

### f3 (P2) — fixed

Correct, and the comment was mine. `verdict-drift.sh:110-111` attributed the empty-`mergeable` branch to "an older gh without the field". `gh` validates `--json` field names before it issues the request, so an older `gh` fails the read and lands in `fail_read` (exit 2) — it never yields an empty value. The empty value comes from an object carrying no `mergeable` key at all, which is exactly what fixtures 1-7 build. The branch was right; only its stated cause was false. Comment corrected.

### f1, f2, f4 — acknowledged, routed as follow-ups

- **f1** — the saturation warning is on stderr, captured to `drift.err` and printed only when `drift_rc >= 2`, but it is emitted at rc 0/1. So on the digest path it is swallowed. The reviewer's own note is the reason this is not blocking: silent truncation *predates* this diff (the limit was 100 with no warning at all), this PR raises and aligns the limit, and the warning does reach its mandated standalone consumer. Already filed as #986.
- **f2** — `drift_map`'s awk selects on `$4 in {no, stale, SHADOW}`, so a line whose only not-ready signal is the new `mergeable=CONFLICTING` suffix produces no digest row; today the coverage holds only through the digest's independent `mergeStateStatus == DIRTY` branch, which is a separate `gh` read of a separate field at a different moment. Fair, and a genuinely better catch than #986 — the R24 field is wired into the drift check but not into the digest's row selector.
- **f4** — none of the three changed fixture suites is referenced by `.github/workflows/` or `lefthook.yml`. Pre-existing for all three, tracked by #927.

### Escalation that cannot be cleared by any fix

Round 1 carried `REVIEW.md changed in this diff — reviewed under the base version`. That escalation is **mandatory** for any diff touching `REVIEW.md` (§6.6) and is not a defect signal; it cannot be removed by changing this PR. It needs adjudication by a qualified engine, or the operator's call.

### Gates

`CI run 34015649019 @ 5252f12008a691b35b1c052d84fe121380db4a69` — all 12 checks SUCCESS.

L0 (this head, Windows):
- `sh -n scripts/fleet/verdict-drift.sh` → exit 0
- `sh scripts/fleet/test-verdict-drift.sh` → 11 PASS, `verdict-drift fixtures passed`

No behaviour changed at this head — the diff is one corrected comment.

fagemx:
## Code Review: Round 2 — PR #980 @ 5252f12008a691b35b1c052d84fe121380db4a69

- elapsed: 174 s (edda review session)
- model_observed: claude-opus-5 (in-band)
- reviewer_session: fc80988a-84d4-5456-8742-099ae2d0de4d
- spec: review-spec-v1.5
- class: code-risk  (REVIEW.md classes: docs-skills, code-risk)
- escalations: D5; REVIEW.md changed in this diff (§6.6); no host-executed gate evidence; the `gh` `--json` claim unprobed — all adjudicated below
- shadow: false
- cost: $2.2882 / 174 s / tool_policy=hard / independence=session (same-model)

### IN SCOPE
Delta round over `89d49dd1..5252f12` (7 files, 292 lines). The delta is comment-only at `verdict-drift.sh:109-113`.

### Rules
|---|---|---|---|---|
| Prior f3 resolved | D5 | P2 | PASS | `verdict-drift.sh:110-113` now states the no-key cause and explicitly rules out the older-gh cause |
| Prior f1, f2, f4 | — | P2 | open, routed | `daily-digest.sh:168-173` and `:177-184` unchanged in the delta — follow-ups #986 / #927, not PR extensions |
| U1 surface | U1 | P0 | PASS | all 7 changed files inside the four GH-958 follow-up items |
| U4 conventional commits | U4 | P1 | PASS | both subjects across `586cb07..5252f12` conform |
| U6 gates READ before RAN | U6 | P0 | PASS | CI Gate green; four Cargo gates `not covered` — correct for a head touching no `crates/**`, `Cargo.toml` or toolchain |
| doneWhen 1 — shared open-PR limit | C1 | P0 | PASS | `verdict-drift.sh:39,49` and `daily-digest.sh:152,153,168`; assertions at `test-verdict-drift.sh:171`, `test-daily-digest.sh:269-272` |
| doneWhen 2 — R24 field (3) | C1 | P0 | PASS | `verdict-drift.sh:114-120` against `docs/fleet/rules.md:32`; cases at `test-verdict-drift.sh:142-157` |
| doneWhen 3 — digest stub applies `--jq` | C1 | P0 | PASS | `test-daily-digest.sh:45-58` and `:248-273` (case 3c) |
| doneWhen 4 — `no-pr-needed` marker attribute | C1 | P0 | PASS | `review-l0.sh:114-118,147,153-162,291-302,397-398`; `REVIEW.md:278,567,612`; `test-review-l0.sh:171,219-250` |
| R2 shell operator precedence | R2 | P0/P1 | PASS | delta adds no mixed line; `test-daily-digest.sh:52` is `(A&&B)||C`, no row runs an R1-list command |
| R3 changed scripts parse | R3 | P0 | N.A.(no execution capability) | delta adds only `#`-prefixed comment lines — see adjudication |
| §5.5 Wiring | 5.5 | P1 | PASS | delta adds no new surface; wiring-scan over `586cb07..5252f12` exit 0, no new pub surfaces |
| D5 wording and structure | D5 | 判斷 | 需升級 | adjudicated below |

### Wiring
|---|---|---|---|---|
| no new surfaces in this delta | — | — | — | Round 1 table re-affirmed |

### Findings
- [P2] `daily-digest.sh:168` — the saturated-enumeration warning is swallowed on the digest path. Carried from Round 1, untouched by this push, pre-existing (base truncated silently at 100 with no warning at all). Routed: #986.
- [P2] `daily-digest.sh:177` — `drift_map`'s awk selects on `$4 in {no, stale, SHADOW}`, so a line whose only not-ready signal is `mergeable=CONFLICTING` produces no digest row; coverage holds today only via the independent `mergeStateStatus == DIRTY` branch. Routed: #986.
- [P2] `test-verdict-drift.sh:1` — none of the three changed fixture suites is referenced by `.github/workflows/` or `lefthook.yml`. Pre-existing for all three. Routed: #927.

P0=0, P1=0.

### FOLLOW-UP ISSUE
#986 (digest saturation + CONFLICTING row selection), #927 (fixture suites in no CI job).

### RAN vs READ
RAN:  nothing — read-only reviewer, `tool_policy=hard`.
READ: `CI Gate` green at the reviewed SHA. Implementer receipt: `CI run 34015649019 @ 5252f12008a691b35b1c052d84fe121380db4a69`, all 12 checks SUCCESS.

### Verdict
LGTM (P0=0, P1=0) — provisional under §6.4 pending the adjudication below.

fagemx:
## Escalation Adjudication — PR #980 @ 5252f12008a691b35b1c052d84fe121380db4a69

Round 2 left four items open. Adjudicated here by `claude-opus-5` via Claude Code — R22 makes it authoritative on all four surfaces, and this PR touches `REVIEW.md`, `scripts/review-l0.sh` and `scripts/fleet/daily-digest.sh`, three 審判面 paths. Separate session from the author and from the round reviewer. These are closed, not re-tagged provisional.

| Item | Verdict |
|---|---|
| D5 — wording and structure | **PASS** |
| §6.6 — `REVIEW.md` changed in this diff | **PASS** — the diff cannot weaken or quietly rewrite the rules judging it |
| `gh` unknown-`--json`-field claim | **PASS** — claim supported |
| No host-executed gate evidence | **PASS** — gates executed, see RAN below |

### §6.6 — the substantive one

The risk that clause exists to surface is closed on five independent grounds:

1. **The spec change is inert as prose.** Stripping ` no-pr-needed` from the head `REVIEW.md` yields a file **byte-identical to base**. No rule text, severity, check-command body, or §6/§8/§9 row changed — only three marker comments.
2. **The attribute can only *add* coverage.** Its sole consumer is `review-l0.sh:294-298`: when a block body contains `$N` and `$N` is empty, `no-pr-needed` lets it run instead of returning `N.A.(needs PR number)`. No path lets an attribute suppress a rule, and `review-l0.sh:401` prints `N.A.(no command block in the spec)` for a routed rule with no block — a rule cannot vanish from the table.
3. **Exact parity with the base gate.** Base decided the same thing by grepping the block body for the literal `REVIEW_FILES` (`origin/main:scripts/review-l0.sh:254`). Enumerating every fenced block in the head spec, the ones containing `REVIEW_FILES` are exactly `{U1, C5, R3}` — precisely the set that gained the attribute. Same three rules run pre-push before and after; `U2/U3/U6` stay `N.A.` before and after.
4. **Inert for this very review.** With `$N` set, the attribute changes nothing — PR #980 is judged identically under base and head.
5. **An unknown attribute fails loudly and early.** `review-l0.sh:153-162` validates against the one allowed value and exits 2 on stderr *before* the classifier or any block runs. Measured: `U1 no-pr-neded` → exit 2, `unknown review-spec:check attribute(s): U1:no-pr-neded`; `C5 no-pr-needed bogus` → exit 2, `C5:bogus`.

It is net-tightening: it removes the runner's coupling to a variable name inside someone else's command text, under which a block mentioning `REVIEW_FILES` for an unrelated reason would have been waved through the `$N` gate.

*Residual, pre-existing, not a finding:* a typo in the rule **id** (`# review-spec:check U1x`) is still unvalidated; the rule then reports `N.A.(no command block in the spec)`, visible in the table. Base behaves identically.

### The `gh` claim

Re-probed on gh 2.87.0. `--json number,bogusfield` → **exit 1, stdout empty**, stderr `Unknown JSON field: "bogusfield"` plus gh's compiled-in `Available fields:` enumeration — i.e. client-side validation, the "before the request" mechanism the comment names. `--json number,mergeable` → exit 0 with real data. Wiring: `verdict-drift.sh:49-52` ends `) || fail_read "pr list"`, and `fail_read` (`:41-44`) exits 2 refusing to print a clean bill — so a gh rejecting `mergeable` lands in `fail_read`, never in a silently empty value, exactly as claimed. The other half holds too: a missing key yields an empty field rather than an error (measured through `jq … | @tsv`), and the pre-GH-958 fixtures at `test-verdict-drift.sh:94-131` carry no `mergeable` key while `:146-162` do.

*Honest caveat:* no older gh exists on this host to probe. What is measured is the version-independent mechanism the claim rests on. Under D1's routing that is a probed runtime claim with its exit code reported.

### D5

Prose checked: `review-l0.sh:30-38`, `:109-113`, `:150-152`, `:283-289`, the `oneline()` back-off note; `verdict-drift.sh:16-32` and `:109-113`; `daily-digest.sh:144-151`; the test comments. Each names the defect it prevents and cites the owning issue; the three mentions of the attribute sit at three distinct mechanisms rather than restating one another. `verdict-drift.sh` cites rules.md R23/R24 as the source and states only the local consequence — a pointer, not a duplicated rule, so no S3-class finding.

- **[P2]** `review-l0.sh:36` runs to 88 columns while the surrounding block wraps at ~76, orphaning `Offline callers can still fake` mid-sentence. Cosmetic; does not change the verdict and does not warrant its own issue. §8: P2 alone never blocks.

### RAN vs READ

**RAN** (Windows 11 / Git Bash, worktree at `5252f12`):
- strip-and-diff of `REVIEW.md` head vs base → byte-identical outside the three markers
- awk enumeration of every fenced block for `$N` / `REVIEW_FILES` → `{U1, C5, R3}` exactly
- `sh scripts/test-review-l0.sh` → exit 0, all assertions held, including *"every routed rule printed a row"*, *"U1/C5/R3 ran without a PR"*, *"an unknown marker attribute exits 2"*, *"the gate reads the marker, not the command text"*
- live pre-push pass on the real spec with no PR number → `U1 … PASS`, `C5 … PASS`
- `sh scripts/fleet/test-verdict-drift.sh` → PASS 1-11, exit 0; `sh scripts/fleet/test-daily-digest.sh` → passed, exit 0
- the gh and jq probes above

**READ:** exact-head CI `CI run 34015649019 @ 5252f12`, all green.

**Coverage note worth keeping:** `scripts/fleet/run-fleet-tests.sh` globs only `scripts/fleet/test-*.sh` and explicitly excludes `scripts/test-*.sh`, so **`scripts/test-review-l0.sh` is run by no CI job** (#927). The local run of it above is therefore load-bearing evidence for the §6.6 item, not a redundant rerun of a recorded gate.

**`escalations:` is now empty. The round-2 LGTM is no longer provisional.**

fagemx:
## Escalation Adjudication — addendum — PR #980 @ 5252f12008a691b35b1c052d84fe121380db4a69

Two additions from the adjudicator's background pre-push pass. **Verdicts unchanged: all PASS, P0=0, P1=0.**

**§6.6 evidence is now direct rather than fixture-only.** A full-table `review-l0` run against the real head `REVIEW.md` with **no PR number** produced exactly the base split:

```
| U1 | any        | P0 | PASS | exit=0; .github/workflows/ci.yml
| C5 | code-plain | P1 | PASS | exit=1; (no output)
| R3 | code-risk  | P0 | PASS | exit=0; scripts/calibrate-canaries.sh -> sh -n exit=0
| U2 | any        | P1 | N.A.(needs PR number)
| U3 | any        | P1 | N.A.(needs PR number)
| U6 | any        | P0 | N.A.(needs PR number)
```

`{U1, C5, R3}` run and `{U2, U3, U6}` stay `N.A.` — coverage parity measured on the real spec, not only through fixtures.

**One R2 candidate, adjudicated rather than suppressed.** The run flagged `test-daily-digest.sh:52` — `[ -n "${1:-}" ] && [ -f "$1" ] || return 0`, an added line of this PR. R2 requires a parse tree and truth table from every engine, and suppressing a candidate is itself a P1, so:

Parse (equal precedence, left-associative): `( ( [ -n "${1:-}" ] && [ -f "$1" ] ) || return 0 )`

| A `-n "${1:-}"` | B `-f "$1"` | A&&B | `return 0` runs? | Intended? |
|---|---|---|---|---|
| false | not evaluated | false | yes | ✓ empty arg → skip |
| true | false | false | yes | ✓ non-file → skip |
| true | true | true | no | ✓ falls through to the `case` |

No row runs a command the line's intent does not want, and the only command on the `||` branch is the builtin `return 0` — nothing from the R1 list. The §5.4 severity lookup yields neither P0 nor P1. This is the `[ … ] && [ … ] || die` guard shape §9's R2 provenance row names as the dominant benign candidate class (10 of 16 on its pinned range); the enumerator's `FAIL` is its conservative default, not a verdict.

**Range caveat, for the record:** `origin/main` has advanced past this PR's merge base, so the adjudicator's `origin/main..HEAD` range pulled in unrelated drift — which is why U1's evidence cell names `.github/workflows/ci.yml`. The R2 candidate is unaffected: it is present in `gh pr diff 980` as an added line of this PR.

fagemx:
## Code Review: Round 2 — PR #980 @ 5252f12008a691b35b1c052d84fe121380db4a69

- elapsed: unmeasured (no pi session file for this lane)
- model_requested: unrecorded (in-process review lane; no dispatch record reached this session)
- model_observed: unverified (no system-readable `model` / `modelUsage` available here; §6 forbids copying `model_requested`)
- reviewer_session: unassigned (no per-PR UUID supplied to this lane)
- spec: review-spec-v1.5, read at the base SHA `586cb07e762170bd0768be8d66b4451b764dca45` (§"Canonical sources"). `REVIEW.md` is changed in this diff, so §6.6's escalation is carried below.
- class: code-risk  (REVIEW.md classes: `code-risk skills` — §3 classifier run on the 7 changed files)
- escalations: (1) `REVIEW.md changed in this diff` — §6.6, mandatory, not clearable by any fix; (2) `D5` adjudicated below with derivation and tagged `provisional` per §6.1
- shadow: false
- cost: one review session, ~50 tool calls; no Cargo gate applies (no crate, `Cargo.*` or toolchain change), so no build lane — `n/a`, nothing compiled

### Record note — Round 1 exists nowhere on this PR

This PR carries a `## Review Response: Round 1` (2026-09-06T06:19:53Z) answering findings `f1`–`f4` plus an escalation, but **no `## Code Review: Round 1` was ever posted here**; `issues/980/timeline` shows it was never published rather than posted-and-deleted. `.claude/CLAUDE.md` is explicit that "Internal verifier reports, task receipts, and CI do not replace PR comments," so the round that response answers is not on the record. This is tracked as **#993** (which names #974, #976, #980, #981 with the same shape). Stated as an observed fact about the record, not as a finding against this diff. Two consequences for this round: it is numbered **Round 2** but is the first *published* round, and I treated every claim in the Round 1 response as an unverified assertion to check rather than as history — so findings a first round would have caught are in scope, and the `loop`'s "later-round blockers must be fix-caused" restriction does not apply.

### IN SCOPE

Full round over `586cb07e..5252f120` — 7 files, +255/−37, all docs/scripts, no crate. Changed behaviour and its direct consumers:

- `scripts/fleet/verdict-drift.sh` — shared `EDDA_OPEN_PR_LIMIT` enumeration limit, a stderr saturation warning, and R24 field (3) `mergeable`. Consumers: `scripts/fleet/daily-digest.sh:168` (only automated caller) and the standalone operator invocation `docs/guides/pi-controller-runbook.md:128`.
- `scripts/fleet/daily-digest.sh` — same limit variable, exported to the drift subprocess. Consumer: board issue #888 via `gh issue comment` (not exercised here — see RAN).
- `scripts/review-l0.sh` + the three `REVIEW.md` markers — the `no-pr-needed` marker attribute and its validation. Consumer: the pre-push L0 pass (`.claude/skills/fleet-worker/SKILL.md:32`, `docs/guides/brief-template.md:206`).
- The three fixture suites.

Acceptance ceiling: issue #958 carries no `## doneWhen` block; its four numbered items are the acceptance list. Current-base integration and the readiness signal's own failure mode (a false clean bill) were weighted heavily, since this diff *is* the machinery that tells an operator which PRs are ready.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 7 files, all inside #958's four items' surface (`verdict-drift.sh`, `daily-digest.sh`, `review-l0.sh`, their fixtures, and the three `REVIEW.md` markers item 4 requires) |
| U2 | any | P1 | PASS | `Closes #958` in body line 1 and in the commit body. #958 has no `## doneWhen`; its four numbered items are all delivered (items 1–4 mapped in Findings/Wiring below), so the closing keyword is correct per `pr.closing-keyword=only-when-all-donewhen-delivered` |
| U3 | any | P1 | **FAIL** | `gh pr view 980 --json body --jq .body \| awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → **empty**. "Empty output is the failure" (§5.0 U3). See [P1] below |
| U4 | any | P1 | PASS | `git log --format=%s origin/main..5252f120 \| grep -Ev '<conventional>'` → empty. Subjects: `fix(fleet): readiness follow-ups …`, `docs(fleet): an empty mergeable means no key in the object, not an older gh` |
| U6 | any | P0 | PASS | READ `gh pr checks 980` — 12/12 pass at this head, `CI run 34015649019 @ 5252f120`. Not deterministically red; no full local rerun taken |
| U7 | any | P1 | PASS | whole scoped audit finished before posting; every blocker batched here (one) |
| D1 | docs | P1 | PASS | claim enumerator over the diff → **0 candidates** (no backticked `edda`/`gh`/`git`/`cargo`/`pi`/`claude` command added). Two source claims checked by hand instead — see Findings note (a) |
| D2 | docs | P1 | N.A.(no ledger claim in the diff) | decision-key enumerator over added lines returned only filenames (`drift.err`, `open.json`, …); no decision value or status is asserted |
| D3 | docs | P1 | PASS | path/link enumerator → **0 candidates**; the four bare script paths named in added comments (`test-verdict-drift.sh`, `test-review-l0.sh`, `daily-digest.sh`, `verdict-drift.sh`) all resolve |
| D4 | docs | P0 | PASS | `gh pr merge\|--delete-branch\|--admin\|--no-verify\|push --force\|force-push` over added lines → empty |
| D5 | docs | [判斷] | 需升級 (adjudicated PASS, `provisional`) | derivation in Findings note (b); escalated per §6.1 |
| S1 | skills | P0 | PASS | spec globs (`*SKILL.md`, `.claude/**`, `skills/**`) match nothing in this diff → empty; re-run widened to `REVIEW.md` (the file the §3 classifier actually puts in `skills`) → also empty. The glob gap itself is routed as a FOLLOW-UP |
| S2 | skills | P1 | PASS | the only `REVIEW.md` change is ` no-pr-needed` appended to three markers (`:278`, `:567`, `:612`); no command or `file:line` claim added |
| S3 | skills | P1 | PASS | no rule owned by `REVIEW.md` or `.claude/CLAUDE.md` is restated; `review-l0.sh`'s added comments describe the runner's own gate, and the rule text stays in the spec |
| C1 | code-plain | P0 | PASS | #958's four items matched to code + test: (1) `verdict-drift.sh:39` / `daily-digest.sh:152,168` + `test-verdict-drift.sh:159-174` + `test-daily-digest.sh:269-272`; (2) `verdict-drift.sh:110-120` + drift cases 8/9; (3) `test-daily-digest.sh:44-79,248-273`; (4) `review-l0.sh:110-162,273-300,397` + `REVIEW.md:278,567,612` + `test-review-l0.sh:219-250`, and the `oneline()` comment at `review-l0.sh:258-266` |
| C2 | code-plain | P0 | PASS | `fix:` with a regression test per item, and the two I could isolate are red before the change (RAN, see below): item 3's `select` injection → RED, item 4's marker drop → RED. Item 1 is covered by drift case 10 + digest case 3c, item 2 by drift cases 8/9 (`mergeable` was not requested at all at base, so neither case can pass there). `grep -cE '^\+.*(#\[test\]\|fn test_)'` over `crates/**` → 0, correct: no crate changed |
| C3 | code-plain | P1 | N.A.(no `crates/**` change) | `git diff … -- 'crates/**'` is empty |
| C4 | code-plain | P1 | PASS | `unsafe {` / `#[allow(clippy::all)]` / `#![allow` over added lines → empty |
| C5 | code-plain | P1 | N.A.(selector empty) | no `crates/` path in the changed-file list; nothing outside the CI Windows 7-crate subset to run |
| R1 | code-risk | P0 | PASS | destructive-command enumerator over added lines → empty. The only `rm -rf` in these files are the pre-existing `trap 'rm -rf "$tmp"' 0 …` sandbox cleanups, unchanged by this diff |
| R2 | code-risk | P0/P1 | PASS | one candidate, adjudicated in full below (`test-daily-digest.sh:52`) |
| R3 | code-risk | P0 | PASS | `sh -n` on all 6 changed shell scripts → exit 0 each |
| R4 | code-risk | P1 | PASS | no new temp dir, file or lock. Both suites' pre-existing `trap … 0 HUP INT TERM` cleanups still cover every exit path; `test-review-l0.sh`'s `trap 'rm -r "$TMP"' EXIT` unchanged |
| R5 | code-risk | P1 | PASS | boundary properties confirmed shape by shape — see Findings note (c) |
| WIRING | any | P1 | PASS | table below; `sh scripts/wiring-scan.sh 586cb07e 5252f120` → `no new pub surfaces`; its three swallow hits are the words "silently"/"silence" inside added *comments*, not code |

**R2 — the one mixed-operator line, full adjudication.**

`scripts/fleet/test-daily-digest.sh:52` — `[ -n "${1:-}" ] && [ -f "$1" ] || return 0`

Parse tree: `( [ -n "${1:-}" ] && [ -f "$1" ] ) || return 0` — `&&` and `||` share precedence and are left-associative.

| A `-n "$1"` | B `-f "$1"` | commands run | effect |
|---|---|---|---|
| false | not evaluated | `return 0` | function returns 0, emits nothing |
| true | false | `return 0` | function returns 0, emits nothing |
| true | true | `||` short-circuits; falls through to the `case` | fixture is emitted |

No row runs any command from the R1 list, and every row matches the stated intent ("skip when there is no fixture file"). The enclosing stub carries no `set -e`, so the guard cannot abort the stub either. **Severity by the §5.4 table lookup: not P0 (no destructive command on any branch) and not P1 (no defect) → PASS.**

### Wiring

|---|---|---|---|---|
| `EDDA_OPEN_PR_LIMIT` env var, default 200 | `verdict-drift.sh:39`; `daily-digest.sh:152`; exported to the subprocess at `daily-digest.sh:168` | both `gh pr list --limit` calls (`verdict-drift.sh:49`, `daily-digest.sh:153`) | none needed — always defined; `${…:-200}` also absorbs a set-but-empty value | env → both enumerations. Asserted at `test-verdict-drift.sh:171` (`--limit 7` reaches argv) and `test-daily-digest.sh:269-272` (one unique `--limit` across both calls) |
| stderr saturation warning | `verdict-drift.sh:53-55`, `>&2` when rows ≥ limit | a direct caller's terminal. **Not the digest**: `daily-digest.sh:168` captures it to `$tmp/drift.err` and `:169-173` prints it only when `drift_rc ≥ 2`; a saturated run exits 0 or 1 | swallowed on the digest path — the operator's report never sees it | script → stderr. Asserted at `test-verdict-drift.sh:182`; **no assertion that a non-saturated run stays silent** (probe below). Tracked as **#986** |
| `mergeable=<v>` line suffix + `not_ready=1` on `CONFLICTING` | `verdict-drift.sh:114-120` | the drift line itself and its exit 1 — the standalone gate #958 item 2 scopes it to. **Not** `daily-digest.sh:178`'s `drift_map` awk, which selects `$4 ∈ {no, stale, SHADOW}`; a `LGTM mergeable=CONFLICTING` line has `$4 = LGTM` and yields no digest row | exit 1 is the standalone signal; digest coverage still comes only from its independent `mergeStateStatus == DIRTY` branch (`daily-digest.sh:189-190`) | `gh --json mergeable` → line + exit code. Asserted by drift cases 8/9 (exact-line match). Not asserted end-to-end through the digest → FOLLOW-UP |
| `no-pr-needed` marker attribute | `REVIEW.md:278,567,612` → extractor `review-l0.sh:109-131` → `@@BLOCK@@ <RULE> <n> <attrs>` → map `:145-147` → `run_block $5` `:398` | `run_block`'s `case ",$attrs,"` at `:293-300`; validated by `bad_attrs` `:150-162` | unknown attribute → `exit 2`, loud (`test-review-l0.sh:228-240`). A *correctly spelled* attribute on a block with no `REVIEW_FILES` fallback is **not** cross-checked — see [P2] | marker → extractor → map → `run_block`. I replicated the extractor at this head: the three attributed markers are exactly the three `REVIEW_FILES`-bearing blocks (U1, C5, R3); all 23 other blocks carry `-` |
| `GH_DRIFT_OPEN_JSON` / `GH_DRIFT_COMMENTS_JSON` (test-only) | `test-daily-digest.sh:261-262` | the stub's `--state open` / `pr view` arms (`:66-79`), routed on whether `GH_OPEN_JSON` is empty | test-only; no product path | fixture env → stub → digest and its drift subprocess |

### Findings

- **[P1] The PR body has no `Issue: #958` line.** `REVIEW.md` §5.0 U3 is explicit and zero-discretion: the closing keyword still supplies the brief's ceiling, "Missing `Issue: #N` is still P1 because the repo wants the issue named in this exact conventional form. Empty output is the failure." — evidence: `gh pr view 980 --json body --jq .body | awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` prints nothing; body line 1 is `Closes #958`. This is a live convention, not a dead letter: #979 (`Issue: #933`), #977 (`Issue: #951`), #973 (`Issue: #934`), #982 (`Issue: #949`), #972 (`Issue: #748`), #946 (`Issue: #914`) all carry it. Fix is one line in the body; the `Closes #958` keyword is correct and should stay (U2 passes).

- **[P2] Declaring `no-pr-needed` and being able to run without `$N` can diverge, and nothing checks it.** The old gate was a conjunction — `[ -n "${REVIEW_FILES:-}" ] && grep -q 'REVIEW_FILES' "$TMP/block.sh"` — so a block could only skip the PR-number guard if it demonstrably read the file list. `review-l0.sh:293-300` now trusts the marker alone; `bad_attrs` (`:150-162`) validates the attribute's *spelling*, never that the block it sits on can actually run with `N=""`. No block is mis-declared today (verified: the three attributed markers are exactly the three `REVIEW_FILES`-bearing blocks), so this is latent, not a present defect. It matters because the failure is silent rather than loud: with `N=""`, `gh` falls back to the current branch's PR — RAN in this worktree, `gh pr checks ""` printed PR #980's 12 checks and `gh pr view "" --json headRefOid` printed `5252f120…`. A future `no-pr-needed` on U2/U3/U6 would therefore report PASS/FAIL for whatever PR the branch happens to have, which is worse than the `N.A.(needs PR number)` it replaced. Cheapest fix consistent with the PR's own argument ("a check quietly not running is the failure mode this whole runner exists to prevent"): keep the `REVIEW_FILES` presence as an additional conjunct, or extend `bad_attrs` to reject `no-pr-needed` on a block whose body does not mention `REVIEW_FILES`.

- **[P2] The new saturation assertion has no negative half — the warning can become unconditional and the suite stays green.** RAN: replacing `verdict-drift.sh:53`'s condition with `if true; then` keeps `scripts/fleet/test-verdict-drift.sh` at 11/11 `verdict-drift fixtures passed`. Case 10 (limit 7, one row) never asserts that `$tmp/err` is silent, and case 11 only asserts the warning is present. The failure mode is noise on every run rather than a false clean bill, hence P2 — but a stderr line that always fires is a line an operator learns to ignore, which is the same erosion #986 is about. One line in case 10 (`grep -q 'hit its limit' "$tmp/err" && fail 10 …`) closes it.

- **[P2] The new shared-limit assertion is blind to a *dropped* `--limit`, which reintroduces #958 item 1 at a smaller number.** `test-daily-digest.sh:269-272` collects `--limit N` values from the stub argv log, `sort -u`s them, and requires exactly one. Three mutations, all RAN:

  | mutation to `daily-digest.sh` | suite | correct? |
  |---|---|---|
  | `:152` → literal `open_pr_limit=100`, propagation at `:168` kept | **green** | **yes** — both enumerations then walk 100, so the invariant the rule exists for (no drift line without a reachable digest row) still holds. The assertion is testing the property, not the literal, which is the right choice |
  | `--limit "$open_pr_limit"` removed from `:153` entirely | **green** | **no** — `sed -n 's/.*--limit \([0-9]*\).*/\1/p'` extracts nothing from that call, the unique set collapses to `{200}` (drift's), the count is 1, and the assertion passes while `gh pr list` falls back to its documented default of **30**. Drift 200 / digest 30 is #958 item 1 exactly |
  | `select(.mergeStateStatus != "CLEAN")` injected into `:155`'s `--jq` | **red**: `FAIL: case 3c: a CLEAN drift-not-ready PR is missing from 擋住什麼` | yes — the jq-applying stub is genuinely load-bearing, which is #958 item 3 delivered |

  Closing the middle row is one line: assert the extracted set has **two** entries before `sort -u` (`[ "$(… | wc -l)" = 2 ]`), so a missing `--limit` is a missing entry rather than a silently agreeing one. Relatedly, nothing asserts that the *digest* honours `EDDA_OPEN_PR_LIMIT` — `test-verdict-drift.sh:167-172` asserts it only for the drift check.

- Clarification, not a finding: the PR body's evidence row "revert `daily-digest.sh` → `FAIL: … different --limit values: 100 200`" is accurate for a **whole-file** revert, which also removes the `EDDA_OPEN_PR_LIMIT="$open_pr_limit"` propagation at `:168`. The narrower single-line mutation in row 1 above does not trip it, and should not.

- **[P2] Case 6's C5 sentinel is inert, and the `undeclared-spec` run's exit code is captured and never asserted.** `test-review-l0.sh:248-249` is a negative assertion — `grep -F '| C5 |' … | grep -Fq 'N.A.(needs PR number)' && fail` — so it goes green whenever its pattern stops matching for *any* reason, including the C5 row disappearing or the runner's N.A. string changing. RAN, to find what actually guards the C5 attribute: dropping ` no-pr-needed` from `REVIEW.md:567` turns the suite **red**, but at **case 4** (`pre-push fixture: C5 still N.A.(needs PR number)`, `:191-197`), not at case 6 — and notably **not** at case 2, which passed unchanged, because `:159-162` only asserts that a row *exists* and accepts `N.A.` as a row. So case 4 is the real guard and case 6's sentinel is redundant with it; it is not a hole, but it is not the protection its comment implies either. Separately, `:244-245` sets `rc=0; run_l0 … || rc=$?` and nothing reads `rc` (contrast `:236`, which does assert `rc -eq 2` for the typo fixture).

- **[P2] The PR body's "Not extended here" claim about CI coverage is stale at the current base.** The body says `scripts/fleet/test-verdict-drift.sh` and `scripts/test-review-l0.sh` "still have no CI consumer — tracked on #927". True at this PR's base `586cb07e` (no `run-fleet-tests.sh`, no `fleet-tests` job), but `main` has since merged #910 (`b19742b`), which adds a `Fleet shell tests` job globbing `scripts/fleet/test-*.sh`. On the merge ref that job **did** run both changed fleet suites green at this head — job `101438822327` logs `RUN scripts/fleet/test-daily-digest.sh` → `daily-digest fixtures passed`, `RUN scripts/fleet/test-verdict-drift.sh` → `PASS 10`, `PASS 11`, `verdict-drift fixtures passed`. Only `scripts/test-review-l0.sh` (one directory up, outside the glob) remains uncovered, so #927 still applies to it. Worth correcting in the body since the sentence understates the delivered coverage.

- Note (a) — **two source claims added by the diff, both checked and both true.** (i) `verdict-drift.sh:109-113` now says an empty `mergeable` means "the object carried no `mergeable` key at all … NOT an older `gh`, which rejects an unknown `--json` field before the request and so lands in `fail_read`". RAN: `gh pr list --repo fagemx/edda --json number,notafield` → `Unknown JSON field: "notafield"` plus the available-field list, exit **1**, client-side, no request — so it does reach `|| fail_read` (exit 2). This is the Round 1 `f3` fix and it is correct. `mergeable` itself is supported by the installed `gh 2.87.0`: a live one-row probe returned `995 f817e8df… main MERGEABLE`. (ii) `review-l0.sh:258-266`'s new `oneline()` comment claims the back-off strips one *complete* character when the byte cap lands on a boundary. True: `cut` takes a byte prefix, so any character with a byte in the cell has its lead byte in the cell too, and `s/[\xC0-\xFF][\x80-\xBF]*$//` therefore always matches the trailing character whole or partial; either way the result decodes. The stated invariant ("must never emit invalid UTF-8") holds in both cases, and case 5's `iconv` round-trip is the standing proof.

- Note (b) — **D5 `[判斷]`, adjudicated `provisional` per §6.1.** Derivation: the added prose is four block comments and one usage-text edit, each stating the *reason* for a behaviour at the site that implements it (`verdict-drift.sh:16-31`, `daily-digest.sh:147-151`, `review-l0.sh:109-114,258-266,283-291`, `test-daily-digest.sh:36-43`). Scope: each names its issue (GH-958) and the failure it prevents, and none restates a rule owned by `REVIEW.md` or `.claude/CLAUDE.md` (S3 passes independently). Duplication: the limit rationale appears in both scripts, which is correct — a reader of either file needs it, and each states the other's number. Clarity: the one place the prose runs ahead of the code is `review-l0.sh:31-37`, which describes `no-pr-needed` blocks as ones that "reach the `$N` only in a fallback branch that stays unreached" — an accurate description of today's three blocks but not an invariant the runner enforces, which is the [P2] above. Adjudicated **PASS**; tagged `provisional` and listed under `escalations:` because this brief does not name me as qualified for this class (§6.1).

- Note (c) — **R5 boundary properties, and the boundary behaviour the brief asked about.** (i) `--limit` interpolation: `"$open_limit"` / `"$open_pr_limit"` are quoted at both call sites; a non-numeric or ≤0 value is rejected by `gh` itself and lands in `fail_read` (exit 2) rather than in a partial enumeration — fail-closed. (ii) Set identity: both enumerations now request the same page size from the same default ordering, so they see the same first-N PRs — that is the actual fix for #958 item 1, and it is what makes a drift line reachable as a digest row. (iii) At exactly the limit: `verdict-drift.sh:53` warns when rows ≥ limit, so N open PRs with limit N warns although nothing was dropped — conservative, correct direction. (iv) One past the limit: `gh` returns N, the drift check warns on stderr, and **the digest prints nothing at all** — it has no saturation check of its own and discards the subprocess's. That is exactly **#986**, filed from this PR's own Round 1 and linked from the Gate receipt comment, so it routes as a follow-up rather than a blocker; it is also unchanged in kind from the base, where the digest truncated at 100 with no warning whatsoever. (v) `printf %s "$open_rows" | grep -c .` counts correctly with no trailing newline and yields `0` (not an error) on an empty enumeration; the `grep` exit 1 is inside an `if` condition and cannot trip `set -e`. (vi) No `$?`-after-a-pipe defect is introduced anywhere in this diff — every added pipeline is either output-signalled (`verdict-drift.sh:53`, `oneline()`) or has its `||` bound to the pipeline whose last stage *is* the assertion (`test-daily-digest.sh:265,267,271`, `test-review-l0.sh:246,248`).

### FOLLOW-UP ISSUE

1. **`daily-digest.sh:178`'s `drift_map` selector drops the new `mergeable=CONFLICTING` state.** The awk matches `$4 == "no" || $4 == "stale" || $4 == "SHADOW"`; a drift line reading `#N <sha> main LGTM mergeable=CONFLICTING` has `$4 = LGTM` and produces no digest row. Today the digest still catches such a PR through its independent `mergeStateStatus == DIRTY` read (`:189-190`), and GitHub computes both fields together, so no false clean bill is reachable — but the R24 field is now wired into the drift check and not into the report R24 calls *the* artefact. #958 item 2 scopes itself to "the drift check … as the sole standalone gate", so this is beyond the acceptance ceiling. Acknowledged as `f2` in the Review Response with no issue number; it needs one.
2. **`scripts/fleet/manager-tick.sh:534` is a third open-PR enumeration, still `--limit 100`, still not reading `EDDA_OPEN_PR_LIMIT`.** It feeds `reap_pass` / `collision_pass` / `assign_pass`, so beyond 100 open PRs the manager and the digest would disagree about which PRs exist — the same shape #958 item 1 fixed for the other two. Outside this PR's surface.
3. **`REVIEW.md`'s S1 path globs do not cover the `skills`-class files the §3 classifier itself names.** S1 greps `-- '*SKILL.md' '.claude/**' 'skills/**'`, but §3 also classifies `REVIEW.md`, `AGENTS.md` and root `CLAUDE.md` as `skills`. A merge-authority instruction added to `REVIEW.md` would be routed to S1 and then scanned by a glob that cannot see it. Pre-existing at base; I re-ran S1 widened to `REVIEW.md` for this diff (empty) so nothing is unchecked here.
4. **`scripts/test-review-l0.sh` still has no CI consumer** (#927) — it sits outside `run-fleet-tests.sh`'s `scripts/fleet/test-*.sh` glob, so all of case 6 is hand-run only. Unchanged by this PR, and now the *only* remaining half of the body's "no CI consumer" note.

### RAN vs READ

**RAN** (all read-only; nothing was posted, labelled, pushed, or written to any branch, worktree, ref or commit status; board issue #888 was never contacted — the digest's `--dry-run` short-circuit at `daily-digest.sh:276-279` `cat`s and `exit 0`s before `gh issue comment`, and every digest run here additionally went through the suite's own `mktemp`-rooted POSIX-path `gh`/`edda` stubs):

- REVIEW.md §3 router on the changed-file list → `classes=code-risk skills`, `canonical_class=code-risk`.
- Rule enumerators U2, U3, U4, U5, D1, D3, D4, R1, R2, R3, C4, C5, S1, WIRING — outputs in the table above.
- Baseline at this head: `sh scripts/fleet/test-verdict-drift.sh` → `PASS 1`…`PASS 11`, `verdict-drift fixtures passed`, rc=0. `sh scripts/fleet/test-daily-digest.sh` → `daily-digest fixtures passed`, rc=0. `sh scripts/test-review-l0.sh` → all six fixtures OK, `all fixture assertions held`, rc=0.
- Five mutation probes, each run against copies in a scratch tree — the PR worktree was never modified and no file in it was written:
  - `select(.mergeStateStatus != "CLEAN")` injected into `daily-digest.sh:155`'s `--jq` → **RED**: `FAIL: case 3c: a CLEAN drift-not-ready PR is missing from 擋住什麼`. The jq-applying stub is load-bearing; #958 item 3 is genuinely delivered, and the old cat-the-TSV stub would have stayed green here.
  - `daily-digest.sh:152` → literal `open_pr_limit=100`, propagation kept → **GREEN**, correctly (both enumerations then walk 100).
  - `--limit "$open_pr_limit"` removed from `daily-digest.sh:153` → **GREEN**, incorrectly (the [P2] above).
  - `verdict-drift.sh:53`'s saturation condition replaced with `if true` → **GREEN**, `PASS 1`…`PASS 11`, `verdict-drift fixtures passed` (the [P2] above).
  - ` no-pr-needed` dropped from `REVIEW.md:567` (C5's marker) → **RED**, at case 4 (`pre-push fixture: C5 still N.A.(needs PR number)`); case 2 passed unchanged, which is how I established that case 4, not case 6's sentinel or case 2's row check, is what guards the attribute.
- Extractor replication at this head — the `@@BLOCK@@` map for all 26 marked blocks; exactly U1, C5 and R3 carry `no-pr-needed`, and exactly those three blocks read `REVIEW_FILES`.
- `gh` behaviour probes: unknown `--json` field → exit 1, client-side (`Unknown JSON field`); `mergeable` supported and live-readable on `gh pr list`; empty `$N` silently resolves the current branch's PR.
- `git diff --check`-adjacent hygiene: `sh -n` on all six changed scripts → 0.

**READ** (not re-run):

- **L1 receipt: `CI run 34015649019 @ 5252f12008a691b35b1c052d84fe121380db4a69`** — all 12 checks SUCCESS: Format; Clippy ×3 OS; Test on ubuntu / macos / windows; MSRV 1.91.0; Detect code changes; **Fleet shell tests**; **Fleet shell tests (windows)**; CI Gate. The four workspace Cargo `gates:` entries are READ from these job results and were never run here (`ladder` L1; C5's selector is empty, so there is no focused Windows-gap run to make either).
- What that covers and does not: the two `Fleet shell tests` jobs come from `main`'s `run-fleet-tests.sh` (merged as #910 after this PR's base) and did execute both changed fleet suites on the merge ref — ubuntu job `101438822327` shows both passing. They do **not** cover `scripts/test-review-l0.sh`, which is outside the glob; that suite is covered only by the local baseline and mutation probes above.
- `scripts/fleet/verdict-drift.sh` reporting `no verdict on head` and exit 1 across the open set is the repo's expected state right now (#993), not something this diff introduced; I did not treat it as a signal about this PR.

### Verdict

Changes Requested, P0=0, P1=1 — the PR body carries no `Issue: #958` line, which §5.0 U3 makes a P1; every other routed rule passes and all four of #958's items are delivered and covered by tests that are red before the change.

PR #1046, opened 2026-09-07
fix(edda-cli): review resolves its subject from a private ref, not FETCH_HEAD (GH-997)

fagemx:
Issue: #997

[...]

PR #1045, opened 2026-09-07
fix(edda-cli): dispatch claim guard bounds bare-CLI claims in time (GH-1018)

fagemx:
Issue: #1018

[...]

PR #1040, opened 2026-09-07
feat(cli): edda fleet order — deterministic health-gated queue, per-file collision, in-product freshness (GH-1015)

fagemx:
## Problem and change

Ordering was the controller's hand-written plan (#932's manual triage of 63
issues), machine selection read labels in `scripts/fleet/next-issue.sh`, and
collision was judged per issue when conflict is per file — the shape that let
#887 be claimed by two machines and built twice. `edda fleet order` moves the
ranking into the product as a deterministic, decomposable computation.

- **Deterministic.** Sort key is `(score descending, issue number ascending)`,
  so equal scores always fall back to the older issue and no input permutation
  can reorder the result. Fixture-tested on serialized output and on both
  renderings.
- **Decomposable.** Every row carries the seven score components that sum to its
  total (`class`, `readiness`, `priority`, `blocker`, `collision`, `freshness`,
  `frozen`), printed in the text and markdown renderings. A bare rank is
  unreviewable; the operator's remaining move is a veto at the queue head.
- **Health-coupled.** RED sets `mechanism_dispatch: freeze` and zeroes every
  mechanism-class row (`status: frozen`, the zeroing recorded as its own
  component so the sum still checks). The one exception is a body citing a red
  run/gate link — a URL containing `/actions/runs/` or `/checks` — which earns
  the `blocker` bonus instead. No link, not a pipeline blocker.
- **Per-file collision (#1005).** Pairwise intersection of `## Predicted
  surface` paths, never labels. Walking the ranked queue leaves the head
  `ready` and holds the rest naming the owner and the shared path; only a
  dispatchable row claims a path, so an in-flight, frozen or stale row never
  blocks the queue behind it.
- **Freshness in the product (#970).** Re-derived against the pinned tree
  (`git ls-tree -r HEAD`) and this binary's own clap verb table — not the `edda`
  on `PATH`, which is what produced the false
  `FAIL edda fleet --help exited nonzero` on this very issue at
  `a4f97990d78a44a95df6f57909e97f7057f104ff`. A path or command declared by a
  section describing what the issue will build (`doneWhen`, any surface heading,
  `改哪裡`) WARNs; an evidence-cited command WARNs; a reference in an
  observation section to something that does not exist FAILs and marks the row
  `stale`. `doneWhen` headings match case-insensitively.
- **Mechanical lane routing.** `governance`/`fleet:goal` label or a `[判斷]`
  marker → controller; empty surface, surface wider than the ledger-keyed cap
[...]

PR #1022, opened 2026-09-06
feat(cli): edda fleet health — path-classified merge/issue mix with ledger thresholds; RED freezes mechanism dispatch

fagemx:
## Problem and change

Queue health (product vs mechanism capacity share, mechanism issue rate) was computed by hand in controller conversation: no command could recompute it, no thresholds existed, and a three-day drift went unwarned (#1014). This makes health a machine state instead of a human-memory state, and gives the ordering layer a mechanical source for the mechanism freeze decision.

Change: new `edda fleet health --window N [--json|--line]` subcommand (new `crates/edda-cli/src/cmd_fleet.rs`, registration in `crates/edda-cli/src/main.rs`):

- Path-classified classification, zero string heuristics on titles/scopes: merged PRs are classified by changed paths (crates/, sdk/ → product; scripts/, docs/fleet/, .github/, REVIEW.md → mechanism; else other); issues by the backticked paths in their `## Predicted surface` section under the same rule.
- Thresholds read from ledger decisions `fleet.health.product-share-floor` and `fleet.health.mech-issues-per-day-ceiling`; missing/unparsed keys fall back to built-in defaults and the output reports the source (`ledger` / `default` / `mixed`). Ledger open failure never fails the command.
- Output: JSON with merged PR mix (product share %), issue mix (mechanism/day, last 24h), thresholds, and `status` GREEN/YELLOW/RED; status is also readable from the process exit code (0/3/4; 1 error, 2 usage). One-line form via `--line` for digest embedding.
- RED emits `mechanism_dispatch: freeze` for the ordering layer; dispatch stays `open` otherwise (fail-open for product work).

One-line anchor update in `COMPATIBILITY.md` (the six inserted main.rs lines moved the cited `run(cli)` error path block from 1092-1099 to 1098-1105).

### RAN

- `cargo fmt --all` → exit 0
- `cargo clippy -p edda --all-targets -- -D warnings` → exit 0, `Finished`
- `cargo test -p edda cmd_fleet` → exit 0, 9 passed
- `cargo run -q -p edda -- fleet health --window 7 --json | jq -r .status` → `RED`
- `cargo run -q -p edda -- fleet health --window 7 --line` → `health: RED | product share 42.1% (floor 50%) | mechanism issues 11.0/day (ceiling 9.0/day) | mechanism dispatch freeze` (exit 4)
- `bash scripts/lint-file-length.sh --tree` → exit 0
[...]

PR #1052, opened 2026-09-07
fix(fleet): review-l0.sh reports ERROR for a block that refused to run (GH-950)

### LINKED ISSUE #950: fix(fleet): review-l0.sh reports PASS on a rule that never ran — a full-SHA base makes every block emit 'fatal:' and none report ERROR
Basis `origin/main` `76692220cb35f478f97ab043531ae533d7f4e277`. Found while reviewing PR #942 (#884); the runner is GH-882/GH-915.

## What happened

`scripts/review-l0.sh` strips one leading `origin/` from its `<base>` argument
and exports the remainder as `BASE`, because every rule block writes
`git diff "origin/$BASE..$SHA"`. Pass a full SHA as `<base>` — the natural thing
to do when pinning a review to an immutable base — and the blocks build
`origin/<40-hex>..HEAD`, which is not a revision.

RAN 2026-09-06 on this checkout:

```
$ sh scripts/review-l0.sh a0862fe8fc19a5c9107215fa9772c40a3854e55c HEAD
| U4 | any       | P1 | FAIL | exit=1; fatal: ambiguous argument 'origin/a0862fe8…..HEAD': unknown revision or path not in the working tree. |
| D3 | docs      | P1 | PASS | exit=0; fatal: ambiguous argument 'origin/a0862fe8…..HEAD': unknown revision or path not in the working tree. |
| R1 | code-risk | P0 | FAIL | exit=1; fatal: ambiguous argument 'origin/a0862fe8…..HEAD': unknown revision or path not in the working tree. |
```

The `D3` row is the defect in its clearest form: **the rule never ran, and the
runner reports `PASS`** with a `fatal:` line sitting in its own evidence column.
`R1` and `U4` report `FAIL`, which is equally wrong — they did not run either —
and no row reports `ERROR`, which is the outcome the runner defines for exactly
this case ("exit 127 (command not found) or exit 128 (bad range)").

## Why it matters

The runner's own header promises `ERROR` for a bad range, and the L0 contract is
that a rule which did not run is never silently dropped. Here a bad range is
laundered into a verdict — and into the *reassuring* verdict for any block whose
tail is a grep. A reviewer or a flash lane pinning L0 to a base SHA gets a table
that looks like a clean pass over rules that never executed. `scripts/review-l0.sh`
is on the judging surface (rules.md R22), so this decides what green means.

## Root cause

`run_block` classifies by exit code and printed lines, but the blocks pipe `git`
into `grep`, so `$?` is grep's status and git's 128 never reaches the classifier
— the same POSIX hazard `REVIEW.md` §0 states in prose ("`$?` after `a | b` is
`b`'s status"). The `fatal:` text arrives as ordinary output and is then read as
a finding signal or its absence, depending on the tail.

## Suspected surface

- `scripts/review-l0.sh` — validate `<base>` before exporting it (a resolvable
  ref after the `origin/` prefix is applied, or an explicit full-SHA mode that
  the blocks can consume), and/or detect `^fatal:` in block output as `ERROR`
  regardless of exit code. The second is the narrower fix and catches the whole
  class, not just this argument shape.
- `scripts/test-review-l0.sh` — a fixture invoking the runner with a full SHA
  base must produce `ERROR` rows and a nonzero runner exit, never `PASS`.

## Predicted surface

`scripts/review-l0.sh`, `scripts/test-review-l0.sh`. No crate, no docs unless
the runner's usage text needs the accepted `<base>` shapes spelled out.

## doneWhen

- `sh scripts/review-l0.sh <full-40-hex-sha> HEAD` either resolves the range
  correctly or reports `ERROR` for every affected rule and exits nonzero. No row
  reports `PASS` or `FAIL` on output containing `fatal:`.
- A block whose output contains a `fatal:` line is `ERROR`, independent of the
  pipeline's exit code — the fix is at the classifier, so it holds for any future
  block that pipes a failing command into a grep.
- `scripts/test-review-l0.sh` covers the full-SHA base case and the generic
  `fatal:`-in-output case, and still exits 0 with all existing assertions held.
- The runner's usage comment states which `<base>` shapes it accepts.

- #942 / #884 — where it surfaced; the reviewer worked around it by running the
  enumerator directly with `git`.
- GH-882 / GH-915 — the runner and its rule-block extraction.
- `REVIEW.md` §0 — the stated hazard this violates.

### LINKED ISSUE #1054: fix(fleet): review-l0.sh reports FAIL for a WIRING block that refused to run — wiring-scan.sh says 'error:', not 'fatal:', so #950's net misses it
Found while fixing #950 (PR #1052, FOLLOW-UP slot). Routed out of that PR because #950's `doneWhen` is written specifically against `fatal:` — this refusal string is a different one, so fixing it there would have been scope creep.

## What happens

#1052 makes `run_block` classify any block whose output carries a `^fatal:` line as `ERROR`, regardless of the pipeline's exit code. That catches every git-backed block, because git refuses with `fatal:`.

`WIRING` does not go through git directly. Its block is

```sh
sh scripts/wiring-scan.sh "origin/$BASE" "$SHA"
```

and `scripts/wiring-scan.sh:32-33` refuses with its own wording:

```sh
echo "error: unknown revision $ref" >&2
exit 2
```

So on the same bad range that made every other rule refuse, `WIRING` produces `rc=2` and output containing `error:`, not `fatal:`. Tracing `run_block` (`scripts/review-l0.sh`, post-#1052):

- the new `fatal:` arm does not fire (wrong string);
- `badline` does not match (`exit=[1-9][0-9]*$|CANDIDATE$|^MISSING`);
- `rc` is neither 0 nor 127 nor 128;
- the tail is `sh scripts/wiring-scan.sh …`, so neither `enumerator` nor `countgrep` is set;

and it falls to the final arm at `scripts/review-l0.sh:412-414` — **`HAS_FAIL=1`, reported as `FAIL`** with `exit=2; error: unknown revision origin/<sha>` in its evidence.

## Why it matters

Same defect class as #950: a rule that refused to run gets a verdict instead of `ERROR`. It is strictly less dangerous than #950's `PASS` — a `FAIL` is loud and blocks — but it is still wrong in the same way, and it puts a finding against a PR for a rule that never executed. `scripts/review-l0.sh` is on the judging surface (`docs/fleet/rules.md` R22).

Note the shape of the gap: #1052 deliberately fixed this at the classifier rather than at the argument, so that it holds for any future block. That reasoning is sound but the *detector* is still string-specific, which is what leaves this hole.

## Suspected surface

`scripts/review-l0.sh` — `run_block`'s refusal detection. Options, roughly in increasing order of blast radius:

1. Widen the pattern to `^(fatal|error):` — one-line, catches this case, but `error:` is a much more common word and some block's legitimate finding output could plausibly begin with it. Needs a check across every check-marked block in `REVIEW.md` before it is safe.
2. Give the runner an explicit "this rule refused" exit convention (e.g. treat `rc=2` from a block whose tail is a bare script invocation as `ERROR`), so refusal is signalled by contract rather than by text.
3. Have `wiring-scan.sh` say `fatal:` for an unknown revision. Narrowest, but it fixes one caller rather than the class, and pushes runner semantics into a standalone tool.

`scripts/test-review-l0.sh` — the full-SHA-base case added by #1052 (§7a) already exercises the exact scenario; it asserts over `fatal:` rows only. A sibling assertion for `WIRING` would pin this.

## Done when

- A `WIRING` block that refused to run reports `ERROR`, not `FAIL`, and the runner's exit reflects an error rather than a finding.
- The chosen mechanism is stated in the runner's header contract alongside the `fatal:` rule, so the next block that refuses by a third wording is not a third silent hole.
- `scripts/test-review-l0.sh` covers it, and its existing assertions still hold.

## Predicted surface

`scripts/review-l0.sh`, `scripts/test-review-l0.sh`, possibly `scripts/wiring-scan.sh`. No crate. No docs unless the header contract change needs mirroring.

fagemx:
Closes #950
Closes #1054
Issue: #950, #1054

## What was wrong

`run_block` classifies a rule by its exit code and its printed lines. Every
git-backed block pipes `git` into a `grep`, so `$?` is the grep's status and
git's 128 never reaches the classifier — the POSIX hazard `REVIEW.md` §0 states
in prose. The `fatal:` text arrives as ordinary output and is then read as a
finding signal, or, for a grep tail that printed nothing, as a clean PASS.

Measured on this branch with the fix disabled, against a full-SHA base (every
block that reads the range builds `origin/<40-hex>..HEAD`, which is not a
revision, so every one of them refuses):

```
| C2     | code-plain | P0 | PASS | exit=1; fatal: ambiguous argument 'origin/3ace0c34…..HEAD': unknown revision or path not in the working tree. |
| U4     | any        | P1 | FAIL | exit=1; fatal: ambiguous argument 'origin/3ace0c34…..HEAD': unknown revision or path not in the working tree. |
| C3     | code-plain | P1 | FAIL | exit=1; fatal: bad revision 'origin/3ace0c34…..HEAD' |
| R1     | code-risk  | P0 | FAIL | exit=1; fatal: ambiguous argument 'origin/3ace0c34…..HEAD': unknown revision or path not in the working tree. |
| WIRING | any        | P1 | FAIL | exit=2; error: unknown revision origin/d4883414… |
```

The `C2` row is the defect in its clearest form: **a P0 rule reports PASS with a
`fatal:` line sitting in its own evidence column.** No row reports ERROR, which
is the outcome the runner's own header defines for exactly this case. A reviewer
or a flash lane pinning L0 to a base SHA gets a table that looks like a clean
pass over rules that never executed, and `scripts/review-l0.sh` is on the
judging surface (`docs/fleet/rules.md` R22).

## The fix — two signals, because neither covers the other

**Text**, for the blocks whose exit code is destroyed by a pipe: a `fatal:`
line in a block's output is `ERROR` regardless of the exit code, checked before
every other classification **and** before `U2`'s reviewer-input row, so no
result is ever built on output the block never produced. It therefore holds for
any future block that pipes a failing command into a grep — the narrower fix
the issue's "Suspected surface" prefers, catching the class rather than one
argument shape. The exit code goes in the evidence rather than the result cell:
it is the grep's, not the failing command's, so `ERROR 0` would read as an error
code rather than as the contradiction it records.

**Exit code**, for the blocks that invoke a tool directly: `2` joins `127`/`128`
as a could-not-run code. Added in round 1 — see below for why `error:` could not
simply be added to the text net.

Two comment surfaces follow: the header's ERROR contract states both signals and
each one's limit, and the `<base>` usage text states that a full SHA is **not**
an accepted shape (the issue's fourth doneWhen item).

## Round 1: the `WIRING` P0

Round 1 found that the first commit did **not** deliver doneWhen item 1 —
"reports `ERROR` for **every affected rule**". `WIRING` invokes
`scripts/wiring-scan.sh` directly, and that tool refuses in its own words
(`scripts/wiring-scan.sh:32-33`):

```sh
echo "error: unknown revision $ref" >&2
exit 2
```

so it escaped the `fatal:` net, fell to `run_block`'s final arm, and reported
**FAIL** for a rule that never ran — dragging the runner's exit to 1 (findings)
instead of 2 (could not run). I had routed this to a follow-up on a selective
reading of my own acceptance criteria: I quoted the second doneWhen bullet,
which is written specifically against `fatal:`, and treated it as the whole
item. It is not. `WIRING` is an affected rule.

**Why not widen the text net to `^(fatal|error):`** — the obvious one-liner, and
wrong here. `error:` is ambiguous in this repo: `wiring-scan.sh` uses it for a
refusal, but `scripts/check-cli-docs.sh` uses the same prefix for genuine
findings (`error: undocumented verb`, `error: undocumented flag`). Matching it
would relabel findings as "did not run", which is the inverse of the bug this
PR fixes.

**Exit 2 instead**, which is safe as a code where the prefix is not safe as
text. It is this repo's could-not-run code — the runner's own exits, `grep`'s
for a bad pattern (a status a pipeline *does* report faithfully), and
`wiring-scan.sh`'s for usage and unknown revisions. I checked every
check-marked block in `REVIEW.md`: none signals a finding with 2 — the
enumerator greps signal with 0/1, and the tools that signal by printed lines
(U5, R3, D1, D3) are caught by `badline`, which is checked first. And
`wiring-scan.sh` itself exits only 0 or 2, never non-zero for a finding.

The fixture's wiring stub was an unconditional `exit 0`, which is why §7a could
not see this: it made `WIRING` report PASS on a range no block can resolve. It
now mirrors the real script's refusal contract, so the stub stands in for the
tool instead of hiding it.

This delivers #1054, which I filed for this gap before round 1 corrected the
routing — closed by this PR rather than left open.

## doneWhen

| Item | Where |
|---|---|
| full-SHA base reports ERROR for every affected rule / exits nonzero, no PASS/FAIL over a refusal | `scripts/test-review-l0.sh` §7a — asserted over both wordings, plus `WIRING` by name, plus `exit == 2` |
| `fatal:` ⇒ ERROR independent of exit code, at the classifier | `scripts/review-l0.sh` `run_block`; `test-review-l0.sh` §7b |
| suite covers both cases and still exits 0 with existing assertions held | run below |
| usage comment states accepted `<base>` shapes | `scripts/review-l0.sh:23-31` |

## Gates RAN (Windows workstation)

```
sh scripts/test-review-l0.sh
  dirty fixture: R1 and R3 rows FAIL, runner exit 1 — OK
  clean fixture: every routed rule printed a row — OK
  clean fixture: all rows PASS/N.A./需升級, runner exit 0 — OK
  unmarked spec: UNMARKED line printed, runner exit 3 — OK
  pre-push fixture: U1/C5/R3 ran without a PR, R3 FAIL, U2/U3/U6 stay N.A. — OK
  cjk fixture: R1 evidence cell survives the cap as valid UTF-8 — OK
  attribute fixture: an unknown marker attribute exits 2 — OK
  attribute fixture: the gate reads the marker, not the command text — OK
  full-SHA base: refused blocks are ERROR, none PASS/FAIL — OK
  generic fatal: an exit-0 block printing 'fatal:' is ERROR — OK
  test-review-l0.sh: all fixture assertions held
  exit 0

sh -n scripts/review-l0.sh scripts/test-review-l0.sh   ok
bash scripts/lint-file-length.sh --tree                exit 0
```

**FAIL-first, both arms, separately:**

- text arm — replacing the classifier's `if` with `if false` turns §7a red and
  prints the `fatal:` table above (`SUITE_EXIT=1`);
- exit-code arm — dropping `|| [ "$rc" -eq 2 ]` turns §7a red on exactly the row
  round 1 found:
  `full-SHA base: a PASS/FAIL row reports a verdict over a refusal: | WIRING | any | P1 | FAIL | exit=2; error: unknown revision origin/d4883414… |`

No crate changed, so no Cargo gate applies locally; `scripts/test-review-l0.sh`
runs in the `fleet-tests` job (it is not in `run-fleet-tests.sh`'s quarantine
list), so CI exercises it on ubuntu/dash too.

## Known limit (stated, not fixed)

The text signal is locale-shaped: git localizes its messages, so a translated
`fatal:` would not match. Measured English on the Git for Windows build here
(2.49.0, which ships no `share/locale`); not guaranteed everywhere. The
exit-code signal is the locale-independent half, but it reaches a refusal only
when a pipeline has not swallowed the status — which is the whole reason the
text signal exists. Both limits are now stated in the header contract.

### DIFF
--- a/scripts/review-l0.sh
+++ b/scripts/review-l0.sh
@@ -24,9 +24,11 @@
 #            blocks diff "origin/$BASE..$SHA", so a leading "origin/" is
 #            stripped before BASE is exported (a plain branch name works too).
 #            A full SHA is NOT an accepted shape: "origin/<40-hex>" is not a
-#            revision, so every block refuses and every routed rule reports
-#            ERROR. Pin an immutable base by naming a remote-tracking ref, or
-#            create one for the SHA (GH-950).
+#            revision, so every block that reads the range refuses and every
+#            rule routed to one of them reports ERROR. (A rule whose block
+#            never touches the range — U5 lints the tree — still runs and
+#            still reports its own verdict.) Pin an immutable base by naming a
+#            remote-tracking ref, or create one for the SHA (GH-950).
 #   <head>   a commit, a ref, or the literal HEAD. With the literal HEAD the
 #            classifier's file list is the working tree against <base>
 #            (staged and unstaged work included) while the blocks diff the
@@ -58,13 +60,28 @@
 #           "CANDIDATE" line (D1), or a "MISSING" line (D3) — or exited
 #           non-zero for any other reason. The script reports exit codes and
 #           printed lines; it does not adjudicate what a finding means.
-#   ERROR   the block failed for a non-finding reason: exit 127 (command not
-#           found), exit 128 (bad range), or output carrying a `fatal:` line —
-#           the command refused to run, so there is no verdict to report. The
-#           last is checked independently of the exit code, because a block
-#           that pipes a failing command into a grep reports the grep's status
-#           (REVIEW.md §0) and the failure would otherwise be read as a
-#           finding, or as a clean PASS (GH-950).
+#   ERROR   the block failed for a non-finding reason — it refused to run, so
+#           there is no verdict to report (GH-950). Two independent signals,
+#           because neither one covers the other:
+#             * output carrying a `fatal:` line, whatever the exit code. A
+#               block that pipes a failing command into a grep reports the
+#               grep's status (REVIEW.md §0), so git's 128 never arrives and
+#               the refusal would otherwise read as a finding, or as a clean
+#               PASS. Printed text is the only surviving evidence there —
+#               which is also this signal's limit: git localizes its messages,
+#               so a translated `fatal:` would not match. Measured English on
+#               the Git for Windows build here (2.49.0, which ships no
+#               share/locale); not guaranteed everywhere. The exit-code signal
+#               below is the locale-independent half, and reaches a refusal
+#               only when a pipeline has not swallowed the status.
+#             * exit 127 (command not found), 128 (bad range) or 2. Exit 2 is
+#               this repo's could-not-run code — the runner's own, grep's for
+#               a bad pattern (which a pipeline does report), and
+#               wiring-scan.sh's for a usage error or an unknown revision.
+#               It is needed because a refusal is not always worded `fatal:`:
+#               wiring-scan.sh writes `error:`, and that prefix must not be
+#               matched on, since check-cli-docs.sh uses the same word for
+#               genuine findings.
 #   N.A.    the rule needs reviewer input (D2's decision key), needs a PR
 #           number, or has no command block in the spec (U7, S2, S3, C1,
 #           R4, R5).
@@ -391,7 +408,21 @@ run_block() { # <rule> <class> <severity> <blocks-file line> <marker attrs>
     fi
     return
   fi
-  if [ "$rc" -eq 127 ] || [ "$rc" -eq 128 ]; then
+  # Exit codes that mean the block could not run. 2 joins 127/128 because a
+  # refusal is not always worded `fatal:` and the text net above cannot be
+  # widened to cover it: WIRING's tool refuses with `error: unknown revision`
+  # (scripts/wiring-scan.sh:32), but `^error:` is ambiguous in this repo —
+  # check-cli-docs.sh writes `error: undocumented verb` for a genuine finding,
+  # so matching that prefix would relabel findings as "did not run", which is
+  # the inverse of the bug GH-950 fixes.
+  #
+  # 2 is safe as a code where the prefix is not safe as text: it is this
+  # repo's could-not-run code (the runner's own exits, grep's for a bad
+  # pattern, wiring-scan.sh's for usage and unknown revisions), and no block
+  # in REVIEW.md signals a finding with it — the enumerator greps signal with
+  # 0/1, and the tools that signal by printed lines (U5, R3, D1, D3) are
+  # caught by `badline` above, which is checked first.
+  if [ "$rc" -eq 127 ] || [ "$rc" -eq 128 ] || [ "$rc" -eq 2 ]; then
     HAS_ERROR=1
     if [ -n "$OUT" ]; then ev=$(printf '%s\n' "$OUT" | oneline); else ev="(no output)"; fi
     print_row "$rule" "$2" "$3" "ERROR $rc" "$ev"
--- a/scripts/review-l0.sh
+++ b/scripts/review-l0.sh
@@ -23,6 +23,10 @@
 #   <base>   base ref the way the blocks write it, e.g. origin/main — the
 #            blocks diff "origin/$BASE..$SHA", so a leading "origin/" is
 #            stripped before BASE is exported (a plain branch name works too).
+#            A full SHA is NOT an accepted shape: "origin/<40-hex>" is not a
+#            revision, so every block refuses and every routed rule reports
+#            ERROR. Pin an immutable base by naming a remote-tracking ref, or
+#            create one for the SHA (GH-950).
 #   <head>   a commit, a ref, or the literal HEAD. With the literal HEAD the
 #            classifier's file list is the working tree against <base>
 #            (staged and unstaged work included) while the blocks diff the
@@ -55,7 +59,12 @@
 #           non-zero for any other reason. The script reports exit codes and
 #           printed lines; it does not adjudicate what a finding means.
 #   ERROR   the block failed for a non-finding reason: exit 127 (command not
-#           found) or exit 128 (bad range).
+#           found), exit 128 (bad range), or output carrying a `fatal:` line —
+#           the command refused to run, so there is no verdict to report. The
+#           last is checked independently of the exit code, because a block
+#           that pipes a failing command into a grep reports the grep's status
+#           (REVIEW.md §0) and the failure would otherwise be read as a
+#           finding, or as a clean PASS (GH-950).
 #   N.A.    the rule needs reviewer input (D2's decision key), needs a PR
 #           number, or has no command block in the spec (U7, S2, S3, C1,
 #           R4, R5).
@@ -306,16 +315,37 @@ run_block() { # <rule> <class> <severity> <blocks-file line> <marker attrs>
   # candidates the reviewer adjudicates against the issue - and this runner
   # never reads the issue. Reporting them as FAIL marks every correctly formed
   # PR failed (GH-882 review round 1), so U2 joins D2 as reviewer input.
+  rc=0
+  OUT=$(sh "$TMP/block.sh" 2>&1) || rc=$?
+
+  # A `fatal:` line is the command refusing to run at all — a bad range, a
+  # missing ref, a broken repo. Inside a pipeline that text arrives as
+  # ordinary output while the failing stage's own status never reaches this
+  # classifier ($? is the last stage's, REVIEW.md §0), so every branch below
+  # reads it as a finding signal, or — for a grep tail that printed nothing —
+  # as a clean PASS. GH-950 measured all three misreadings in one run against
+  # a full-SHA base: D1 and D3 PASS, U4 and R1 FAIL, no row ERROR. The rule
+  # did not run, and only ERROR says that.
+  #
+  # Checked before every other classification, and before U2's reviewer-input
+  # row, so no result is ever built on output the block never produced. It is
+  # deliberately at the classifier rather than at the argument: it holds for
+  # any future block that pipes a failing command into a grep, whatever made
+  # the command fail.
+  if printf '%s\n' "$OUT" | grep -q '^fatal:'; then
+    HAS_ERROR=1
+    # The exit code goes in the evidence, not the result cell: it is the
+    # grep's, not the failing command's, so "ERROR 0" would read as an error
+    # code rather than as the contradiction it records.
+    print_row "$rule" "$2" "$3" 'ERROR' "exit=$rc; $(printf '%s\n' "$OUT" | oneline)"
+    return
+  fi
+
   if [ "$1" = "U2" ]; then
-    rc=0
-    OUT=$(sh "$TMP/block.sh" 2>&1) || rc=$?
     if [ -n "$OUT" ]; then ev=$(printf '%s\n' "$OUT" | oneline); else ev='(no closing keyword)'; fi
     print_row "$rule" "$2" "$3" 'N.A.(needs reviewer input)' "$ev"
     return
   fi
-
-  rc=0
-  OUT=$(sh "$TMP/block.sh" 2>&1) || rc=$?
 
   # Finding signals carried by printed lines (REVIEW.md §0: a piped check
   # signals by its output; U5 and R3 print their exit, D1 prints CANDIDATE,
--- a/scripts/test-review-l0.sh
+++ b/scripts/test-review-l0.sh
@@ -123,12 +123,15 @@ make_fixture() {
 run_l0() { # <fixture-dir> <spec> <out-file> [PR-number] — runner exit code via $?
   # With no PR-number argument the runner runs in the pre-push shape
   # (GH-922): U1/C5/R3 must still run, from the runner's own file list.
+  # L0_BASE overrides the <base> argument for the GH-950 cases; every other
+  # caller gets the ordinary origin/main shape.
   (cd "$1" \
     && PATH="$TMP/bin:$PATH" \
        GH_STUB_DIR="$TMP" \
        REVIEW_L0_SPEC="$2" \
-       sh "$RUNNER" origin/main HEAD ${4:-}) > "$3" 2>&1
+       sh "$RUNNER" "${L0_BASE:-origin/main}" HEAD ${4:-}) > "$3" 2>&1
 }
+L0_BASE=
 
 # ---- 1. dirty fixture: R1 + R3 FAIL, exit 1 ---------------------------------
 make_fixture dirty
@@ -248,5 +251,60 @@ grep -F '| U1 |' "$TMP/undeclared.out" | grep -Fq 'N.A.(needs PR number)' \
 grep -F '| C5 |' "$TMP/undeclared.out" | grep -Fq 'N.A.(needs PR number)' \
   && fail "attribute fixture: C5 keeps its attribute and must still run"
 echo "attribute fixture: the gate reads the marker, not the command text — OK"
+
+# ---- 7. a block that never ran is ERROR, never PASS or FAIL (GH-950) -------
+# 7a. A full SHA as <base>. The blocks diff "origin/$BASE..$SHA", so
+# "origin/<40-hex>" is not a revision and every git-backed block refuses. The
+# refusal text arrives as ordinary output — the failing stage's status never
+# reaches the classifier ($? is the pipeline's last stage) — so before the fix
+# D1 and D3 reported PASS with a `fatal:` line sitting in their own evidence,
+# and U4, C2, C4, R1 reported FAIL. None of them ran.
+make_fixture clean
+# A prefix assignment on a function call persists in POSIX sh, so set and
+# clear it explicitly rather than relying on that.
+L0_BASE=$(cat "$TMP/base-sha")
+rc=0
+run_l0 "$FIXDIR" "$SPEC" "$TMP/sha-base.out" || rc=$?
+L0_BASE=
+[ "$rc" -ne 0 ] || fail "full-SHA base: runner exited 0 over rules that never ran"
+grep -Fq 'fatal:' "$TMP/sha-base.out" \
+  || fail "full-SHA base: fixture produced no refusal at all — the case is not being exercised"
+# The defect in one assertion: no row may carry a verdict over `fatal:` output.
+if grep -E '\| (PASS|FAIL) \|' "$TMP/sha-base.out" | grep -Fq 'fatal:'; then
+  fail "full-SHA base: a PASS/FAIL row reports a verdict over 'fatal:' output: $(grep -E '\| (PASS|FAIL) \|' "$TMP/sha-base.out" | grep -F 'fatal:')"
+fi
+grep -F '| ERROR |' "$TMP/sha-base.out" | grep -Fq 'fatal:' \
+  || fail "full-SHA base: no ERROR row for the refused blocks"
+echo "full-SHA base: refused blocks are ERROR, none PASS/FAIL — OK"
+
+# 7b. The generic case, with no bad range anywhere: a block that prints a
+# `fatal:` line and still EXITS 0 through a grep tail. Before the fix this was
+# the reassuring misreading — an enumerator tail with output reads as FAIL, and
+# with none as PASS; either way the runner reports a verdict for a rule that
+# refused to run. The fix is at the classifier, so it holds for any block and
+# any cause, not only for git and not only for this argument shape.
+sub_cmd='echo "fatal: simulated refusal" >&2; echo ok | grep -v ZZZ'
+awk -v s="$sub_cmd" '
+  /^# review-spec:check U5( |$)/ { print; print "```sh"; print s; print "```"; incut = 1; next }
+  incut && /^# review-spec:check-end$/ { incut = 0; print; next }
+  incut { next }
+  { print }
+' "$SPEC" > "$TMP/fatal-spec.md"
+grep -Fq 'fatal: simulated refusal' "$TMP/fatal-spec.md" \
+  || fail "generic fatal: the substituted spec copy was not produced"
+make_fixture clean
+rc=0
+run_l0 "$FIXDIR" "$TMP/fatal-spec.md" "$TMP/fatal.out" || rc=$?
+u5=$(grep -F '| U5 |' "$TMP/fatal.out")
+case "$u5" in
+  *'| ERROR |'*) : ;;
+  *) fail "generic fatal: U5 exited 0 with a 'fatal:' line and was not ERROR: $u5" ;;
+esac
+case "$u5" in
+  *'exit=0'*) : ;;
+  *) fail "generic fatal: U5's evidence does not record the exit-0 contradiction: $u5" ;;
+esac
+[ "$rc" -eq 2 ] || fail "generic fatal: expected runner exit 2 (ERROR, no FAIL), got $rc"
+echo "generic fatal: an exit-0 block printing 'fatal:' is ERROR — OK"
 
 echo "test-review-l0.sh: all fixture assertions held"
--- a/scripts/test-review-l0.sh
+++ b/scripts/test-review-l0.sh
@@ -85,7 +85,21 @@ make_fixture() {
   git -C "$fix" config user.name fixture
   printf '[workspace.package]\nversion = "0.0.0-fixture"\n' > "$fix/Cargo.toml"
   printf '# fixture lint stub\nexit 0\n' > "$fix/scripts/lint-markdown-content.sh"
-  printf '# fixture wiring stub\nexit 0\n' > "$fix/scripts/wiring-scan.sh"
+  # The wiring stub mirrors the one thing about the real script the runner
+  # has to classify: its refusal (scripts/wiring-scan.sh:30-34). An
+  # unconditional `exit 0` would make WIRING report PASS on a range no block
+  # can resolve, which is exactly the GH-950 defect the §7a case exists to
+  # catch — the stub would hide it rather than stand in for the tool.
+  cat > "$fix/scripts/wiring-scan.sh" <<'STUB'
+# fixture wiring stub — mirrors the real script's refusal contract
+for ref in "$1" "$2"; do
+  if ! git rev-parse --verify --quiet "${ref}^{commit}" >/dev/null 2>&1; then
+    echo "error: unknown revision $ref" >&2
+    exit 2
+  fi
+done
+exit 0
+STUB
   git -C "$fix" add -A
   git -C "$fix" commit -q -m "chore(fleet): fixture base"
   git -C "$fix" rev-parse HEAD > "$TMP/base-sha"
@@ -269,12 +283,28 @@ L0_BASE=
 [ "$rc" -ne 0 ] || fail "full-SHA base: runner exited 0 over rules that never ran"
 grep -Fq 'fatal:' "$TMP/sha-base.out" \
   || fail "full-SHA base: fixture produced no refusal at all — the case is not being exercised"
-# The defect in one assertion: no row may carry a verdict over `fatal:` output.
-if grep -E '\| (PASS|FAIL) \|' "$TMP/sha-base.out" | grep -Fq 'fatal:'; then
-  fail "full-SHA base: a PASS/FAIL row reports a verdict over 'fatal:' output: $(grep -E '\| (PASS|FAIL) \|' "$TMP/sha-base.out" | grep -F 'fatal:')"
-fi
+# The defect in one assertion: no row may carry a verdict over a refusal —
+# stated over both wordings, because the two arrive by different routes. A
+# git-backed block's `fatal:` survives only as text (its status is eaten by
+# the grep it feeds); WIRING's tool is invoked directly, so its `error:` line
+# comes with a truthful exit 2. Asserting only the first would leave the
+# doneWhen's "ERROR for every affected rule" half-tested.
+verdicts=$(grep -E '\| (PASS|FAIL) \|' "$TMP/sha-base.out" | grep -E 'fatal:|error: unknown revision' || true)
+[ -z "$verdicts" ] \
+  || fail "full-SHA base: a PASS/FAIL row reports a verdict over a refusal: $verdicts"
 grep -F '| ERROR |' "$TMP/sha-base.out" | grep -Fq 'fatal:' \
   || fail "full-SHA base: no ERROR row for the refused blocks"
+# WIRING by name: it is the rule the `fatal:` net alone does not reach, and
+# the one that proved the runner still had to key on the exit code too.
+wiring=$(grep -F '| WIRING |' "$TMP/sha-base.out")
+case "$wiring" in
+  *'| ERROR '*) : ;;
+  *) fail "full-SHA base: WIRING refused and was not ERROR: $wiring" ;;
+esac
+# Exit 2 (a rule could not run), not 1 (a rule failed): with every affected
+# rule ERROR there is no finding left to report, and the two exits mean
+# different things to a caller.
+[ "$rc" -eq 2 ] || fail "full-SHA base: expected runner exit 2 (could not run), got $rc"
 echo "full-SHA base: refused blocks are ERROR, none PASS/FAIL — OK"
 
 # 7b. The generic case, with no bad range anywhere: a block that prints a


fagemx:
FOLLOW-UP filed: #1054 — `WIRING` refuses with `error: unknown revision` + exit 2 from `scripts/wiring-scan.sh:32-33`, so it escapes this PR's `fatal:` net and falls to `run_block`'s final arm (`scripts/review-l0.sh:412-414`), reporting **FAIL** for a rule that never ran. Same class as #950, different refusal string; #950's doneWhen is written against `fatal:` specifically, so it is out of this PR's frozen surface.

fagemx:
review: started on 1aaaddebfa4993a3355b255a609686ea3493a3d1

fagemx:
## Code Review: Round 1 — PR #1052 @ 1aaaddebfa4993a3355b255a609686ea3493a3d1

- elapsed: unmeasured (no pi session for this lane)
- model_observed: unverified (no session file / `modelUsage` reachable from this lane; §6 forbids copying `model_requested`)
- reviewer_session: unset (independent review lane, no per-PR UUID assigned)
- spec: review-spec-v1.5
- escalations: none (no `docs`/`skills` class ⇒ D5 not routed; REVIEW.md unchanged in this diff)
- shadow: false
- cost: ~35 tool calls; **no Cargo command run** (no crate in the diff — C5 selector empty); two full runs of `scripts/test-review-l0.sh` (green, then FAIL-first), two real-repo runs of `scripts/review-l0.sh`

### IN SCOPE

Changed behaviour: `run_block`'s result classification in `scripts/review-l0.sh` — one new branch that maps `^fatal:` in a block's captured output to `ERROR` + `HAS_ERROR=1`, hoisted above the U2 reviewer-input row and above every finding-signal branch; two comment surfaces (usage `<base>` shapes, the `ERROR` contract); and two new fixture cases in `scripts/test-review-l0.sh` (§7a full-SHA base, §7b generic exit-0 `fatal:`).

Direct consumers: the runner's own exit-code contract (`HAS_ERROR` → `exit 2`), `docs/guides/brief-template.md:206` / `.claude/skills/fleet-worker/SKILL.md:32` (pre-push step 20), and the `fleet-tests` CI job, which runs `scripts/test-review-l0.sh` on ubuntu/dash. Acceptance ceiling: #950 `## doneWhen` (4 items). No data-loss or authority surface; `scripts/review-l0.sh` is on the judging surface (`docs/fleet/rules.md` R22), so what it calls green is itself the risk.

**The mechanism is right.** The fix is at the classifier, which is the layer #950's "Suspected surface" prefers; it sits above every branch that could build a verdict on output the block never produced; and both new assertions are real detectors — I reproduced the FAIL-first myself (below), not from the author's account. The blockers below are about **completeness and claims**, not about the classifier.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `scripts/review-l0.sh`, `scripts/test-review-l0.sh` — exactly #950's "Predicted surface" |
| U2 | any | P1 | **FAIL** | body line 1 `Closes #950` while doneWhen item 1 is undelivered — see [P0]/[P1] below |
| U3 | any | P1 | **FAIL** | `gh pr view 1052 --json body --jq .body \| awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → empty. PRs #1045/#1046/#1040/#1022/#1016 all carry the line |
| U4 | any | P1 | PASS | `git log --format=%s origin/main..1aaadde \| grep -Ev '<conventional>'` → no output |
| U5 | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → exit 0 |
| U6 | any | P0 | PASS | READ `gh pr checks 1052` → 12/12 pass at head. Both U6 blocks: `edda --version` → `edda 0.5.0 (3306c1a2d9a0 2026-09-07)`; base `Cargo.toml` version read |
| U7 | any | P1 | PASS | whole scoped audit finished and batched before posting; round 1 |
| WIRING | any | P1 | PASS | `sh scripts/wiring-scan.sh origin/main 1aaadde` → exit 0; table below |
| C1 | code-plain | P0 | **FAIL** | doneWhen item 1 — see [P0] |
| C2 | code-plain | P0 | PASS | `fix:` with a regression test that fails without the change — FAIL-first reproduced for **both** new assertions (RAN below) |
| C3 | code-plain | P1 | PASS | no `crates/**` in the diff; no `.unwrap(`/`.expect(` added |
| C4 | code-plain | P1 | PASS | no `unsafe {`, no `clippy::all`, no `#![allow` in added lines |
| C5 | code-plain | P1 | PASS | selector prints nothing (no `crates/` path) — no Windows gap, no Cargo gate applies |
| R1 | code-risk | P0 | PASS | destructive-command grep over added lines → no output. The only `rm` in either file is the pre-existing `trap 'rm -r "$TMP"' EXIT` on a `mktemp -d` |
| R2 | code-risk | P0/P1 | PASS | mixed `\|\|`/`&&` grep over added `*.sh` lines → no output |
| R3 | code-risk | P0 | PASS | `sh -n scripts/review-l0.sh` → 0; `sh -n scripts/test-review-l0.sh` → 0. Also `sh -n` under dash in CI via `run-fleet-tests.sh` |
| R4 | code-risk | P1 | PASS | every new artefact (`$TMP/fatal-spec.md`, `$TMP/sha-base.out`, `$TMP/fatal.out`) lands inside the existing `mktemp -d` covered by `trap … EXIT` |
| R5 | code-risk | P1 | PASS | `$OUT` is only pattern-matched, never evaluated; `oneline` escapes `\|`; `L0_BASE` is a fixture-local SHA read from `$TMP/base-sha`; `$sub_cmd` reaches awk via `-v` as data and is executed only inside the test's own `$TMP` spec copy |

### Wiring

|---|---|---|---|---|
| `ERROR` result for a `fatal:`-bearing block | `scripts/review-l0.sh:335-341` — `grep -q '^fatal:'` → `HAS_ERROR=1` + `print_row … 'ERROR' "exit=$rc; …"` | `scripts/review-l0.sh:449-450` (`HAS_ERROR` → `exit 2`); `scripts/test-review-l0.sh:277`, `:300-307` | not swallowed: sets `HAS_ERROR`, drives the process exit; asserted at `test-review-l0.sh:307` (`[ "$rc" -eq 2 ]`) | classifier → `HAS_ERROR` → process exit; both ends asserted (row shape at `:277`, exit code at `:307`) |
| `L0_BASE` — test-harness `<base>` override | `scripts/test-review-l0.sh:265` set, `:134`/`:268` init+clear | `scripts/test-review-l0.sh:132` `"${L0_BASE:-origin/main}"` | n/a (test harness); anti-vacuity guard at `:270-271` fails the suite if the fixture produces no refusal at all | variable → `( … )` subshell → runner argv → block `origin/$BASE..$SHA`; proven by the `fatal:` rows landing in `$TMP/sha-base.out` |
| usage / ERROR-contract comment text | `scripts/review-l0.sh:26-29`, `:61-66` | human + agent readers (`docs/guides/brief-template.md:206`) | n/a — but see [P1] #2: one clause is false | doc-only |

### Findings

- **[P0] doneWhen item 1 is claimed but not delivered: with the fix applied, a full-SHA base still makes a rule that never ran carry a P1 `FAIL`, and the runner still exits 1 (findings) rather than 2 (could not run).** #950's first doneWhen bullet is *"`sh scripts/review-l0.sh <full-40-hex-sha> HEAD` either resolves the range correctly or reports `ERROR` for **every affected rule** and exits nonzero."* Measured on this head, in the real repo, with the fix in place:

  ```
  $ sh scripts/review-l0.sh e613c3c53460fe40ac56c8282d18df1571f07aeb HEAD
  | U4 | any | P1 | ERROR | exit=1; fatal: ambiguous argument 'origin/e613c3c5…..HEAD': …
  | C2 | code-plain | P0 | ERROR | …            ← the headline defect: fixed
  | R1 | code-risk | P0 | ERROR | …
  | WIRING | any | P1 | FAIL | exit=2; error: unknown revision origin/e613c3c5…
  RUNNER_EXIT=1
  ```

  `WIRING` is a routed rule for **every class, every round** (`scripts/review-l0.sh` RULE_TABLE `WIRING|any|P1`). It is affected by the full-SHA base — `scripts/wiring-scan.sh:32` refuses with `echo "error: unknown revision $ref" >&2` and exits 2 — and it reports `FAIL`. #950's own text calls this the *same* defect, not a lesser one: *"`R1` and `U4` report `FAIL`, which is equally wrong — they did not run either."* Because `HAS_FAIL` outranks `HAS_ERROR` at `scripts/review-l0.sh:446-450`, that one row also drags the runner's exit from 2 back to 1, so the very caller #950 describes — "a reviewer or a flash lane pinning L0 to a base SHA" — still gets "there is a blocking finding" instead of "this could not run".

  No test covers the clause either: §7a asserts only *nonzero* exit, *some* `ERROR` row, and *no* PASS/FAIL over `fatal:` — and it cannot assert more, because the fixture's `scripts/wiring-scan.sh` is a stub (`test-review-l0.sh:88` writes `# fixture wiring stub\nexit 0`), so `WIRING` reports `PASS` there and the real refusal shape is never exercised.

  I do **not** think the FOLLOW-UP routing is scope-dodging — the body discloses the gap precisely and the mechanism is genuinely `error:` not `fatal:`. But the disclosure does not license the two claims that say otherwise: `Closes #950` (U2) and the usage comment (below). Either resolution clears this: extend the net to the refusal the runner's own sibling script emits, **or** drop the closing keyword to `Issue: #950` and correct the comment, leaving #950 open for the remainder (REVIEW.md U2 explicitly provides for that).

- **[P1] `scripts/review-l0.sh:27-28` states as fact the outcome the PR knows it does not produce.** The added usage text reads *"so every block refuses and **every routed rule reports ERROR**"*. Measured above: `WIRING` reports `FAIL`. The PR body's own FOLLOW-UP paragraph says so in the same breath — *"a refused WIRING block is still classified by exit code alone"* — so the file and the body contradict each other, and the file is the one the next reader will trust. This is independently wrong however the P0 is resolved. (`REVIEW.md` §5 severity: *P1 — a false claim*.)

- **[P1] U2 — `Closes #950` while doneWhen item 1 is undelivered.** `pr.closing-keyword=only-when-all-donewhen-delivered`: a closing keyword is written exactly when the diff delivers **every** doneWhen item, "so merging genuinely finishes the issue". Merging as-is auto-closes #950 with the `WIRING` case live and the runner's exit still 1 for a run in which nothing ran. Evidence: `gh pr view 1052 --json body --jq .body | grep -Ein '…close…'` → `1:Closes #950`, against the measured run above.

- **[P1] U3 — the PR body carries no `Issue: #950` line.** `gh pr view 1052 --json body --jq .body | awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → empty; the runner's own row is `| U3 | any | P1 | FAIL | exit=0; (no output) — empty output is the failure`. This is the repo's live convention, not a dead letter: #1045 (`Issue: #1018`), #1046 (`Issue: #997`), #1040 (`Issue: #1015`), #1022 (`Issue: #1014`), #1016 (`Issue: #764`) all carry it. Swapping `Closes #950` → `Issue: #950` would clear U2 and U3 together.

- **[P2] `grep -q '^fatal:'` is locale-shaped, and I could not measure the failure here.** git's `fatal: ` prefix goes through gettext, so a build with translation catalogs installed and a non-English `LC_ALL` would print a translated prefix and slip the net. On this workstation the failure is **not reproducible** — `LC_ALL=zh_TW.UTF-8` and `LC_ALL=de_DE.UTF-8` both still printed `fatal: ambiguous argument …`, because this Git for Windows build ships no locale catalogs — and CI ubuntu runs in the C locale. So: a property worth hardening (`LC_ALL=C` around the block, or matching git's exit 128 at the source), not a defect I can evidence. Stated as unmeasured, per §5's evidence threshold.

Explicitly **not** findings, having looked for them:

- *Could the new branch misclassify a genuine finding as ERROR?* Audited every check-marked block in `REVIEW.md`. Every enumerator that prints free text prints it **prefixed** — `grep -n` line numbers (U2, D4, S1, C3, R1, R2), `OK `/`MISSING ` (D3), `<cmd> <flag> -> ` (D1), `exit=` (U5, R3) — so no finding line can begin at column 0 with `fatal:`. The only unprefixed free text is a commit subject (U4, C2 `git log --format=%s`); a subject literally starting `fatal:` would flip a `FAIL` to an `ERROR`, both non-zero, no false green. Not worth a finding.
- *exit 2 collision with the argument/spec-error exits.* Real, but **pre-existing**: `git show origin/main:scripts/review-l0.sh` already has `HAS_ERROR → exit 2` at `:419-420` alongside the argument-error `exit 2`s. No caller reads the code numerically — `docs/guides/brief-template.md:206` and `.claude/skills/fleet-worker/SKILL.md:32` are agent-read, and `crates/edda-cli/src/cmd_review/qualification.rs:63,255` only names the path as judging surface. Not introduced here, not in #950's doneWhen.
- *`L0_BASE` leaking into later cases.* The comment at `test-review-l0.sh:263` is right that a prefix assignment on a function call persists in POSIX sh — and the code avoids the prefix form entirely: plain assignment at `:265`, explicit clear at `:268`, read through `${L0_BASE:-origin/main}` at `:132`. `:-` (not `-`) substitutes on **null as well as unset**, so the cleared empty string correctly falls back to `origin/main`; the initialiser at `:134` keeps `set -u` happy for §1–§6. Sound.
- *POSIX/dash safety of the §7 additions.* No bashisms (no `local`, `[[`, arrays, `+=`, `${var^^}`, `pipefail`); the nested-quote `$( … "$TMP/…" … )` inside a double-quoted `fail` argument is the POSIX `$()` form dash parses. Not a code read alone — CI **measured** it: `scripts/test-review-l0.sh` is neither quarantined nor SKIPped in `scripts/fleet/run-fleet-tests.sh`, and the ubuntu `Fleet shell tests` job at this head logs `RUN scripts/test-review-l0.sh` → `full-SHA base: … — OK` → `generic fatal: … — OK` → `PASS scripts/test-review-l0.sh`.

### FOLLOW-UP ISSUE

- **`review-l0.sh` classifies every non-`fatal:` refusal by exit code alone.** Beyond `WIRING`/`wiring-scan.sh:32`, the same hole is open for any future block whose command refuses with its own wording (`error:`, `usage:`, a non-git tool). The durable fix is probably not a second string to grep but a per-block way to distinguish "refused" from "ran and found things" — worth its own issue independent of how the P0 above is closed. Named here as the general form; the specific `WIRING` instance is in scope, per the P0.
- **Locale-hardening the refusal match** ([P2] above) — `LC_ALL=C` for block execution, so the classifier never depends on git's message language.

### RAN vs READ

RAN (all in the existing worktree `C:\ai_agent\edda-wt-gh950` at `1aaadde`; no Cargo, no new worktree, no ad-hoc `CARGO_TARGET_DIR`):

- `git rev-parse HEAD` → `1aaaddebfa4993a3355b255a609686ea3493a3d1`; `git status --porcelain` → clean before and after.
- `sh -n scripts/review-l0.sh` → 0; `sh -n scripts/test-review-l0.sh` → 0.
- `sh scripts/test-review-l0.sh` → all 10 assertion lines OK, `SUITE_EXIT=0`. Matches the body's quoted block (the body omits git's `LF will be replaced by CRLF` warnings; cosmetic).
- **FAIL-first, reproduced independently.** Backed up the file, `sed -i "335s/if printf.*then/if false; then/"` (line 335 is exactly the `if` the body cites), reran the suite → red at §7a, `SUITE_EXIT=1`, printing the defect table including `| C2 | code-plain | P0 | PASS | exit=1; fatal: ambiguous argument …` — a **P0 rule reporting PASS with `fatal:` in its own evidence column**. The body's measured table is reproduced exactly for the four rows it quotes (C2 PASS, U4 FAIL, C3 FAIL, R1 FAIL); my run additionally showed C4 and R2 FAIL, i.e. the body's table is a truthful subset, not an overstatement.
- **§7b proved to be an independent detector too**, not shielded by §7a: with the classifier still disabled and §7a's `fail` calls suppressed in a throwaway copy, §7b fired on its own — `generic fatal: U5 exited 0 with a 'fatal:' line and was not ERROR: | U5 | any | P1 | FAIL | exit=0; fatal: simulated refusal`.
- Restored with `git checkout -- scripts/review-l0.sh`, removed the throwaway copy; `git status --porcelain` → empty, `grep -n 'fatal:' scripts/review-l0.sh` → 62, 321, 335. **The worktree is clean at the reviewed SHA.**
- `sh scripts/review-l0.sh e613c3c5…(full SHA) HEAD` → the P0 measurement above, `RUNNER_EXIT=1`.
- `sh scripts/review-l0.sh origin/main HEAD 1052` → the routed-rule table (U3 FAIL; U6's `gh pr checks` row was exit 8 = a then-pending job, now green).
- `sh scripts/wiring-scan.sh "origin/<full-sha>" HEAD` → `error: unknown revision …`, exit 2; `sh scripts/wiring-scan.sh origin/main 1aaadde` → exit 0.
- `sh scripts/lint-markdown-content.sh` → 0; `sh scripts/lint-file-length.sh --tree` → 0 (both body claims confirmed).
- `LC_ALL=zh_TW.UTF-8` / `LC_ALL=de_DE.UTF-8 git rev-list origin/<bogus>..HEAD` → English `fatal:` both times (P2 above, unmeasurable here).
- `gh` reads: PR body/head SHA, issue #950 doneWhen, `gh pr checks 1052`, `gh api …/jobs/101739407117/logs`, `Issue:` lines on #1045/#1046/#1040/#1022/#1016.

READ (not rerun): **CI run 34121147243 @ 1aaaddebfa4993a3355b255a609686ea3493a3d1** — Format, Clippy ×3 OS, Test on ubuntu + macOS (full workspace) + windows (7-crate subset), MSRV 1.91.0, Fleet shell tests (ubuntu **and** windows), CI Gate — **12/12 pass**. What it covers here: the dash execution of `scripts/test-review-l0.sh` including both new §7 cases (quoted from the job log above), which is the coverage claim I most wanted independent of my Windows box. What it does not cover: the real-repo full-SHA-base behaviour of `review-l0.sh` against the unstubbed `wiring-scan.sh` — no CI job runs that shape, which is why I ran it, and it is where the P0 lives. No Cargo gate applies to this PR (C5 selector empty), so nothing Cargo-shaped was RAN or needed.

### Verdict

Changes Requested, P0=1, P1=3 — the classifier fix is correct, well-placed and genuinely FAIL-first proven, but `Closes #950` and `review-l0.sh:27-28` both assert an outcome the head does not produce: a full-SHA base still lands a `FAIL` verdict on `WIRING`, a rule that refused to run, and still exits 1 rather than 2.

fagemx:
## Review Response: Round 1

**SHA: `d0437dea0fa82560a6effee695ef03e3b1a6dc87`**

Every blocking finding is fixed. The P0 was real and my routing of it was wrong in a way worth naming plainly.

### [P0] `WIRING` still FAIL, doneWhen item 1 undelivered — accepted, fixed

You are right, and the failure was in how I read my own acceptance criteria. doneWhen item 1 says "reports `ERROR` for **every affected rule**". I quoted the *second* bullet — which is written specifically against `fatal:` — and used it to argue the whole item was met, then routed the remainder to a follow-up. `WIRING` is an affected rule. Selectively quoting the ceiling to make the work fit it is exactly the move the ceiling exists to prevent.

**The obvious fix is wrong here.** Widening the text net to `^(fatal|error):` would catch `WIRING`, but `error:` is ambiguous in this repo: `scripts/wiring-scan.sh:32` uses it for a refusal, and `scripts/check-cli-docs.sh` uses the same prefix for genuine findings — `error: undocumented verb` (`:154`), `error: undocumented flag` (`:144`). Matching it would relabel findings as "did not run", which is the inverse of the bug this PR fixes.

**So the second signal is the exit code**, safe as a code where the prefix is not safe as text. `2` joins `127`/`128`: it is this repo's could-not-run code — the runner's own exits, `grep`'s for a bad pattern (a status a pipeline *does* report faithfully), and `wiring-scan.sh`'s for usage and unknown revisions. I checked it against every check-marked block in `REVIEW.md` before adding it: none signals a finding with 2 — the enumerator greps signal with 0/1, and the tools that signal by printed lines (U5, R3, D1, D3) are caught by `badline`, which is checked first. `wiring-scan.sh` itself exits only 0 or 2, never non-zero for a finding, so `rc != 0` from it always means refusal.

Your point about `HAS_FAIL` outranking `HAS_ERROR` resolves with it: with no FAIL left in the full-SHA case the runner now exits **2**, and §7a pins that (`[ "$rc" -eq 2 ]`) rather than the weaker `-ne 0` it had.

**The fixture was hiding this.** `make_fixture` wrote an unconditional `exit 0` wiring stub, so `WIRING` reported PASS on a range no block can resolve — the GH-950 defect itself, inside the test meant to catch it. The stub now mirrors the real script's refusal contract.

§7a additionally asserts over **both** refusal wordings, and names `WIRING`:

```sh
verdicts=$(grep -E '\| (PASS|FAIL) \|' "$TMP/sha-base.out" | grep -E 'fatal:|error: unknown revision' || true)
[ -z "$verdicts" ] || fail "…a PASS/FAIL row reports a verdict over a refusal: $verdicts"
```

**FAIL-first for the new arm, separately from the text arm.** Dropping `|| [ "$rc" -eq 2 ]` turns §7a red on exactly the row you found:

```
test-review-l0.sh: full-SHA base: a PASS/FAIL row reports a verdict over a refusal: | WIRING | any | P1 | FAIL | exit=2; error: unknown revision origin/d4883414e17d79d1437f06a42f9a356ebf7cabcc |
SUITE_EXIT=1
```

### [P1] `review-l0.sh:27-28` asserts an outcome it does not produce — accepted, fixed

Fixed twice over: the outcome is now produced, and the sentence was overstated independently of the bug. A rule whose block never reads the range still runs — U5 lints the tree. The text now says "every block **that reads the range** refuses and every rule routed to one of them reports ERROR", with the U5 exception stated.

### [P1] U2: `Closes #950` while undelivered — resolved by delivering it

Kept, now honest. `Closes #1054` added: that follow-up is delivered by this change, so it is closed by the PR rather than left open as a gap that no longer exists.

### [P1] U3: no `Issue:` line — accepted, fixed

```
$ gh pr view 1052 --json body --jq .body | awk 'tolower($0) ~ /^issues?[[:space:]]*:/'
Issue: #950, #1054
```

### [P2] locale — accepted, stated and measured

Now in the header contract rather than left implicit, and measured rather than asserted: this Git for Windows build (2.49.0) ships no `share/locale`, so `LC_ALL` has nothing to switch to and `fatal:` is always English here — which is why your probe saw English. That is a property of the build, not a guarantee. The contract now says so, and says which half of the detection is locale-independent (the exit code) and why it cannot cover the pipeline case on its own.

I did not force `LC_ALL=C` for block execution: it would also change collation and character-class behaviour for every block, and the suite's CJK case exists precisely because evidence cells carry non-ASCII. Trading a stated limit for an unstated regression is not an improvement.

### Gates RAN @ `d0437de`

```
sh scripts/test-review-l0.sh              exit 0, 11/11 assertions incl. both §7 cases
sh -n scripts/review-l0.sh scripts/test-review-l0.sh   ok
bash scripts/lint-file-length.sh --tree   exit 0
```

CI is running on the new head; I will not claim it until it lands.

### One thing you flagged that I want to keep visible

You called the FOLLOW-UP routing "not scope-dodging" because the disclosure was precise — but the disclosure being accurate is what made it *look* fine, and it still shipped an undelivered doneWhen under a closing keyword. The lesson I am taking is narrower than "don't route follow-ups": when a follow-up covers ground the doneWhen names, the closing keyword is the thing that has to go first, and that check belongs before writing the body, not after.

fagemx:
**Gate receipt — CI run 34124548741 @ `d0437dea0fa82560a6effee695ef03e3b1a6dc87`: 12/12 green.**

The load-bearing job for this change is `Fleet shell tests` (ubuntu, 57s), which executes `scripts/test-review-l0.sh` under **dash** — so the round-1 additions (`verdicts=$(… || true)`, the `case "$wiring"` block, `[ "$rc" -eq 2 ]`, and the heredoc'd fixture stub) are POSIX-clean by measurement, not by code read. `Fleet shell tests (windows)` 1m26s, `Format` (which runs `lint-file-length.sh --tree` and `lint-markdown-content.sh`) 17s, MSRV 44s, Clippy ×3 and Test ×3 all pass.

No crate changed, so the Cargo jobs are incidental coverage rather than the gate for this diff.

fagemx:
## Code Review: Round 2 — PR #1052 @ d0437dea0fa82560a6effee695ef03e3b1a6dc87

- elapsed: unmeasured (no pi session for this lane)
- model_observed: unverified (no session file / `modelUsage` reachable from this lane; §6 forbids copying `model_requested`)
- reviewer_session: unset — **replacement reviewer**: the round-1 session cannot be resumed in this build. Per `.claude/CLAUDE.md` § "Verification cost" ("a replacement reads receipts and CI before running anything") I READ the round-1 verdict, the `Review Response: Round 1`, and exact-head CI before running anything, and re-derived only what they do not cover.
- spec: review-spec-v1.5
- class: code-risk  (REVIEW.md classes: code-risk — classifier printed `classes=code-risk`, `canonical_class=code-risk`)
- escalations: none (no `docs`/`skills` class ⇒ D5 not routed; `REVIEW.md` unchanged in this diff — `git diff --name-only origin/main..HEAD` = the two scripts)
- shadow: false
- cost: ~20 tool calls; **no Cargo command, no new worktree, no ad-hoc `CARGO_TARGET_DIR`** (C5 selector empty); 3 full runs of `scripts/test-review-l0.sh` (baseline + two FAIL-first mutations), 2 real-repo runs of `scripts/review-l0.sh`, 4 substituted-spec adversarial probes

### IN SCOPE

Delta round on `git diff 1aaadde..d0437de` (77+/16-, two files), plus RAN confirmation that each round-1 blocker is closed.

Changed behaviour: one new arm in `run_block`'s exit-code classification — `2` joins `127`/`128` as could-not-run (`scripts/review-l0.sh:425`); the fixture's wiring stub replaced by one mirroring the real script's refusal contract (`scripts/test-review-l0.sh:93-101`); three new §7a assertions (`:292-294`, `:299-304`, `:307`); the usage `<base>` clause corrected (`:26-31`) and the header ERROR contract restated with both signals and each one's limit (`:63-84`); PR body now carries `Issue: #950, #1054` and `Closes #1054`.

Direct consumers: `HAS_ERROR` → the runner's process exit (`:480-482`); `.claude/skills/fleet-worker/SKILL.md:32` and `docs/guides/brief-template.md:206` (pre-push step 20); `crates/edda-cli/src/cmd_review/qualification.rs:63` (names the path as judging surface only). Acceptance ceiling: `#950` doneWhen (4 items) **and** `#1054` doneWhen (3 items) — both are closing-keyword targets now, so both were checked item by item.

**The round-1 P0 is genuinely closed, and the load-bearing new claim survived an adversarial audit.** I did not take the exit-2 safety argument on the author's word: I read every check-marked block in `REVIEW.md`, read `scripts/wiring-scan.sh` end to end, and then tried to break the invariant with substituted-spec probes. It held. Details under RAN.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `scripts/review-l0.sh`, `scripts/test-review-l0.sh` — exactly the "Predicted surface" of **both** #950 and #1054 |
| U2 | any | P1 | PASS *(was FAIL)* | body hits `1:Closes #950`, `2:Closes #1054`; every doneWhen item of both issues verified delivered (below). Commit trailers `43:Closes #1054`, `76:Closes #950` consistent |
| U3 | any | P1 | PASS *(was FAIL)* | `gh pr view 1052 --json body --jq .body \| awk 'tolower($0) ~ /^issues?[[:space:]]*:/'` → `Issue: #950, #1054` |
| U4 | any | P1 | PASS | `git log --format=%s origin/main..d0437de \| grep -Ev '<conventional>'` → empty. Both subjects `fix(fleet): …` |
| U5 | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → exit 0 |
| U6 | any | P0 | PASS | READ `gh pr checks 1052` → **12/12 pass** at head. Both U6 blocks PASS in the runner (`edda 0.5.0 (3306c1a2d9a0 2026-09-07)`; base `Cargo.toml` version read) |
| U7 | any | P1 | PASS | whole delta audited and batched before posting; round-1 blockers each re-measured, not assumed |
| U8 | any | P1 | N.A.(no `Decision:` line) | DECISIONS block → empty output; the rule is inapplicable, never failed |
| WIRING | any | P1 | PASS | `sh scripts/wiring-scan.sh origin/main HEAD` → exit 0, `no new pub surfaces` (via the runner's WIRING row); table below |
| C1 | code-plain | P0 | PASS *(was FAIL)* | #950 items 1–4 and #1054 items 1–3 each matched to code + the assertion that proves it — measured, below |
| C2 | code-plain | P0 | PASS | `fix:` with a regression test that fails without the change. FAIL-first reproduced **independently for both arms**, by me, on this head |
| C3 | code-plain | P1 | PASS | no `crates/**` in the diff; no `.unwrap(`/`.expect(` added |
| C4 | code-plain | P1 | PASS | delta grep for `unsafe {` / `clippy::all` / `#![allow` → empty |
| C5 | code-plain | P1 | PASS | selector over the delta prints nothing (no `crates/` path) — no Windows gap, no Cargo gate applies |
| R1 | code-risk | P0 | PASS | destructive-command grep over the delta's added lines → empty. The only `rm` in either file is the pre-existing `trap 'rm -r "$TMP"' EXIT` on a `mktemp -d` |
| R2 | code-risk | P0/P1 | PASS | mixed `\|\|`/`&&` grep over the delta's added `*.sh` lines → empty |
| R3 | code-risk | P0 | PASS | `sh -n scripts/review-l0.sh` → 0; `sh -n scripts/test-review-l0.sh` → 0. Dash execution **measured** in CI, not inferred (job log below) |
| R4 | code-risk | P1 | PASS | the new artefact is the stub heredoc written into `$fix/scripts/wiring-scan.sh`, inside the existing `mktemp -d` covered by `trap … EXIT` (`:122`). No new temp root, no new lock |
| R5 | code-risk | P1 | PASS | the stub's `"$1"`/`"$2"` are quoted and only reach `git rev-parse --verify --quiet "${ref}^{commit}"`; `$verdicts`/`$wiring` are pattern-matched and interpolated into a double-quoted `fail` argument, never evaluated |

*(D1–D5, S1–S3 not routed: §4 routes them from the `docs`/`skills` classes, and the classifier prints `code-risk` only.)*

### Wiring

|---|---|---|---|---|
| exit `2` as a could-not-run code | `scripts/review-l0.sh:425` — `[ "$rc" -eq 127 ] \|\| [ "$rc" -eq 128 ] \|\| [ "$rc" -eq 2 ]` → `HAS_ERROR=1`, `print_row … "ERROR $rc"` (`:426-428`) | `scripts/review-l0.sh:480-482` (`HAS_ERROR` → `exit 2`); `scripts/test-review-l0.sh:292-294`, `:299-304` (WIRING by name), `:307` (`[ "$rc" -eq 2 ]`) | not swallowed: sets `HAS_ERROR`, drives the process exit. **Measured end-to-end on the real repo**, not only in the fixture: `\| WIRING \| any \| P1 \| ERROR 2 \|`, runner `exit=2` | classifier → `HAS_ERROR` → process exit; both ends asserted **and** both ends measured |
| fixture wiring stub refusal contract | `scripts/test-review-l0.sh:93-101` — heredoc mirroring `scripts/wiring-scan.sh:30-35` (same rev-parse guard, same `error: unknown revision`, same `exit 2`) | REVIEW.md's `WIRING` block via the runner; asserted at `test-review-l0.sh:292-304` | n/a — it is the **stimulus**, not the detector. Two of the three §7a assertions (`case "$wiring" in *'\| ERROR '*` and `[ "$rc" -eq 2 ]`) are wording-independent, so stub drift cannot fake a green | stub → WIRING block → `run_block` → row + runner exit; proven red by the FAIL-first mutation below |
| header ERROR contract + `<base>` usage text | `scripts/review-l0.sh:63-84` (both signals, each with its stated limit), `:26-31` (`<base>` shapes) | human + agent readers (`docs/guides/brief-template.md:206`, `.claude/skills/fleet-worker/SKILL.md:32`) | n/a — doc surface. It is #1054 doneWhen item 2, which requires the mechanism be *stated* alongside the `fatal:` rule; it is | doc-only |

### Findings

**none blocking.** P0 = 0, P1 = 0.

The three things I most expected to be wrong, and what I actually measured:

**1. Is exit 2 safe? — yes, verified two ways, with a stated bound.**

Claim (b) first, because it is checkable outright: `scripts/wiring-scan.sh` has exactly three exit points — `exit 2` at `:23` (usage) and `:33` (unknown revision), and an explicit `exit 0` at `:95`. Its findings are *printed lines* (`… — N outside references`, `file: line`) emitted on the exit-0 path. So `rc != 0` from it is always a refusal. Claim (b) holds.

Claim (a) — "no check-marked block in `REVIEW.md` signals a finding with exit 2" — I re-derived rather than accepted. The runner executes only blocks whose marker id is in `RULE_TABLE` (`:244-269`, iterated at `:468-475`), so `FETCH`/`ISSUES`/`DECISIONS`/`DIFF`/`ROUTER`/`ELAPSED` never run — the `node` in ELAPSED is not on this surface at all. Of the 20 rule blocks that do run:

- **grep-tailed blocks** (U2, U4, C3, C4, C5, D4, S1, R1, R2) exit 0/1. `grep` exits 2 only when *an error occurred*, which is a could-not-run by definition — and every one of these patterns is a fixed literal, so there is no bad-pattern path. The one block with an interpolated pattern (D3's `"^($tops)(/|$)"`) does not have that grep as its last stage.
- **`grep -c` tail** (C2) → 0/1, handled by `countgrep` at `:431`.
- **`while`-tailed pipelines** (D1, D3) exit 0; they signal by `CANDIDATE`/`MISSING` lines.
- **print-their-own-exit blocks** (U5 `; echo "exit=$?"`, R3 `sh -n "$f"; echo …`) always end on `echo`, so `rc` is 0 and the signal is `badline`. R3 in particular: `sh -n` on a broken script cannot become the block's `rc`.
- **U1** ends on `cat` (0/1, `REVIEW_FILES` is always exported by the runner at `:212`). **U6** ends on `gh pr checks` — I probed it: bogus PR → **1**, unknown flag → **1**; `gh`'s exit 2 is `exitCancel` (interrupt), itself a could-not-run. **U2** never reaches the exit-code arm at all: its reviewer-input row returns at `:361-365`, above it.
- **U7, C1, R4, R5, S2, S3** have no block; **D2** carries a `<decision-key>` placeholder and returns N.A. at `:304-307`.

Then I tried to break it, by substituting the U5 block in a scratch spec copy and running the real runner against it (`REVIEW_L0_SPEC`, worktree untouched):

```
[echo "MISSING scripts/nope"; exit 2]              → | U5 | any | P1 | FAIL | exit=2; MISSING scripts/nope        runner_exit=1
[echo "exit=1"; exit 2]                            → | U5 | any | P1 | FAIL | exit=2; exit=1                      runner_exit=1
[echo "edda … -> exit=2 CANDIDATE"; exit 2]        → | U5 | any | P1 | FAIL | exit=2; … CANDIDATE                 runner_exit=1
[echo "12:+rm -rf /"; exit 2]                      → | U5 | any | P1 | ERROR 2 | 12:+rm -rf /                     runner_exit=2
```

The first three are the author's "caught by `badline`, which is checked first" claim, proven mechanically: `badline` is at `:391`, the exit-2 arm at `:425`, and a finding marker wins every time. The fourth is the residual, and it is the honest bound on the claim: a block that exits 2 **and** prints a finding with no `badline` marker would be relabelled "did not run". No block in `REVIEW.md` has that shape — the blocks whose findings are bare `grep -n` lines exit **0** when they print them, never 2 — so the invariant holds today. It is an invariant on a file the runner reads at run time, though, which nothing enforces; see the P2 and the follow-up.

**2. Is the U5 carve-out honest, or a loophole? — honest, and if anything understated.** RAN in the worktree at this head:

```
$ sh scripts/review-l0.sh e613c3c53460fe40ac56c8282d18df1571f07aeb HEAD
| U4     | any        | P1    | ERROR   | exit=1; fatal: ambiguous argument 'origin/e613c3c5…..HEAD': …
| C2     | code-plain | P0    | ERROR   | exit=1; fatal: ambiguous argument …          ← round 1's headline row
| C3     | code-plain | P1    | ERROR   | exit=1; fatal: bad revision …
| C4     | code-plain | P1    | ERROR   | exit=1; fatal: ambiguous argument …
| R1     | code-risk  | P0    | ERROR   | exit=1; fatal: ambiguous argument …
| R2     | code-risk  | P0/P1 | ERROR   | exit=1; fatal: bad revision …
| WIRING | any        | P1    | ERROR 2 | error: unknown revision origin/e613c3c5…      ← round 1's P0 row
| U1     | any        | P0    | PASS    | exit=0; scripts/review-l0.sh
| U5     | any        | P1    | PASS    | exit=0; exit=0
| C5     | code-plain | P1    | PASS    | exit=1; (no output)
| R3     | code-risk  | P0    | PASS    | exit=0; scripts/review-l0.sh -> sh -n exit=0
exit=2
```

**Seven of seven rules whose block reads the range are ERROR. Zero rows report PASS or FAIL over a refusal. The runner exits 2, not 1.** doneWhen item 1 of #950 and item 1 of #1054 are both delivered, measured, not claimed.

The four PASS rows are not the carve-out being abused. U1, C5 and R3 read `$REVIEW_FILES`, which the runner computes at `:202-206` from the *unprefixed* `$ARG_BASE` — a full SHA resolves there — so they get a correct file list and their verdicts are meaningful, not vacuous; U5 lints the tree. The usage text's general clause ("every block **that reads the range** refuses and every rule routed to one of them reports ERROR") is exactly right; U5 is named as an instance, not as an exhaustive list. See P2 #2 for the one wording nit.

**3. Is the fixture stub a tautology? — no.** The stub (`test-review-l0.sh:93-101`) is a faithful copy of `wiring-scan.sh:30-35` — same guard, same message, same exit code — and it supplies the *stimulus*; the assertions test `review-l0.sh`'s *classification* of it. Two of the three are wording-independent (`case "$wiring" in *'| ERROR '*` matches both `| ERROR |` and `| ERROR 2 |` result cells; `[ "$rc" -eq 2 ]` reads no text at all), so drift in the stub's wording cannot fake a pass on a runner regression. I proved both directions:

- **exit-code arm** — dropped `|| [ "$rc" -eq 2 ]` from `:425`, reran: **red**, on exactly round 1's row —
  `full-SHA base: a PASS/FAIL row reports a verdict over a refusal: | WIRING | any | P1 | FAIL | exit=2; error: unknown revision origin/554f6347… ` , `SUITE_EXIT=1`.
- **text arm** — replaced `:352` with `if false; then`, reran: still **red**, reproducing the original GH-950 defect in full, including the headline
  `| C2 | code-plain | P0 | PASS | exit=1; fatal: ambiguous argument …` alongside U4/C3/C4/R1/R2 FAIL.

So §7a is a live detector for both the new regression and the old one, and neither arm is shielded by the other. Restored with `git checkout -- scripts/review-l0.sh` after each; `git status --porcelain` empty, worktree at `d0437de`.

**doneWhen, item by item — both closing keywords are now honest (U2):**

| Issue | Item | Verified |
|---|---|---|
| #950 | 1. full-SHA base → ERROR for every affected rule, exits nonzero, no PASS/FAIL over `fatal:` | the measured table above: 7/7 ERROR, exit 2 |
| #950 | 2. `fatal:` ⇒ ERROR independent of exit code, at the classifier | `review-l0.sh:352-359`, above every other branch; §7b proves the exit-**0** case (`generic fatal: … — OK`) |
| #950 | 3. suite covers both cases, still exits 0 with existing assertions held | baseline `SUITE_EXIT=0`, all 10 case lines OK including the 8 pre-existing |
| #950 | 4. usage comment states the accepted `<base>` shapes | `review-l0.sh:23-31` |
| #1054 | 1. refused `WIRING` is ERROR, runner exit reflects an error not a finding | `\| WIRING \| any \| P1 \| ERROR 2 \|`, `exit=2` |
| #1054 | 2. mechanism stated in the header contract alongside the `fatal:` rule | `review-l0.sh:77-84`, adjacent to `:66-76`, including why `^error:` may not be matched |
| #1054 | 3. `test-review-l0.sh` covers it, existing assertions still hold | `:299-304` (WIRING by name), `:307` (exit 2); baseline green |

**The `error:` rejection is correct.** `grep -n 'error:' scripts/check-cli-docs.sh` → `:144` `error: undocumented flag for '$verb'` and `:154` `error: undocumented verb` are genuine findings on that prefix (its refusals at `:31/:39/:45/:56` use it too). Widening the text net to `^(fatal|error):` really would relabel findings as "did not run". The reasoning holds. (Those findings exit **1**, `:168` — so they would not have tripped the exit-2 arm either, had that script been on a block; it is not.)

**Blast radius of the new arm is contained.** The only classification that changes is `rc == 2`: FAIL → ERROR. `HAS_FAIL` still outranks `HAS_ERROR` (`:477-482`), so any run with a genuine finding still exits 1. No caller branches numerically on 1 vs 2: `.claude/skills/fleet-worker/SKILL.md:32` keys on "FAIL 列" (FAIL *rows*), and `crates/edda-cli/src/cmd_review/qualification.rs:63` only names the path as judging surface. On the real repo at normal range the table is unchanged — `sh scripts/review-l0.sh origin/main HEAD 1052` → exit 0, no ERROR row, no FAIL row.

### P2 (non-blocking)

- **[P2] The exit-2 arm creates an invariant on `REVIEW.md` that nothing enforces.** `review-l0.sh:419-424` asserts "no block in REVIEW.md signals a finding with it" — true today, and I verified it — but the runner reads the spec at run time (`:104`), so a future block whose tail tool exits 2 for a finding without a `badline` marker would be silently relabelled "did not run". Probe 4 above is that shape, demonstrated. Cheapest guard would be a fixture case asserting the property, or the badline convention stated in `REVIEW.md` §0 next to the exit-code prose. Follow-up filed below; not a defect in this diff.
- **[P2] `review-l0.sh:28-30` names only U5 as a rule that still runs on a full-SHA base.** The general clause is correct, but `U1`, `C5` and `R3` are in the same position (they read `REVIEW_FILES`, not the range) and all three PASS in the measured table. A next reader could take U5 as the exhaustive exception. One word — "for example" — would close it.
- **[P2] `Review Response: Round 1` says the suite is "11/11 assertions"; it prints 10 case lines.** The PR body's own quoted gate block lists exactly 10 `— OK` lines plus the summary. Receipt arithmetic only — no product claim depends on it, and everything else in the response reproduced exactly.
- **Round 1's [P2] (locale) is addressed**, in the form round 1 asked for: the limit is now stated in the header contract (`:70-76`) with the measurement behind it — I confirmed `git version 2.49.0.windows.1` and that `share/locale` does not exist on this build, so the English `fatal:` here is a property of the build, exactly as written. The rejection of `LC_ALL=C` is sound (it would move collation and character classes under every block, and the CJK evidence case exists precisely because cells carry non-ASCII). One narrower lever was not considered, though: **`LC_MESSAGES=C`** forces git's message language without touching `LC_COLLATE` or `LC_CTYPE`. Worth the follow-up rather than this PR.

### FOLLOW-UP ISSUE

- **Enforce the "no block signals a finding with exit 2" invariant** that `review-l0.sh:419-424` now depends on — a fixture case, or the convention written into `REVIEW.md` §0 beside the existing pipeline-exit prose, so the next block author cannot break the classifier by choosing an exit code. Evidence: probe 4 above (`echo "12:+rm -rf /"; exit 2` → `ERROR 2`, runner exit 2).
- **Locale-harden the refusal match with `LC_MESSAGES=C`** rather than `LC_ALL=C`, which the response correctly rejects. Narrower than the option that was weighed; keeps collation and character classes untouched, so the CJK evidence case is unaffected. (Supersedes round 1's locale follow-up.)
- Round 1's general follow-up stands: **`review-l0.sh` still classifies a refusal by two hardcoded signals**; a third refusal wording carrying an exit code outside `{2,127,128}` remains a hole. `#1054`'s doneWhen asked only that the chosen mechanism be *stated*, which it is.

### RAN vs READ

RAN (all in the existing worktree `C:\ai_agent\edda-wt-gh950` at `d0437de`; **no Cargo, no new worktree, no ad-hoc `CARGO_TARGET_DIR`** — C5 selector empty, so no Cargo gate applies):

- `git rev-parse HEAD` → `d0437dea0fa82560a6effee695ef03e3b1a6dc87`; `git status --porcelain` empty before, between every mutation, and after.
- `sh scripts/review-l0.sh $(git rev-parse origin/main) HEAD` → the full-SHA table above, `exit=2`. **This is the shape no CI job runs, and it is where round 1's P0 lived.**
- `sh scripts/review-l0.sh origin/main HEAD 1052` → exit 0, no ERROR/FAIL row; `U3` now `PASS | exit=0; Issue: #950, #1054`.
- `sh scripts/test-review-l0.sh` → `SUITE_EXIT=0`, 10 case lines OK.
- **FAIL-first ×2, independently, each restored.** Exit-code arm: `sed` dropped `|| [ "$rc" -eq 2 ]` at `:425` → §7a red on the WIRING row, `SUITE_EXIT=1`. Text arm: `:352` → `if false; then` → §7a red reproducing the original `fatal:` table including `C2 … PASS`. `git checkout -- scripts/review-l0.sh` after each; `sed -n '352p;425p'` confirms both lines restored.
- **4 adversarial substituted-spec probes** via `REVIEW_L0_SPEC` pointed at a scratchpad copy (worktree never modified) — the badline-vs-exit-2 ordering table above.
- Read in full: `scripts/wiring-scan.sh` (95 lines, 3 exit points), `scripts/review-l0.sh` (483 lines), `REVIEW.md`'s 27 `# review-spec:check` markers and every rule block they wrap.
- `grep -n 'error:' scripts/check-cli-docs.sh` → `:144`/`:154` findings confirmed; `grep -n 'exit '` → findings exit 1 at `:168`, refusals exit 2.
- `gh pr checks 999999` → exit 1; `gh pr checks 1052 --badflagxyz` → exit 1 (neither is 2).
- U2/U3/U4/U5/R1/R2/R3/C4/C5/D4 enumerators on `origin/main..d0437de` and on the delta `1aaadde..d0437de`; DECISIONS block → empty.
- `sh -n` ×2 → 0; `sh scripts/lint-markdown-content.sh` → 0; `sh scripts/lint-file-length.sh --tree` → 0 (all three body gate claims reproduced).
- `git --version` → `2.49.0.windows.1`; `ls .../share/locale` → does not exist (the response's locale measurement reproduced).
- `gh` reads: PR body/head SHA, `#950` and `#1054` doneWhen, round-1 verdict and `Review Response`, `gh pr checks 1052`, and the ubuntu job log.

READ (not rerun): **CI run 34124548741 @ `d0437dea0fa82560a6effee695ef03e3b1a6dc87` — 12/12 pass.** Format; Clippy ×3 OS; Test on ubuntu + macOS (full workspace) + windows (7-crate subset); MSRV 1.91.0; Fleet shell tests (ubuntu **and** windows); Detect code changes; CI Gate. What it covers that I therefore did not rerun: the **dash** execution of `scripts/test-review-l0.sh` with all three new §7a assertions — `gh api …/jobs/101750058296/logs` shows `RUN scripts/test-review-l0.sh` → `test-review-l0.sh: all fixture assertions held` → `PASS scripts/test-review-l0.sh`, and the script is neither quarantined nor SKIPped (the job's own log names the 5 that are, and it is not among them). That is the POSIX/dash-safety measurement for `verdicts=$(… || true)`, the quoted-`|` `case` pattern, and `[ "$rc" -eq 2 ]` under `set -u` with no `set -e` (`test-review-l0.sh:27` sets only `-u`). What it does not cover: the real-repo full-SHA-base run against the **unstubbed** `wiring-scan.sh`, and the badline/exit-2 ordering — no CI job runs either shape, which is why I ran them, and they are where round 1's P0 lived. No Cargo gate applies to this PR.

### Verdict

LGTM (P0=0, P1=0)

Round 1's P0 and all three P1s are closed and each was re-measured here rather than accepted: the full-SHA base now reports ERROR for all seven range-reading rules including `WIRING`, and the runner exits 2 rather than 1. The load-bearing new claim — that exit 2 can be read as could-not-run without relabelling a finding — is correct as scoped, and I verified both halves myself (every rule block in `REVIEW.md`, every exit path in `wiring-scan.sh`) and then failed to break it with four adversarial probes; `badline` precedes the new arm, so every finding-marked block still reports FAIL. `Closes #950` and `Closes #1054` are both honest against all seven doneWhen items. Three P2s and three follow-ups recorded; none blocks.

Issue #1071, opened 2026-09-07
test(edda-cli): cmd_review::evidence::tests::deadline_stops_descendant_and_does_not_start_next_gate is flaky under load on windows — three failures in one day, green on rerun and 5/5 in isolation


fagemx:
## What

`cmd_review::evidence::tests::deadline_stops_descendant_and_does_not_start_next_gate` fails intermittently under concurrent load on Windows, and passes on rerun.

Observed three times on 2026-09-07 during PR #1060's review rounds, each time in a full `cargo test -p edda` run while another session was compiling against the same `target/` directory. Each time:

- the failure was the only one in the run (`649 passed; 1 failed`),
- an immediate rerun of the same suite was green,
- `cargo test -p edda --bin edda deadline_stops_descendant` passed **5/5** in isolation.

The test is in `crates/edda-cli/src/cmd_review/evidence*`, which PR #1060 does not touch; the branch it flaked on changed only `cmd_claim`/`claim_standing`/`dispatch_claim`/`peers` and docs.

## Why it matters

It is a deadline test, so it asserts on wall-clock behavior — the class that fails when the machine is busy rather than when the code is wrong. Same shape as #1031, different test. Left alone it costs a rerun each time it fires and, worse, trains readers to rerun a red suite instead of reading it. On a shared-`target/` workstation (several sessions compiling at once) the load that triggers it is normal, not exceptional.

## Where

- `crates/edda-cli/src/cmd_review/evidence.rs` (or `evidence/`) — the test and the deadline it asserts.

## doneWhen

- The test's timing assertion tolerates a busy machine, or the deadline is injected rather than read from the real clock (the pattern `edda review due` uses for its debounce: a `--now`/`Facts` seam that lets the policy be tested without waiting).
- Running the suite under deliberate CPU contention does not fail it.
- If the flake turns out to be a real race in the deadline path rather than a slow machine, it is fixed there instead — a test that only fails under load is sometimes reporting a defect that only appears under load.

## Provenance

Declared in the `Review Response: Round 2` on #1060 rather than left as an unexplained red. Not caused by that PR; filed so the next reader of a one-test red in `cargo test -p edda` can recognize it.

fagemx:
More evidence, and the frequency has gone up.

[...]

PR #1070, opened 2026-09-07
feat(edda-cli): edda review due — the trigger policy in the product (GH-763)

fagemx:
Issue: #763

## What

"Is this PR worth reviewing again" is the main cost switch in the whole review
system, and it lived in `scripts/pr-review-watch.sh`'s awk with one rule: *head
differs from the last reviewed SHA*. That is **one review per push**. A round-1
Opus review measured $1.28–$2.57 on #754; the resumed delta round of the same
PR measured $0.22 and then $0.02. Five pushes on one PR bought five round-1
prices for what a debounce would have charged once.

It is also a judgement, not forge glue (#766 D8). So `edda review due` decides
and the daemon gathers facts:

```
edda review due --head <sha> [--pr N] [--pushed-at T] [--ready]
                [--response-at T] [--draft] [--last-reviewed SHA]
                [--unreviewed-label]
  exit 0  REVIEW <reason> [--resume <session>]
  exit 1  SKIP <reason>
  exit 2  cannot judge
```

Both lines print `review cost so far: $X over N rounds`, so the switch is
visible at the moment it is thrown.

## The policy, in order

| Condition | Answer |
|---|---|
| `--draft` | `SKIP draft` |
| `--unreviewed-label` and the head has not demonstrably moved | `SKIP review-unreviewed` |
| `--ready` | `REVIEW ready` |
| `--response-at` newer than the last verdict | `REVIEW response` |
| head moved, push settled | `REVIEW push` |
| head moved, inside the debounce | `SKIP debounce <n>s` |
| otherwise | `SKIP reviewed` |

`ready` and `response` are one-time events the operator is waiting on, so
neither is debounced; only `push` repeats, so only `push` waits. A draft is
refused first and no trigger setting can open it.

## Two facts the verb cannot see

- **`--last-reviewed`** — the daemon's own reviewed-head record. A round
  published through the §7 comment path writes no `review_verdict` event, so
  the ledger would read it as never-reviewed and make every cycle due. This is
  the same source split as #769's `--verdicts`, and for the same reason: the
  rule is the product's, the facts are the caller's.
- **`--unreviewed-label`** — `review:unreviewed` marks a head the watcher gave
  up on. It is released by the head demonstrably moving, and only by that: with
  no recorded SHA there is no way to tell that it moved, and guessing in the
  reviewing direction buys a round-1 price every cycle, forever.

## `--resume` is published, not re-wired

The REVIEW line carries ` --resume <round-1 session id>` from round 2 on, as
[...]

Issue #1072, opened 2026-09-07
docs(fleet): a reverted mutation still lives in target/debug/edda.exe — integration suites rebuild the lane binary, so reviewer probes read the mutated build


fagemx:
## What

A source mutation left in the tree contaminates any later hand-run probe against `target/debug/edda.exe`, because the integration-test suites rebuild that binary as a side effect.

`crates/edda-cli/tests/*.rs` spawn the CLI via `env!("CARGO_BIN_EXE_edda")`, so `cargo test -p edda --test claim_e2e_contract` builds `target/debug/edda.exe` fresh. That is the GH-789 property working as designed — for the *tests*. The unintended consequence is that the lane binary is now built **from whatever is in the tree at that moment**, including a mutation a reviewer introduced deliberately to check a test's strength.

The sequence that bites:

1. Reviewer mutates a production rule to confirm a test fails on revert (correct practice).
2. Reviewer runs the suite; the suite rebuilds `edda.exe` from the mutated tree.
3. Reviewer reverts the mutation.
4. Reviewer runs `./target/debug/edda.exe claim check ...` as a product probe — **still the mutated binary**, because nothing rebuilt it after the revert.

Step 4 produces confident, reproducible, entirely false measurements. The failure mode is a *clean* false positive: the probe agrees with itself across repeated runs, so the usual smell of "this looks wrong, run it again" never fires.

## Observed

Round 3 of PR #1060. The reviewer mutated `claim_standing.rs`, ran the e2e suite, reverted, then ran six product probes against the docs table and got a perfect false positive against it. They caught it, rebuilt from the reverted tree, re-ran everything, and reported only the re-run numbers — but only because they thought to ask why the binary disagreed with the source.

## Why the existing note does not cover it

`.claude/CLAUDE.md` (Testing Standards) documents the GH-789 hazard in the warm-lane direction: `cargo test -p <crate> --bin <bin>` runs pure unit tests and does not depend on the binary, so process-spawning tests live in `tests/*.rs` where Cargo guarantees a fresh build. That is about the binary being *stale relative to the source*. This is the mirror case — the binary is fresh relative to a source state that no longer exists, and it is specific to the mutation-testing workflow the review spec asks reviewers to perform.

## doneWhen

- The verification guidance names the hazard where a reviewer will meet it — mutation testing followed by product probes — with the one-line remedy: rebuild (`cargo build -p edda`) after reverting a mutation, before any probe against `target/debug/edda.exe`.
[...]

Issue #1029, opened 2026-09-06
test(fleet): test-pr-review-watch.sh r22 live case 2 is red on ubuntu — glm-on-shipping writes a status (1 call), the fixture expects none


fagemx:
## What happened

#927's gate is the first machine run of `scripts/test-pr-review-watch.sh`. After the dash parse defect was fixed (process substitution), the test now executes on ubuntu and fails at the r22 live block, case 2: `live: r22 case 2 — glm on shipping must not write a status, got 1 calls`.

The r22 live cases drive `scripts/pr-review-watch.sh` against fixture `gh` calls; case 2 records one call where the fixture expects the watcher to write no status. Red on ubuntu in CI run 34045480617 (job 101519672815); the same test passes on the 4090 workstation, which is where it has always been hand-run — #896's premise measured again.

## Done when

Diagnose the divergence (environment-dependent path in the watcher's live-probe branch, or a fixture assumption), fix it, and remove the QUARANTINE entry in `scripts/fleet/run-fleet-tests.sh` in the same change, per the #963/#964/#987/#1024 pattern.

fagemx:
依 #1010 裁定 B（strangler：shell 過渡層性質併入產品化單）與 2026-09-07 治理憲法 board triage（#1032）：本單性質併入 #1030 範疇後結案（載體結案，性質不滅）。

保留性質：The property — 'an engine R22 does not name as authoritative for the surface never writes the Independent Review status' — is already carried verbatim in #1030's doneWhen (non-authoritative-engine fixtures: zero GitHub writes), and #1030's follow-up adapter PR deletes the watcher status blocks plus this very test coverage, so debugging the ubuntu divergence of soon-deleted shell is not worth a lane.

PR #1074, opened 2026-09-07
fix(edda-cli): edda review due — the response trigger, and the resume layer's tests (GH-763 follow-up)

fagemx:
Follow-up to #1070, which merged at `98ab748` while its round-2 review was still running. The review landed nine minutes after the merge and found three P1s; `main` carries all three today. This PR is those fixes plus round 3's.

Issue: #763

## Why this is a separate PR

#1070's head is frozen at `98ab748` and `refs/pull/1070/head` cannot move. The fixes were pushed to the branch after the merge, so they existed only there. Rebased onto current `main` (`--onto origin/main 98ab748`, dropping the two commits the squash already applied) and reopened here so they get exact-head CI and a review against the base they will actually merge into.

## What was on `main` and is fixed here

**The `response` trigger re-fires on a head already reviewed.** #1070's round-1 fix required a ledger verdict to *exist*, which is not the same as requiring it to be *current* — and `response` is the one rule that never reads the recorded head. Measured on `main`: head equal to the daemon's own `--last-reviewed`, an old verdict, a newer response → `REVIEW response`, exit 0, on every poll. The branch is undebounced and nothing marks a response consumed, so this is one round-1 price per cycle, indefinitely.

The rule now fires only when the response is newer than **every** record of having reviewed: the ledger verdict, and a new `--last-reviewed-at` fed from a new fourth column on the daemon's state row (`state_set` is its only writer; rows without it read as "no record", which is what they meant already).

**`--pushed-at`'s help states the behavior #1070 replaced** — "without it a moved head is due at once", when it now waits.

**The resume-id layer has no test.** Three separate mutations each left `main`'s suite green: deleting the daemon's decision-log line, changing round-1 selection to `Some(7)`, and dropping the ` --resume {session}` suffix. All three are pinned now, two through a ledger-backed fixture that appends real `review_verdict` events and reads them back through `history()`.

## Round 3's finding, also fixed here

The `refs.pr` pre-filter over-corrected. "Skip only what is positively another PR's" meant a verdict with `refs.pr` **absent** — what `edda review --base X --head Y` writes when no PR is named — counted as *every* PR's history: another subject's rounds, cost and resume session, and a `last_verdict_at` high enough to answer a genuine Review Response with `SKIP reviewed`. The raw gate stays a cheap pre-filter; the decision is now the typed equality check after deserializing, so an unreadable `refs.pr` still fails loudly rather than being silently attributed.

Also: `first_session` falls back when no event carries a round at all, and the operator guide documents the state row's fourth column — the script header was updated and the guide was not, which is the same drift class the round-2 review blocked on.

## Tests

| Test | Pins |
|---|---|
| `a_response_older_than_the_callers_own_review_is_already_answered` | The exact shape measured on `main`: an old verdict, a newer response, a head the daemon already reviewed |
| `the_callers_own_record_counts_even_with_no_verdict_at_all` | The common case — a §7-comment round writes no ledger event |
| `a_verdict_with_no_pr_belongs_to_no_pr` | A PR-less verdict contributes no rounds, no cost, no resume id, and cannot age this PR's verdict |
| `the_session_to_resume_is_round_ones_by_number_not_by_ledger_position` | Round 1 by `refs.round`, through a real ledger |
| `a_ledger_without_round_one_resumes_the_lowest_round_it_has` | A machine that imported only later rounds still resumes, rather than paying round-1 price |
| `the_resume_suffix_names_round_ones_session_and_only_from_round_two` | The emitted suffix |
| shell: decision line logged; `--last-reviewed-at` on the argv; state field 4 is RFC3339; cold-`$SCRATCH` `decide` | The wiring — each verified by deleting what it claims to pin |

Every fix in this PR was verified by re-running the reviewer's own mutation and confirming it now fails.

## Gates

L0 on Windows, `worker-1` lane: `cargo fmt --all --check`; `cargo clippy -p edda --all-targets -- -D warnings`; `cargo test -p edda`; `sh scripts/test-pr-review-watch.sh` clean and under six separate mutations; `scripts/lint-file-length.sh --tree`; `scripts/lint-markdown-content.sh`; `scripts/lint-doc-citations.sh --tree`; `sh -n` on both shell files.

`cmd_review::evidence::tests::deadline_stops_descendant_and_does_not_start_next_gate` is a known flake (#1071) in a file this PR does not touch — it now fails in most full-suite runs and passes in isolation.

### DIFF
--- a/crates/edda-cli/src/cmd_review/due.rs
+++ b/crates/edda-cli/src/cmd_review/due.rs
@@ -170,7 +170,13 @@ pub(crate) fn history(repo: &Path, pr: u64) -> Result<History> {
         }
         let payload: ReviewVerdictPayload = serde_json::from_value(event.payload.clone())
             .with_context(|| format!("review_verdict event {}", event.event_id))?;
-        if payload.verdict == "unreviewed" {
+        // The raw gate above is a cheap pre-filter, not the decision. A
+        // verdict with no `refs.pr` at all — what `edda review --base X --head
+        // Y` writes when no PR is named — reaches here, and counting it as
+        // this PR's history would attribute another subject's rounds, cost and
+        // resume session to it, and raise `last_verdict_at` enough to answer a
+        // genuine Review Response with `SKIP reviewed`.
+        if payload.verdict == "unreviewed" || payload.refs.pr != Some(pr) {
             continue;
         }
         history.rounds += 1;
@@ -186,8 +192,11 @@ pub(crate) fn history(repo: &Path, pr: u64) -> Result<History> {
         // present is the best available resume target; answering "no session"
         // would send a continuing review back to round-1 price for want of an
         // event this machine never received.
+        // A verdict that records no round at all still names a reviewer worth
+        // resuming; `u32::MAX` would leave `first_session` empty for a ledger
+        // where no event carries a round.
         let round = payload.refs.round.unwrap_or(u32::MAX);
-        if round < history.first_round {
+        if round < history.first_round || history.first_session.is_none() {
             history.first_round = round;
             history.first_session = Some(payload.reviewer.session_id.clone());
         }
@@ -636,6 +645,44 @@ mod tests {
         );
         assert_eq!(history.rounds, 2);
         assert_eq!(history.resume_suffix(), " --resume round-1-session");
+    }
+
+    #[test]
+    fn a_verdict_with_no_pr_belongs_to_no_pr() {
+        // `edda review --base X --head Y` with no --pr writes a verdict whose
+        // refs.pr is absent. Counting it here would attribute another
+        // subject's rounds, cost and resume session to this PR — and raise
+        // last_verdict_at enough to answer a real Review Response with
+        // "reviewed", which is the under-review failure this policy exists to
+        // avoid.
+        let temp = tempfile::tempdir().expect("tempdir");
+        let root = temp.path().to_path_buf();
+        let ledger = edda_ledger::Ledger::open_or_init(&root).expect("ledger");
+        let mut value = verdict_payload(42, 1, "orphan", "orphan-session");
+        value["refs"] = serde_json::json!({ "round": 1 });
+        let payload: edda_core::ReviewVerdictPayload =
+            serde_json::from_value(value).expect("payload");
+        let event = edda_core::event::new_review_verdict_event(
+            "main",
+            ledger.last_event_hash().expect("hash").as_deref(),
+            &payload,
+            None,
+            None,
+            &[],
+        )
+        .expect("event");
+        ledger.append_event(&event).expect("append");
+
+        let history = history(&root, 42).expect("history");
+        assert_eq!(
+            history.rounds, 0,
+            "a PR-less verdict is not this PR's round"
+        );
+        assert_eq!(history.resume_suffix(), "");
+        assert!(
+            history.last_ts.is_none(),
+            "and it cannot age this PR's verdict"
+        );
     }
 
     #[test]
--- a/crates/edda-cli/src/cmd_review/due.rs
+++ b/crates/edda-cli/src/cmd_review/due.rs
@@ -39,7 +39,7 @@ pub struct DueArgs {
     /// The PR's current head, as a full lowercase 40-hex SHA
     #[arg(long, value_name = "SHA")]
     pub head: String,
-    /// When that head was pushed (RFC3339); without it a moved head is due at once
+    /// When that head was pushed (RFC3339); without it a moved head waits
     #[arg(long, value_name = "RFC3339")]
     pub pushed_at: Option<String>,
     /// The PR left draft during this cycle
@@ -48,6 +48,14 @@ pub struct DueArgs {
     /// Timestamp of the newest `Review Response: Round N` comment (RFC3339)
     #[arg(long, value_name = "RFC3339")]
     pub response_at: Option<String>,
+    /// When the caller last reviewed this PR (RFC3339), from its own record
+    ///
+    /// A ledger verdict answers the same question but only when one exists:
+    /// verdicts do not cross machines (GH-671) and a round published through
+    /// the §7 comment path writes none at all. Without this, a response older
+    /// than a review that did happen still reads as unanswered.
+    #[arg(long, value_name = "RFC3339")]
+    pub last_reviewed_at: Option<String>,
     /// The PR is a draft
     #[arg(long)]
     pub draft: bool,
@@ -77,8 +85,11 @@ pub(crate) struct History {
     pub(crate) last_sha: Option<String>,
     /// Timestamp of that verdict's event.
     pub(crate) last_ts: Option<String>,
-    /// Reviewer session of round 1 — what a later round resumes.
+    /// Reviewer session of the lowest round present — what a later round
+    /// resumes. Round 1 when this ledger has it.
     pub(crate) first_session: Option<String>,
+    /// The round `first_session` came from; `u32::MAX` until one is seen.
+    pub(crate) first_round: u32,
     /// Rounds that produced a verdict.
     pub(crate) rounds: u32,
     /// Summed cost of those rounds, and whether every one of them was measured.
@@ -94,6 +105,19 @@ impl History {
     /// quietly contributing zero: a sum that silently drops a round reads as
     /// cheaper than the truth, which is the one direction a cost display must
     /// never fail in.
+    /// The ` --resume <session>` suffix on a REVIEW line, or empty.
+    ///
+    /// A second round resumes the first round's session: the reviewer already
+    /// holds the diff, the spec and its own findings, which is the difference
+    /// between a $2 round and a $0.02 one. Round 1's session is the one to
+    /// resume — a later round's reviewer holds a delta, not the whole subject.
+    pub(crate) fn resume_suffix(&self) -> String {
+        match (self.rounds >= 1, self.first_session.as_deref()) {
+            (true, Some(session)) => format!(" --resume {session}"),
+            _ => String::new(),
+        }
+    }
+
     pub(crate) fn cost_line(&self) -> String {
         let rounds = if self.rounds == 1 { "round" } else { "rounds" };
         if self.rounds == 0 {
@@ -124,6 +148,7 @@ pub(crate) fn history(repo: &Path, pr: u64) -> Result<History> {
     let ledger = Ledger::open(repo)?;
     let mut history = History {
         all_measured: true,
+        first_round: u32::MAX,
         ..Default::default()
     };
     for event in ledger.iter_events_by_type("review_verdict")? {
@@ -134,8 +159,14 @@ pub(crate) fn history(repo: &Path, pr: u64) -> Result<History> {
         // `refs.pr` from the raw JSON keeps one malformed event from turning
         // the whole daemon into `SKIP due-unknown` until the ledger is
         // repaired.
-        if event.payload.pointer("/refs/pr").and_then(|v| v.as_u64()) != Some(pr) {
-            continue;
+        // Skip only what is *positively* another PR's. An event whose
+        // `refs.pr` is absent or unreadable falls through to the full
+        // deserialize below, which either reads it or fails loudly — skipping
+        // it here would be the same blindness this gate exists to avoid, just
+        // moved one line up.
+        match event.payload.pointer("/refs/pr").and_then(|v| v.as_u64()) {
+            Some(other) if other != pr => continue,
+            _ => {}
         }
         let payload: ReviewVerdictPayload = serde_json::from_value(event.payload.clone())
             .with_context(|| format!("review_verdict event {}", event.event_id))?;
@@ -145,11 +176,19 @@ pub(crate) fn history(repo: &Path, pr: u64) -> Result<History> {
         history.rounds += 1;
         history.last_sha = Some(payload.subject.head_sha.clone());
         history.last_ts = Some(event.ts.clone());
-        // Round 1's reviewer is the one to resume, and the payload says which
-        // round it is. Taking the first event seen would name whichever round
-        // happens to sit earliest in this ledger — not the same thing on a
-        // machine that only imported later rounds.
-        if payload.refs.round == Some(1) {
+        // Round 1's reviewer is the one to resume — it holds the whole
+        // subject, where a later round holds a delta — and the payload says
+        // which round it is. Selecting by ledger position instead would name
+        // whichever round happens to sit earliest here, which is not the same
+        // thing on a machine that imported only later rounds.
+        //
+        // When no round 1 is present at all, the lowest round that *is*
+        // present is the best available resume target; answering "no session"
+        // would send a continuing review back to round-1 price for want of an
+        // event this machine never received.
+        let round = payload.refs.round.unwrap_or(u32::MAX);
+        if round < history.first_round {
+            history.first_round = round;
             history.first_session = Some(payload.reviewer.session_id.clone());
         }
         match (payload.cost.measured, payload.cost.usd) {
@@ -182,6 +221,8 @@ pub(crate) struct Facts<'a> {
     pub(crate) now: i64,
     pub(crate) last_sha: Option<&'a str>,
     pub(crate) last_verdict_at: Option<i64>,
+    /// When the caller itself last reviewed, if it keeps such a record.
+    pub(crate) last_reviewed_at: Option<i64>,
 }
 
 /// The trigger policy. Pure: same facts, same answer, on every machine.
@@ -211,23 +252,31 @@ pub(crate) fn decide(facts: &Facts, config: &DueConfig) -> Due {
     }
 
     // A Review Response is the implementer saying "look again", and it counts
-    // only if it arrived after the verdict it answers. An older one is the
+    // only if it arrived after the review it answers. An older one is the
     // previous round's, already answered.
     //
-    // A response with **no** verdict to compare against is not a trigger. It
-    // reads like one — the comment is right there — but this branch is not
-    // debounced and nothing records that a response was acted on, so answering
-    // "review" to an uncomparable response answers it again on the next poll,
-    // and every poll after that, each at round-1 price. Verdicts do not cross
-    // machines yet (GH-671: `review_verdict` is not among the event types the
-    // committed mirror imports) and a round published only through the §7
-    // comment path writes no event at all, so "no verdict in this ledger" is
-    // the common case rather than the exotic one. Such a PR falls through to
-    // the push rule below, which is debounced and reads the daemon's own
-    // recorded head — the record that actually exists.
+    // This branch is not debounced and nothing marks a response as consumed,
+    // so getting "already answered" wrong does not cost one wasted round — it
+    // costs one per poll, forever, each at round-1 price. It therefore
+    // measures against **every** record of having reviewed that the caller
+    // has, and fires only if the response is newer than all of them:
+    //
+    // - the ledger verdict, which exists only sometimes (verdicts do not cross
+    //   machines — GH-671 — and a round published through the §7 comment path
+    //   writes no event at all), and
+    // - the caller's own `--last-reviewed-at`, which is the record that does
+    //   exist for a round this daemon ran.
+    //
+    // With neither, there is nothing to be newer than and this is not a
+    // trigger: the PR falls through to the push rule below, which is debounced
+    // and reads the caller's recorded head.
     if config.enabled("response") {
-        if let (Some(response_at), Some(verdict_at)) = (facts.response_at, facts.last_verdict_at) {
-            if response_at > verdict_at {
+        let reviewed_at = match (facts.last_verdict_at, facts.last_reviewed_at) {
+            (Some(a), Some(b)) => Some(a.max(b)),
+            (only, None) | (None, only) => only,
+        };
+        if let (Some(response_at), Some(reviewed_at)) = (facts.response_at, reviewed_at) {
+            if response_at > reviewed_at {
                 return Due::Review("response".into());
             }
         }
@@ -260,7 +309,10 @@ pub(crate) fn decide(facts: &Facts, config: &DueConfig) -> Due {
     if settled_for >= debounce {
         Due::Review("push".into())
     } else {
-        Due::Skip(format!("debounce {}s", debounce - settled_for))
+        Due::Skip(format!(
+            "debounce {}s",
+            debounce.saturating_sub(settled_for)
+        ))
     }
 }
 
@@ -313,6 +365,11 @@ fn judge(args: DueArgs, cwd: &Path) -> Result<()> {
         .as_deref()
         .map(|value| epoch(value, "--response-at"))
         .transpose()?;
+    let last_reviewed_at = args
+        .last_reviewed_at
+        .as_deref()
+        .map(|value| epoch(value, "--last-reviewed-at"))
+        .transpose()?;
 
     // The repo is only needed once a PR number gives the ledger something to
     // look up, so the flag-only cases work outside a checkout.
@@ -346,6 +403,7 @@ fn judge(args: DueArgs, cwd: &Path) -> Result<()> {
             now,
             last_sha,
             last_verdict_at,
+            last_reviewed_at,
         },
         &config,
     );
@@ -355,11 +413,7 @@ fn judge(args: DueArgs, cwd: &Path) -> Result<()> {
             // A second round resumes the first round's session: the reviewer
             // already holds the diff, the spec and its own findings, which is
             // the difference between a $2 round and a $0.02 one.
-            let resume = match (history.rounds >= 1, history.first_session.as_deref()) {
-                (true, Some(session)) => format!(" --resume {session}"),
-                _ => String::new(),
-            };
-            println!("REVIEW {reason}{resume}");
+            println!("REVIEW {reason}{}", history.resume_suffix());
             println!("{}", history.cost_line());
             Ok(())
         }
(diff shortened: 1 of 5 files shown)


fagemx:
## Code Review: Round 1 — PR #1074 @ 4d2df56ae8e0c200d419c9bdad5f3d58e891a4ce

- elapsed: unmeasured (no pi session file — this lane ran as a Claude Code session)
- model_observed: unverified (no session JSONL / `modelUsage` key reachable from this lane)
- reviewer_session: unset (lane launched without a per-PR UUID)
- spec: review-spec-v1.5
- class: code-risk  (REVIEW.md classes: code-risk code-plain docs — base-SHA classifier at `REVIEW.md@bedf1d3`)
- escalations: D5 (wording/structure — the spec's only `[判斷]` item; adjudicated below and marked `provisional`)
- shadow: false
- cost: 1 session, ~40 tool calls. Local RAN: 1 baseline + 6 mutant runs of `cargo test -p edda --bin edda cmd_review::due`; 1 baseline + 6 mutant runs of `sh scripts/test-pr-review-watch.sh` (~5 min each); 12 probes of the binary built at this head against four throwaway `edda init` repos seeded with synthetic `review_verdict` rows. Detached worktree `%LOCALAPPDATA%\fleet-workstation\review-wt\wt-review-pr1074`, `verifier` lane, `CARGO_INCREMENTAL=0`. Every mutation reverted; `git status --porcelain` empty after each battery; the binary rebuilt from the clean tree before the probe block (#1072). Nothing written to either shared checkout.

---

### IN SCOPE

Changed behaviour and paths: `crates/edda-cli/src/cmd_review/due.rs` (the `response` rule, the `refs.pr` scoping, the round-1 resume selection, `resume_suffix` extraction), `scripts/pr-review-watch.sh` (state row field 4, the `--last-reviewed-at` argv, one log wording), `scripts/test-pr-review-watch.sh` (four new cases), `docs/guides/pr-review-watcher.md` and `docs/reference/cli.md` (the trigger table and the state-file schema). Direct consumers: `decide()` and `scan_open_prs` in the same script; `judge()` in the same module. Acceptance ceiling: issue #763's doneWhen **plus** the four blocking findings raised on #1070 — round 2's three P1s (which are on `main` today) and round 3's one P1. Security / data-loss: the state file gains a column; `rm -rf` appears in the new test case. Current-base integration: this is the rescue PR for a branch stranded by #1070's merge, so the rebase itself is in scope.

Not in scope, routed below: `--ready` having no writer (observable at #1070 rounds 1–3, routed to follow-up there), the per-PR `gh pr view` cost, a future `committedDate`, and #1029.

### The four blocking findings this PR exists to close

Each verified by measurement on the binary built at this head, not by reading the PR body.

**[Round 2 P1-A] the `response` trigger re-fires on an already-reviewed head — CLOSED.** Round 2's probe shape, reproduced against a throwaway ledger carrying a PR-42 verdict at `2026-09-07T10:00:00Z`:

```
edda review due --head 1111…11 --pr 42 --last-reviewed 1111…11 \
    --last-reviewed-at 2026-09-07T11:30:00Z --response-at 2026-09-07T11:00:00Z
SKIP reviewed
review cost so far: $1.00 over 1 round
exit=1
```

Head equal to `--last-reviewed`, a verdict older than the response, and the daemon's own record newer: the answer is now `SKIP`, exit 1, where round 2 measured `REVIEW response`, exit 0. The trigger still survives its own hardening — the same call with `--response-at 2026-09-07T11:30:01Z` gives `REVIEW response --resume pr42-round-1-session`, exit 0. Pinned by `a_response_older_than_the_callers_own_review_is_already_answered` and `the_callers_own_record_counts_even_with_no_verdict_at_all`: replacing `max(last_verdict_at, last_reviewed_at)` (`due.rs:282-285`) with `facts.last_verdict_at` fails both.

**[Round 2 P1-B] `--pushed-at --help` states the behaviour the fix replaced — CLOSED.** `due.rs:42` now reads `without it a moved head waits`, and the head-built binary (`edda 0.5.0 (4d2df56ae8e0 2026-09-07)`) prints exactly that. It matches `:314-315` (`SKIP debounce push-time-unknown`).

**[Round 2 P1-C] the resume-id layer is proven by no test — CLOSED, all three of round 2's mutations.** Re-ran each one:

| Round 2's mutation | Result now |
|---|---|
| delete the decision-line log (`pr-review-watch.sh:200-201`) | `sh scripts/test-pr-review-watch.sh` **exit 1** — `decide: the decision line was not logged` |
| `refs.round == Some(1)` → `Some(7)` (`due.rs:199`) | **3 FAILED** — `a_ledger_without_round_one_…`, `another_prs_verdicts_…`, `the_session_to_resume_is_round_ones_by_number_…` |
| drop the ` --resume {session}` string (`due.rs:115`) | **4 FAILED** — the three above plus `the_resume_suffix_names_round_ones_session_and_only_from_round_two` |

Two more of the delta's own wirings are pinned as well: dropping `${prev_at:+--last-reviewed-at "$prev_at"}` (`:189`) and reverting `state_set` to three columns (`:514-515`) each turn the shell suite red, as does reading field `$3` instead of `$4` at `:156`.

**[Round 3 P1-A] a `refs.pr`-less verdict counted as every PR's history — CLOSED.** Built the ledger round 3 described and measured all four consequences. A PR-42 verdict at 10:00 (`$1.00`, session `pr42-round-1-session`) followed by a **PR-less** verdict at 12:00 (`$5.00`, session `orphan-session`), then a genuine Review Response at 11:00:

```
edda review due --head 1111…11 --pr 42 --last-reviewed 1111…11 --response-at 2026-09-07T11:00:00Z
REVIEW response --resume pr42-round-1-session
review cost so far: $1.00 over 1 round
exit=0
```

Rounds 1 not 2; `$1.00` not `$6.00`; the resume id is PR 42's, not the orphan's; and `last_verdict_at` stays at 10:00, so the Review Response fires instead of being answered `SKIP reviewed` — round 3's consequence (3), the under-review class, is gone. `last_sha` is not hijacked either: with **only** a PR-less verdict whose `head_sha` equals the queried head, the answer is `SKIP debounce push-time-unknown` / `$0.00 over 0 rounds` (i.e. `last_sha == None`), while the identical event with `refs.pr: 42` gives `SKIP reviewed` / `$5.00 over 1 round`. Pinned by `a_verdict_with_no_pr_belongs_to_no_pr`: removing `|| payload.refs.pr != Some(pr)` (`due.rs:179`) fails it.

**[Round 3 P1-B] the published state-file schema said three columns — CLOSED.** `docs/guides/pr-review-watcher.md:143` now reads `pr<TAB>reviewed_sha<TAB>round<TAB>reviewed_at` and states that rows without the field read as "no record", matching `scripts/pr-review-watch.sh:29,514-515`.

**[Round 3 P1-C / P1-D] no CI, and the branch could not reach `main`** — both dissolved by this PR existing: exact-head CI is 14/14 green at `4d2df56`, and the delta now sits against current `main`.

**The rebase is honest.** `git diff origin/main...4d2df56` is 5 files / +377 −38 and contains only the fixes; `git log origin/main..4d2df56` is two commits, both conventional `fix(edda-cli):`, so `--onto origin/main 98ab748` did drop the pair the squash `bedf1d3` already applied. Nothing the squash landed is reintroduced, and no unrelated file is touched.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 5 changed files, all inside #763's "改哪裡"/Predicted surface (`due.rs`, `pr-review-watch.sh`, `pr-review-watcher.md`, `cli.md`) plus that script's own fixture file |
| U2 | any | P1 | PASS | no closing keyword in the body or in either commit message — correct: doneWhen 3's second clause (daemon launches with `--resume`, `Session observed:` equals it) is still undelivered |
| U3 | any | P1 | PASS | `Issue: #763` |
| U4 | any | P1 | PASS | enumerator empty — `fix(edda-cli): …` ×2 |
| U5 | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → exit 0 |
| U6 | any | P0 | PASS | `gh pr checks 1074` → 14/14 pass; L1 receipt below |
| U7 | any | P1 | PASS | whole scoped audit finished before posting; every P2 carries evidence |
| U8 | any | P1 | N.A.(no `Decision:` line in the body) | DECISIONS block printed nothing |
| D1 | docs | P1 | PASS | one candidate, `edda review`. The PATH runtime is `edda 0.5.0 (3306c1a2d9a0)`, which predates the verb (`edda review due --help` → `unexpected argument 'due'`), so every claim was routed as a source claim and probed on the binary built at this head: `--pushed-at` help = `without it a moved head waits`, matching `due.rs:314-315` |
| D2 | docs | P1 | N.A.(the diff states no decision value or status) | no `Decision:` key; no ledger claim in the added text |
| D3 | docs | P1 | PASS | enumerator returned 0 repo-anchored path candidates; the added text cites runtime state (`.edda/review/due.json`, `review-state.tsv`), not repo paths |
| D4 | docs | P0 | PASS | enumerator empty |
| D5 | docs | `[判斷]` | 需升級 | adjudicated in Findings, tagged `provisional` (§6.1) |
| S1–S3 | skills | — | N.A.(no `skills` path in the diff) | classifier printed `code-risk code-plain docs` |
| C1 | code-plain | P0 | PASS | partial delivery, no closing keyword; each of the four blocking findings matched to code **and** to a test that fails when that code is reverted (table above) |
| C2 | code-plain | P0 | PASS | both subjects are `fix:`; 8 added `#[test]`/`fn test_` lines plus 4 shell cases; 9 of 11 mutations turn a named test red |
| C3 | code-plain | P1 | N.A.(all candidates in `#[cfg(test)]`) | 14 `.expect(` hits, all at `due.rs:611+`; `#[cfg(test)]` opens at `due.rs:437` |
| C4 | code-plain | P1 | PASS | enumerator empty |
| C5 | code-plain | P1 | N.A.(selector empty) | only `crates/edda-cli` is touched, i.e. `-p edda`, already in the CI Windows subset |
| R1 | code-risk | P0 | PASS | 1 hit — `test-pr-review-watch.sh:428` `rm -rf "$EDDA_FLEET_SCRATCH"`. Trigger: the harness **exports its own** `EDDA_FLEET_SCRATCH="$tmp/scratch"` at `:27` from `tmp=$(mktemp -d)` (`:17`) with `trap 'rm -rf "$tmp"' 0 HUP INT TERM` (`:18`), before any use — so an inherited value cannot be reached and the command can only remove that tempdir subtree. `$SCRATCH` is recreated by `pr-review-watch.sh:441` on the next `decide` dispatch |
| R2 | code-risk | P0 | PASS | no added line mixes `\|\|` and `&&`; nothing to parse |
| R3 | code-risk | P0 | PASS | `sh -n` → exit 0 on both changed scripts |
| R4 | code-risk | P1 | PASS | `set -eu` + `trap … 0 HUP INT TERM` unchanged; `state_set` still writes `$STATE.tmp` then `mv`; a failed `date -u` yields an empty field 4, which the verb reads as "no record" rather than as garbage |
| R5 | code-risk | P1 | PASS | `${prev_at:+--last-reviewed-at "$prev_at"}` (`:189`) keeps the inner quoting, so a value with whitespace stays one argument; the value is validated at the verb boundary by `epoch()` (`due.rs:328-335`) — measured: `--last-reviewed-at garbage` → exit 2, and `decide` logs it at `:218`. No new filename encoding, no new SHA interpolation |
| §5.5 | any | — | see table | one row per new surface |

### Wiring

|---|---|---|---|---|
| `--last-reviewed-at <RFC3339>` / `DueArgs::last_reviewed_at` (`due.rs:57-58`) | `pr-review-watch.sh:189` `${prev_at:+--last-reviewed-at "$prev_at"}`, fed from state field 4 read at `:156` | `judge()` `due.rs:377-381` → `Facts::last_reviewed_at` `:233-234` → `decide()` `:282-285` | loud — non-RFC3339 → `epoch()` error → exit 2 → `decide` logs `pr<n> due check could not judge:` (`:218`) and prints `SKIP <n> due-unknown` (measured) | flag→argv→policy, both halves pinned: dropping `:189` turns the suite red (`the verb was not handed --last-reviewed-at`), and ignoring the fact in `decide` fails 2 Rust tests |
| `review-state.tsv` field 4 `reviewed_at` (`pr-review-watch.sh:514-515`) | `state_set`, `printf '%s\t%s\t%s\t%s\n' … "$(date -u '+%Y-%m-%dT%H:%M:%SZ')"`, at round settlement (`:830,848,863,883,918,931,973`) | `decide()` `:156` (awk `$4`); harness `state_get_at` (`test-pr-review-watch.sh:300`) | empty reads as "no record" — documented at `pr-review-watcher.md:143` and `:29` | shell→file→verb, pinned twice (3-column `state_set` → `live: the recorded review time is not RFC3339`; reading `$3` → argv assertion). **No other reader breaks**: the only other consumer is the field-indexed `state_field` used at `:1119-1120`, and the harness's `state_get` now reads `cut -f1-3` |
| `History::resume_suffix()` (`due.rs:114-119`) | extracted from `judge`'s inline `match` | `judge()` `:425`; the printed id's only consumer is the daemon's decision log `:200-201` → `watch.log` | present-or-absent, REVIEW path only by construction (`sed -n '1{/^REVIEW /p;}'`) | proven at both ends now — this is round 2's [R2-P1-C] pair, and each half is red under its own mutation |
| `History::first_round: u32` (`due.rs:92`) | `history()` `:198-201` | same function's `first_session` selection | none of its own — a wrong pick publishes a wrong resume id, never an error | primary rule pinned (`Some(1)` → 1 FAIL, `Some(7)` → 3 FAIL); the `\|\| history.first_session.is_none()` half is pinned by nothing → [P2] below |
| `history()` no longer counts a `refs.pr`-less verdict (`due.rs:179`) | `history()` | `judge()` `:400-425` → rounds, cost line, `--resume`, `last_verdict_at` | silent by design — such an event is simply not this PR's | pinned in Rust (M2 → `a_verdict_with_no_pr_belongs_to_no_pr` FAIL) and measured end-to-end (probes above) |

`sh scripts/wiring-scan.sh origin/main 4d2df56` reports `last_reviewed_at (field)` and `resume_suffix (fn)` as "0 outside references" and no swallow patterns; both reference counts are the expected false positives — each consumer is in the same file, and the table above names it with a line number.

### Findings

No P0 and no P1. The four blocking findings raised on #1070 are each closed by code **and** by a test that fails when that code is reverted; the rebase carries nothing else. Everything below is P2 and does not block.

- [P2] **The module header is still a third policy table, and it now contradicts both published ones.** `due.rs:20` reads ``| `--response-at` newer than the last verdict | `REVIEW response` |`` — the exact sentence this delta replaced at `cli.md:1294` and `pr-review-watcher.md:171` — and still carries no `push-time-unknown` row and no `--unreviewed-label` row. It is `//!` module documentation on a private module, so it reaches no `--help` output (verified: the shipped `about` is "Is this PR worth reviewing again? The trigger policy (GH-763)"), which is why this is P2 rather than round 2's [R2-P1-B]. Raised at #1070 round 3 as a P2 and carried unchanged. — evidence: `crates/edda-cli/src/cmd_review/due.rs:14-28` vs `:282-289`.

- [P2] **Two adjacent comments in `history()` state opposite policies, and this delta is what made the older one false — measured.** `due.rs:155-161` (retained) says reading `refs.pr` from the raw JSON "keeps one malformed event from turning the whole daemon into `SKIP due-unknown` until the ledger is repaired"; `:162-166` (new) says an unreadable `refs.pr` "falls through to the full deserialize below, which either reads it or fails loudly". The second describes the code. Seeded one `review_verdict` whose `refs.pr` is the JSON string `"77"` into a throwaway ledger and asked about an unrelated PR 42:

  ```
  edda review due --head 1111…11 --pr 42 --last-reviewed 1111…11
  edda review due: review_verdict event evt_drift: invalid type: string "77", expected u64
  exit=2
  ```

  Exit 2 is `SKIP due-unknown` for *every* PR on that machine, which is precisely what the retained sentence promises cannot happen; `main`'s `!= Some(pr) → continue` skipped that event silently. It stays P2 because no writer in this repository can produce it — `ReviewRefs.pr` is `Option<u64>` (`crates/edda-core/src/review.rs:52-54`) and every writer goes through the typed constructor — so reachability is schema-version drift, not a live defect. Delete the stale paragraph. — evidence: `due.rs:155-166`; probe output above.

- [P2] **`resume_suffix`'s rustdoc is `cost_line`'s, and `cost_line` now has none.** The new function was inserted between `cost_line`'s doc comment and `cost_line` itself, so `cargo doc` renders "`review cost so far: …` — the line that makes the cost switch visible" and "An unmeasured round makes the whole total unmeasured…" as `resume_suffix`'s documentation. Compiles, lints clean, wrong in the rendered docs. Raised at round 3, unchanged. — evidence: `due.rs:100-121`.

- [P2] **`cli.md` still says the resume id is round 1's, and this delta is what made that only usually true.** `docs/reference/cli.md:1324-1325`: "A second or later round appends ` --resume <session id>` naming round 1's reviewer session." `due.rs:198-201` now selects the **lowest round present**, and — via the new `|| history.first_session.is_none()` clause — the first verdict's session when no event carries a round at all. That is reachable without any cross-machine import: a round 1 published through the §7 comment path writes no `review_verdict`, so a ledger can legitimately hold round 2 and no round 1, and the daemon then publishes round 2's session under a sentence promising round 1's. It is P2 rather than P1 because no operator action follows from the sentence and the fallback is the cheaper of the two answers — but the same commit rewrote both in-code statements of the rule (`due.rs:88-89` and `:185-197`) and left this one. — evidence: `docs/reference/cli.md:1324-1325` vs `crates/edda-cli/src/cmd_review/due.rs:198-201`.

- [P2] **Two of the delta's own changes are pinned by nothing.** Dropping the `|| history.first_session.is_none()` half of the selection rule (`due.rs:199`) leaves all 22 `cmd_review::due` tests green, and so does reverting `debounce.saturating_sub(settled_for)` (`:320-324`) to plain subtraction. The first is this PR's code answer to round 3's "the lowest-round fallback does not cover a ledger where `refs.round` is absent everywhere"; the second is round 2's own P2 fix, arriving for the second review round without a regression test. Neither is a `doneWhen` layer, so §5.5 routes both to follow-up rather than P1.

- [P2] **Three under-review edges on the new comparison, all measured, none new in kind.** (a) `response_at > reviewed_at` is strict and `state_set` records at 1-second resolution, so a Review Response landing in the same second a round settles reads as answered — `--last-reviewed-at 2026-09-07T11:30:00Z --response-at 2026-09-07T11:30:00Z` → `SKIP reviewed`, exit 1. (b) Field 4 comes from the daemon's own `date -u` while `response_at` is a forge timestamp, so a row dated ahead of the forge answers every response for the whole skew — `--last-reviewed-at 2099-01-01T00:00:00Z --response-at 2026-09-07T11:00:00Z` → `SKIP reviewed`, exit 1 — and nothing logs it, because `scan_open_prs` discards every non-`REVIEW` line at `:1104-1105`. (c) `state_set` runs at round *settlement*, so a Review Response posted while a round is in flight is older than the record that round writes. All three are the price of comparing timestamps instead of marking a response consumed, which is the design round 2 offered as the alternative; the ledger path already had the same shape, so the delta widens the class rather than creating it.

- [P2] **A legacy 3-column state row still reproduces round 2's probe once.** Same ledger, same call, without `--last-reviewed-at`: `REVIEW response --resume pr42-round-1-session`, exit 0 — the verb can only use the record the caller has. It self-heals the moment that round settles and writes field 4, and `pr-review-watcher.md:143` documents it ("早期沒有這欄的列讀成「沒有記錄」"), so the cost is one extra round per PR carrying a legacy row. Recorded because the guide also tells the operator to hand-write rows (`:99`, `:215`), and a hand-written 3-column row lands in the same state deliberately.

- [P2] **The quoting artifact in the new cold-machine case survived.** `test-pr-review-watch.sh:429-430` is written `expect_decide_wired     '''name'''     …     ''''''     '''42<TAB>def456<TAB><TAB>2026-09-02T00:00:00Z'''`, with a literal newline and literal tabs, where every neighbouring case uses `\n`/`\t` through `printf '%b'`. It is correct (`'''X'''` concatenates to `X`, `''''''` is the empty string) and non-vacuous — removing `mkdir -p "$SCRATCH"` from `pr-review-watch.sh:441` turns the suite red with `expected REVIEW 42 def456 / got SKIP 42` — but it is the only line in the file written that way and it reads as a paste artifact. Raised at round 3, unchanged.

- [P2/D5, `provisional`] **Wording and structure.** The added prose is again the strongest part of the change: the `response` block (`due.rs:263-281`) states *why* an undebounced wrong answer costs one round per poll rather than one round, the `refs.pr` block (`:173-179`) names the exact writer that produces a PR-less verdict, and `cli.md`'s two new paragraphs explain why two rows exist rather than only that they do. The structural failure is the same one round 3 named and it is now on its second round: the added text is careful and the *adjacent existing* text is not re-read — a third policy table left stale one screen above the code (`:14-28`), a doc comment welded onto the wrong function (`:100-121`), a retained sentence asserting the opposite of the code eight lines below it (`:155-166`), and a published reference sentence (`cli.md:1325`) describing the selection rule the same commit replaced. Each is individually cosmetic; the pattern is what makes it worth stating, because it is the mechanism that produced round 2's [R2-P1-B] and round 3's [R3-P1-B], both of which *were* blocking. Tagged `provisional` and escalated per §6.1, because this brief does not name me qualified for this class.

### FOLLOW-UP ISSUE

- **`--ready` still has no writer.** `grep -rn -- "--ready" scripts/ crates/ docs/` returns exactly two hits, both tables that document it (`due.rs:19`, `cli.md:1293`); `decide()` passes `--head --pr --last-reviewed --last-reviewed-at --pushed-at --response-at --draft --unreviewed-label` (`pr-review-watch.sh:187-192`) and never `--ready`, so one of the three defaults in `due.json` cannot fire in the daemon. Observable at #1070 rounds 1–3 and routed to follow-up at both 2 and 3; `loop` item 2 keeps it out of this PR. Still has no issue number.
- **The `first_session.is_none()` fallback and the `saturating_sub` bound both need a regression test** (the [P2] above). One issue covering both.
- **`scripts/test-pr-review-watch.sh` runs in no CI job.** Confirmed at this head: `scripts/fleet/run-fleet-tests.sh:84-92` still quarantines it under `#1029` ("r22 live case 2 red on ubuntu, green on the workstation"). `Fleet shell tests` and `Fleet shell tests (windows)` therefore mean `sh -n` only for the file carrying a third of this delta, and every shell measurement in this round had to be made by hand.
- **Bounding `--pushed-at` and `--last-reviewed-at` against `now`** would close the (b) edge above and the pre-existing future-`committedDate` hold in one place. Both are hold-rather-than-spend directions today, which is why neither blocks.
- **The per-PR `gh pr view`** the daemon adds per poll — untouched by this delta, carried from rounds 1–3.
- Recording the negative: **#1071 did not reproduce** in this round's `cargo test -p edda` (684 passed, 0 failed, plus 20 other test binaries all green).

### RAN vs READ

RAN (detached worktree at `4d2df56`, `CARGO_TARGET_DIR` = `verifier` lane, `CARGO_INCREMENTAL=0`; every mutation reverted, `git status --porcelain` empty after each battery):

- `cargo test -p edda` → **exit 0**, 684 passed / 0 failed in the bin plus 20 other binaries, 0 failed anywhere. Stated reason for running a gate CI also covers: every mutation control below has to be read against a known-green base on this machine.
- `sh scripts/test-pr-review-watch.sh` at head → **exit 0**, `pr-review-watch fixtures passed`. Stated reason: no CI job runs this file (#1029).
- Rust mutations, `cargo test -p edda --bin edda cmd_review::due` each. Red where they should be: `max(...)` → `last_verdict_at` → 2 FAILED; `|| payload.refs.pr != Some(pr)` removed → 1 FAILED; ` --resume {session}` dropped → 4 FAILED; round selection → `== Some(1)` → 1 FAILED; → `== Some(7)` → 3 FAILED. Green where nothing pins the code: `|| first_session.is_none()` dropped → 22 passed; `saturating_sub` → `-` → 22 passed.
- Shell mutations, one full suite run each: delete the decision-line log → exit 1; drop `--last-reviewed-at` from the argv → exit 1; `state_set` back to three columns → exit 1 (`live: the recorded review time is not RFC3339`); remove `mkdir -p "$SCRATCH"` → exit 1; `prev_at` reads `$3` → exit 1. Five for five.
- Probes of the binary built at this head (`edda 0.5.0 (4d2df56ae8e0 2026-09-07)`) against four throwaway `git init` + `edda init` repos under `%TEMP%`, each seeded with synthetic `review_verdict` rows written directly into its own `.edda/ledger.db` — never a shared checkout, never the workstation ledger: `--help`; the round-2 probe with and without `--last-reviewed-at`; the same-second boundary; a future-dated field 4; garbage in field 4; the round-3 PR-less-verdict battery (three variants plus the `refs.pr: 42` control); and the `refs.pr: "77"` type-drift probe.
- `sh scripts/lint-markdown-content.sh` → exit 0 (U5). `sh scripts/wiring-scan.sh origin/main 4d2df56`. U2/U3/U4/D1/D3/D4/C2/C3/C4/C5/R1/R2/R3 enumerators; base-SHA classifier (`REVIEW.md@bedf1d3`) → `classes=code-risk code-plain docs`, `canonical_class=code-risk`.
- `edda --version` → `edda 0.5.0 (3306c1a2d9a0 2026-09-07)`; `gh pr --help` → 0.

READ:

- L1 receipt: **`CI run 34153026142 @ 4d2df56ae8e0c200d419c9bdad5f3d58e891a4ce`** (plus `34153026155` sdk-contract and `34153026255` publish dry-run on the same SHA) — Format; Clippy ubuntu/macos/**windows**; Test ubuntu/macos/**windows**; MSRV 1.91.0; Fleet shell tests (+windows); Detect code changes; CI Gate; sdk-contract; Workspace publish dry-run — **14/14 pass**. That is the whole `gates:` set: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets`, and `cargo test --workspace` on Linux and macOS plus the 7-crate Windows subset, which contains `edda` — the only crate this PR touches. The C5 selector is empty, so no focused Windows run was owed; I ran none.
- What CI does **not** cover: `scripts/test-pr-review-watch.sh` (#1029, quarantined at `run-fleet-tests.sh:84-92`), which is why every shell measurement above is a local RAN rather than a READ.
- `.claude/CLAUDE.md` (loop, ladder), `REVIEW.md@bedf1d3`, issue #763 body and doneWhen, PR #1074 body, and the full #1070 thread — Code Review rounds 1, 2 and 3 with their `Review Response` comments — since this PR's acceptance bar is those rounds' findings.

### Verdict

LGTM (P0=0, P1=0) — provisional under §6.4 while `escalations: D5` stands.

All four blocking findings from #1070 are closed by measurement rather than by claim: the `response` trigger now answers `SKIP reviewed` on the exact input round 2 measured `REVIEW response` on; `--pushed-at --help` matches the code; all three of round 2's mutations are caught by a test that names them; and a `refs.pr`-less verdict contributes no round, no cost, no resume id and no `last_verdict_at`. Nine of eleven mutations I ran turn a named test red. The rebase carries only those fixes. The eight P2s are documentation and comment drift plus two unpinned hardenings, and none of them changes what the verb answers.

PR #1055, opened 2026-09-07
feat(edda-cli): edda review gate — union rule and window check in the product (GH-769)

fagemx:
Issue: #769
Closes #769.

## What

The merge gate's two rules — the same-SHA verdict **union** and the **base
window** check — lived in `scripts/pr-review-watch.sh` behind a
`# D8-debt(#769)` marker, because #652 had not landed when #742 needed them.
They are verdict semantics, not forge glue (#766 D8), so this moves them into
the binary as `edda review gate <sha>` and leaves the shell an adapter.

```
edda review gate <sha> [--base <ref>] [--verdicts <path|->] [--json]
  exit 0  PASS <sha> verdicts=<n>
  exit 1  FAIL <sha> union | window
  exit 2  NONE <sha>          (no verdict — an inability to judge)
```

The verb reads only: no ledger write, no lane, no GitHub call.

## Two sources, one rule

`union()` is a pure function over `Standing` values, and two adapters produce
them:

- **the ledger's `review_verdict` events** — the source the issue specifies, and
- **`--verdicts <path|->`** — the caller's facts in the tab-separated shape the
  watcher already derives from §7 comments.

The second exists because the verdict *source* is a different debt from the
union *rule*. `D8-debt(#671)` — reading verdicts out of PR comments — stands
until the ledger carries verdicts across machines, and **it does not yet**:
`review_verdict` is not among the event types `#1017`'s committed mirror
imports, and this workstation's ledger holds only its own 15 verdict events. A
fleet that reviews on one machine and merges from another would, on the ledger
alone, see no verdict at all and read `NONE` as `error`. Feeding the caller's
facts to the same rule keeps that correct today without a second copy of the
rule, and the ledger path is ready for the day #671 is paid.

## What the shell keeps

`gate_state()` is now only an exit-code-to-status-word adapter and logs the
receipt the issue's verify step asks for (`gate <sha> exit <n>`). Forge facts —
`CI Gate`, `mergeStateStatus`, head == reviewed sha — stay in shell as
specified. The issue's "merge step" half needs no change: the watcher never
merges (`pr-review-watch.sh:84`, merge stays behind operator authorization).

The shell suite's union matrix moved with the rule. Its 17 union cases became
4 exit-code mapping cases plus 2 receipt cases, and the live status cases now
declare the gate's answer (`EDDA_GATE_EXIT`) instead of deriving it — that is
exactly what the issue's verify step specifies ("stub `edda review gate` 回
[...]

PR #1082, opened 2026-09-08
chore(fleet): retire the review shell branch (GH-1061)

fagemx:
Issue: #1061

## What

Retires the review shell branch wholesale: the watcher (`pr-review-watch.sh`), `review-pr.sh`, `review-round.sh`, the launcher, the queue helper, their seven test files and the watcher guide — 13 files, −6652 lines in the first commit — then a full Edit pass removing every reference from the surviving tree (REVIEW.md, `run-fleet-tests.sh`, `ci.yml`, runbooks, rules, skills, and six crate files). 46 files total, +223/−7135, six commits.

Deletion = completion: the review loop now runs controller-direct against the product verbs (`edda review due` / `gate` / `deliver`), per `review.dispatch-transport=controller-subagent-direct` and `review.deliver-adapter=superseded-by-gh1061-retirement` (ledger). Proven end-to-end on PR #1077 and PR #1081 before this deletion.

Two defects found beyond the issue's Edit list, fixed here because leaving them makes CI red at every run:
- The `fleet-tests-windows` CI job's only test command was `sh scripts/test-review-capabilities.sh` — a file the first commit deletes.
- `test-next-loop.sh` and `test-issue-freshness.sh` asserted against the deleted `next-review.sh`.

Spec notes (stated per the issue's doneWhen):
- The doneWhen asks for both a clean repo-wide grep and byte-identical REVIEW.md §7/§8, but §8 itself named two deleted scripts. §7 is byte-identical to main (sha256 match); §8 differs by exactly the four dangling-reference sentences.
- The doneWhen names `edda review brief <pr>`, which does not exist — the shipped verbs are `due`/`gate`/`deliver` plus bare `edda review --pr`. The runbook documents what ships.

## RAN

- Repo-wide grep for all deleted filenames: clean (exit 1, nothing outside exempt paths).
- REVIEW.md §7 byte-identical to main (sha256 match).
- Machine cleanup (doneWhen): `edda-pr-review-watcher` scheduled task deleted and verified absent; 20 `wt-review-pr*` worktrees removed and pruned, 0 remaining.
- Live GhCli smoke (folded-in prerequisite): `edda review deliver --pr 1081 --json` → exit 0, union `success`, label and status already settled — the real `gh` argv seam exercised against this repo.
- `cargo fmt --all --check` and `cargo clippy -p edda --all-targets -- -D warnings` green (pre-commit hook).
[...]

Issue #1061, opened 2026-09-07
chore(fleet): retire the review shell branch — watcher, review-pr.sh, review-round.sh, six-green receipt and status writer have no consumer after review.watcher / review.merge-gate


fagemx:
## Why

After 2026-09-07 the review branch of the shell layer has no consumer:

- `review.merge-gate=ci-gate-only-independent-review-status-removed` — the `Independent Review` required status is out of the ruleset. The watcher's status writer (`pr-review-watch.sh:616`, the repo's only writer of that context) now writes a status nothing reads.
- `review.watcher=polling-stopped-independent-rounds-dispatched-on-demand` — the poller is stopped. Nothing calls `review-pr.sh` any more; the last three launches it made after that ruling (PR #1051 docs 1 file / #1052 / #1055) were all still in the 1077-file snapshot when they were stopped.
- `fleet.review-engine-model` (2026-09-02) — review runs on Opus via Claude Code. There is no flash review tier, so "keep the review shell for flash" keeps it for a consumer that was ruled out a week ago. The flash-reserved shell named by `review.default-path` is the **dispatch** shell (brief render, `brief-validate`, oracle bundle, qualification, SHADOW); none of that lives in the files below.
- The two PRs the 2026-09-07 review (`docs/…/2026-09-07-兩張PR大檢討`) examined were merged by direct review in ~5 minutes per round; the same rounds through this branch cost 50–100 minutes of snapshot per round plus the six-green receipt refusing a clean verdict over a `git worktree remove` that lost to an AV handle (PR #1040 R1). 17 orphaned `wt-review-pr*` worktrees under `~/.edda/fleet` are that race's residue.

`fleet.mechanism-layer` (ratified 2026-09-06) already sentences replaced scripts to deletion on landing. It assumed the shell keeps executing until the product lands. For this branch execution reached zero **before** #769/#1030 landed, so this issue sequences the deletion now rather than after them. Decision: `review.shell-branch=retire-ahead-of-landing-zero-consumers`.

Meanwhile fixes keep being filed against the files being retired (#1058, #953's review-pr.sh half, #1057's caller side) — exactly the fix-one-spawns-two pattern the strangler ruling measured (35/50 merged PRs 09-04..09-06 were mechanism fixes). Deleting the files closes that tap.

## Predicted surface

Delete:
- `scripts/pr-review-watch.sh` — poller, six-green receipt gate, verdict publisher, `Independent Review` status writer
[...]

Issue #1057, opened 2026-09-07
fix(fleet): merge-reviewed-pr.sh selects the latest review instead of the union — wire it to `edda review gate`


fagemx:
Split out of #769 review round 1 (PR #1055), which added `edda review gate <sha>` as the single implementation of the union rule and the base-window check.

## The gap

`scripts/merge-reviewed-pr.sh` is the operator's merge entrypoint (docs/guides/operator-runbook.md). Its verdict selection is latest-trusted-review-wins:

```
scripts/merge-reviewed-pr.sh:36
  $trusted | sort_by([.updated_at, .id]) | last | .body
```

It then requires that one body's Verdict line to be `LGTM (P0=0, P1=0)`. So an earlier `Changes Requested` on the same head is ignored the moment a later LGTM comment exists — exactly the failure GH-742 identified and the union rule was written to prevent. The union does not reach it indirectly either: ruleset 18852689 lists required contexts as `[{"context":"CI Gate"}]`, so `Independent Review` is not required and `:45`'s `gh pr checks --required` never sees it.

The window check has no consumer at the merge step at all.

## What to do

Add a union gate over **every** trusted review pinned to the head, as an additional refusal — the existing head-pinning, `escalations: none`, and `--required` checks all stay:

```sh
sh scripts/pr-review-watch.sh collect-verdicts "$pr" "$head"   | "${EDDA_BIN:-edda}" review gate "$head" --verdicts - || die 'union rule refuses this head'
```

Reuse `collect-verdicts` rather than writing a third verdict parser. Strictly tightening: it can refuse more, never fewer.

Then decide the window step. `edda review gate --base <ref>` needs both commits present locally, and `merge-reviewed-pr.sh` today runs anywhere (`gh --repo`), so wiring it changes the script's contract to require a checkout. Either accept that and fetch head+base fail-closed, or state in the script why GitHub's `--match-head-commit` plus `mergeStateStatus` is deemed sufficient there.

## Why it was not done in #1055

PR #1055's frozen surface is the new verb and the watcher's status step. `merge-reviewed-pr.sh` is a pre-existing, adjacent defect on the operator's merge path; per REVIEW.md's loop it is a follow-up rather than an extension of that PR. #1055 round 1 recorded it as P1-2/P1-3 and its REVIEW.md wording was corrected to stop claiming the merge step already asks the gate.

## doneWhen

[...]

PR #1083, opened 2026-09-08
fix(edda-cli): review gate window sees renames — --no-renames on both diffs (GH-1062)

### LINKED ISSUE #1062: fix(edda-cli): review gate's window check reads clear across a rename — `git diff --name-only` reports only the destination path
Found in PR #1055 round 2 review (GH-769), routed as a follow-up because it is inherited from the shell implementation the PR replaced rather than caused by it.

## The gap

`window_moved` in `crates/edda-cli/src/cmd_review/gate.rs` lists the subject's changed files and feeds them back as pathspecs against the base's advance. `git diff --name-only` applies rename detection, so a renamed file is reported as its **destination** path only. If the base moved the same file under its **old** name — or the subject renamed a file the base then edited — the second diff matches nothing and the window reads *clear*.

Same failure direction as the quoted-path bug fixed in #1055 (fail open on the rule that exists to catch a moved base), and a sibling of it: both come from treating `--name-only` output as an exact file list.

## What to do

Pass `--no-renames` to both diffs in `window_moved`, so a rename appears as a delete of the old path plus an add of the new one and both names enter the pathspec set.

```
git diff --name-only -z --no-renames <base>...<sha>
git --literal-pathspecs diff --name-only --no-renames <merge-base>..<base> -- <files>
```

## doneWhen

- A test in `gate.rs` alongside `a_path_git_would_quote_is_still_seen_by_the_window`: the subject renames `a.txt` to `b.txt`, the base then edits `a.txt`, and `window_moved` returns true. It must fail on the current code.
- The existing window tests stay green — an ordinary edit with no rename is unchanged.

fagemx:
Issue: #1062
Closes #1062

## Summary

`window_moved` in `crates/edda-cli/src/cmd_review/gate.rs` fed
`git diff --name-only`'s output back as pathspecs to check whether the base
advanced over a file the subject touched. `git diff --name-only` applies
rename detection, so a renamed file is reported only under its
**destination** path. If the base moved the same file under its **old**
name, the second diff's pathspec set never contained that name and the
window falsely read *clear* — the same fail-open direction as the
quoted-path bug fixed in #1055, and a sibling of it.

Fix: pass `--no-renames` to both diffs in `window_moved`, so a rename
splits into a delete of the old path plus an add of the new one, and both
names enter the pathspec set.

## RAN

Focused module test, run FAIL-first before the fix (new test only):

```
$ cargo test -p edda cmd_review::gate::tests::a_rename_the_base_then_edits_under_the_old_name_is_still_seen
running 1 test
thread 'cmd_review::gate::tests::a_rename_the_base_then_edits_under_the_old_name_is_still_seen' panicked at crates\edda-cli\src\cmd_review\gate.rs:483:9:
the base's edit under the pre-rename name must not read as clear
test cmd_review::gate::tests::a_rename_the_base_then_edits_under_the_old_name_is_still_seen ... FAILED

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 721 filtered out; finished in 9.00s
```

After the fix, gates (L0, touched crate only, warm lane
`CARGO_TARGET_DIR=$LOCALAPPDATA/fleet-workstation/lanes/worker-2`):

- `cargo fmt --all --check` — clean, no output
- `cargo clippy -p edda --all-targets -- -D warnings` — clean, `Finished` with no warnings
- `cargo test -p edda cmd_review::gate` — `test result: ok. 10 passed; 0 failed`, including
  the new `a_rename_the_base_then_edits_under_the_old_name_is_still_seen` test and both
  pre-existing window tests (`a_path_git_would_quote_is_still_seen_by_the_window`,
  `window_is_clear_until_the_base_moves_on_a_changed_file`)

Exact-head CI on the pushed SHA is the authoritative full run (Format; Clippy
×3 OS; Test on Linux + macOS full workspace; Test on the Windows 7-crate
subset; MSRV check).

### DIFF
--- a/crates/edda-cli/src/cmd_review/gate.rs
+++ b/crates/edda-cli/src/cmd_review/gate.rs
@@ -140,10 +140,24 @@ pub(crate) fn union(standing: &[Standing]) -> Union {
 /// so the second diff is empty and the window reads *clear* for exactly the
 /// case the rule exists to catch. `--literal-pathspecs` closes the other half:
 /// a filename containing `*` or `[` is a filename here, never a glob.
+///
+/// Both diffs also pass `--no-renames`: rename detection reports a renamed
+/// file as only its *destination* path, so if the base moved the same file
+/// under its *old* name — or the subject renamed a file the base then edited
+/// under the old name — the name that would have matched never enters the
+/// pathspec set and the window reads clear. `--no-renames` turns a rename
+/// into a delete of the old path plus an add of the new one, so both names
+/// enter, matching whichever side the base actually touched.
 fn window_moved(repo: &Path, sha: &str, base: &str) -> Result<bool> {
     let changed = git(
         repo,
-        &["diff", "--name-only", "-z", &format!("{base}...{sha}")],
+        &[
+            "diff",
+            "--name-only",
+            "-z",
+            "--no-renames",
+            &format!("{base}...{sha}"),
+        ],
     )?;
     let files: Vec<&str> = changed
         .split('\0')
@@ -158,6 +172,7 @@ fn window_moved(repo: &Path, sha: &str, base: &str) -> Result<bool> {
         "--literal-pathspecs",
         "diff",
         "--name-only",
+        "--no-renames",
         range.as_str(),
         "--",
     ];
@@ -461,6 +476,28 @@ mod tests {
         assert!(
             window_moved(&root, &sha, "main").unwrap(),
             "a quoted path must not read as an untouched file"
+        );
+    }
+
+    #[test]
+    fn a_rename_the_base_then_edits_under_the_old_name_is_still_seen() {
+        // `git diff --name-only` applies rename detection: the subject's
+        // rename of a.txt to b.txt comes back as only "b.txt". If the
+        // second diff's pathspec set is built from that output, the base's
+        // edit under the *old* name ("a.txt") matches nothing and the
+        // window falsely reads clear — the same fail-open direction as the
+        // quoted-path bug above, but from treating a rename as an ordinary
+        // single-path change instead of a delete-plus-add.
+        let (_temp, root) = testrepo::init();
+        testrepo::run(&root, &["checkout", "-q", "-b", "subject"]);
+        testrepo::run(&root, &["mv", "a.txt", "b.txt"]);
+        testrepo::run(&root, &["commit", "-q", "-m", "rename a.txt to b.txt"]);
+        let sha = testrepo::run(&root, &["rev-parse", "HEAD"]);
+        testrepo::run(&root, &["checkout", "-q", "main"]);
+        testrepo::commit_file(&root, "a.txt", "base edit\n", "base advance");
+        assert!(
+            window_moved(&root, &sha, "main").unwrap(),
+            "the base's edit under the pre-rename name must not read as clear"
         );
     }
 


fagemx:
## Code Review: Round 1 — PR #1083 @ 858eef8a3250a09f473cffd6d24f344f6af8f8fe

- elapsed: unmeasured (no pi session file — this round ran in a Claude Code session, not via `pi`)
- model_requested: opus (Claude Code review lane)
- model_observed: unverified — this round ran inside an interactive Claude Code session, not a `claude -p --output-format json` invocation, so no top-level `modelUsage` receipt exists to read. Per `brief-v2` §7 the value is written `unverified` rather than copied from `model_requested` or self-labelled.
- spec: review-spec-v1.5
- escalations: none
- shadow: false
- cost: read-only round; 9 tool calls; RAN only `git` / `gh` / `sh scripts/*` (see RAN vs READ). No Cargo run — C5 selector is empty and exact-head CI covers `edda` on Windows, so a rerun had no named gap to close.

### IN SCOPE

**Changed behaviour.** `window_moved` (`crates/edda-cli/src/cmd_review/gate.rs:151`) now passes `--no-renames` to both `git diff` invocations: the subject-side file list (`gate.rs:158`) and the base-advance pathspec check (`gate.rs:175`). One new test (`gate.rs:483`) and one doc-comment paragraph (`gate.rs:144-150`). One file, no other path touched.

**Direct callers/consumers.** `window_moved` has exactly one caller — the `--base` arm of `run` — which turns `true` into `Union::Fail` with `reason = "window"`, surfacing as `FAIL <sha> window` / the JSON `reason` field / exit 1. That surface is unchanged by this diff. `scripts/pr-review-watch.sh` is the one caller of the verb (REVIEW.md §8); it reads the same answer it read before, only now correct across a rename.

**The bug and the fix, reproduced independently.** I did not take the PR body's FAIL-first claim on trust. Built the issue's exact scenario in a scratch repo (git 2.46.0.windows.1) and ran the two diffs in both shapes:

```
--- OLD first diff (rename detection on):     b.txt|
--- NEW first diff (--no-renames):            a.txt|b.txt|
--- OLD second diff (pathspec b.txt):         (empty)   -> window reads CLEAR = the bug
--- NEW second diff (pathspec a.txt b.txt):   a.txt     -> window reads MOVED = fixed
```

So the pre-fix code genuinely fails open on this input and the post-fix code does not. This also independently confirms the new test is FAIL-first: its assertion (`gate.rs:498`) is exactly the `window_moved(...) == true` that the old first diff makes unreachable.

**The FAIL-first receipt is internally consistent.** The PR body's panic cites `gate.rs:483:9`. The fix adds 15 lines ahead of the test block (8 doc lines, net +6 in the first arg list, +1 in the second); 483 + 15 = 498, which is the `assert!` line at head. The quoted panic is therefore from a tree with the test present and the fix absent, not a transcription.

**Test mechanics read, not trusted.** `testrepo::init` (`crates/edda-cli/src/cmd_review/git.rs:236-244`) commits `a.txt` on `main`, so `git mv a.txt b.txt` has a real subject. `git mv` stages the rename, so the bare `commit -q -m` is sufficient. `merge-base(subject, main)` is the initial commit, the second diff's range is `initial..main`, and the base's `a.txt` edit lands inside it. The test asserts the true branch; the pre-existing `window_is_clear_until_the_base_moves_on_a_changed_file` (`gate.rs:505`) still asserts both false branches, so this does not become a test that can only pass.

**Adversarial: does `--no-renames` change any other consumer of these diffs' output?**

- *First diff, `-z` interaction with the #1055 quoted-path fix.* None, and slightly better than none: the output is still NUL-separated and unquoted, and `--no-renames` removes the only `--name-only` shape whose field count per change could differ from one, so the `split('\0')` in `gate.rs:163-166` cannot be desynchronised by a rename. The two fixes are orthogonal and compose.
- *Second diff, `--literal-pathspecs` with the enlarged set.* Unchanged semantics — every added entry is one more literal path, and `--literal-pathspecs` treats each as a filename either way. The result is consumed only as `!trim().is_empty()` (`gate.rs:181`), so nothing downstream parses the enlarged output.
- *Direction of the change.* `--no-renames` can only add entries to the pathspec set and only add rows to the second diff, so the window can move clear→moved and never moved→clear. That is the fail-closed direction, and it is the direction the rule exists to protect. No false-clear is introduced.
- *Could it produce a false "moved"?* Only if a base rename of a file the subject changed should not count as the base advancing — it should; that rename changes the merge tree.
- *Second-diff `--no-renames`: load-bearing or defensive?* Measured, not assumed. With the subject editing `a.txt` and the base renaming `a.txt`→`c.txt`, the pathspec-limited second diff already reports `a.txt` with and without the flag (rename detection cannot pair a destination outside the limited set). So the second flag is defensive symmetry rather than the operative half — harmless, matched to the command pair the issue prescribes, and it removes the dependency on that pathspec-limiting detail. Not a finding; recorded so a later round need not re-derive it.
- *Command-line growth.* The enlarged file set lengthens the second `git` invocation. This is a pre-existing scaling property (the old code already passed every changed file), amplified only by the count of renamed files, and its failure mode is fail-closed: an over-long argv surfaces as an `io::Error` from `Command::output()`, which `git()` propagates through `?` (`git.rs:8-17`) and `run` propagates again — an error, never a silent clear. Not a finding.

**Consistency with #1082 (71e0654).** That PR retensed this module's docs. The added paragraph (`gate.rs:144-150`) is present-tense, describes only `git`'s own behaviour, and makes no claim about the watcher or the merge step; the module-level doc block (`gate.rs:1-36`), including the GH-1061 retirement sentence and the `#671`/`#1057` standing-debt wording, is untouched by this diff. Nothing retensed is reintroduced.

**Issue acceptance.** Both doneWhen items of #1062 delivered (see C1). `Closes #1062` is therefore correct under `pr.closing-keyword=only-when-all-donewhen-delivered`.

**Security / data loss.** None introduced or exposed: no new writes, no process spawn change, no path construction, no deletion. The change strictly tightens a gate that guards merges.

**Current-base integration.** Merge-base with `origin/main` is 71e0654, which is the current `main` tip, so the reviewed tree is the tree that would merge. The base has not advanced over `gate.rs` since the head was pushed.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 1083 --name-only` → `crates/edda-cli/src/cmd_review/gate.rs` only; issue #1062 names exactly `window_moved` in that file |
| U2 | any | P1 | PASS | body line 2 `Closes #1062`; both doneWhen items delivered (C1) — the keyword is correct, not a finding. Commit-message scan: empty |
| U3 | any | P1 | PASS | `Issue: #1062` present in the body |
| U4 | any | P1 | PASS | `git log --format=%s origin/main..858eef8a \| grep -Ev '<conventional>'` → empty (1 commit: `fix(edda-cli): review gate window sees renames across --no-renames (GH-1062)`) |
| U6 | any | P0 | PASS | `gh pr checks 1083` — 12 pass, 2 `skipping` (Fleet shell tests, correct under `ci.path-filter`: no shell file changed). Run 34265064814 `headSha=858eef8a…`, `conclusion=success` |
| U6 (subject) | any | — | N.A.(no CLI probe made) | No backticked runtime command claim is added by this diff, so no installed-vs-source subject needs recording |
| U7 | any | P1 | PASS | Whole scoped audit finished before posting; Round 1, nothing batched for a later round |
| U8 | any | P1 | N.A.(no `Decision:` line) | DECISIONS block output empty |
| C1 | code-plain | P0 | PASS | doneWhen 1 → `gate.rs:483` test, placed directly after `a_path_git_would_quote_is_still_seen_by_the_window` (`gate.rs:465`) as the issue asks, renames `a.txt`→`b.txt`, base edits `a.txt`, asserts `window_moved(...)` true (`gate.rs:498`), and fails on pre-fix code (reproduced at git level, see IN SCOPE). doneWhen 2 → `window_is_clear_until_the_base_moves_on_a_changed_file` (`gate.rs:505`) and `a_path_git_would_quote_is_still_seen_by_the_window` unmodified and green on all three CI Test jobs |
| C2 | code-plain | P0 | PASS | `fix:` subject with a regression test: `git diff … \| grep -cE '^\+.*(#\[test\]\|fn test_)'` → 1, and it is FAIL-first (independently reproduced) |
| C3 | code-plain | P1 | N.A.(the one candidate is in `#[cfg(test)]`) | Candidate `+ window_moved(&root, &sha, "main").unwrap(),` is `gate.rs:499`, inside `mod tests` which opens at `gate.rs:373` |
| C4 | code-plain | P1 | PASS | grep for `unsafe {` / `#[allow(clippy::all)]` / `#![allow` on added lines → empty |
| C5 | code-plain | P1 | N.A.(selector empty) | Selector prints nothing: the only touched crate is `edda-cli`, whose package `edda` is inside CI's Windows 7-crate subset, and `Test (windows-latest)` passed at this SHA. No Windows gap to close, so no local Cargo run was owed or made |
| D1–D5 | docs | — | N.A.(§4 does not route `code-plain` to §5.1) | No `*.md`, `docs/**` or `*.txt` in the changed-file list |
| S1–S3 | skills | — | N.A.(§4 does not route `code-plain` to §5.2) | No skill path in the changed-file list |
| R1–R5 | code-risk | — | N.A.(§4 does not route `code-plain` to §5.4) | `gate.rs` is not on §3.1's enumerated risk surface (only `crates/edda-cli/src/cmd_dispatch.rs` is, for this crate). R1's enumerator was run anyway as a cross-check and printed nothing |
| §5.5 | any | P1 | PASS | See Wiring — no new surfaces; `sh scripts/wiring-scan.sh` reports no new `pub` surfaces and no swallow patterns |

### Wiring

|---|---|---|---|---|
| no new surfaces | — | — | — | — |

No new `pub` fn/field/enum variant, no CLI flag, no config key, no event payload field, no written file or side-file. `--no-renames` is an argument passed to an already-spawned `git`, not a surface this crate exposes. `sh scripts/wiring-scan.sh origin/main 858eef8a…` → `no new pub surfaces` / `no swallow patterns on added lines`. The changed behaviour reaches its layer through the existing chain `window_moved` (`gate.rs:151`) → `run`'s `--base` arm → `Union::Fail` + `reason="window"` → `FAIL <sha> window` / JSON `reason` / exit 1, and #1062's doneWhen pins acceptance at the `window_moved` level, which the new test proves directly.

### Findings

**Zero findings.** Explicitly stated per §6.5, with what was checked: U1–U8 and C1–C5 as tabled above; the wiring slot; the FAIL-first claim reproduced at git level rather than read; the new test's fixture, ranges and assertion direction read line by line; `--no-renames` interaction with `-z`, with `--literal-pathspecs`, and with the #1055 quoted-path fix; the direction of the behaviour change (fail-closed only); whether the second diff's flag is load-bearing (measured — defensive); argv growth and its failure mode; doc-comment tense against #1082; and current-base integration. No P0, no P1, no P2.

### FOLLOW-UP ISSUE

none

### RAN vs READ

RAN (read-only; `git`, `gh`, `sh scripts/*` only — no Cargo, no checkout, no write to the tree):
- `gh pr view 1083 …`, `gh issue view 1062 …` — head 858eef8a…, base 71e0654…, doneWhen loaded
- `git diff --name-only origin/main...858eef8a` → 1 file; `git diff origin/main...858eef8a` → the diff
- `git show 858eef8a:crates/edda-cli/src/cmd_review/{gate.rs,git.rs,subject.rs}` — read at the frozen SHA, not from the working checkout
- Classifier block from `REVIEW.md` at the base → `classes=code-plain`, `canonical_class=code-risk`
- U2/U3/U4/DECISIONS enumerators → `Closes #1062` / `Issue: #1062` / empty / empty
- `sh scripts/lint-markdown-content.sh` → `exit=0` (U5)
- C2/C3/C4/C5 and R1 enumerators → `1` / one `#[cfg(test)]` candidate / empty / empty / empty
- `sh scripts/wiring-scan.sh origin/main 858eef8a…` → no new pub surfaces, no swallow patterns
- **Independent reproduction** of the issue's scenario and of the second-diff shape, in two throwaway repos, with `git diff --name-only [-z] [--no-renames]` in both variants — the outputs quoted in IN SCOPE. This is the check CI cannot do: it establishes the pre-fix fail-open and the post-fix behaviour without trusting the PR body.
- `git log --oneline origin/main -- crates/edda-cli/src/cmd_review/gate.rs` → `71e0654 (#1082)`, `c0ef0b9 (#1055)` — the #1082 retensing baseline
- `gh run view 34265064814 --json headSha,conclusion,status` → pinned to 858eef8a…, success

READ — L1 receipt: **CI run 34265064814 @ 858eef8a3250a09f473cffd6d24f344f6af8f8fe**, conclusion `success`.
- Covers: Format (`cargo fmt --all --check`); Clippy `--workspace --all-targets` on ubuntu, macos **and windows**; Test `--workspace` on ubuntu and macos; Test on the Windows 7-crate subset — which **includes** the `edda` package, i.e. the crate this PR touches, so `cmd_review::gate`'s tests ran on Windows; MSRV 1.91.0 check; plus `Workspace publish dry-run` and `sdk-contract`.
- Does not cover: the 16 non-subset crates' own test targets on Windows — irrelevant here, the C5 selector is empty. `Fleet shell tests` / `Fleet shell tests (windows)` show `skipping`, correct under `ci.path-filter` for a diff with no shell file.
- No workspace Cargo gate was rerun locally: exact-head CI is green and pinned, and no coverage gap was found to justify one (`ladder` L2).
- Also READ: the implementer's L0 receipts in the PR body (fmt clean, `clippy -p edda` clean, `cargo test -p edda cmd_review::gate` 10 passed) — corroborated by, and subordinate to, the exact-head CI result above.

Per this lane's read-only brief the only write made is this comment; the `fleet:reviewed` label (§8) is left to the operator, and merge remains the operator's action (GATE-01).

### Verdict

LGTM (P0=0, P1=0)

PR #1084, opened 2026-09-08
fix(fleet): merge-reviewed-pr.sh asks the union gate, not the latest review (GH-1057)

fagemx:
Issue: #1057

Closes #1057

Decision: mechanism.shell-role

## What was wrong

`scripts/merge-reviewed-pr.sh` is the operator's merge entrypoint. Its verdict
selection was latest-trusted-review-wins:

```
$trusted | sort_by([.updated_at, .id]) | last | .body
```

so an earlier standing `Changes Requested` on the head stopped counting the
moment a later `LGTM (P0=0, P1=0)` existed on the same SHA — the exact failure
the union rule (GH-742, REVIEW.md §8) was written to prevent, and the shape PR
#1055 was merged over on 2026-09-07. The union did not reach it indirectly
either: ruleset 18852689 requires only `CI Gate`, so `Independent Review` never
reached `gh pr checks --required`.

## What this does

One additional refusal, wired to the product:

```sh
report=$(GH_REPO="$repo" "${EDDA_BIN:-edda}" review deliver --pr "$pr" --sha "$head" --json) || true
state=$(printf '%s\n' "$report" | jq -er '.status') || die ...
malformed=$(printf '%s\n' "$report" | jq -er '.malformed | length') || die ...
[ "$state" = success ] || die "union over every trusted review pinned to $head is '$state', not a pass; ..."
[ "$malformed" = 0 ] || die "$malformed verdict comment(s) on $head are malformed and outside the union; refusing"
```

Strictly tightening. Head-pinning, `escalations: none`, the latest-review LGTM
requirement and `gh pr checks --required` all stay; the union is checked before
the required-checks network call so a refused head costs nothing. The accept
path for a head whose only trusted review is a qualifying LGTM is
behavior-identical — same stdout line, same exit code, same `--required` call
(fixture case 2 asserts all three).

`malformed` is its own refusal because a §7 heading below line 1 (#917) is a
round the union never saw: the union can read `success` precisely because a
blocking round is invisible to it.

## Spec adaptation (stated, per the controller's adjudication)

The issue's "What to do" says to pipe `scripts/pr-review-watch.sh
collect-verdicts` into `edda review gate`. **That file no longer exists** —
GH-1061 deleted the review shell (merged today as 71e0654). It was not
resurrected and no third verdict parser was written in shell.

**`edda review deliver`, not `edda review gate` directly.** `gate` is the union
rule and the only implementation of it, but its *input contract* does not fit
[...]

PR #1081, opened 2026-09-08
feat(edda-cli): edda review deliver — the writing half: verdict comment, labels, Independent Review status (GH-1030)

fagemx:
Issue: #1030

This is the **write half** of `edda review deliver` (GH-1030). The read half — turning §7 verdict comments into the verdict facts the union rule consumes — merged as #1077 / 8d838e5e. This PR performs the GitHub writes that follow: the `review:*` label, the `Independent Review` commit status, and the malformed-comment notice.

## What it does

```
$ edda review deliver --pr 1030
success 65f135b66305a815a80bc195e5958a9dc63d4137 verdicts=1
status success done
label review:lgtm done
```

`delivery::deliver` performs, or correctly withholds, three kinds of write over the SHA the read half already extracted:

- **The `Independent Review` commit status**, by the union rule (`edda review gate`, GH-769) — `success` for a clean qualifying LGTM, `failure` while any non-qualifying verdict stands (GH-742: a later LGTM never overrides an earlier Changes Requested on the same SHA), `error` when no verdict stands at all. Written regardless of whether the PR's current head has since moved past the reviewed SHA.
- **The `review:lgtm` / `review:changes-requested` label**, matching the same union state — but **only** while the current head still equals the reviewed SHA. A moved head still gets the status; it never gets the label.
- **A one-shot `review: malformed verdict comment <id>` notice**, for a comment whose §7 heading is not on line 1 (#867/#917) — with no status and no label that run.

A round whose only standing signal is a self-declared ` (SHADOW)` verdict (REVIEW.md §8, rules.md R22) gets **zero GitHub writes** — not even the `error` status an unreviewed SHA would get, because a SHADOW round is calibration evidence, not a judgement to erase.

Every write is idempotent by construction, not by a local marker file: `deliver` reads the label already on the PR, the latest `Independent Review` state already posted for the SHA, and the comments already present, before writing anything — so re-running deliver for an already-delivered verdict performs zero new GitHub calls, from any machine, with no local state to go stale.

## The seam

`cmd_review::delivery::Gh` is a small trait — `head`, `labels`, `add_label`, `latest_status`, `post_status`, `post_comment` — implemented for real by `GhCli` (over two new `github.rs` primitives, `gh_write` for the non-JSON write calls and `pr_head` for a head-only PR read) and by an in-module `FakeGh` for tests. Every doneWhen fixture is a `deliver()` call against `FakeGh`; none touch the network.

## Two divergences from the issue's original "What" — stated, not silently resolved

[...]

PR #1077, opened 2026-09-07
feat(edda-cli): read §7 verdict comments in the product — the parsing half of edda review deliver (GH-1030)

fagemx:
Issue: #1030

Delivery is the last review-loop capability with no product verb. The §7 verdict comment, the `review:*` labels and the `Independent Review` commit status exist only in `scripts/pr-review-watch.sh` — which is the layer that broke on 2026-09-06/07, when the product could judge but could not publish.

This PR is the **read half**, and deliberately only that: turning §7 comments into the verdict facts the union rule consumes. It is the half that has to be exactly right before anything writes to GitHub.

## What it does

`edda review deliver --pr <N> [--sha <sha>] [--json]` reads the PR's comments, extracts the verdicts standing on one reviewed SHA, and reports what they amount to:

```
$ edda review deliver --pr 1030
failure 65f135b66305a815a80bc195e5958a9dc63d4137 verdicts=2
malformed 5573431960
```

The extracted records are the tab-separated `verdict<TAB>p0<TAB>p1` shape `gate::from_lines` already parses, so the union rule stays exactly where #769 put it and this module never re-decides what a verdict means. That reuse is the design: `edda review gate --verdicts -` has been able to judge a caller's verdict lines since #769, but the only thing that could *produce* those lines was the watcher's awk.

## The rules it moves out of shell

Each one has a test that fails without it.

- **A verdict comment BEGINS with the §7 heading.** A heading anywhere else in the body is a transcript dump (#867) — no verdict, no status, no label — and earns one `malformed <id>` report so the round is not lost silently (#917).
- **` (SHADOW)` is accepted in the heading, in both recorded positions, but not in the SHA pin.** A SHADOW round is therefore *well formed* — it raises no malformed notice, so calibration rounds do not spam one — and contributes nothing, because it is never a verdict (REVIEW.md §8). Well-formed and non-contributing are different states; conflating them breaks one of the two.
- **Classification is ordered, not keyword-matched:** `Changes Requested`, then the `Provisional — …` line written for an unqualified LGTM (exit 3, §6.4, #998), then plain `LGTM`. A blocking line reading "the LGTM cannot stand" contains the word LGTM.
- **An absent `P0=`/`P1=` count stays absent.** `gate::from_lines` reads a missing count as non-zero and fails closed; writing `0` here would invent a clean round out of a silent one.
- **A SHA that is not 40 lowercase hex is not a heading** (REVIEW.md R5).

## Not in this PR

[...]

Issue #917, opened 2026-09-05
fix(fleet): verdict publication posts the report only, heading first; watcher rejects transcript dumps and skips non-open PRs


fagemx:
## Evidence

PR #867, comment `2026-09-05T02:37:15Z`: one comment holding roughly twelve paragraphs of the reviewer's own narration ("I'll start by loading the project guide…", "Rerunning with a backslash-free harness…"), then `---`, then the complete `## Code Review: Round 3 — PR #867 @ 3eadccf3…` report. The reviewer (`wave867-glm-final`, read-only per contract) wrote the report inline "for the controller to publish to the PR"; the controller published the entire session output as the verdict comment.

What followed:

| time (Z) | event |
|---|---|
| 02:37:15 | transcript-dump comment posted; the verdict heading is mid-body |
| 02:37:18 | `Independent Review=success` set on `3eadccf3` — `pr-review-watch.sh:200` matches the heading as a *line*, so it still found the LGTM |
| 02:37:53 | merged to `main` (`f675991`) |
| 02:38:36, 02:38:41 | `review: started on 3eadccf3…` posted twice — by the watcher, on an already-merged PR |

Any tool that anchors on "the comment begins with `## Code Review: Round`" — the natural shape for a drift check (#914) or a compare script (#907) — does not see this verdict at all. The operator session's drift scan over the six in-window merges reported #867's latest verdict as sol's Round 2 at `9407f8b9` for exactly that reason, and had to be corrected by reading the comment by hand.

## Gap

1. Nothing checks the shape of what gets posted as a verdict. A verdict comment is a contract surface (REVIEW.md §7 fixes its fields and table shape); a transcript dump around it satisfies the watcher's line match while breaking every parser that treats the comment as the unit.
2. Nothing stops the watcher launching a review on a PR that is no longer open.

## doneWhen

1. The publish step — whichever script posts a verdict body (`scripts/review-pr.sh` lane done-handling, `scripts/fleet/next-review.sh` once #899 merges, `scripts/fleet/manager-tick.sh`) — posts only the text from the first `## Code Review: Round` line onward. If that line is absent it posts nothing and exits non-zero naming the PR.
2. `scripts/pr-review-watch.sh` treats a verdict comment whose first line is not the heading as malformed: no status, no label, one `review: malformed verdict comment <comment id>` comment, non-zero exit for that PR.
[...]

PR #1086, opened 2026-09-09
fix(edda-cli): deliver read path — argv regression guard on the gh seam, and --sha validates PR-vs-issue (GH-1079)

### LINKED ISSUE #1079: fix(edda-cli): edda review deliver read path — argv-level regression guard on the gh seam, and --sha bypasses PR-vs-issue validation
## What was observed

Two follow-ups routed by PR #1077's review rounds (Round 2, https://github.com/fagemx/edda/pull/1077#issuecomment-5580247270, pinned to `8df257f9`):

1. **Nothing asserts the `gh` argv on the comment-source seam.** A revert from `gh api --paginate repos/{owner}/{repo}/issues/{pr}/comments` back to `gh pr view --json comments` would reintroduce Round 1's P1 (base64 node ids where the #917 notice contract needs numeric, plus an unpaginated union input on the merge-gate path) with all module tests still green — the tests feed parsed JSON and never see the argv.
2. **`--sha` bypasses PR-vs-issue validation.** `run` calls `resolve_pr` only when `--sha` is absent, so `--pr <issue-number> --sha <40hex>` reaches `comments()` unvalidated. Measured: `gh pr view 1030 --json comments` exits 1 ("Could not resolve to a PullRequest") while `gh api .../issues/1030/comments` exits 0 with 7 comments. Fails closed today (an issue thread has no §7 heading pinned to the SHA → union `None` → `error`; it cannot manufacture `success`) and is parity with `pr-review-watch.sh:365` — hence P2, not a blocker.

## Predicted surface

`crates/edda-cli/src/cmd_review/deliver.rs` and its tests. The argv assertion wants the injectable-gh seam GH-1030's writing half introduces.

## doneWhen

- A test asserts the exact argv of the comment-source `gh` call (REST endpoint + `--paginate`), and fails on a revert to `gh pr view --json comments`.
- The `--sha` path validates the number is a PR before fetching comments; a fixture proves `--pr <issue> --sha <hex>` errors with the PR-vs-issue distinction instead of reading issue comments.
- Tests FAIL-first; `cargo test -p edda` green; L0 on the touched crate.

## Ordering

After GH-1030's writing half (shares `deliver.rs` and needs its gh injection seam; same-file cluster — sequential).

fagemx:
Issue: #1079

Closes #1079

## What

Two follow-ups from PR #1077's Round 2 review, both in
`crates/edda-cli/src/cmd_review/deliver.rs`:

1. **Argv-level regression guard on the comment-source `gh` seam.** Every
   existing test in this module feeds `extract`/`parse_comments` already-
   parsed JSON and never sees what `comments()` actually asked `gh` for — so
   a revert of the REST endpoint back to `gh pr view --json comments` (base64
   node ids where the #917 notice contract needs numeric ids, plus an
   unpaginated union input on the merge-gate path) would have left every
   module test green. Split the argv construction into `comments_argv(pr)`,
   which `comments()` now actually consumes (not a second, independent
   literal standing beside it), and added a test that locks its exact shape.
   Because `comments()` has no other source for its argv, a revert either
   changes what `comments_argv` returns (test goes red) or leaves it unused
   (dead code under `-D warnings`).

2. **`--sha` bypassed PR-vs-issue validation.** `deliver_inner` only calls
   `resolve_pr` (which fails closed on an issue number) when `--sha` is
   absent, so `--pr <issue-number> --sha <hex>` reached `comments()`
   unvalidated: `gh`'s REST issue-comments endpoint answers success on an
   issue number too. Added `validate_pr_for_sha`, which reuses
   `github::pr_head` (already a PR-only `gh pr view` call) as the one `gh`
   call that stands in for `resolve_pr`'s own check on this path. The
   ordinary (no `--sha`) path gets no new call — `resolve_pr` already does
   this validation as a side effect of its own fetch.

   This matters more than the original P2 severity suggested: PR #1084 made
   `merge-reviewed-pr.sh` call `edda review deliver --pr N --sha <head>
   --json`, so `--sha` is now on a write path (labels/status, since #1081)
   exercised by the operator's real merge entrypoint — not just a read-only
   curiosity.

## Design note: why a closure seam instead of extending the `Gh` trait

`delivery.rs`'s injectable `Gh` trait (head/labels/add_label/remove_label/
latest_status/post_status/post_comment) exists so union/idempotency/failure-
contract fixtures never touch the network. I considered folding the
comment-read and the PR-vs-issue probe into it, but both `comments()` and
`validate_pr_for_sha()` are used in `deliver_inner` *before* a `GhCli` is
constructed, and `Gh`'s own writes are keyed to a reviewed SHA already in
hand — reshaping it to also own the argv-level read half would only be
justified if this needed the same idempotency/failure-contract matrix
`delivery::tests` builds. It doesn't: both new guards want to assert one
call's exact input, not sequence multiple calls against seeded state. A
`comments_argv(pr) -> Vec<String>` pure function and a
`validate_pr_for_sha(pr, probe: impl FnOnce(u64) -> Result<String>)` closure
seam are the minimal shapes that make each testable in isolation, and both
are actually consumed (not just parallel literals) by the production call
sites, which is what keeps a revert from leaving the guard test green.

### FAIL-first (captured before the fix; both new guards go red on a
simulated pre-fix `deliver.rs`, every pre-existing test stays green)

Simulated the exact regressions doneWhen names — `comments_argv` reverted to
the `gh pr view --json comments` shape, `validate_pr_for_sha` reverted to a
no-op `Ok(())` — then ran the focused suite:

```
test cmd_review::deliver::tests::comments_argv_is_the_paginated_rest_endpoint_not_the_graphql_view ... FAILED
test cmd_review::deliver::tests::comments_argv_is_not_the_old_pr_view_json_comments_shape ... FAILED
test cmd_review::deliver::tests::sha_path_rejects_a_number_gh_cannot_resolve_as_a_pull_request ... FAILED
test cmd_review::deliver::tests::sha_path_probe_is_called_with_the_pr_number_argument ... FAILED
test result: FAILED. 34 passed; 4 failed; 0 ignored; 0 measured; 689 filtered out
```

Only the 4 new tests failed; every other test in `cmd_review::deliver` and
`cmd_review::delivery` stayed green — confirming the old suite really would
have missed both regressions, and the new tests target exactly the intended
gap. Reverted the simulation and re-ran to confirm green (below).

### L0, touched crate (`edda`), lane `worker-2`

- `cargo fmt --all --check` — clean (one formatting diff from the new test
  fixed with `cargo fmt --all`, then re-verified clean)
- `cargo clippy -p edda --all-targets -- -D warnings` — clean, zero warnings
- `cargo test -p edda cmd_review::deliver` — `test result: ok. 38 passed; 0
  failed; 0 ignored; 0 measured; 689 filtered out` (module tests only; the
  crate's other integration suites that need a live Codex server / `sh` on
  PATH were not run, per the brief)

Head SHA: `def6e0e651d70ce672cf22d721c387a5809e63c8`

## Scope note

Both findings were confined to `crates/edda-cli/src/cmd_review/deliver.rs`;
`delivery.rs` (the write half, `Gh` trait, `FakeGh` fixtures) is unchanged —
its own 30+ tests still pass and were re-run as part of the focused filter
above.

### DIFF
--- a/crates/edda-cli/src/cmd_review/deliver.rs
+++ b/crates/edda-cli/src/cmd_review/deliver.rs
@@ -333,9 +333,10 @@ pub fn run(args: DeliverArgs, cwd: &Path) -> Result<()> {
 fn validate_pr_for_sha(pr: u64, probe: impl FnOnce(u64) -> Result<String>) -> Result<()> {
     probe(pr).map(|_| ()).with_context(|| {
         format!(
-            "--pr {pr} did not resolve as a pull request (gh could not find a PR #{pr} — is \
-             {pr} an issue instead? --sha only skips the PR head fetch, not this PR-vs-issue \
-             check, so it would otherwise read issue #{pr}'s comments)"
+            "--pr {pr} did not resolve as a pull request (either {pr} is an issue rather than \
+             a PR, or gh failed to confirm it — see the gh error this is chained to; --sha \
+             only skips the PR head fetch, not this PR-vs-issue check, so an issue number \
+             would otherwise read issue #{pr}'s comments)"
         )
     })
 }
@@ -345,25 +346,68 @@ fn validate_pr_for_sha(pr: u64, probe: impl FnOnce(u64) -> Result<String>) -> Re
 /// status/label withheld under R23/#917 — never returned as an `Err`, since
 /// something either succeeded or was deliberately withheld this round; see
 /// [`delivery::Delivery::exit_code`]).
+///
+/// Thin production wrapper over [`deliver_inner_with`]: the real PR-vs-issue
+/// probe is [`super::github::pr_head`] and the real comment source is
+/// [`comments`]. Split out so a fixture can drive the same sequence with both
+/// seams faked — see that function's doc comment (GH-1079 Round 1 P1).
 fn deliver_inner(args: &DeliverArgs, cwd: &Path) -> Result<i32> {
+    deliver_inner_with(
+        args,
+        cwd,
+        |number| super::github::pr_head(cwd, number),
+        comments,
+    )
+}
+
+/// [`deliver_inner`] with its `gh` seams injected.
+///
+/// This is what makes the `--pr <issue-number> --sha <hex>` wiring and
+/// ordering testable without shelling out: a fixture can fake `probe` to
+/// fail the way `gh` fails on an issue number, and fake `comments_fn` to
+/// record whether it was ever reached, then assert both that the failure
+/// surfaces naming `args.pr` and that `comments_fn` was never called — GH-1079
+/// Round 1 P1. The prior test at this call site (`validate_pr_for_sha`
+/// invoked directly) only proved that helper hands its own argument to its
+/// own closure, which stays green even if the call below were deleted,
+/// reordered after the comment read, or hardcoded to a number other than
+/// `args.pr`; the seam here lets a fixture reach the call site itself.
+fn deliver_inner_with(
+    args: &DeliverArgs,
+    cwd: &Path,
+    probe: impl FnOnce(u64) -> Result<String>,
+    comments_fn: impl FnOnce(&Path, u64) -> Result<Vec<Comment>>,
+) -> Result<i32> {
     let sha = match &args.sha {
         Some(sha) => {
+            // GH-1079 Round 1 P2: the free local shape check runs first, so
+            // a malformed --sha is rejected without spending the gh
+            // round-trip below.
+            anyhow::ensure!(
+                is_full_sha(sha),
+                "--sha must be a full 40-character lowercase hex commit"
+            );
             // GH-1079: without --sha, resolve_pr's own PR-only GraphQL fetch
             // already fails closed on an issue number. --sha skips that
             // fetch entirely, so this is the one gh call that stands in for
             // it here — gh's REST issue-comments endpoint below (unlike
             // resolve_pr) answers success on an issue number too, so nothing
             // downstream would otherwise notice #<pr> was never a PR.
-            validate_pr_for_sha(args.pr, |number| super::github::pr_head(cwd, number))?;
+            validate_pr_for_sha(args.pr, probe)?;
             sha.clone()
         }
         None => super::github::resolve_pr(cwd, args.pr)?.head,
     };
+    // Redundant with the Some(sha) arm's own check above, but kept here too:
+    // it is the only shape guard on the None arm's resolved head (resolve_pr
+    // bails on a wrong-length or non-hex head, but not on stray uppercase —
+    // is_full_sha does), and it stays cheap enough that checking it twice on
+    // the --sha path costs nothing worth removing it for.
     anyhow::ensure!(
         is_full_sha(&sha),
         "--sha must be a full 40-character lowercase hex commit"
     );
-    let existing_comments = comments(cwd, args.pr)?;
+    let existing_comments = comments_fn(cwd, args.pr)?;
     let extracted = extract(&sha, &existing_comments);
     let union = gate::union(&gate::from_lines(&extracted.lines.join("\n")));
     let state = match union {
@@ -690,14 +734,57 @@ mod tests {
     }
 
     #[test]
-    fn sha_path_probe_is_called_with_the_pr_number_argument() {
-        // Guards the wiring itself: deliver_inner must probe the *same*
-        // number `--pr` carries, not a hardcoded or mismatched one.
-        let seen = std::cell::Cell::new(0u64);
-        let _ = validate_pr_for_sha(1030, |number| {
-            seen.set(number);
-            Ok("deadbeef".repeat(5))
-        });
-        assert_eq!(seen.get(), 1030);
+    fn deliver_inner_probes_the_pr_number_before_reading_comments_and_names_it_on_failure() {
+        // GH-1079 Round 1 P1: the test that used to stand here
+        // (`sha_path_probe_is_called_with_the_pr_number_argument`) called
+        // `validate_pr_for_sha` directly and asserted its closure saw the
+        // number it was given — true no matter what `deliver_inner` does,
+        // since it only proves that helper hands its own argument to its own
+        // closure (`validate_pr_for_sha`'s entire body is `probe(pr)`). This
+        // drives `deliver_inner_with` instead — the real `--pr <issue-number>
+        // --sha <hex>` layer — through its injected seams, and checks all
+        // three ways the wiring could break: the validation call deleted,
+        // moved after the comment read, or wired to a number other than the
+        // one `--pr` carries.
+        let args = DeliverArgs {
+            pr: 4242,
+            sha: Some(SHA.to_owned()),
+            json: false,
+        };
+        let seen_probe = std::cell::Cell::new(None);
+        let comments_reached = std::cell::Cell::new(false);
+
+        let result = deliver_inner_with(
+            &args,
+            Path::new("."),
+            |number| {
+                seen_probe.set(Some(number));
+                Err(anyhow::anyhow!(
+                    "gh: GraphQL: Could not resolve to a PullRequest with the number of \
+                     {number}. (repository.pullRequest)"
+                ))
+            },
+            |_cwd, _pr| {
+                comments_reached.set(true);
+                Ok(Vec::new())
+            },
+        );
+
+        let err = result.expect_err("a PR that gh cannot resolve must fail deliver_inner");
+        let message = format!("{err:#}");
+        assert!(
+            message.contains("4242"),
+            "the surfaced error must name the number --pr carries: {message}"
+        );
+        assert_eq!(
+            seen_probe.get(),
+            Some(4242),
+            "the probe must be called with the number --pr carries, not a hardcoded \
+             or mismatched one"
+        );
+        assert!(
+            !comments_reached.get(),
+            "comments must never be fetched once PR-vs-issue validation has failed"
+        );
     }
 }
--- a/crates/edda-cli/src/cmd_review/deliver.rs
+++ b/crates/edda-cli/src/cmd_review/deliver.rs
@@ -243,26 +243,37 @@ pub(crate) fn extract(sha: &str, comments: &[Comment]) -> Extracted {
     out
 }
 
-/// The §7 comments GitHub holds for one PR.
+/// The exact `gh` argv [`comments`] shells out with — REST (`gh api`), not
+/// GraphQL (`gh pr view --json comments`): the GraphQL shape carries only
+/// base64 node ids, which the #917 malformed-notice contract cannot use — it
+/// needs the numeric id a human can resolve in the UI. `--paginate` alone (no
+/// `--slurp`) is enough: `gh` combines an array-shaped REST endpoint's pages
+/// into one JSON array before this ever sees it (verified live against this
+/// repo, forcing multiple pages with `per_page=2`), so a comment list longer
+/// than one page is never silently truncated on the merge-gate path.
 ///
-/// REST (`gh api`), not GraphQL (`gh pr view --json comments`): the GraphQL
-/// shape carries only base64 node ids, which the #917 malformed-notice
-/// contract cannot use — it needs the numeric id a human can resolve in the
-/// UI. `--paginate` alone (no `--slurp`) is enough: `gh` combines an
-/// array-shaped REST endpoint's pages into one JSON array before this ever
-/// sees it (verified live against this repo, forcing multiple pages with
-/// `per_page=2`), so a comment list longer than one page is never silently
-/// truncated on the merge-gate path.
+/// Split out from [`comments`] — and actually consumed by it below, not just
+/// left standing beside it — so the argv is asserted directly (GH-1079):
+/// every prior test here feeds `extract`/`parse_comments` parsed JSON and
+/// never sees what was actually requested, so Round 1's P1 (this endpoint
+/// reverted back to `gh pr view --json comments`) left every module test
+/// green. Because `comments` builds its call from this function's return
+/// value rather than a second, independent literal, a revert has nowhere to
+/// hide: either it changes what this returns (the test below goes red), or
+/// it leaves `comments` not calling this at all (dead code, `-D warnings`).
+fn comments_argv(pr: u64) -> Vec<String> {
+    vec![
+        "api".to_owned(),
+        "--paginate".to_owned(),
+        format!("repos/{{owner}}/{{repo}}/issues/{pr}/comments"),
+    ]
+}
+
+/// The §7 comments GitHub holds for one PR.
 pub(crate) fn comments(repo: &Path, pr: u64) -> Result<Vec<Comment>> {
-    let value = gh(
-        repo,
-        &[
-            "api",
-            "--paginate",
-            &format!("repos/{{owner}}/{{repo}}/issues/{pr}/comments"),
-        ],
-    )
-    .with_context(|| format!("read comments of PR #{pr}"))?;
+    let argv = comments_argv(pr);
+    let args: Vec<&str> = argv.iter().map(String::as_str).collect();
+    let value = gh(repo, &args).with_context(|| format!("read comments of PR #{pr}"))?;
     Ok(parse_comments(&value))
 }
 
@@ -307,6 +318,28 @@ pub fn run(args: DeliverArgs, cwd: &Path) -> Result<()> {
     }
 }
 
+/// Confirm `pr` is a pull request before trusting an operator-supplied
+/// `--sha` to skip [`super::github::resolve_pr`]'s own PR-only fetch.
+///
+/// `--pr <issue-number> --sha <hex>` used to reach [`comments`] unvalidated:
+/// `resolve_pr` is the only place that ever asked GitHub "is this really a
+/// PR?", and `--sha` bypasses it (measured: `gh pr view 1030 --json comments`
+/// exits 1 while `gh api .../issues/1030/comments` exits 0 with real
+/// comments — GH-1079). `probe` is the injected PR-only check (the real path
+/// passes [`super::github::pr_head`], which already exists for a lighter
+/// reason and happens to fail exactly the way `resolve_pr` does on an issue
+/// number); its `Ok` value is discarded here — this only wants its failure
+/// mode. Tests fake `probe` to reproduce that failure without shelling out.
+fn validate_pr_for_sha(pr: u64, probe: impl FnOnce(u64) -> Result<String>) -> Result<()> {
+    probe(pr).map(|_| ()).with_context(|| {
+        format!(
+            "--pr {pr} did not resolve as a pull request (gh could not find a PR #{pr} — is \
+             {pr} an issue instead? --sha only skips the PR head fetch, not this PR-vs-issue \
+             check, so it would otherwise read issue #{pr}'s comments)"
+        )
+    })
+}
+
 /// The read, decide, and write sequence; returns the exit code `run` should
 /// use on success (0 delivered, 1 partially delivered, 2 failed, 3 a due
 /// status/label withheld under R23/#917 — never returned as an `Err`, since
@@ -314,7 +347,16 @@ pub fn run(args: DeliverArgs, cwd: &Path) -> Result<()> {
 /// [`delivery::Delivery::exit_code`]).
 fn deliver_inner(args: &DeliverArgs, cwd: &Path) -> Result<i32> {
     let sha = match &args.sha {
-        Some(sha) => sha.clone(),
+        Some(sha) => {
+            // GH-1079: without --sha, resolve_pr's own PR-only GraphQL fetch
+            // already fails closed on an issue number. --sha skips that
+            // fetch entirely, so this is the one gh call that stands in for
+            // it here — gh's REST issue-comments endpoint below (unlike
+            // resolve_pr) answers success on an issue number too, so nothing
+            // downstream would otherwise notice #<pr> was never a PR.
+            validate_pr_for_sha(args.pr, |number| super::github::pr_head(cwd, number))?;
+            sha.clone()
+        }
         None => super::github::resolve_pr(cwd, args.pr)?.head,
     };
     anyhow::ensure!(
@@ -585,5 +627,77 @@ mod tests {
     fn parse_comments_on_an_unexpected_shape_yields_no_comments_rather_than_panicking() {
         let got = parse_comments(&serde_json::json!({"not": "an array"}));
         assert!(got.is_empty());
+    }
+
+    // ---- GH-1079: argv-level regression guard on the comment-source seam --
+
+    #[test]
+    fn comments_argv_is_the_paginated_rest_endpoint_not_the_graphql_view() {
+        // Locks the exact call `comments` shells out with. A revert back to
+        // `gh pr view <pr> --json comments` (Round 1's P1: base64 node ids
+        // where #917 needs numeric ones, plus an unpaginated union input on
+        // the merge-gate path) must change this function's return value —
+        // `comments` has no other source for its argv — so this goes red on
+        // that revert instead of staying green like every test above it,
+        // which only ever sees already-parsed JSON.
+        assert_eq!(
+            comments_argv(1030),
+            vec![
+                "api",
+                "--paginate",
+                "repos/{owner}/{repo}/issues/1030/comments"
+            ],
+        );
+    }
+
+    #[test]
+    fn comments_argv_is_not_the_old_pr_view_json_comments_shape() {
+        // The specific shape a revert would reintroduce, spelled out so the
+        // guard fails for the right, legible reason rather than merely any
+        // mismatch.
+        let argv = comments_argv(1030);
+        assert_ne!(argv, vec!["pr", "view", "1030", "--json", "comments"]);
+    }
+
+    // ---- GH-1079: --sha validates PR-vs-issue before reading comments -----
+
+    #[test]
+    fn sha_path_rejects_a_number_gh_cannot_resolve_as_a_pull_request() {
+        // Reproduces gh's real failure mode on an issue number
+        // (`gh pr view 1030 --json headRefOid` => "Could not resolve to a
+        // PullRequest with the number of 1030.") without shelling out.
+        let err = validate_pr_for_sha(1030, |_| {
+            Err(anyhow::anyhow!(
+                "gh: GraphQL: Could not resolve to a PullRequest with the number of 1030. \
+                 (repository.pullRequest)"
+            ))
+        })
+        .expect_err("an issue number must not validate as a PR");
+        let message = format!("{err:#}");
+        assert!(
+            message.contains("pull request") || message.contains("PR"),
+            "message must name the PR-vs-issue distinction: {message}"
+        );
+        assert!(
+            message.contains("1030"),
+            "message must name the offending number: {message}"
+        );
+    }
+
+    #[test]
+    fn sha_path_accepts_a_number_gh_resolves_as_a_pull_request() {
+        assert!(validate_pr_for_sha(1030, |_| Ok("deadbeef".repeat(5))).is_ok());
+    }
+
+    #[test]
+    fn sha_path_probe_is_called_with_the_pr_number_argument() {
+        // Guards the wiring itself: deliver_inner must probe the *same*
+        // number `--pr` carries, not a hardcoded or mismatched one.
+        let seen = std::cell::Cell::new(0u64);
+        let _ = validate_pr_for_sha(1030, |number| {
+            seen.set(number);
+            Ok("deadbeef".repeat(5))
+        });
+        assert_eq!(seen.get(), 1030);
     }
 }


fagemx:
## Code Review: Round 1 — PR #1086 @ def6e0e651d70ce672cf22d721c387a5809e63c8

- elapsed: unmeasured (controller-spawned subagent; no pi session file)
- model_requested: claude (Opus, via `review.dispatch-transport=controller-subagent-direct`)
- model_observed: unverified (no session file / `--output-format json` envelope to read)
- reviewer_session: n/a (controller-spawned subagent, not a launcher-issued per-PR UUID)
- spec: review-spec-v1.5
- class: code-risk  (REVIEW.md classes: code-plain — up-classed for §5.4, see IN SCOPE)
- escalations: none
- shadow: false
- cost: 1 file / +92 −13; ~16 tool calls; zero Cargo runs (C5 selector empty — reason below)

### IN SCOPE

Changed behaviour and paths: `crates/edda-cli/src/cmd_review/deliver.rs` only — (a) the
comment-source `gh` argv split into `comments_argv(pr)` and consumed by `comments()`;
(b) a new `validate_pr_for_sha(pr, probe)` wired into `deliver_inner`'s `Some(sha)` arm.

Direct callers/consumers: `scripts/merge-reviewed-pr.sh:107`
(`edda review deliver --pr "$pr" --sha "$head" --json`, GH-1057/#1084) is the only
production caller of the `--sha` path; `crates/edda-cli/src/cmd_review/mod.rs:74` is the
dispatch. `delivery.rs` unchanged.

Acceptance ceiling: issue #1079 body + its 2026-09-08 escalation comment (`--sha` became a
write-path entry after #1081).

Security / data-loss surface: the `--sha` path can publish an `Independent Review` status
and a `review:*` label on an arbitrary commit — the reason the escalation raised this
issue's priority. The change is in the fail-closed direction throughout.

Current-base integration: `git merge-base origin/main def6e0e6` = `16f7b4b4` =
`origin/main` tip; `git log 16f7b4b4..origin/main -- crates/edda-cli/src/cmd_review/deliver.rs`
empty. Window clean. Head re-read at posting time: still `def6e0e6`.

Up-class rationale (§3, "you may up-class and must state why"): the mechanical router puts
this at `code-plain` — `deliver.rs` is not on §3.1's enumerated risk surface. I ran §5.4
anyway because this file's output is what `merge-reviewed-pr.sh` turns into a merge
refusal, so a defect here escapes the review gate. R1/R2/R3 are vacuous (no shell files);
R4/R5 judged below.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 1086 --name-only` → `crates/edda-cli/src/cmd_review/deliver.rs` only; inside #1079's "Predicted surface" |
| U2 | any | P1 | PASS | `Closes #1079` at body line 3. doneWhen 1 ✓, doneWhen 3 ✓ (CI, below), doneWhen 2's *code* ✓ — only its fixture layer is short, which is this round's P1 and blocks the merge, so the auto-close hazard cannot materialise on this SHA. Keyword is correct for the intended full delivery |
| U3 | any | P1 | PASS | `Issue: #1079` present |
| U4 | any | P1 | PASS | offender grep empty; 1 commit, `fix(edda-cli): deliver read path — argv guard on the gh seam, --sha validates PR-vs-issue (GH-1079)` |
| U6 | any | P0 | PASS | `gh pr checks 1086` → 12 pass, 2 `skipping` (Fleet shell tests, `ci.path-filter` — correct for a non-fleet path). `check-runs` `total_count=14` @ `def6e0e6`; run `34297458946` `head_sha=def6e0e6`, `conclusion=success`. READ, not RAN |
| U7 | any | P1 | PASS | whole scoped audit finished before posting; the single blocker is batched here |
| U8 | any | P1 | N.A.(no `Decision:` line in the PR body — DECISIONS block printed nothing) | |
| D1–D5 | docs | P1/[判斷] | N.A.(class is code-plain/code-risk; §4 does not route §5.1) | The one new *factual* claim inside the added doc comment was probed anyway: `gh pr view 1030 --json comments` → exit 1; `gh api repos/fagemx/edda/issues/1030/comments` → exit 0, 8 comments. Claim holds (`deliver.rs:326-330`) |
| S1–S3 | skills | P0/P1 | N.A.(no skills/instruction files in the diff) | |
| C1 | code-plain | P0 | PASS | doneWhen 1 → `deliver.rs:264` + `:274` + tests `:634`/`:653`. doneWhen 2 → `deliver.rs:333` wired at `:357`, ahead of `comments()` at `:372`. doneWhen 3 → CI. Every checkbox has code behind it; the test-proof shortfall on item 2 is §5.5's clause, routed there rather than double-counted as a C1 P0 |
| C2 | code-plain | P0 | PASS | `fix:` with regression tests — 5 added (`git diff … \| grep -cE '^\+.*(#\[test\]\|fn test_)'` → 5). FAIL-first verified by inspection, not re-run: reverting `comments_argv` to the `pr view` shape reds `:634` and `:653`; reverting `validate_pr_for_sha` to `Ok(())` reds `:668` (`expect_err` panics) and `:692` (`seen` stays 0) while `:687` stays green — exactly the body's "34 passed; 4 failed" against 38 total (33 pre-existing + 5 new). Arithmetic and revert semantics both consistent |
| C3 | code-plain | P1 | PASS | grep empty (`.expect_err(` does not match; it is inside `#[cfg(test)] mod tests` regardless) |
| C4 | code-plain | P1 | PASS | grep empty — no `unsafe`, no `clippy::all`, no crate-level `allow` |
| C5 | code-plain | P1 | N.A.(selector empty) | `crates/edda-cli` → package `edda`, which is inside CI's Windows 7-crate subset (`.github/workflows/ci.yml`, windows matrix `args:` lists `-p edda`). Nothing for the verifier to run; a rerun would be over-verification with no stated reason (`ladder` L2) |
| R1 | code-risk | P0 | PASS | grep empty — no destructive command added |
| R2 | code-risk | P0/P1 | PASS | grep empty — no shell file in the diff, and no added Rust line mixes `\|\|` with `&&` |
| R3 | code-risk | P0 | N.A.(no `*.sh`/`*.bash` in the diff) | |
| R4 | code-risk | P1 | PASS | no temp dir, file or lock created; nothing to release |
| R5 | code-risk | P1 | PASS | `format!("repos/{{owner}}/{{repo}}/issues/{pr}/comments")` interpolates a `u64`, and `gh` is invoked through `Command::args` — no shell, no quoting surface. `--sha` still gated by `is_full_sha` (40 lowercase hex) at `deliver.rs:365` before any write. Two distinct inputs cannot collide on a filename here — no file is written |
| §5.5 Wiring | any | P1 | **FAIL** | see the Wiring table and the P1 below |

### Wiring

`scripts/wiring-scan.sh origin/main def6e0e6` → `no new pub surfaces`. Both additions are
module-private, so the table below is the substantive read rather than the scanner's.

|---|---|---|---|---|
| `comments_argv(pr) -> Vec<String>` (`deliver.rs:264`) | pure fn, returns `["api", "--paginate", "repos/{owner}/{repo}/issues/{pr}/comments"]` | `comments()` at `deliver.rs:274-276` — sole production call site (`git grep comments_argv def6e0e6` → one non-test use); `gh()` bubbles a non-zero exit as `Err` | not swallowed: `gh()` at `github.rs:16` `bail!`s on non-success; `.with_context("read comments of PR #{pr}")` | argv → `gh()` → `Command`. Asserted at `deliver.rs:634` against the exact vec. **Sound**: `comments()` has no second literal, so a revert either changes the asserted return value or leaves `comments_argv` unused → `dead_code` under `-D warnings`. Both prongs of the PR's claim 1 verified |
| `validate_pr_for_sha(pr, probe)` (`deliver.rs:333`) | returns `Result<()>`; `probe`'s `Ok` discarded | `deliver_inner` at `deliver.rs:357`, sole production call site, inside the `Some(sha)` arm and ahead of `comments()` at `:372` | fails closed: any probe `Err` → `?` → `run()` exits 2, and `anyhow`'s `{error:#}` keeps `gh`'s own stderr in the chain | closure → `github::pr_head` → `gh pr view <n> --json headRefOid`. **Gap**: no test reaches `deliver_inner`; `:692` claims to guard that reach and does not (P1 below). Deletion of the call site is caught by `dead_code`; a *mismatched argument* at the call site is caught by nothing |

### Findings

- **[P1]** `sha_path_probe_is_called_with_the_pr_number_argument` does not guard what its
  comment says it guards, and nothing else does. `deliver.rs:694-695` states *"Guards the
  wiring itself: deliver_inner must probe the **same** number `--pr` carries, not a
  hardcoded or mismatched one."* The test body (`deliver.rs:696-701`) calls
  `validate_pr_for_sha(1030, |number| { seen.set(number); … })` directly and asserts
  `seen == 1030` — i.e. it asserts that `validate_pr_for_sha` passes its own `pr` argument
  to its own `probe`, which is the one line of that function (`deliver.rs:334`,
  `probe(pr)`). It is tautological with respect to the helper and says nothing about
  `deliver_inner`. Change `deliver.rs:357` to `validate_pr_for_sha(0, …)` or to a probe
  that closes over a constant and this test stays green.
  Consequences, both of them blocking:
  (1) it is a **false claim inside the diff** (§5 P1), and a conspicuous one in a PR whose
  entire thesis is that a test which never sees the real call can look green through a
  regression;
  (2) §5.5's fixed adjudication — *"`doneWhen` requires reaching a layer and no test proves
  it → P1"* — applies directly: #1079's doneWhen asks that *"a fixture proves `--pr <issue>
  --sha <hex>` errors with the PR-vs-issue distinction **instead of reading issue
  comments**"*. The delivered fixtures prove the **helper** errors with the distinction
  (`deliver.rs:668-685`); no fixture reaches the `--sha` branch, so neither the wiring nor
  the *ordering* half of that sentence ("instead of reading issue comments" — that
  `validate_pr_for_sha` precedes `comments()`) is proven by a test.
  The `dead_code`/`-D warnings` argument the module leans on (`deliver.rs:262-263`) covers
  *deletion* of the call site; it does not cover mis-wiring or reordering, which is exactly
  what `:694-695` promises and does not deliver.
  Cheap shape, if useful: give `deliver_inner` an injectable probe
  (`fn deliver_inner_with(args, cwd, probe: impl FnOnce(u64) -> Result<String>)`, with
  `deliver_inner` passing `|n| github::pr_head(cwd, n)`), then assert from `deliver_inner_with`
  that (a) the probe receives `args.pr` and (b) an `Err` probe short-circuits before
  `comments()` is reached. No network, no process spawn. Alternatively, if the wiring test
  is judged not worth the seam, the honest minimum is to delete the false comment and say
  plainly that the call site is guarded only by `dead_code` — but that leaves doneWhen 2's
  fixture clause unmet.

- **[P2]** The `--sha` path now makes two identical `gh pr view <n> --json headRefOid`
  round-trips per run: `validate_pr_for_sha` → `github::pr_head` (`deliver.rs:357`), and
  later `delivery::deliver`'s head-moved gate → `GhCli::head` → the same
  `github::pr_head` (`delivery.rs`, `match gh.head(pr)`). The ordering justifies a *check*
  before `comments()`, but the result is fetched twice and discarded once. Hoisting the head
  once and threading it would drop a network call; not required by doneWhen.

- **[P2]** New hard-stop on a transient `gh` failure, correctly directed but worth naming.
  `delivery::deliver` deliberately treats a head-read failure as conservative-but-survivable
  (`Err(_) => Write::Withheld(REASON)`), whereas `validate_pr_for_sha` turns the *same*
  failure into `?` → exit 2 with no JSON at all. The union answer itself depends only on the
  comment read, so a flaky `gh pr view` now blocks `merge-reviewed-pr.sh`'s union question
  (`jq -er '.status'` fails → `die "union gate gave no readable answer"`) where previously it
  would have been answered. That is the fail-closed direction and I am **not** treating it as
  a defect — but the top-line message ("`--pr N` did not resolve as a pull request … is N an
  issue instead?") asserts a diagnosis that a network/auth failure does not support. The real
  `gh` stderr does survive in `anyhow`'s chain via `{error:#}` at `deliver.rs:314`, so the
  operator is not misled beyond the first line; softening that line ("could not confirm #N is
  a pull request") would cost nothing.

- **[P2]** `validate_pr_for_sha` runs *before* the `is_full_sha` ensure (`deliver.rs:357`
  vs `:364-367`), so `--pr <n> --sha notahex` now spends a network round-trip before the
  free local shape check rejects it, and reports the PR-vs-issue framing first. Checking the
  cheap local invariant first is the usual ordering.

### FOLLOW-UP ISSUE

none. Two adjacent things were checked and are **not** findings, recorded so a later round
need not re-derive them:

- `scripts/merge-reviewed-pr.sh:68-71` justifies `--sha` on the grounds that it "skips
  deliver's own PR resolution, which fetches commits and would therefore require a checkout"
  and that "`GH_REPO` … does not narrow where the script may run". Both still hold: the new
  call is `gh pr view`, not `resolve_pr`'s `git fetch`. Probed from a non-repo directory with
  only `GH_REPO=fagemx/edda` set — `gh pr view 1086 --json headRefOid` → exit 0
  (`def6e0e6…`), and the pre-existing `gh api repos/{owner}/{repo}/issues/1086/comments` →
  exit 0. The merge entrypoint's legitimate `--pr N --sha <head>` call is **not** broken; it
  gains one round-trip.
- The failure mode the fixture at `deliver.rs:668-685` reproduces is faithful to the real
  one: `gh pr view 1079 --json headRefOid` (1079 is an issue) → exit 1,
  `GraphQL: Could not resolve to a PullRequest with the number of 1079. (repository.pullRequest)`
  — the string the test fakes.

Note on the spec itself, since I was asked to treat it as reviewable: #1079's doneWhen is
**not** the under-specified half here. Its item 2 names the fixture at the command layer
(`--pr <issue> --sha <hex>`) and adds the ordering clause ("instead of reading issue
comments"). The implementation satisfies the letter of the *code* half and the implementer
evidently knew the fixture half was owed — which is why `:692` exists and why its comment
claims the wiring. The spec asked for the right thing; the delivery stopped one seam short
of it.

### RAN vs READ

RAN (all read-only, in `C:\ai_agent\edda` on `main`, nothing checked out or modified):
- §3 router on `gh pr diff 1086 --name-only` → `classes=code-plain`, `canonical_class=code-risk`
- U1/U2/U3/U4 greps (outputs in the Rules table); `sh scripts/lint-markdown-content.sh` → exit 0
- `gh pr checks 1086`; `gh api repos/fagemx/edda/commits/def6e0e6…/check-runs` → `total_count=14`;
  `gh api …/actions/runs/34297458946` → `head_sha=def6e0e6…`, `conclusion=success`
- `edda --version` → `edda 0.5.0 (16f7b4b41cde-dirty 2026-09-09)`; base `Cargo.toml` → `version = "0.5.0"`
- C2/C3/C4/C5 and R1/R2/R3 enumerators; `sh scripts/wiring-scan.sh origin/main def6e0e6…`
- `git grep -n 'comments_argv|validate_pr_for_sha|deliver_inner' def6e0e6 -- crates/**` → one
  production call site each (`:274`, `:357`)
- `gh` probes: `gh pr view 1030 --json comments` → exit 1; `gh api …/issues/1030/comments` →
  exit 0, 8 comments; `gh pr view 1079 --json headRefOid` → exit 1 with the GraphQL message
- No-checkout probe with `GH_REPO` only (see FOLLOW-UP ISSUE) — three results, all as needed
- Window: `git merge-base origin/main def6e0e6` = `16f7b4b4` = `origin/main`;
  `git log 16f7b4b4..origin/main -- crates/edda-cli/src/cmd_review/deliver.rs` empty; head
  re-read → still `def6e0e6…`

READ (the L1 receipt — **CI run 34297458946 @ def6e0e651d70ce672cf22d721c387a5809e63c8**):
- `Format` pass (`cargo fmt --all --check`); `Clippy` pass on ubuntu, macOS **and** windows
  (`cargo clippy --workspace --all-targets`, `-Dwarnings`) — this is what makes the
  `dead_code` argument above load-bearing; `Test (ubuntu-latest)` and `Test (macos-latest)`
  pass = `cargo test --workspace`, all 23 crates; `Test (windows-latest)` pass = the 7-crate
  subset, whose matrix `args:` includes `-p edda` — so #1079's doneWhen 3 (`cargo test -p edda`
  green) is covered on all three OSes; `MSRV (1.91.0)` pass; `sdk-contract`,
  `Workspace publish dry-run`, `CI Gate` pass.
- What CI does **not** cover here: nothing for this crate. C5's selector is empty because
  `edda` is inside the Windows subset, so there is no coverage gap to close and no stated
  reason to spend a local Cargo run (`ladder` L2). `Fleet shell tests` / `Fleet shell tests
  (windows)` show `skipping` — correct under `ci.path-filter` for a diff with no fleet path.
- Zero Cargo commands were run by this reviewer.

### Verdict

Changes Requested, P0=0, P1=1 — the `--sha` wiring is correct in the code but unproven by any
test, and the test that claims to prove it (`deliver.rs:692-702`) only re-asserts that
`validate_pr_for_sha` hands its own argument to its own closure; that comment at `:694-695` is
a false claim, and #1079's doneWhen 2 asks for a fixture at the `--pr <issue> --sha <hex>`
layer. The `comments_argv` half (finding 1 of the issue) is delivered and verified on both
prongs; #917's withhold-and-exit-3 semantics and the ordinary no-`--sha` path are untouched.

fagemx:
## Review Response: Round 1 — PR #1086 @ 4359ab1437a0e95ab8e600ff9376ffe4b8381a6c

### [P1] Tautological wiring test (`deliver.rs:692-702`)

Fixed. `deliver_inner` (`deliver.rs:349-358`) is now a thin production
wrapper over a new `deliver_inner_with` (`deliver.rs:375-410`), which takes
the PR-vs-issue `probe` and the comment source (`comments_fn`) as injected
parameters instead of closing over them internally. The old test
`sha_path_probe_is_called_with_the_pr_number_argument` — which called
`validate_pr_for_sha` directly and only proved that function forwards its
own argument to its own closure — is replaced by
`deliver_inner_probes_the_pr_number_before_reading_comments_and_names_it_on_failure`
(`deliver.rs:736-786`), which drives `deliver_inner_with` itself with
`pr: 4242` and asserts:
- the surfaced error names `4242`,
- the fake probe was called with `4242` (not a hardcoded/mismatched number),
- the fake comment source was never reached.

Proved all three required reds locally (each simulated, run, reverted, `git
diff` confirmed clean before restoring):
1. **Validation call deleted** from the `Some(sha)` arm → `result.expect_err`
   panics because `deliver_inner_with` now returns `Ok`.
2. **Validation moved after the comment fetch** (moved the call to after
   `comments_fn(cwd, args.pr)?`) → the `!comments_reached.get()` assertion
   fails, since the fake comment source runs before validation can short-
   circuit.
3. **Wired to a hardcoded/wrong number** (`validate_pr_for_sha(0, probe)`
   instead of `validate_pr_for_sha(args.pr, probe)`) → `seen_probe.get() ==
   Some(0)`, failing the `assert_eq!(seen_probe.get(), Some(4242), …)`.

All three reverted cleanly and the suite is back to 38/38 green on the fix
commit.

### P2 1 — duplicate `pr_head` round-trip

**Declined**, with cost reasoning. `validate_pr_for_sha` (`deliver.rs:357`)
and `delivery::deliver`'s head-moved gate (`delivery.rs`, `gh.head(pr)` at
two call sites) both resolve to `github::pr_head`. Threading the validated
head through would mean changing `delivery::deliver`'s public signature
and/or the `Gh` trait it is built against (`delivery.rs:60-81`) to accept a
pre-known current head instead of always calling `gh.head(pr)` itself — that
trait and its ~20 existing fixtures in `delivery.rs` are outside this PR's
scoped surface (`deliver.rs` only, per Round 1's own IN SCOPE), and the
payoff is one avoided `gh pr view` round-trip on a path that has already
succeeded once this run. Not required by doneWhen; the reach outweighs the
saving.

### P2 2 — "not a PR" diagnosis on a possibly-transient `gh` failure

Fixed. `validate_pr_for_sha`'s context message (`deliver.rs:333-341`) now
reads:

> `--pr {pr} did not resolve as a pull request (either {pr} is an issue
> rather than a PR, or gh failed to confirm it — see the gh error this is
> chained to; --sha only skips the PR head fetch, not this PR-vs-issue
> check, so an issue number would otherwise read issue #{pr}'s comments)`

States both possibilities instead of asserting the PR-vs-issue diagnosis
outright; the real `gh` stderr still survives in the `anyhow` chain via
`{error:#}` at the `run()` print site, unchanged. Fail-closed behavior is
untouched — this is a message-only change, covered by the existing
`sha_path_rejects_a_number_gh_cannot_resolve_as_a_pull_request` test (still
green — it only asserts the message contains "pull request"/"PR" and the
offending number, both still true).

### P2 3 — validation runs before the free `is_full_sha` check

Fixed. In `deliver_inner_with`'s `Some(sha)` arm (`deliver.rs:381-398`), the
`is_full_sha` shape check now runs first, before `validate_pr_for_sha`'s
network probe — a malformed `--sha` is rejected without spending the
round-trip. The original post-match `ensure!` is kept as well (now
redundant on the `--sha` path, but it remains the only shape guard on the
`None` arm's `resolve_pr`-resolved head, which checks length/hex-digit-ness
but not lowercase the way `is_full_sha` does) — see the comment at
`deliver.rs:401-405` for why it stays rather than being removed.

### Gates ran (L0, warm lane `worker-2`)

```
export CARGO_TARGET_DIR="$LOCALAPPDATA/fleet-workstation/lanes/worker-2"
cargo fmt --all --check                              # clean
cargo clippy -p edda --all-targets -- -D warnings    # clean
cargo test -p edda cmd_review::deliver               # 38 passed; 0 failed
```

New full SHA: `4359ab1437a0e95ab8e600ff9376ffe4b8381a6c`

fagemx:
## Code Review: Round 2 — PR #1086 @ 4359ab1437a0e95ab8e600ff9376ffe4b8381a6c

- elapsed: unmeasured (controller-spawned subagent; no pi session file)
- model_requested: claude (Opus, via `review.dispatch-transport=controller-subagent-direct`)
- model_observed: unverified (no session file / `--output-format json` envelope to read)
- reviewer_session: n/a (replacement verifier continuing this thread; controller-spawned subagent, not a launcher-issued per-PR UUID)
- spec: review-spec-v1.5 (read at the base with `git show origin/main:REVIEW.md`)
- class: code-risk  (REVIEW.md classes: code-plain — up-classed for §5.4, same rationale as Round 1: this file's output is what `merge-reviewed-pr.sh` turns into a merge refusal)
- escalations: none
- shadow: false
- cost: delta round — `git diff def6e0e6..4359ab14` = 1 file / +101 −14 (whole PR +221 −20); ~14 tool calls; **zero Cargo runs** (C5 selector empty — `edda` is inside CI's Windows subset)

### IN SCOPE

Delta round per §2: the target is `git diff def6e0e651d70ce672cf22d721c387a5809e63c8..4359ab1437a0e95ab8e600ff9376ffe4b8381a6c` plus confirmation that Round 1's single blocking finding is resolved. The whole PR is not re-reviewed (`loop` items 2 and 6); Round 1's PASS rows on the unchanged surface stand.

Changed behaviour and paths (delta): `crates/edda-cli/src/cmd_review/deliver.rs` only — (a) `deliver_inner` split into a thin production wrapper over a new `deliver_inner_with(args, cwd, probe, comments_fn)` (`:349-361` wrapper, `:363-410` seam fn); (b) the old tautological test replaced by `deliver_inner_probes_the_pr_number_before_reading_comments_and_names_it_on_failure` (`:736-789`); (c) `validate_pr_for_sha`'s context message softened (`:333-341`); (d) `is_full_sha` hoisted ahead of the network probe inside the `Some(sha)` arm (`:386-389`), with the post-match `ensure!` retained and its reason documented (`:401-405`).

Direct callers/consumers: unchanged from Round 1 — `scripts/merge-reviewed-pr.sh:107` is the only production caller of the `--sha` path; `crates/edda-cli/src/cmd_review/mod.rs:74` dispatches. `delivery.rs` untouched.

Acceptance ceiling: issue #1079 doneWhen (3 items) plus its 2026-09-08 escalation comment.

Security / data-loss surface: unchanged in kind — the `--sha` path can publish an `Independent Review` status and a `review:*` label on a commit. The delta only moves a cheap local shape check earlier and changes wording; both directions are fail-closed or neutral.

Current-base integration: `git merge-base origin/main 4359ab14` = `16f7b4b41cde` = `origin/main` tip; `git log 16f7b4b4..origin/main -- crates/edda-cli/src/cmd_review/deliver.rs` empty. Window clean. Head re-read at posting time: still `4359ab14`.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 1086 --name-only` → `crates/edda-cli/src/cmd_review/deliver.rs` only; inside #1079's "Predicted surface" |
| U2 | any | P1 | PASS | `Closes #1079` at body line 3. All three doneWhen items now delivered (see C1) — the keyword is correct on this SHA, which it was not on `def6e0e6` |
| U3 | any | P1 | PASS | `Issue: #1079` present |
| U4 | any | P1 | PASS | offender grep empty; 2 commits, both `fix(edda-cli): …` |
| U6 | any | P0 | PASS | `gh pr checks 1086` → 12 pass, 2 `skipping` (Fleet shell tests, correct under `ci.path-filter`). `check-runs` `total_count=14` @ `4359ab14`; run `34300571470` `head_sha=4359ab14…`, `conclusion=success`. READ, not RAN |
| U7 | any | P1 | PASS | whole delta audited before posting; no blocker withheld |
| U8 | any | P1 | N.A.(no `Decision:` line in the PR body — DECISIONS block printed nothing) | |
| D1–D5 | docs | P1/[判斷] | N.A.(class is code-plain/code-risk; §4 does not route §5.1) | D1 enumerator on the delta printed **no** backticked command candidates. The one new *source* claim in the delta was read anyway: `deliver.rs:401-405` says `resolve_pr` "bails on a wrong-length or non-hex head, but not on stray uppercase — `is_full_sha` does". Confirmed: `github.rs:100-104` checks `len()!=40 \|\| !is_ascii_hexdigit()`; `deliver.rs:105-110` additionally rejects `is_ascii_uppercase()`. Claim holds |
| S1–S3 | skills | P0/P1 | N.A.(no skills/instruction files in the diff) | |
| C1 | code-plain | P0 | PASS | doneWhen 1 → `comments_argv` (`:264`) consumed at `:274`, tests `:679-696`/`:698-704`. **doneWhen 2 → now fully met**: `validate_pr_for_sha` (`:333`) wired at `:396`, ahead of `comments_fn` at `:410`, and a fixture at `:736-789` drives the `--pr <issue> --sha <hex>` layer itself and asserts the comment source was never reached. doneWhen 3 → CI green on all three OSes (`-p edda` is in the Windows matrix) |
| C2 | code-plain | P0 | PASS | `fix:` with regression tests. Whole-PR `grep -cE '^\+.*(#\[test\]\|fn test_)'` → 5. The delta's own count is 0 only because the `#[test]` attribute line is unchanged context around a *replaced* function body — net coverage rises, not falls. FAIL-first re-derived by inspection below (Findings → resolved blocker) |
| C3 | code-plain | P1 | PASS | grep empty on delta and whole PR (`.expect_err(` does not match the pattern, and every hit is inside `#[cfg(test)] mod tests` at `:505` regardless) |
| C4 | code-plain | P1 | PASS | grep empty — no `unsafe`, no `clippy::all`, no crate-level `allow`. Clippy `-Dwarnings` green on ubuntu/macOS/windows @ this SHA |
| C5 | code-plain | P1 | N.A.(selector empty) | `crates/edda-cli` → package `edda`, inside CI's Windows 7-crate subset. Nothing for the verifier to run; a local Cargo run here would be over-verification with no stated reason (`ladder` L2) |
| R1 | code-risk | P0 | PASS | grep empty on the delta |
| R2 | code-risk | P0/P1 | PASS | grep empty — no shell file, and no added Rust line mixes `\|\|` with `&&` |
| R3 | code-risk | P0 | N.A.(no `*.sh`/`*.bash` in the diff) | |
| R4 | code-risk | P1 | PASS | no temp dir, file or lock created by the delta; nothing to release |
| R5 | code-risk | P1 | PASS | `--sha` is now matched against `is_full_sha` (`^[0-9a-f]{40}$` semantics, `:105-110`) **before** any network call on the `--sha` path, which strengthens R5's own "a caller-supplied SHA must be matched before use". No shell interpolation added; `gh` still goes through `Command::args` |
| §5.5 Wiring | any | P1 | **PASS** | Round 1's FAIL is resolved — see the Wiring table |

### Wiring

`sh scripts/wiring-scan.sh def6e0e6 4359ab14` → `no new pub surfaces`; `no swallow patterns on added lines`. Both additions are module-private, so the substantive read is below.

|---|---|---|---|---|
| `deliver_inner_with(args, cwd, probe, comments_fn) -> Result<i32>` (`deliver.rs:375`) | private fn; the whole read/decide/write sequence with its two `gh` seams as `impl FnOnce` parameters | exactly two call sites (`git grep deliver_inner_with 4359ab14 -- crates/**`): production `deliver_inner` at `:355`, and the fixture at `:757`. `deliver_inner` is itself called only from `run()` at `:311` | unchanged: probe `Err` → `?` → `run()` prints `{error:#}` and exits 2; `comments_fn` `Err` likewise. Nothing new is swallowed | **No default drift**: `deliver_inner` passes the REAL probe `\|number\| super::github::pr_head(cwd, number)` and the REAL comment source `comments` (`:356-360`) — the same two functions the pre-split body called inline — the delta shows the very same `super::github::pr_head(cwd, number)` closure and the same `comments` call moving out of the inline body and into the wrapper, unchanged. Neither seam is `pub`, neither is reachable from a CLI flag, config key or trait object, so no test-only bypass exists in production |
| `deliver_inner_probes_the_pr_number_before_reading_comments_and_names_it_on_failure` (`deliver.rs:736`) | test fixture driving `deliver_inner_with` with `pr: 4242`, `sha: Some(SHA)` (`SHA` = 40 lowercase hex at `:508`, so the hoisted `is_full_sha` passes and the probe is genuinely reached) | asserts (i) the surfaced error names `4242` (`:773-778`), (ii) `seen_probe == Some(4242)` (`:779-783`), (iii) `!comments_reached` (`:785-788`) | n/a (test) | Reaches the real call site, which is exactly what Round 1 said the old test did not. **Hermetic**: validation fails first, so `comments_fn`, `delivery::GhCli` and `delivery::deliver` are never constructed — no process spawn, no network, and `cwd = Path::new(".")` is never touched |

### Findings

**Round 1's blocking P1 — resolved.** I judged the new test on Round 1's own stated criterion (does it go red under each of the three mis-wirings?) by reading the mechanics at `deliver.rs:375-410` and `:736-789`, rather than taking the response's word for it:

- **(a) validation call deleted** (`:396` removed) → `probe` goes unused, and the sequence runs on to `comments_fn` → `Ok(vec![])` → `extract`/`union` → `Ok(exit_code)`. `result.expect_err(...)` at `:773` panics. **Red.** (Second, independent net: an unused `probe` parameter is a `-D warnings` failure under CI's three-OS Clippy.)
- **(b) validation moved after the comment fetch** → `comments_fn` sets `comments_reached = true` before validation can short-circuit; the first two assertions still pass, and `assert!(!comments_reached.get(), …)` at `:785-788` fails. **Red.** This is the half of #1079's doneWhen 2 ("instead of reading issue comments") that nothing proved on `def6e0e6`.
- **(c) probes a number other than the one `--pr` carries** (e.g. `validate_pr_for_sha(0, probe)`) → the faked probe interpolates `{number}` into its error, so `message.contains("4242")` at `:776` fails *and* `assert_eq!(seen_probe.get(), Some(4242))` at `:779-783` fails. **Red** twice over.

All three are observable because the assertions are made about `deliver_inner_with`'s own execution, not about `validate_pr_for_sha` in isolation — the tautology Round 1 named is gone. The stale test name survives only inside the new test's explanatory comment (`:739`), where it is accurate.

**Behaviour-neutrality of the refactor on the ordinary path.** The `None` arm is byte-for-byte the pre-split `resolve_pr(cwd, args.pr)?.head`, followed by the same `ensure!(is_full_sha(&sha))` and the same comment read (production `comments_fn` *is* `comments`). Same calls, same order, same error types and messages. The only externally observable change anywhere is on the `--sha` path and is the requested P2 fix: a malformed `--sha` now exits 2 on the local shape check without spending a `gh` round-trip, and — when `--sha` is malformed *and* `--pr` is an issue — reports the shape error rather than the PR-vs-issue one. Both were exit 2 before and after.

**P2 dispositions (adjudicated).**

- *Reorder (`is_full_sha` before the network probe)* — **correct as landed** (`:386-389`). The retained post-match `ensure!` at `:406-409` is redundant on the `--sha` path but is the only shape guard on the `None` arm's resolved head, and the reason is documented at `:401-405` with a claim I verified against `github.rs:100-104` (see D1 row). Keeping it is the right call, not a leftover.
- *Softened error message* — **correct as landed** (`:333-341`). It now states both possibilities ("either {pr} is an issue rather than a PR, or gh failed to confirm it — see the gh error this is chained to") instead of asserting a diagnosis a network/auth failure does not support, and the real `gh` stderr still survives via `{error:#}`. The neighbouring test at `:708-730` asserts only `contains("pull request") \|\| contains("PR")` and `contains("1030")` — both still true of the new wording, so it stays green for the right reason rather than by accident.
- *Duplicate `pr_head` round-trip — **DECLINED by the implementer; the decline holds.*** Adjudicating under §5.5 and `loop` item 1: this is a P2, it is not named by any doneWhen item, and it is not a wiring failure (no dead surface, no swallowed error, no unproven layer). Removing the second call means threading a pre-known head into `delivery::deliver` and therefore into the `Gh` trait at `delivery.rs:60-81` and its ~20 existing fixtures — a file Round 1 froze **outside** `IN SCOPE`. Trading a scoped-surface expansion of that size for one avoided `gh pr view` on an already-successful run is exactly the cost the `Verification cost` section tells us not to spend. `loop` item 1 routes wants beyond the ceiling to follow-up, not to the PR; and this one is small enough that I am not opening an issue for it either — recorded here so a later round need not re-derive it.

**New adversarial pass on the delta — zero blocking findings.** Explicitly checked and clean: no test-only bypass reachable in production (both seams are hard-coded in the sole production wrapper; nothing `pub`, no flag, no env var, no `cfg(test)` divergence in production code); `deliver_inner_with` does not leak into public API (`wiring-scan` → `no new pub surfaces`; the fn is bare `fn`, and `deliver_inner`/`run` are unchanged in visibility); no new `unwrap`/`expect`/`unsafe`/blanket-allow; no error newly swallowed (`let _ =`/`.ok()`/`unwrap_or_default()` scan clean on added lines); the #917 withhold-and-exit-3 contract, `extract`, `gate::union` and the JSON shape are untouched by the delta; clippy-cleanliness under `-D warnings` is CI-witnessed on all three OSes at this exact SHA, which is what keeps the `dead_code` net above load-bearing.

No P2 is raised this round.

### FOLLOW-UP ISSUE

none. Round 1's two recorded non-findings (the `merge-reviewed-pr.sh --sha` justification, and the fidelity of the faked `gh` failure string) are unaffected by this delta and still hold; the duplicate-`pr_head` P2 is adjudicated above rather than routed, for the cost reason stated there.

### RAN vs READ

RAN (all read-only, in `C:\ai_agent\edda` on `main`; nothing checked out, edited or pushed):
- `git fetch origin`; §3 router on `gh pr diff 1086 --name-only` → `classes=code-plain`, `canonical_class=code-risk`
- Delta: `git diff def6e0e6..4359ab14` (+101 −14) and `git diff --shortstat origin/main..4359ab14` (+221 −20)
- U2/U3/U4 greps; `sh scripts/lint-markdown-content.sh` → `exit=0`
- `gh pr checks 1086`; `gh api …/commits/4359ab14…/check-runs` → `total_count=14`; `gh api …/actions/runs/34300571470` → `head_sha=4359ab1437a0e95ab8e600ff9376ffe4b8381a6c`, `conclusion=success`
- `edda --version` → `edda 0.5.0 (16f7b4b41cde-dirty 2026-09-09)`; DECISIONS block → empty
- C2/C3/C4/C5, R1/R2/R3, D1 and D4 enumerators on the delta (and C2/C3 on the whole PR)
- `sh scripts/wiring-scan.sh def6e0e6 4359ab14` → no new pub surfaces, no swallow patterns
- `git grep -n 'deliver_inner_with' 4359ab14 -- crates/**` → two call sites (`:355` production, `:757` test); `git grep sha_path_probe_is_called_with_the_pr_number_argument 4359ab14` → one hit, inside the new test's comment
- Source reads at the reviewed SHA via `git show 4359ab14:<path>`: `deliver.rs:250-410`, `:660-790`, `:30-46`, `:100-110`; `github.rs:1-140`
- Window: `git merge-base origin/main 4359ab14` = `16f7b4b41cde` = `origin/main`; `git log 16f7b4b4..origin/main -- crates/edda-cli/src/cmd_review/deliver.rs` empty; head re-read → still `4359ab14…`

READ (the L1 receipt — **CI run 34300571470 @ 4359ab1437a0e95ab8e600ff9376ffe4b8381a6c**):
- `Format` pass — `cargo fmt --all --check` **and** `bash scripts/lint-file-length.sh --tree` (`.github/workflows/ci.yml:117-119`), so the GH-779 ratchet is green on a file that grew by ~90 lines; `Clippy` pass on ubuntu, macOS **and** windows (`--workspace --all-targets`, `-Dwarnings`); `Test (ubuntu-latest)` and `Test (macos-latest)` pass = `cargo test --workspace`, all 23 crates; `Test (windows-latest)` pass = the 7-crate subset whose matrix `args:` includes `-p edda`, so #1079's doneWhen 3 is covered on all three OSes; `MSRV (1.91.0)` pass; `Detect code changes`, `sdk-contract`, `Workspace publish dry-run`, `CI Gate` pass. `Fleet shell tests` / `(windows)` `skipping` — correct under `ci.path-filter` for a diff with no fleet path.
- What CI does **not** cover here: nothing for this crate. The C5 selector is empty because `edda` is inside the Windows subset, so there is no coverage gap and no stated reason to spend a local Cargo run (`ladder` L2).
- Round 1 verdict (#issuecomment-5594378548) and `Review Response: Round 1` (#issuecomment-5594526531) read as receipts; the response's FAIL-first claims were re-derived from the code above rather than accepted on assertion.
- Zero Cargo commands were run by this reviewer.

### Verdict

LGTM (P0=0, P1=0) — Round 1's blocking P1 is genuinely resolved: the fixture now drives `deliver_inner_with` itself and goes red under all three of Round 1's own counterfactuals (call deleted, call moved after the comment read, call wired to a different number), which is what #1079's doneWhen 2 asked for and what the replaced test never proved. The production wrapper passes the real `github::pr_head` probe and the real `comments` source — no default drifted — and the refactor is behaviour-neutral on the ordinary no-`--sha` path. Two P2s are fixed, the third is declined on cost grounds I agree with. Merge remains the operator's action.

PR #1097, opened 2026-09-09
fix(review): REVIEW.md's diff surface enumeration — two-dot vs three-dot, and R3's file-order short-circuit (GH-1003, GH-992)

fagemx:
## Summary

Two defects in `REVIEW.md`'s own surface enumeration, on disjoint sections, bundled per the controller's brief (one fixture file covers both):

- **GH-1003**: `REVIEW.md`'s content-enumerating `git diff "origin/$BASE..$SHA"` calls used two-dot (tip-to-tip) instead of three-dot (from the merge base). On a branch behind its base, two-dot folds in every commit the base gained since the branch point, inflating the reviewed surface (measured on PR #988: 39 files reported for 3 actually changed). The three `git log "origin/$BASE..$SHA"` commit enumerators (U2, U4, C2) are correctly left two-dot — `git log` two-dot means something different (the PR's own commits) and three-dot there would pull in the base's commits too, breaking the commit-message rules.
- **GH-992**: `REVIEW.md`'s R3 block (every changed shell script parses, P0) short-circuited on `[ -f "$f" ] && { sh -n "$f"; ...; }` inside a `for` loop — when the *alphabetically last* changed `.sh` was simply absent from the reviewer's working tree (e.g. checked out at `$BASE`, not `$SHA`), the loop's own exit status flipped to 1 regardless of every other file parsing clean, reporting a P0 FAIL on a file R3 never actually checked, with the evidence column showing an unrelated clean file's `exit=0` line.

## Provenance — this PR is not one author's work

The first fix lane on these issues died with its edits on disk and nothing committed. The controller preserved that state verbatim as commit `3e22ee8c8a2d35b0aa421d1298d8ea25644ae629` ("wip(review): GH-1003 + GH-992 in progress — uncommitted lane work, preserved") and pushed it. That commit was explicitly unverified, ungated, and unreviewed — no gate had been run against it, and the lane that wrote it could not be asked to explain it.

[...]

PR #1095, opened 2026-09-09
docs(fleet): rules.md has an R18-R20 hole — R21 and R22 cite rules that do not exist, and R6 still has no Independent Review condition (PR #760 re-filed)

fagemx:
## Problem and change

`docs/fleet/rules.md` on `origin/main` numbers R1–R17 then R21–R29: R18, R19, R20
are a hole. Two existing rules cite into the hole — R21's source line ends
"…R19 的同一原則用在實作側" (R19 does not exist) and R22's body reads
"權威輪次寫 status（帳本 `review.verdict-carrier`；#760 的 R18（未合併））"
(R18 does not exist; the carrier is a closed, unmerged PR). A third gap is a
missing condition, not a dangling reference: R6 (merge) never gained an
`Independent Review` status condition, so the text is looser than the
ruleset it describes — #962 is a live case of a PR merging on a hand-written
status with no such condition written down anywhere.

Content for all three missing rules and the R6 gap was drafted once already
in closed PR #760 (branch `docs/fleet-rules-status-claim`, head `50d45bb5`),
but that diff targets the 2026-09-02 rules.md and no longer applies now that
R21–R29 exist on top of it. This PR re-derives the same content against the
current file: R18/R19/R20 inserted between R17 and R21 (no renumbering —
every existing R21–R29 cross-reference in `REVIEW.md`, `.claude/CLAUDE.md`
and dozens of issues stays valid), R6 gets the missing condition, and R22's
dangling parenthetical now points at this file's own R18 instead of the
closed PR.

**Controller ruling carried over from the issue (quoted, not chosen by this
lane):** "控制者裁定：不等 #959，也不引 `review.verdict-carrier` —— 該 key
在 2026-09-06 仍回 `No results found.`，#959 未落地，這一張不得因它而卡住。"
R6 and R18's source lines cite `REVIEW.md` §8 "Step 8 — the verdict" (the
`Independent Review` status semantics live there, not §3, which is
classify), this file's own rule numbers, and issue numbers — never
`review.verdict-carrier`. See the FOLLOW-UP ISSUE note below: on this
lane's machine that key does in fact resolve via `edda ask`, which
contradicts the issue's stated *reason* but not the ruling itself — R25
already legislates that cross-machine ledger disagreement is settled by
this file on `origin/main`, not by any one machine's local answer, which is
exactly why the ruling holds regardless of what my `edda ask` returns here.

R7 was read and deliberately left untouched — main's R7 already carries the
[...]

Issue #1003, opened 2026-09-06
fix(review): REVIEW.md enumerates the reviewed surface with two-dot git diff — 39 files instead of 3 on a branch behind its base (but the git log ranges are correct and must stay)


fagemx:
## What happened

`REVIEW.md`'s own rule blocks enumerate the reviewed surface with **two-dot** `git diff "origin/$BASE..$SHA"`. Two-dot diffs the two tips; three-dot diffs from the merge base. Whenever the branch is behind its base — the common case for a PR opened an hour ago — two-dot folds every commit `main` gained since into the reviewed delta.

Measured on PR #988 (`dd10481c`, base `origin/main` at `85202b81`, merge base `586cb07e`):

```text
git diff origin/main..dd10481c --name-only   ->  39 files
git diff origin/main...dd10481c --name-only  ->   3 files
```

The PR touches **3** files. A reviewer following the spec literally audits 36 files it never changed. On a differently shaped branch the same defect hides files instead of adding them, which is the worse direction.

Found by the independent reviewer on PR #988, which used three-dot ranges and noticed the spec disagreed with it.

## Scope — the `git log` lines are correct and must NOT be changed

This is the part worth getting right before anyone edits the file. Two-dot and three-dot mean **different things** to `git log` than to `git diff`:

| | two-dot | three-dot | correct for the rule |
|---|---|---|---|
| `git diff` | tip-to-tip | from the merge base | **three-dot** |
| `git log` | commits in `$SHA` not in base | symmetric difference | **two-dot** |

Same PR, measured:

```text
git log --format=%s origin/main..dd10481c   ->  1 commit   (the PR's own — correct)
git log --format=%s origin/main...dd10481c  ->  9 commits  (adds main's — wrong)
```

So a blanket "replace `..` with `...`" would fix the file enumeration and simultaneously break U4 (commit-message conventions) and every other rule that walks the PR's commits.

## The affected lines

**Change to three-dot** — these enumerate changed content:

- `REVIEW.md:427`, `:470`, `:488`, `:506`, `:532` — `git diff "origin/$BASE..$SHA"` in the U/D/S/C rule blocks

**Leave as two-dot** — these enumerate the PR's commits:

- `REVIEW.md:299`, `:327`, `:531` — `git log --format=%B|%s "origin/$BASE..$SHA"`

`scripts/wiring-scan.sh` takes `$BASE_SHA` and `$SHA` as separate arguments; check which form it builds internally and route it the same way.

[...]

PR #1098, opened 2026-09-09
fix(edda-cli): ratify cited-authority stops inferring supersession from a mention (GH-1066)

### LINKED ISSUE #1066: fix(edda-cli): ratify --by-rule cited-authority infers supersession from a name mention and has no domain/date guard — the sweep holds the operator's plain-flow rulings and would swear in d-033.merge_authority (2026-08-14)
## What was observed

`edda ratify --by-rule cited-authority --dry-run` on 2026-09-07 (session `0c87c653`, main): **176 rows, 97 `ratify`, 79 `hold`.** Two defects in the rule engine make the sweep unsafe to run and, worse, invert its purpose:

**1. "A later decision names this key" is treated as supersession.** The hold reason reads `superseded — a later decision '<key>' names it`. But a later decision that *builds on* an earlier one cites it in its reason without replacing it. Result on today's table:

| held key | held because | reality |
|---|---|---|
| `review.default-path` | `review.shell-branch` names it | `review.shell-branch` depends on it |
| `review.watcher` | `review.shell-branch` names it | same |
| `review.merge-gate` | `review.default-path` names it | `review.default-path` calls it its 「成對條目」 |
| `review.readonly-proof` | `fleet.ledger-sync-trigger` names it | unrelated citation |
| `fleet.lane-launch` | `ledger.cross-machine-projection` names it | unrelated citation |

So the sweep systematically **holds the most-cited rulings** (the load-bearing ones every later decision refers to) and **ratifies the leaves**. The operator's 2026-09-07 plain-flow rulings, which quote him verbatim, were all held; the controller's downstream `review.shell-branch` would have been ratified.

**2. No domain or date guard against superseded-in-substance entries.** `d-033.merge_authority = required` (2026-08-14: 「Mandatory issues remain open only because the formation has no explicit merge authority… Resume only on an explicit operator merge delegation」) and `d-034.merge_authority`, `d-032.fresh_premerge_gate`, `d-013.final_gate` are all `ratify` — they cite an issue, nothing later *names* them, and the engine cannot see that `review.auto-merge` (ratified 2026-09-03) contradicts them in substance. A blind sweep would swear the superseded constitution back in, on the same day the carriers are being cut over (#1032, #1063, #1064).

`decision.auto-ratify` (binding) says a decision auto-ratifies when it cites operator/issue/decision authority **and contradicts no binding decision**. The engine implements the first clause and approximates the second by name-mention, which is wrong in both directions.

## Predicted surface

- `crates/edda-cli/src/cmd_ratify.rs` (and whatever module holds the `cited-authority` sweep; a peer has claimed cmd_ratify per #1032 — coordinate via `edda claim` first)
- `crates/edda-core` or `crates/edda-ledger` if the supersession relation lives there
- tests: fixtures reproducing the five held keys above and the four old-constitution rows

## doneWhen

- Supersession is explicit, never inferred from a mention: a decision is superseded only by a later decision on the **same key**, or one whose reason/notes carry an explicit `supersedes:<key>` marker (or the existing 「顯式取代」 / `SUPERSEDES` phrasing the ledger already uses — the lane inventories the phrasings on main and picks one canonical form, migrating none).
- A `hold` reason distinguishes `superseded-by-same-key`, `superseded-explicit`, `no-citation`, and the new `older-than-binding-in-domain`.
- Domain guard: an unratified decision whose key shares a domain prefix (`review.`, `fleet.`, …) with a **binding** decision recorded later than it is held with reason `older-than-binding-in-domain` and listed in the dry-run for the operator; it is never auto-ratified.
- `--by-rule` accepts `--key <KEY>` (repeatable) and `--since <date>` so a sweep can be bounded; the unbounded form prints the counts and refuses without `--yes` when the ratify set exceeds a threshold (default 20).
- Dry-run fixture test: on the 2026-09-07 table snapshot (checked in as a fixture, 176 rows) the five held keys become `ratify` (they cite the operator by date) and the four old-constitution rows become `hold: older-than-binding-in-domain`.
- Unit tests in `#[cfg(test)]`, fixture-based, no ledger network; L0 on `edda` (package name) and any touched core crate; L1 exact-head CI.

## Sequencing

Product-class, on the constitution line: blocks a safe sweep, which is half of landing the plain flow (the other half is #1063/#1064). Until it lands, ratify per key with `--by 'auto-ratify rule (operator 2026-09-03)'` as done today for `review.merge-gate`, `review.default-path`, `review.watcher`.

## Predicted flash routing

`strong` — the supersession semantics are the judgment.

fagemx:
## Summary

`edda ratify --by-rule cited-authority` used to infer supersession from a
name mention: if a later decision's reason merely *named* an earlier key
while building on it, the earlier (correctly-cited) key was held. On the
2026-09-07 sweep (176 rows) this held the operator's five most-cited,
plain-flow rulings and would have ratified four rows of a superseded
merge-authority regime that a later, binding decision had overtaken in
substance but never named.

- Supersession is now explicit, never inferred from a mention: only a
  later decision under the **same key**, or one carrying an **explicit**
  `supersedes:<key>` marker (plus the pre-existing `SUPERSEDES` / `顯式取代`
  phrasings already on the ledger — recognised, not migrated), counts.
- Four distinguishable hold reasons: `superseded-by-same-key`,
  `superseded-explicit`, `no-citation`, `older-than-binding-in-domain`
  (the new domain guard: an unratified decision whose key shares a domain
  prefix with a later **binding** decision is held and shown, never
  auto-ratified).
- `--by-rule` now accepts `--key <KEY>` (repeatable) and `--since <date>`
  to bound a sweep; the unbounded form refuses past a 20-ratify threshold
  without `--yes`.

## This PR carries two authors' work — here is the split

The first fix lane on this issue died with its edits uncommitted. The
controller preserved that state as commit `de79b22` (683 insertions / 36
deletions across `cmd_ratify/{mod,rule,tests}.rs` and
`tests/ratify_exit_codes.rs`) and pushed it — **not verified, not gated,
not reviewed; no test had ever been run against it.** I picked up from
there as a replacement lane, per the addendum's instruction to audit
before adding anything.

**`cargo test -p edda` on the preserved commit: 744 passed, 2 failed.**
Both failures were real bugs in the fixture, not flakiness:

1. `gh1066_same_key_stopgap_holds_the_original_d033_row` assumed
   `active_decisions()` can return two active rows for one `(branch,
   key)`. It cannot — the ledger deactivates a key's prior row in the
   same write that appends a same-key successor (`UPDATE decisions SET
   is_active = FALSE ... WHERE key = ? AND branch = ?`,
   `sqlite_store/events.rs:136-141`), which predates GH-1066 entirely.
   The fixture's fabricated "same-key stopgap" scenario for
   `d-033.merge_authority` could never produce the two rows the test
   expected. I removed the fabricated block (its target coverage — one
   ratify event per re-decided key — was already proven by the
   pre-existing `a_re_decided_key_is_ratified_once_not_once_per_row`)
   and corrected the stale `collect()` comment that stated the false
   premise.
2. `gh1066_since_bound_keeps_the_rail_closeout_batch_out_of_the_sweep`
   asserted `--since 2026-08-20` keeps the 2026-08 rail-closeout batch
   out of the sweep, but the fixture's `decide()` helper always stamps
   `ts` as real wall-clock "now" — every row in the fixture carried
   *today's* date regardless of its narrative comment, so `--since`
   filtered nothing. Added a `decide_at` test helper that stamps a
   caller-given RFC3339 timestamp (re-finalizing the event, since the
   hash covers `ts`) and used it for the four 2026-08 rows.

Fixed and pushed as a second commit on this branch; `rule.rs` and
`evidence.rs` were untouched by either lane.

## One doneWhen nuance for review

The issue's acceptance bullet says the four old-constitution rows
(`d-013.final_gate`, `d-032.fresh_premerge_gate`, `d-033.merge_authority`,
`d-034.merge_authority`) become `hold: older-than-binding-in-domain`.
In this fixture they do not — they are kept out of an *unscoped* sweep by
the `--since` date guard instead, and ratify (via the existing
issue-number citation fallback, out of this issue's scope) when `--since`
is omitted. Reason: `domain` is `text before the first '.'` throughout
the codebase (`edda_core::decision::extract_domain`, mirrored here by
`rule.rs::domain_of`), so `d-013`/`d-032`/`d-033`/`d-034` are each their
own domain — none of them is `review`. Connecting them to
`review.auto-merge` would mean inferring a relationship the ledger never
recorded, which is the exact kind of inference GH-1066 removes. The
issue's own bullet title is "No domain **or** date guard" — this batch is
the case the date guard covers; the domain guard is proven separately
(`domain_guard_holds_older_candidate_behind_a_later_binding_sibling` and
`four_case_ledger`'s case 3) against keys that do share a real domain
prefix with a later binding sibling.

Flagging this rather than silently reinterpreting the doneWhen text or
fabricating a domain match in the fixture — this is the "supersession
semantics are the judgment" call the issue itself predicted.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy -p edda --all-targets -- -D warnings` — clean
- [x] `cargo test -p edda` — 745 unit tests + all `edda-cli` integration
      binaries (including `tests/ratify_exit_codes.rs`, which never ran
      against the preserved commit because `cargo test` stops after the
      first failing binary) — 0 failed
- [x] `scripts/lint-file-length.sh --tree` — clean
- [x] `supersede_only_looks_forward` and the other pre-existing
      `rule.rs` tests pass unmodified

Issue: #1066
Closes #1066

### DIFF
--- a/crates/edda-cli/src/cmd_ratify/mod.rs
+++ b/crates/edda-cli/src/cmd_ratify/mod.rs
@@ -178,19 +178,54 @@ fn append_ratify(
 /// deliberately, so the threshold does not apply once either is given.
 const RATIFY_THRESHOLD: usize = 20;
 
-/// `--since <date>` must be `YYYY-MM-DD`: cheap to check at the boundary, so
-/// a typo is a usage error (exit 2) rather than a silently-wrong comparison.
-fn validate_since(s: &str) -> String {
+/// True when `s` is a real calendar date in `YYYY-MM-DD` form — not merely
+/// shaped like one (PR #1098 Round 1 P2): the original shape-only check
+/// accepted `2026-99-99`. The `--since` comparison itself stays
+/// lexicographic and was never unsafe on a malformed value, but the flag
+/// advertises a date it should actually parse.
+fn is_valid_since_date(s: &str) -> bool {
     let bytes = s.as_bytes();
-    let ok = bytes.len() == 10
-        && bytes[4] == b'-'
-        && bytes[7] == b'-'
-        && bytes.iter().enumerate().all(|(i, b)| match i {
-            4 | 7 => true,
-            _ => b.is_ascii_digit(),
-        });
-    if !ok {
-        usage_exit(&format!("--since must be YYYY-MM-DD — got '{s}'"));
+    if bytes.len() != 10 || bytes[4] != b'-' || bytes[7] != b'-' {
+        return false;
+    }
+    if !bytes
+        .iter()
+        .enumerate()
+        .all(|(i, b)| matches!(i, 4 | 7) || b.is_ascii_digit())
+    {
+        return false;
+    }
+    let Ok(year) = s[0..4].parse::<i32>() else {
+        return false;
+    };
+    let Ok(month) = s[5..7].parse::<u32>() else {
+        return false;
+    };
+    let Ok(day) = s[8..10].parse::<u32>() else {
+        return false;
+    };
+    if !(1..=12).contains(&month) {
+        return false;
+    }
+    let leap = (year % 4 == 0 && year % 100 != 0) || year % 400 == 0;
+    let max_day = match month {
+        1 | 3 | 5 | 7 | 8 | 10 | 12 => 31,
+        4 | 6 | 9 | 11 => 30,
+        2 if leap => 29,
+        2 => 28,
+        _ => unreachable!("month already bounded to 1..=12"),
+    };
+    (1..=max_day).contains(&day)
+}
+
+/// `--since <date>` must be a real `YYYY-MM-DD` calendar date: cheap to
+/// check at the boundary, so a typo is a usage error (exit 2) rather than a
+/// silently-wrong comparison.
+fn validate_since(s: &str) -> String {
+    if !is_valid_since_date(s) {
+        usage_exit(&format!(
+            "--since must be a real YYYY-MM-DD date — got '{s}'"
+        ));
     }
     s.to_string()
 }
--- a/crates/edda-cli/src/cmd_ratify/mod.rs
+++ b/crates/edda-cli/src/cmd_ratify/mod.rs
@@ -319,16 +319,17 @@ fn collect(
         });
     }
     candidates.sort_by_key(|c| c.order);
-    // Every active row reaches the rule, same-key predecessors included:
-    // `active_decisions` can return more than one active row for a `(branch,
-    // key)` — `Ledger::ratified_decisions_map` groups them for exactly that
-    // reason. Before GH-1066 this function collapsed to "keep only the
-    // latest row per key" so an older row was invisible to the rule and its
-    // supersession was silently assumed rather than shown; now the rule's
-    // own same-key check (rule.rs) holds every predecessor explicitly and
-    // names why, and only the newest row in a same-key chain can ever clear
-    // that check and reach the citation check — so the write loop below
-    // still appends at most one ratify event per key.
+    // `active_decisions` never returns two rows for one `(branch, key)`: the
+    // ledger deactivates a key's prior row in the same write that appends a
+    // same-key successor (`UPDATE decisions SET is_active = FALSE ... WHERE
+    // key = ? AND branch = ?`, sqlite_store/events.rs) — a redecided key's
+    // old value is gone from this candidate set before the rule ever runs,
+    // not merely held by it. `evaluate`'s own same-key check
+    // (rule.rs::same_key_superseded) stays correct regardless and is
+    // exercised directly there with hand-built same-key candidates, so it
+    // still guards the pure function itself if a future caller ever feeds it
+    // a candidate set that is not pre-deduplicated this way. Either way, the
+    // write loop below appends at most one ratify event per key.
     Ok((candidates, binding))
 }
 
--- a/crates/edda-cli/src/cmd_ratify/mod.rs
+++ b/crates/edda-cli/src/cmd_ratify/mod.rs
@@ -53,6 +53,21 @@ pub struct RatifyArgs {
     /// With --by-rule: print the table without writing anything
     #[arg(long = "dry-run", requires = "by_rule")]
     pub dry_run: bool,
+    /// With --by-rule: only judge these keys (repeatable). Bounds the sweep
+    /// (GH-1066) — with this set, the unbounded-sweep size refusal below
+    /// does not apply.
+    #[arg(long = "key", requires = "by_rule")]
+    pub keys: Vec<String>,
+    /// With --by-rule: only judge decisions recorded on or after this date
+    /// (`YYYY-MM-DD`). Bounds the sweep the same way `--key` does (GH-1066).
+    #[arg(long, requires = "by_rule")]
+    pub since: Option<String>,
+    /// With --by-rule and neither --key nor --since: proceed even though the
+    /// ratify set exceeds the safety threshold (GH-1066). Without it, an
+    /// unbounded sweep that would ratify more than the threshold prints the
+    /// table and counts, and writes nothing.
+    #[arg(long, requires = "by_rule")]
+    pub yes: bool,
     /// Session ID (uses EDDA_SESSION_ID; --session required when identity is ambiguous)
     #[arg(long)]
     pub session: Option<String>,
@@ -156,23 +171,77 @@ fn append_ratify(
     Ok(())
 }
 
+/// Default cap on an unbounded `--by-rule` sweep's ratify count before it
+/// refuses without `--yes` (GH-1066): a sweep nobody scoped is a sweep
+/// nobody dares run, so the default protects against acting on more than a
+/// screenful of decisions sight-unseen. `--key`/`--since` bound the sweep
+/// deliberately, so the threshold does not apply once either is given.
+const RATIFY_THRESHOLD: usize = 20;
+
+/// `--since <date>` must be `YYYY-MM-DD`: cheap to check at the boundary, so
+/// a typo is a usage error (exit 2) rather than a silently-wrong comparison.
+fn validate_since(s: &str) -> String {
+    let bytes = s.as_bytes();
+    let ok = bytes.len() == 10
+        && bytes[4] == b'-'
+        && bytes[7] == b'-'
+        && bytes.iter().enumerate().all(|(i, b)| match i {
+            4 | 7 => true,
+            _ => b.is_ascii_digit(),
+        });
+    if !ok {
+        usage_exit(&format!("--since must be YYYY-MM-DD — got '{s}'"));
+    }
+    s.to_string()
+}
+
 /// `--by-rule <RULE>`: evaluate every active decision on this branch and
 /// print one row per unratified key. `--dry-run` prints the same table and
 /// writes nothing, so the operator can read what the machine is about to do
 /// on their behalf before it does it.
+///
+/// `--key`/`--since` bound *which* verdicts are shown and acted on, not what
+/// the rule sees while deciding: same-key and domain-guard comparisons
+/// (rule.rs) still see every active decision on the branch, so scoping a
+/// sweep can never blind it to an older, still-relevant ruling.
 fn sweep(repo_root: &Path, rule_name: &str, args: &RatifyArgs) -> anyhow::Result<()> {
     if rule_name != RULE_CITED_AUTHORITY {
         usage_exit(&format!(
             "unknown rule '{rule_name}' — known rules: {RULE_CITED_AUTHORITY}"
         ));
     }
+    let since = args.since.as_deref().map(validate_since);
 
     let ledger = edda_ledger::Ledger::open(repo_root).context("cmd_ratify: opening ledger")?;
     let _lock = edda_ledger::lock::WorkspaceLock::acquire(&ledger.paths)?;
     let branch = ledger.head_branch()?;
 
     let (candidates, binding) = collect(&ledger, &branch)?;
-    let verdicts = evaluate(&candidates, &binding);
+    let dates: HashMap<&str, &str> = candidates
+        .iter()
+        .map(|c| (c.key.as_str(), c.date.as_str()))
+        .collect();
+    let all_verdicts = evaluate(&candidates, &binding);
+
+    let bounded = !args.keys.is_empty() || since.is_some();
+    let verdicts: Vec<Verdict> = all_verdicts
+        .into_iter()
+        .filter(|v| args.keys.is_empty() || args.keys.iter().any(|k| k == &v.key))
+        .filter(|v| {
+            since
+                .as_deref()
+                .is_none_or(|s| dates.get(v.key.as_str()).is_some_and(|d| *d >= s))
+        })
+        .collect();
+
+    let ratify_count = verdicts.iter().filter(|v| v.is_ratify()).count();
+    if !args.dry_run && !bounded && !args.yes && ratify_count > RATIFY_THRESHOLD {
+        print_table(&verdicts, rule_name, true);
+        println!(
+            "refusing: {ratify_count} decisions would ratify unscoped (threshold {RATIFY_THRESHOLD}) — bound with --key/--since, or pass --yes; nothing written."
+        );
+        return Ok(());
+    }
 
     print_table(&verdicts, rule_name, args.dry_run);
 
@@ -244,19 +313,22 @@ fn collect(
             cites: cites_by_event.remove(&d.event_id).unwrap_or_default(),
             ratified: ratified.contains_key(&d.event_id),
             order: ledger.rowid_for_event_id(&d.event_id)?.unwrap_or(0),
+            date: d.ts.clone().unwrap_or_default(),
             key: d.key,
             reason: d.reason,
         });
     }
     candidates.sort_by_key(|c| c.order);
-    // One row per key: `active_decisions` can return more than one active row
-    // for a `(branch, key)` — `Ledger::ratified_decisions_map` groups them for
-    // exactly that reason — and two rows would otherwise write two ratify
-    // events for one key. The later row is the current value, so keep it.
-    let mut seen = BTreeSet::new();
-    candidates.reverse();
-    candidates.retain(|c| seen.insert(c.key.clone()));
-    candidates.reverse();
+    // Every active row reaches the rule, same-key predecessors included:
+    // `active_decisions` can return more than one active row for a `(branch,
+    // key)` — `Ledger::ratified_decisions_map` groups them for exactly that
+    // reason. Before GH-1066 this function collapsed to "keep only the
+    // latest row per key" so an older row was invisible to the rule and its
+    // supersession was silently assumed rather than shown; now the rule's
+    // own same-key check (rule.rs) holds every predecessor explicitly and
+    // names why, and only the newest row in a same-key chain can ever clear
+    // that check and reach the citation check — so the write loop below
+    // still appends at most one ratify event per key.
     Ok((candidates, binding))
 }
 
--- a/crates/edda-cli/src/cmd_ratify/rule.rs
+++ b/crates/edda-cli/src/cmd_ratify/rule.rs
@@ -38,6 +38,11 @@ pub struct Candidate {
     pub ratified: bool,
     /// Ledger insertion order. Only ever compared, never displayed.
     pub order: i64,
+    /// Ledger timestamp (RFC3339), for `--since` bounding (GH-1066). Empty
+    /// when unknown. `evaluate` never reads this field — `--since` filters
+    /// the caller's output (`mod.rs::sweep`), not a verdict here, so an
+    /// empty date never changes what this pure function decides.
+    pub date: String,
 }
 
 /// What the rule decided, and why. The why is not decoration: it is the
@@ -109,28 +114,99 @@ fn verdict_for(c: &Candidate, all: &[Candidate], binding_keys: &BTreeSet<String>
             why: format!("held domain '{prefix}' — operator ratifies these"),
         };
     }
-    if let Some(later) = superseding(c, all) {
+    // Supersession is explicit, never inferred from a mention (GH-1066): a
+    // decision that merely comes up while explaining something else has not
+    // overtaken it. Only two things count as "superseded" now, checked in
+    // this order because same-key is unambiguous and free, and an explicit
+    // marker is still cheaper to trust than reading the domain guard below.
+    if same_key_superseded(c, all) {
         return Outcome::Hold {
-            why: format!("superseded — a later decision '{later}' names it"),
+            why: format!(
+                "superseded-by-same-key — a later decision recorded under '{}' replaces it",
+                c.key
+            ),
+        };
+    }
+    if let Some(newer) = all
+        .iter()
+        .filter(|o| o.order > c.order)
+        .find(|o| names_explicit_supersession(&o.reason, &c.key))
+    {
+        return Outcome::Hold {
+            why: format!("superseded-explicit — '{}' marks it superseded", newer.key),
+        };
+    }
+    // Domain guard (GH-1066): a valid citation is not immunity. A newer,
+    // *binding* ruling in the same governance domain can have overtaken this
+    // one in substance even though it never names it — decision.auto-ratify's
+    // "contradicts no binding decision" clause, approximated the only way
+    // that does not require reading prose.
+    if let Some(later) = domain_superseded_by(c, all) {
+        return Outcome::Hold {
+            why: format!(
+                "older-than-binding-in-domain — '{later}' is binding and newer in domain '{}'",
+                domain_of(&c.key)
+            ),
         };
     }
     match citation(c, binding_keys) {
         Some(citation) => Outcome::Ratify { citation },
         None => Outcome::Hold {
-            why: "no citation — add --cite operator:<when> | issue:#<n> | decision:<key>"
+            why: "no-citation — add --cite operator:<when> | issue:#<n> | decision:<key>"
                 .to_string(),
         },
     }
 }
 
-/// The key of a later decision whose reason names `c`, if any. This is the
-/// soft supersede the ledger's own `supersedes` edge does not catch: a new
-/// decision that explains itself by pointing at an older one has already
-/// overtaken it, whatever its key.
-fn superseding(c: &Candidate, all: &[Candidate]) -> Option<String> {
+/// True when a later decision was recorded under the identical key — the
+/// most explicit supersession there is: nothing needs inferring when the
+/// ledger already carries a newer value for the same key.
+fn same_key_superseded(c: &Candidate, all: &[Candidate]) -> bool {
+    all.iter().any(|o| o.key == c.key && o.order > c.order)
+}
+
+/// Marker tokens this rule reads as an *explicit* supersession claim in a
+/// decision's own reason — never inferred from a mention anywhere in the
+/// prose, only from the target key appearing immediately after one of these
+/// tokens. `supersedes:` is the canonical form this rule writes going
+/// forward (GH-1066 doneWhen); `SUPERSEDES` (bare, all-caps) and `顯式取代`
+/// are phrasings already on the ledger before this rule existed — recognised
+/// so existing decisions need no migration, and never rewritten.
+const SUPERSEDES_MARKERS: [&str; 3] = ["supersedes:", "SUPERSEDES", "顯式取代"];
+
+/// True when `reason` explicitly names `target_key` as superseded: one of
+/// `SUPERSEDES_MARKERS`, immediately followed (after optional separator
+/// whitespace or a colon) by `target_key` at a word boundary. `target_key`
+/// appearing anywhere else in the sentence does not count — that is the
+/// exact shape of the GH-1066 bug this replaces.
+fn names_explicit_supersession(reason: &str, target_key: &str) -> bool {
+    SUPERSEDES_MARKERS.iter().any(|marker| {
+        reason.find(marker).is_some_and(|idx| {
+            let after = reason[idx + marker.len()..].trim_start_matches([' ', ':', '\t', '　']);
+            after.strip_prefix(target_key).is_some_and(|rest| {
+                rest.chars()
+                    .next()
+                    .is_none_or(|ch| !(ch.is_alphanumeric() || matches!(ch, '_' | '-' | '.')))
+            })
+        })
+    })
+}
+
+/// The domain segment of a key: everything before the first `.` — the same
+/// split the ledger uses to populate `decisions.domain`.
+fn domain_of(key: &str) -> &str {
+    key.split('.').next().unwrap_or(key)
+}
+
+/// The key of a later, **binding** decision in the same domain, if any
+/// (GH-1066 domain guard). Only a ratified sibling counts: an unratified
+/// newer decision in the same domain has no more authority than `c` itself
+/// yet, so it cannot be what overtook it.
+fn domain_superseded_by(c: &Candidate, all: &[Candidate]) -> Option<String> {
+    let domain = domain_of(&c.key);
     all.iter()
-        .filter(|o| o.order > c.order && o.key != c.key)
-        .find(|o| o.reason.contains(&c.key))
+        .filter(|o| o.ratified && o.order > c.order && o.key != c.key)
+        .find(|o| domain_of(&o.key) == domain)
         .map(|o| o.key.clone())
 }
 
@@ -206,6 +282,7 @@ mod tests {
             cites: Vec::new(),
             ratified: false,
             order,
+            date: String::new(),
         }
     }
 
@@ -239,14 +316,19 @@ mod tests {
     }
 
     #[test]
-    fn decision_named_by_a_later_reason_is_held() {
+    fn mention_alone_no_longer_holds_it() {
+        // GH-1066: this is the exact shape of the bug. `review.pool`'s
+        // reason names `review.engine` in passing (it builds on it, it does
+        // not replace it), so a later decision *mentioning* an earlier key
+        // must not hold it — only a same-key redecision or an explicit
+        // marker does. `review.engine` has a real citation and nothing
+        // superseded it, so it ratifies.
         let mut old = cand("review.engine", "cited in issue #900", 1);
         old.cites = vec!["issue:#900".to_string()];
         let newer = cand("review.pool", "replaces review.engine after the window", 2);
         let v = evaluate(&[old, newer], &none());
-        let held = v.iter().find(|x| x.key == "review.engine").unwrap();
-        assert!(!held.is_ratify());
-        assert!(held.columns().1.contains("superseded"), "{held:?}");
+        let engine = v.iter().find(|x| x.key == "review.engine").unwrap();
+        assert!(engine.is_ratify(), "{engine:?}");
     }
 
     #[test]
@@ -260,11 +342,163 @@ mod tests {
         assert!(v.iter().find(|x| x.key == "b.two").unwrap().is_ratify());
     }
 
+    // ── same-key supersession (GH-1066) ─────────────────────────────────
+
+    #[test]
+    fn same_key_later_decision_holds_the_older_row() {
+        let mut old = cand("db.engine", "postgres, per #1", 1);
+        old.cites = vec!["issue:#1".to_string()];
+        let mut newer = cand("db.engine", "sqlite instead, embedded is simpler", 2);
+        newer.cites = vec!["issue:#2".to_string()];
+        let v = evaluate(&[old, newer], &none());
+        // Both verdicts share the key "db.engine" here — same-key chains are
+        // exactly the case where key alone cannot disambiguate — so pick out
+        // each side by outcome instead. Both rows carry a valid citation, so
+        // only same-key supersession explains why one is held.
+        assert_eq!(v.len(), 2);
+        let held = v.iter().find(|x| !x.is_ratify()).unwrap();
+        assert!(
+            held.columns().1.starts_with("superseded-by-same-key"),
+            "{held:?}"
+        );
+        assert!(v.iter().any(|x| x.is_ratify()), "{v:?}");
+    }
+
+    #[test]
+    fn newest_row_in_a_same_key_chain_still_reaches_citation() {
+        let old = cand("db.engine", "postgres", 1);
+        let mut newer = cand("db.engine", "sqlite instead", 2);
+        newer.cites = vec!["issue:#2".to_string()];
+        let v = evaluate(&[old, newer], &none());
+        // Two candidates share one key here (mod.rs::collect no longer
+        // dedups before the rule sees them — GH-1066), so `evaluate` returns
+        // two verdicts under the same key; the newest one must ratify.
+        assert_eq!(v.len(), 2);
+        assert!(v.iter().any(|x| x.is_ratify()), "{v:?}");
+        assert!(v.iter().any(|x| !x.is_ratify()), "{v:?}");
+    }
+
+    // ── explicit-marker supersession (GH-1066) ──────────────────────────
+
+    #[test]
+    fn explicit_canonical_marker_holds_the_named_key() {
+        let mut old = cand("review.old-gate", "cited in issue #900", 1);
+        old.cites = vec!["issue:#900".to_string()];
+        let newer = cand(
+            "review.new-gate",
+            "supersedes:review.old-gate — the gate moved to the aggregate check",
+            2,
+        );
+        let v = evaluate(&[old, newer], &none());
+        let held = v.iter().find(|x| x.key == "review.old-gate").unwrap();
+        assert!(!held.is_ratify());
+        assert!(
+            held.columns().1.starts_with("superseded-explicit"),
+            "{held:?}"
+        );
+    }
+
+    #[test]
+    fn legacy_all_caps_supersedes_marker_is_recognized() {
+        // The exact phrasing already on the ledger before this rule existed
+        // (fleet.merge-authority's own reason text): `SUPERSEDES <key>` with
+        // no colon. Recognised, not migrated.
+        let mut old = cand("fleet.old-policy", "per #12", 1);
+        old.cites = vec!["issue:#12".to_string()];
+        let newer = cand(
+            "fleet.new-policy",
+            "SUPERSEDES fleet.old-policy=old-value ONLY in its gate clause",
+            2,
+        );
+        let v = evaluate(&[old, newer], &none());
+        let held = v.iter().find(|x| x.key == "fleet.old-policy").unwrap();
+        assert!(!held.is_ratify());
+        assert!(
+            held.columns().1.starts_with("superseded-explicit"),
+            "{held:?}"
+        );
+    }
+
+    #[test]
+    fn legacy_chinese_explicit_marker_is_recognized() {
+        let mut old = cand("review.old-carrier", "per #7", 1);
+        old.cites = vec!["issue:#7".to_string()];
+        let newer = cand(
+            "review.new-carrier",
+            "顯式取代review.old-carrier，理由如下",
+            2,
+        );
+        let v = evaluate(&[old, newer], &none());
+        let held = v.iter().find(|x| x.key == "review.old-carrier").unwrap();
+        assert!(!held.is_ratify());
+        assert!(
+            held.columns().1.starts_with("superseded-explicit"),
+            "{held:?}"
+        );
+    }
+
+    #[test]
+    fn explicit_marker_key_must_be_a_whole_word_not_a_prefix() {
+        // `review.merge-gate` must not be caught by a marker that actually
+        // names the longer key `review.merge-gate-v2` — the whole point of
+        // GH-1066 is that a partial textual match never counts as explicit.
+        let mut short = cand("review.merge-gate", "per #55", 1);
+        short.cites = vec!["issue:#55".to_string()];
+        let newer = cand(
+            "review.merge-gate-v2",
+            "supersedes:review.merge-gate-v2 — restates the v1 record",
+            2,
+        );
+        let v = evaluate(&[short, newer], &none());
+        let short_verdict = v.iter().find(|x| x.key == "review.merge-gate").unwrap();
+        assert!(short_verdict.is_ratify(), "{short_verdict:?}");
+    }
+
+    // ── domain guard (GH-1066) ───────────────────────────────────────────
+
+    #[test]
+    fn domain_guard_holds_older_candidate_behind_a_later_binding_sibling() {
+        let mut old = cand("review.legacy-gate", "per #100", 1);
+        old.cites = vec!["issue:#100".to_string()];
+        let mut newer = cand("review.auto-merge", "gate is now mechanical, per Tim", 2);
+        newer.ratified = true;
+        let v = evaluate(&[old, newer], &none());
+        let held = v.iter().find(|x| x.key == "review.legacy-gate").unwrap();
+        assert!(!held.is_ratify());
+        assert!(
+            held.columns().1.starts_with("older-than-binding-in-domain"),
+            "{held:?}"
+        );
+    }
+
+    #[test]
+    fn domain_guard_does_not_cross_domains() {
+        let mut old = cand("fleet.legacy-gate", "per #100", 1);
+        old.cites = vec!["issue:#100".to_string()];
+        let mut newer = cand("review.auto-merge", "gate is now mechanical", 2);
+        newer.ratified = true;
+        let v = evaluate(&[old, newer], &none());
+        let verdict = v.iter().find(|x| x.key == "fleet.legacy-gate").unwrap();
+        assert!(verdict.is_ratify(), "{verdict:?}");
+    }
+
+    #[test]
+    fn domain_guard_requires_the_sibling_to_be_binding_not_merely_newer() {
+        // An unratified newer decision in the same domain has no more
+        // authority than the candidate itself — it must not hold it.
+        let mut old = cand("review.legacy-gate", "per #100", 1);
+        old.cites = vec!["issue:#100".to_string()];
+        let newer = cand("review.auto-merge", "gate is now mechanical", 2); // not ratified
+        let v = evaluate(&[old, newer], &none());
+        let verdict = v.iter().find(|x| x.key == "review.legacy-gate").unwrap();
+        assert!(verdict.is_ratify(), "{verdict:?}");
+    }
+
     #[test]
     fn uncited_decision_is_held() {
         let v = evaluate(&[cand("cache.ttl", "seems about right", 1)], &none());
         assert!(!v[0].is_ratify());
-        assert!(v[0].columns().1.contains("no citation"));
+        assert!(v[0].columns().1.contains("no-citation"));
     }
 
     #[test]
--- a/crates/edda-cli/src/cmd_ratify/rule.rs
+++ b/crates/edda-cli/src/cmd_ratify/rule.rs
@@ -179,9 +179,14 @@ const SUPERSEDES_MARKERS: [&str; 3] = ["supersedes:", "SUPERSEDES", "顯式取�
 /// whitespace or a colon) by `target_key` at a word boundary. `target_key`
 /// appearing anywhere else in the sentence does not count — that is the
 /// exact shape of the GH-1066 bug this replaces.
+///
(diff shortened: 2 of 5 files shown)


fagemx:
## Code Review: Round 1 — PR #1098 @ b9eea1031e3f76b81f9706d27052d46bb345e93a

- elapsed: unmeasured (controller-spawned subagent; no pi session file)
- model_observed: unverified
- spec: review-spec-v1.5
- class: code-risk  (REVIEW.md classes: code-plain; fleet surface 出貨面 `crates/**`)
- escalations: none
- cost: 16 tool calls, read-only; no cargo, no checkout, no build (C5 selector empty)

### IN SCOPE

Changed behaviour and paths: `crates/edda-cli/src/cmd_ratify/{mod,rule,tests}.rs` and
`crates/edda-cli/tests/ratify_exit_codes.rs` — 4 files, +717/-38, two commits
(`de79b22` preserved, `b9eea10` fixture repair). Consumer of the changed rule is
`cmd_ratify::sweep` alone (`evaluate` has exactly two callers: `mod.rs:224` and
`tests.rs:498`); no other crate reads this engine. Base `bb60501` is unmoved and
equals the merge-base, so current-base integration is clean.

**Provenance accounted for.** `git diff de79b22..b9eea10 --name-only` touches only
`mod.rs` and `tests.rs` — `rule.rs` and `evidence.rs` are untouched by the second
commit, exactly as the PR body claims. The rule engine is therefore entirely the
dead lane's work, verified by nobody who wrote it, so I audited it by reading
rather than by trusting the body.

**The core defect is genuinely fixed, not narrowed.** The name-mention inference is
gone: `superseding()` (base `rule.rs`, `o.reason.contains(&c.key)`) is deleted, not
guarded. It is replaced by three predicates that cannot fire on a mention —
`same_key_superseded` (`rule.rs:164`), `names_explicit_supersession` (`rule.rs:182`,
marker-anchored with a word-boundary check), and `domain_superseded_by`
(`rule.rs:205`). No surviving path infers supersession from prose.

**The domain guard reaches production.** `collect` keeps already-ratified rows in
the candidate slice and `evaluate` filters only its *output* (`rule.rs:102-108`), so
a binding newer sibling is present in `all` when `domain_superseded_by` looks for
one. This is the wiring question that mattered most and it holds.

**Both fixture repairs rest on true premises, verified at the base SHA.**
(1) The same-key stopgap: `crates/edda-ledger/src/sqlite_store/events.rs` at base runs
`UPDATE decisions SET is_active = FALSE, status = 'superseded' WHERE key = ?1 AND
branch = ?2 AND is_active = TRUE` in the same write that appends a same-key
successor, so `active_decisions()` cannot return two active rows for one
`(branch,key)`. The deleted fixture asserted something the storage layer forbids;
removing it is correct, and its stated coverage is genuinely carried by the
surviving `a_re_decided_key_is_ratified_once_not_once_per_row`.
(2) The `--since` repair: `decide_at` sets `event.ts` and then calls
`finalize_event` before `append_event`, which recomputes the hash the `ts` is
covered by. The fixture straddles the boundary for real — four rows at
2026-08-14/15, the rest at wall-clock now — and `gh1066_without_since_the_rail_
closeout_batch_would_have_ratified` is a genuine negative control proving `--since`
is load-bearing. All three OS test jobs are green on these fixtures, which is
empirical proof the re-finalized events append cleanly.

**Pre-existing tests were not weakened.** `supersede_only_looks_forward` is
byte-identical at base and head. The only removed test is
`decision_named_by_a_later_reason_is_held`, which asserted the bug itself.

**Security / data-loss:** no ledger data is rewritten — the diff touches no ledger,
no migration, and no `.edda` content; the three legacy marker phrasings are
recognised in a const, never rewritten. No new `unsafe`, no destructive command,
no shell.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 4 changed files, all under `crates/edda-cli/` — inside the issue's predicted surface (`cmd_ratify` + tests) |
| U2 | any | P1 | **FAIL** | `Closes #1066` at body line 99, while the body's own "One doneWhen nuance" section reports a doneWhen bullet undelivered — see [P1] |
| U3 | any | P1 | PASS | `Issue: #1066` present |
| U4 | any | P1 | PASS | Only offender is `wip(edda-cli): GH-1066 in progress…`, exempt per `CONTRIBUTING.md` (wip checkpoint) |
| U5 | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → exit=0 |
| U6 | any | P0 | **FAIL** | `CI Gate` = failure @ this SHA; `Test (ubuntu-latest)` failed deterministically — see [P0] |
| U7 | any | P1 | PASS | Whole scoped audit finished before posting; all blocking findings batched in this round |
| U8 | any | P1 | N.A.(no `Decision:` line in the PR body) | DECISIONS block printed empty |
| C1 | code-plain | P0 | **FAIL** | doneWhen bullet 5 (acceptance fixture) not met on its PR-side parts — see [P0] |
| C2 | code-plain | P0 | PASS | `fix:` subject; 19 added `#[test]` lines, including regression tests that fail without the change (`mention_alone_no_longer_holds_it`, `mention_alone_does_not_hold_it_through_the_cli`) |
| C3 | code-plain | P1 | PASS | 20 `unwrap`/`expect` candidates, every one inside `#[cfg(test)]` (`rule.rs mod tests`; `tests.rs` is declared `#[cfg(test)] mod tests` at `mod.rs:23-24`) |
| C4 | code-plain | P1 | PASS | no `unsafe`, no `clippy::all`, no crate-level `#![allow]` — grep empty |
| C5 | code-plain | P1 | N.A.(selector empty) | Only touched crate dir is `crates/edda-cli`, which the selector excludes; CI's Windows `Test` job covers package `edda` directly and passed. No local cargo owed on this SHA |
| D1–D5 | docs | — | N.A.(class not routed — no `*.md`/`docs/**` in the diff) | classifier: `classes=code-plain` |
| S1–S3 | skills | — | N.A.(class not routed) | classifier: `classes=code-plain` |
| R1 | code-risk | P0 | PASS (run anyway) | destructive-command grep empty on the diff |
| R2 | code-risk | P0/P1 | PASS (run anyway) | mixed `&&`/`\|\|` grep empty — no shell in this diff |
| R3–R5 | code-risk | — | N.A.(no `*.sh`/`*.ps1` in the diff; class not routed) | — |
| §5.5 | any | P1 | PASS (with one note) | table below; `sh scripts/wiring-scan.sh` RAN |

### Wiring

|---|---|---|---|---|
| `--key` (`RatifyArgs.keys`, `mod.rs:59-60`) | clap `Vec<String>`, `requires = by_rule` | `mod.rs:229` filter, applied before the write loop at `mod.rs:251` | empty vec = no filter; not an error path | flag→sweep filter→write loop; asserted by `key_bound_sweep_ignores_the_threshold` |
| `--since` (`RatifyArgs.since`, `mod.rs:63-64`) | clap `Option<String>`, validated by `validate_since` (`mod.rs:183`, exit 2) | `mod.rs:230-234` filter via the `dates` map built at `mod.rs:220-223` | malformed → exit 2 (`ratify_exit_codes.rs::malformed_since_exits_2`); unknown ts → row excluded, fail-safe | flag→validate→date map→filter→write loop; proven end to end by `gh1066_since_bound_…` plus its negative control |
| `--yes` (`RatifyArgs.yes`, `mod.rs:69-70`) | clap `bool` | `mod.rs:238` threshold guard | refusal prints count + table and returns without writing (`mod.rs:239-243`) | flag→guard→early return; asserted by both threshold tests |
| `Candidate.date` (`rule.rs:45`) | `collect`, `d.ts.clone().unwrap_or_default()` (`mod.rs`) | `sweep`'s `dates` map only; `evaluate` never reads it (documented `rule.rs:41-44`) | missing ts → `""` → excluded by `--since`; a read default, not a swallowed write | field→collect→sweep filter |
| hold `superseded-explicit` / `older-than-binding-in-domain` / `no-citation` | `verdict_for` (`rule.rs:130-158`) | `print_table` → operator | n/a | reachable in product; domain guard reaches it because `collect` keeps ratified rows in `all` |
| hold `superseded-by-same-key` (`rule.rs:122-128`) | `verdict_for` | `print_table` → operator | n/a | **not reachable through `sweep`** — `collect` feeds only `active_decisions`, ≤1 row per `(branch,key)`; see [P2] |

### Findings

- [P0] Exact-head CI is deterministically red, and the cause is in this diff: three new flags are undocumented. `CI Gate` = failure; `Test (ubuntu-latest)` failed at step "Check CLI docs coverage (GH-650)" — `cargo test --workspace` itself passed on all three OSes. This is a content check, not an environmental flake, so re-running will not clear it. — evidence: run 34347321953 job 102451969278: `error: undocumented flag for 'ratify': --key` / `--since` / `--yes`; `check-cli-docs: FAIL — missing flags: ratify:--key ratify:--since ratify:--yes`. Fix: document the three flags in the `### edda ratify` section of `docs/reference/cli.md` (line 205), or add them to `scripts/cli-docs-ignore.txt` if genuinely internal.
- [P0] doneWhen bullet 5 (the acceptance fixture) is not met on the parts that are the PR's to deliver. The bullet asks that "the five held keys become `ratify`". (a) The fifth key, `fleet.lane-launch`, is present in the fixture but written with reason `"per the lane launcher design"` and no `cites` (`tests.rs:438-441`), so it carries no citation of any kind and holds as `no-citation` — the opposite of the required outcome — and `gh1066_plain_flow_rulings_ratify_despite_being_named_by_a_later_decision` (`tests.rs:494`) silently omits it, asserting only four keys. (b) Of the four it does assert, only `review.default-path` and `review.watcher` are named by a later decision (`review.shell-branch`, reason `"builds on review.default-path and review.watcher"`, `tests.rs:484-486`); nothing in the fixture names `review.readonly-proof` or `review.merge-gate`, so those two assertions pass identically against the pre-fix engine and prove nothing about the bug — while the test's own name claims all four are "being named by a later decision". (c) The bullet asks for the 176-row snapshot checked in as a fixture; the fixture is an 11-row reconstruction. (a)–(c) are undisclosed — the PR body's "One doneWhen nuance" section raises only the domain-guard point. **Carve-out:** I am *not* charging the PR for the bullet's other clause (the four `d-NNN` rows becoming `hold: older-than-binding-in-domain`) — see the adjudication below. — evidence: `tests.rs:395-511`; `rule.rs:225-245` (`citation_from_reason`: no `#`/`GH-` number, no binding key, and no "operator"/「操作者」 in that reason, so all three arms return `None`).
- [P1] `Closes #1066` is present while the PR reports a doneWhen bullet as undelivered. U2 permits a closing keyword exactly when the diff delivers every doneWhen item. Merging as-is auto-closes #1066 with the fixture bullet unresolved and with the residual behaviour (below) recorded nowhere. — evidence: PR body line 99 `Closes #1066`; PR body section "One doneWhen nuance for review". Fix either way: amend the issue's bullet 5 first (my recommendation — see FOLLOW-UP) and keep the keyword, or drop the keyword to `Issue: #1066` alone.
- [P2] `names_explicit_supersession` inspects only the **first** occurrence of each marker, so a reason that explicitly supersedes two keys is honoured for one of them. `reason.find(marker)` (`rule.rs:184`) returns the first hit; for `"supersedes:a.one and supersedes:b.two"` the check against `b.two` reads `after = "a.one and supersedes:b.two"`, `strip_prefix("b.two")` fails, and `b.two` ratifies despite an explicit marker naming it. Forward-looking only: `supersedes:` is the canonical form this PR introduces, so no ledger data has this shape today, and the legacy `SUPERSEDES` row on main names a single key. Iterating `match_indices` instead of `find` closes it.
- [P2] `validate_since` (`mod.rs:183-195`) checks shape but not calendar validity, so `--since 2026-99-99` is accepted. The comparison stays lexicographic and therefore safe, but the flag advertises a date it never parses.
- [P2] One of the four required hold reasons cannot be produced by the product. `same_key_superseded` (`rule.rs:164`) needs two candidates sharing a key, and `collect` feeds `evaluate` only `active_decisions`, which the ledger guarantees carries at most one row per `(branch,key)` — the same fact the PR correctly used to delete the same-key fixture. `evaluate` has no other caller, so `superseded-by-same-key` is unreachable outside unit tests. The *outcome* the doneWhen wants is still delivered (a redecided key's old row is deactivated before the sweep sees it), and the lane documents the situation honestly in the `collect` comment, so this is a note rather than a blocker — but "four distinguishable hold reasons" is true of the rule and not of the sweep's output, and that is worth stating in the issue's close-out.

### FOLLOW-UP ISSUE

- **Amend #1066's doneWhen bullet 5, rather than changing this PR's engine.** I verified the lane's mechanical claim independently: `edda_core::decision::extract_domain` is `key.split('.').next()` (`crates/edda-core/src/decision.rs:101-103`) and `rule.rs::domain_of` (`rule.rs:197-199`) mirrors it exactly, so `d-033.merge_authority` is in domain `d-033` and `review.auto-merge` is in domain `review`. They share no domain. Making the domain guard hold those four rows would require asserting a relationship the ledger never recorded — precisely the inference GH-1066 exists to delete. The issue's bullet contradicts the issue's own thesis, and the lane was right to flag it instead of fabricating a domain match. The issue text should be corrected to say the four `d-NNN` rows are covered by the date guard, matching the bullet's own title "No domain **or** date guard".
- **Record the residual, which the issue close-out should not lose.** Judging the doneWhen in substance, as asked: the required property is that a blind sweep cannot swear the superseded constitution back in. After this PR an *unbounded* sweep still ratifies `d-033`/`d-034` — the PR's own `gh1066_without_since_the_rail_closeout_batch_would_have_ratified` (`tests.rs:539`) asserts exactly that. What actually stops the blind case is the `--yes` threshold: the real table's 97 ratifies exceed 20, so an unscoped sweep refuses, prints the table with those rows visible in the ratify column, and writes nothing. I judge that adequate in substance — the sweep is no longer silently destructive and the operator sees the rows — but it is a threshold plus operator attention, not a guard, and `--yes` still walks straight past it. Worth an evidenced follow-up on whether `d-NNN.*` legacy keys should carry an explicit `supersedes:` marker (which this PR now honours) so the protection stops depending on a flag the operator must remember.
- **Reviewer-process note, not blocking:** the lane's stated L0 set does not include `scripts/check-cli-docs.sh`, which is why a green local L0 still produced a red CI. Adding it to the L0 line for any diff that adds a clap flag would have caught the P0 above before the push.

### RAN vs READ

RAN:
- `sh scripts/lint-markdown-content.sh` → exit=0 (U5)
- `sh scripts/wiring-scan.sh bb60501…@b9eea10…` → 4 new fields, 9 swallow candidates (all test-cleanup `let _ = remove_dir_all` plus the fail-safe `unwrap_or_default()` on a read; none on a write path)
- REVIEW.md §3 classifier on the changed-file list → `classes=code-plain`, `canonical_class=code-risk`
- U1/U2/U3/U4/C2/C3/C4/C5/R1/R2 enumerators (git greps over `origin/main..b9eea10`)
- `edda ask fleet.merge-authority`, `edda ask decision.auto-ratify` — D2-style check of the code's source claims. Confirms the `SUPERSEDES` phrasing the fixture cites is real ledger text: `fleet.merge-authority`'s reason literally begins `SUPERSEDES fleet.merge-authority=lgtm-is-the-merge-trigger-no-per-pr-ask ONLY in its gate clause`. The inventory claim in `SUPERSEDES_MARKERS` (`rule.rs:175`) is accurate and nothing was migrated.
- Source reads at the base SHA: `crates/edda-ledger/src/sqlite_store/events.rs` (same-key deactivation), `crates/edda-core/src/decision.rs:101-103` (`extract_domain`), `crates/edda-core/src/types.rs:181-216` (`DecisionPayload` has no `notes` field, so the doneWhen's "reason/notes" is satisfied by reading `reason` — there is no notes field to miss)
- `git show bb60501:…/rule.rs` vs head for `supersede_only_looks_forward` → byte-identical, PR body claim confirmed
- `git diff de79b22..b9eea10 --name-only` → `mod.rs`, `tests.rs` only; provenance split confirmed

READ:
- L1 receipt: **CI run 34347321953 @ b9eea1031e3f76b81f9706d27052d46bb345e93a** — exact head. Format ✅, Clippy ubuntu/macos/windows ✅, Test macos ✅, Test windows (7-crate subset, includes package `edda`) ✅, MSRV 1.91.0 ✅, **Test ubuntu ❌**, **CI Gate ❌**. Covers: fmt, clippy on all three OSes, `cargo test --workspace` on Linux and macOS, the 7-crate Windows subset, MSRV compile. Does not cover: the 16 non-subset crates' Windows test targets — irrelevant here, the C5 selector is empty because the only touched crate is `edda-cli`/package `edda`, which the Windows job runs directly and which passed.
- No Cargo gate was RAN by me: the workspace gates are READ from the run above per `ladder` L1, and C5 owed nothing on this SHA.

### Verdict

Changes Requested, P0=2, P1=1 — exact-head CI is deterministically red on three undocumented `ratify` flags, and the acceptance fixture leaves the issue's fifth held key unasserted (it holds as `no-citation`) with two of its four assertions vacuous; the rule engine itself is sound and the mention-inference bug is genuinely removed.

fagemx:
## Review Response: Round 1

Answering every blocking finding from Round 1 (`b9eea1031e3f76b81f9706d27052d46bb345e93a`). New head: `501d308` — gate: `CI run 34352040457 @ 501d308`.

### P0-1 — exact-head CI red (undocumented `--key`/`--since`/`--yes`)

Fixed in `5d9e6aa`. `docs/reference/cli.md`'s `### \`edda ratify\`` Options
table now documents all three flags. Also corrected the "Rule
`cited-authority`" subsection, which still described the removed
mention-based inference ("anything a later decision's reason names") as
current behavior and showed a stale example row with the old "superseded"
wording — replaced with the four hold outcomes the engine actually
produces (`superseded-by-same-key`, `superseded-explicit`,
`older-than-binding-in-domain`, `no-citation`).

Verified locally against a fresh build:

```
$ EDDA_BIN=<built edda.exe> bash scripts/check-cli-docs.sh
check-cli-docs: OK — all 65 verbs; flags verified in docs/reference/cli.md
```

### P0-2 — acceptance fixture omitted the fifth key and two assertions were vacuous

Fixed in `501d308`. Both defects Round 1 found are addressed:

- `fleet.lane-launch` now carries a real citation (`cites:
  ["operator:2026-09-02"]`) instead of none, and is now included in
  `gh1066_plain_flow_rulings_ratify_despite_being_named_by_a_later_decision`'s
  assertion list — all five held keys are present and asserted, not four.
- `review.readonly-proof` and `review.merge-gate` are now genuinely named
  by a later decision: `review.shell-branch`'s reason was extended from
  naming two of the four review keys to naming all four. `fleet.lane-launch`
  is named by `review.auto-merge`'s reason (paraphrased, as the surrounding
  reasons already are) — this is that key's own mention-inference trip,
  since nothing previously named it at all.

**Falsification proof, as asked.** Reverted just the fix — reinstated base
`bb60501`'s exact `superseding()` (mention-based inference) in place of
the three current predicates — on a throwaway filesystem copy of the
worktree (not a git worktree; copied via `tar`, built with its own scratch
`CARGO_TARGET_DIR`, both deleted immediately after — ~2.1 GB, nothing
retained, the real worktree untouched throughout). Running only the
repaired acceptance test against that revert:

```
test cmd_ratify::tests::gh1066_plain_flow_rulings_ratify_despite_being_named_by_a_later_decision ... FAILED
thread '...' panicked at crates\edda-cli\src\cmd_ratify\tests.rs:554:9:
review.readonly-proof: Verdict { key: "review.readonly-proof", outcome: Hold { why: "superseded — a later decision 'review.shell-branch' names it" } }
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 748 filtered out
```

`review.readonly-proof` — one of the two previously-vacuous assertions —
fails first and aborts the `for` loop, which is sufficient to falsify: the
fixed engine is required for this test to pass. (By the same
construction, `review.merge-gate` and `fleet.lane-launch` are also newly
tripped by the reverted engine, and `review.default-path`/`review.watcher`
as before Round 1 — the loop just panics on the first failure, so only
`review.readonly-proof`'s message is captured above.)

Also corrected the file-level comment above `gh1066_snapshot_ledger`,
which claimed `review.auto-merge` had "overtaken the whole 2026-08
merge-authority regime in substance" — the exact claim the controller's
amendment withdrew as mechanically impossible. It now states plainly that
`review.auto-merge` does not, and is not claimed to, domain-guard the four
`d-NNN.*` rows, and points at where the domain guard is actually proven.

### Amended doneWhen bullet 5 — read and confirmed

Read the issue's newest comment (controller amendment, 2026-09-09) before
doing this round's work. Addressed all three parts:

1. **All five held keys present and asserted, none vacuous** — P0-2 above.
2. **Domain guard proven on genuinely shared domains, positive and
   negative, `review.` and `fleet.`** — the `review.` pair already existed
   (`domain_guard_holds_older_candidate_behind_a_later_binding_sibling` /
   `domain_guard_requires_the_sibling_to_be_binding_not_merely_newer`,
   both unmodified). Added the missing `fleet.` pair in `rule.rs`:
   `domain_guard_holds_older_fleet_candidate_behind_a_later_binding_fleet_sibling`
   and `domain_guard_requires_the_fleet_sibling_to_be_binding_not_merely_newer`.
3. **Four old-constitution rows, outcome-oriented** — already satisfied
   and left unchanged: `gh1066_without_since_the_rail_closeout_batch_would_have_ratified`
   (negative control: an unbounded sweep does ratify them, honestly) +
   `gh1066_since_bound_keeps_the_rail_closeout_batch_out_of_the_sweep`
   (`--since` keeps them out) + the pre-existing generic
   `unbounded_sweep_refuses_past_the_threshold_without_yes` (the real
   table's 97 ratifies exceed the default cap of 20, so a truly unscoped
   sweep refuses and prints the table rather than writing silently). The
   residual — an unbounded sweep still *shows* `d-033`/`d-034` as ratify
   candidates before the threshold refuses — is explicitly accepted by the
   amendment; no engine change was needed or made.

`Closes #1066` (already in the PR body) is now honest: every part of the
amended bullet 5 is delivered, and the residual is recorded rather than
hidden.

### P1 — `Closes #1066` while a doneWhen bullet was undelivered

Resolved as a consequence of P0-2 and the amendment above: the doneWhen
bullet is now delivered in full, so the existing `Closes #1066` in the PR
body is accurate as written. No PR-body edit was needed.

### P2s

- **`names_explicit_supersession` first-occurrence-only** — fixed
  (`3ccbb2b`): `reason.find(marker)` → `reason.match_indices(marker)`, so a
  reason superseding two keys with the same marker holds both. Regression
  test: `explicit_marker_used_twice_holds_both_named_keys`.
- **`validate_since` accepted `2026-99-99`** — fixed (`3ccbb2b`): added
  `is_valid_since_date` (calendar-correct, leap years included, no new
  dependency — `Cargo.toml` is outside this lane's writable surface
  regardless) and wired `validate_since` through it. Coverage:
  `since_calendar_validity_rejects_impossible_dates` (the pure predicate)
  and `ratify_exit_codes.rs::since_with_an_impossible_calendar_date_exits_2`
  (the `usage_exit(2)` boundary end to end, spawned-binary — `usage_exit`
  calls `std::process::exit` directly and cannot run in-process, the same
  reason the pre-existing `well_formed_since_passes_through_unchanged`
  cannot exercise the malformed path either).
- **`superseded-by-same-key` unreachable through the product path** —
  routed, not fixed. Round 1 itself characterized this as a note rather
  than a blocker: the *outcome* the doneWhen wants (a redecided key is
  ratified once) is already delivered by the ledger's own same-key
  deactivation before the rule ever sees a candidate set, and the rule's
  own guard is directly exercised by its dedicated unit tests
  (`same_key_later_decision_holds_the_older_row`,
  `newest_row_in_a_same_key_chain_still_reaches_citation`). Making the
  branch reachable through `sweep()` would mean feeding `evaluate` a
  non-deduplicated candidate set from the product path — a design change
  outside this PR's scope, not a cheap fix. Left as Round 1's own
  documented FOLLOW-UP; no new issue filed beyond that note.

### Engine untouched, per Round 1's principle

`rule.rs`'s three supersession predicates (`same_key_superseded`,
`names_explicit_supersession`'s marker-detection shape,
`domain_superseded_by`) are unmodified from `b9eea103` except the
`match_indices` loop fix above, which is strictly a superset (finds more
matches, never fewer) — every pre-existing `rule.rs` test still passes
unmodified, plus the four new ones.

### Verification

- `cargo fmt --all --check` — clean (all three commits)
- `cargo clippy -p edda --all-targets -- -D warnings` — clean (all three commits)
- `cargo test -p edda --bin edda cmd_ratify` — 56 passed, 0 failed (the repaired fixture and both `gh1066_*` sweep tests individually confirmed)
- `cargo test -p edda` (full package) — 858 passed, 0 failed
- `bash scripts/check-cli-docs.sh` (against a freshly built binary) — OK
- `sh scripts/lint-file-length.sh --tree` — clean (default 1000-line ceiling; no per-file override for any touched file; `docs/reference/cli.md` is `.md`, outside this ratchet's `*.rs` scope)
- Falsification proof for P0-2, above
- Did **not** run `cargo test --workspace` (out of this lane's L0 verification budget)
- Stray in-worktree `target/` from the pre-commit hook's own cargo invocation (known gap, #1099) deleted after each commit, ~791 MB each time; the falsification scratch copy (~2.1 GB, source + its own throwaway target) fully deleted after use

Gate: **CI run 34352040457 @ 501d308**

New head: `501d308` — three commits on top of Round 1's `b9eea103`, each
pushed immediately after its own local L0 pass:
- `5d9e6aa` — P0-1 (CLI docs)
- `3ccbb2b` — P2s (multi-marker supersession, real calendar dates, fleet-domain guard coverage)
- `501d308` — P0-2 / P1 (acceptance fixture repair)

Not merging.

---

*Posted by the controller on the fix lane's behalf. The lane wrote this response and pushed all three commits, then stopped while waiting on the CI watch rather than posting. Before relaying, the controller verified independently at `501d308`: all five held keys appear in the assertion list of `gh1066_plain_flow_rulings_ratify_despite_being_named_by_a_later_decision`; `fleet.lane-launch` carries a real citation and is named by a later decision's reason in the fixture; the two new `fleet.` domain-guard tests exist in `rule.rs`; and the diff is confined to the five declared files. The falsification output and the local gate results above are the lane's, not re-run by the controller.*

fagemx:
**Follow-up to the Review Response above: CI confirmed green at the full head SHA**

The comment above was relayed from this lane's draft while it was still
waiting on the CI watch — it named the run but not yet its conclusion,
and used the abbreviated SHA. For the record, per this project's
gate-receipt convention (full SHA, not abbreviated):

**CI run 34352040457 @ `501d3088c48170c7deed08446993462d7ef72d52` — conclusion: success.**

All required jobs are green: Format, Clippy (ubuntu/macos/windows), Test
(ubuntu/macos/windows), MSRV (1.91.0), Detect code changes, CI Gate —
including `Test (ubuntu-latest)` → `Check CLI docs coverage (GH-650)`,
the exact step Round 1 found red. `git rev-parse HEAD` on the lane's
worktree matches this SHA exactly (verified independently, not assumed).

No further changes. Not merging.

fagemx:
## Code Review: Round 2 — PR #1098 @ 501d3088c48170c7deed08446993462d7ef72d52

- elapsed: unmeasured (controller-spawned subagent; no pi session file)
- model_observed: claude-opus-5 (self-reported by the running session)
- reviewer_session: 41362b6e-78e8-478f-8fb5-214516e1f6e6 (replacement session; Round 1's is gone — receipts and CI READ before anything was RAN)
- spec: review-spec-v1.5 (REVIEW.md at the reviewed SHA). §5 **content** enumerators run in the three-dot form: this branch is behind its base, and two-dot `git diff` would fold `origin/main`'s own commits into the PR's surface. That is the GH-1003 correction already on main as v1.6; `git log` stayed two-dot, as both versions require.
- class: code-risk  (REVIEW.md classes: `code-plain docs` — `docs/reference/cli.md` joined the surface this round, so §5.1 D1–D5 route for the first time. No path on §3.1's risk surface, so §5.4 is not routed; R1/R2 run anyway.)
- escalations: none
- shadow: false
- cost: 18 tool calls, read-only. No checkout, no worktree, no cargo, no build, no lane (C5 selector empty — nothing owed on this SHA).

### IN SCOPE

Delta round per §2: the target is `git diff b9eea103..501d308` — 250 insertions across five files in three commits (`5d9e6aa` P0-1 docs, `3ccbb2b` P2s, `501d308` P0-2/P1 fixture) — plus confirmation that each Round 1 blocking finding is resolved. Round 1's passing checks are not reworked.

**Surface.** Five files, every one inside the declared surface (`crates/edda-cli/src/cmd_ratify/**`, `crates/edda-cli/tests/ratify_exit_codes.rs`, `docs/reference/cli.md`). Nothing outside it changed. `cli.md` was added to the surface between rounds by the controller; Round 1's P0-1 existed because the original brief excluded it while the change added three clap flags.

**Current-base integration.** Merge-base is still `bb60501`; base `main` advanced to `5d71458` during Round 1 (#1097, #1095) touching only `REVIEW.md`, `docs/fleet/rules.md`, `scripts/test-review-l0.sh`, `scripts/wiring-scan.sh` — no overlap with this PR's five files, no `crates/**`. `mergeable: MERGEABLE`.

**Round 1's engine findings survived Round 2's edits.** `3ccbb2b` touches `rule.rs`, so I re-checked rather than assumed: the mention inference is still deleted, not narrowed (`superseding()` with `o.reason.contains(&c.key)` is absent at head); the three replacement predicates still cannot fire on a mention (`same_key_superseded` `rule.rs:164`, `names_explicit_supersession` `rule.rs:187`, `domain_superseded_by` `rule.rs:210`); the domain guard still reaches production (`collect` keeps ratified rows (`mod.rs:309`), `evaluate` filters only its output (`rule.rs:101-108`)); and the `review.` domain-guard pair is **byte-identical** `b9eea103` → `501d308`.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 5 changed files, all inside the (widened) declared surface; nothing outside it |
| U2 | any | P1 | **PASS (Round 1 P1 closed)** | `Closes #1066` at body line 99 is now honest — the amended bullet 5 is delivered in full (see Findings); bullets 1–4 and 6 unchanged and verified (bullet 6's L1 leg = exact-head CI, all green) |
| U3 | any | P1 | PASS | `Issue: #1066` present |
| U4 | any | P1 | PASS | Only offender `wip(edda-cli): GH-1066 in progress…`, exempt per `CONTRIBUTING.md`; the three new subjects are conventional `fix(edda-cli):` |
| U5 | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → exit=0 |
| U6 | any | P0 | **PASS (Round 1 P0-1 closed)** | `CI run 34352040457 @ 501d308` — exact head, **completed/success, every job green**, `Test (windows-latest)` included (14m10s). The step that failed in Round 1 now reports `check-cli-docs: OK — all 65 verbs; flags verified in docs/reference/cli.md` |
| U7 | any | P1 | PASS | Whole scoped audit finished before posting; every Round 1 finding adjudicated |
| U8 | any | P1 | N.A.(no `Decision:` line in the PR body) | DECISIONS block empty |
| D1 | docs | P1 | PASS | Enumerator over added lines returns **no** backticked-command candidates (the `edda ratify` examples sit in fenced blocks). Flag existence checked as a *source* claim instead — `--key`/`--since`/`--yes` all exist in clap at `mod.rs:59-70`, each `requires = "by_rule"` — and by CI's own freshly-built binary above, which is stronger than my installed runtime (see RAN vs READ for the recorded version mismatch) |
| D2 | docs | P1 | PASS | `edda ask` run from the checkout for all five held keys and both namers. Doc/fixture claims check out; details under Findings |
| D3 | docs | P1 | PASS | Enumerator returns no repo-anchored paths added — empty list, no `MISSING` |
| D4 | docs | P0 | PASS | authority-boundary grep empty on the diff |
| D5 | docs | `[判斷]` | PASS | The `cited-authority` prose now matches the shipped engine reason-for-reason; see [P0-1] under Findings. One wording note routed as [P2] |
| C1 | code-plain | P0 | **PASS (Round 1 P0-2 closed)** | Amended doneWhen bullet 5 delivered on all three parts, established by reading; see Findings |
| C2 | code-plain | P0 | PASS | `fix:` subjects; 24 added `#[test]` lines across the PR, 5 of them this round, each failing without its change |
| C3 | code-plain | P1 | PASS | 3 new `unwrap` candidates this round, all inside `rule.rs`'s `#[cfg(test)] mod tests` (`rule.rs:279-280`); `tests.rs` is `#[cfg(test)] mod tests` (`mod.rs:23-24`) |
| C4 | code-plain | P1 | PASS | no `unsafe`, no `clippy::all`, no crate-level `#![allow]` — grep empty |
| C5 | code-plain | P1 | N.A.(selector empty) | Only touched crate dir is `crates/edda-cli`, which the selector excludes; CI's Windows `Test` job runs package `edda` directly and passed. No local cargo owed, none run |
| R1 | code-risk | P0 | PASS (class not routed; run anyway) | destructive-command grep empty on the diff |
| R2 | code-risk | P0/P1 | PASS (class not routed; run anyway) | mixed `&&`/`\|\|` grep empty — no shell in this diff |
| R3–R5 | code-risk | — | N.A.(class not routed; no `*.sh`/`*.ps1` in the diff) | — |
| S1–S3 | skills | — | N.A.(class not routed) | classifier: `classes=code-plain docs` |
| §5.5 | any | P1 | PASS | table below; `sh scripts/wiring-scan.sh` RAN |

### Wiring

|---|---|---|---|---|
| `--key` (`RatifyArgs.keys`, `mod.rs:59-60`) | clap `Vec<String>`, `requires = by_rule` | `mod.rs:264` filter, before the write loop | empty vec = no filter; not an error path | flag→sweep filter→write loop; asserted by `key_bound_sweep_ignores_the_threshold` |
| `--since` (`RatifyArgs.since`, `mod.rs:63-64`) | clap `Option<String>`, validated by `validate_since` (`mod.rs:224`, exit 2) | `mod.rs:265-269` filter via the `dates` map | malformed **or impossible** → exit 2; unknown ts → row excluded, fail-safe | flag→validate→date map→filter→write loop; end to end by `gh1066_since_bound_…` + its negative control + `since_with_an_impossible_calendar_date_exits_2` |
| `--yes` (`RatifyArgs.yes`, `mod.rs:69-70`) | clap `bool` | `mod.rs:273` threshold guard | refusal prints count + table, returns without writing | flag→guard→early return; both threshold tests |
| `Candidate.date` (`rule.rs:45`) | `collect`, `d.ts.clone().unwrap_or_default()` | `sweep`'s `dates` map only; `evaluate` never reads it | missing ts → `""` → excluded by `--since`; a read default, not a swallowed write | field→collect→sweep filter |
| `is_valid_since_date` (`mod.rs:186-219`) — **new this round**, private | `validate_since` (`mod.rs:224`) is its only caller | `sweep` (`mod.rs:248`) | `false` → `usage_exit` → exit 2 | predicate→boundary→process exit; both a pure-predicate test and a spawned-binary exit-code test |
| hold-reason vocabulary in `docs/reference/cli.md:265-277` — **new this round** | `verdict_for` (`rule.rs:111-159`) format strings | operator reading `--dry-run` | n/a | four of the five documented reasons are reachable through `sweep`; `superseded-by-same-key` is not — see [P2] |

Swallow scan: 13 candidates, all test-cleanup `let _ = remove_dir_all`, prose containing the word "silently", or the fail-safe `unwrap_or_default()` on a **read**. None on a write path. Unchanged from Round 1's adjudication.

### Findings

- **[P0-1 — CLOSED]** Exact-head CI is green. `CI run 34352040457 @ 501d308` completed `success` with every job passing, and the step that failed deterministically in Round 1 now prints `check-cli-docs: OK — all 65 verbs; flags verified in docs/reference/cli.md`. I also checked the *second* half of `5d9e6aa`, which CI cannot judge: the "Rule `cited-authority`" prose no longer describes the deleted inference. The stale bullet ("anything a later decision's reason names") and the stale example row (`superseded — a later decision 'review.pool' names it`) are both gone, replaced by the four hold outcomes — and each documented reason matches the engine's actual format string verbatim, including the example row `older-than-binding-in-domain — 'review.auto-merge' is binding and newer in domain 'review'` (`cli.md:286` vs `rule.rs:147`). The three new Options rows are true of the code as well, not just present: `--key` "the unbounded-sweep size refusal does not apply" ⇢ `bounded` at `mod.rs:261` gates the refusal at `mod.rs:273`; `--yes` "and neither `--key` nor `--since` given" ⇢ the same `!bounded` term. **No stale prose describing the deleted inference survives on the PR that deletes it.**
- **[P0-2 — CLOSED, and non-vacuous by reading, not by report]** I judged this statically rather than accepting the lane's falsification run, which I cannot re-execute read-only. Two things had to hold, and both do.
  **(1) All five would trip the deleted inference.** At base, `verdict_for` consults `superseding()` **before** `citation()` (`bb60501:rule.rs:106-133`), and `superseding()` is `all.iter().filter(|o| o.order > c.order && o.key != c.key).find(|o| o.reason.contains(&c.key))` — any strictly later row whose reason contains the key as a substring, so a trip yields `Hold` whatever the citation. In the head fixture: `review.shell-branch` (last row) has reason `"builds on review.readonly-proof, review.merge-gate, review.default-path and review.watcher"` (`tests.rs:523`), naming four; `review.auto-merge` has `"…keeping fleet.lane-launch's rollout order"` (`tests.rs:480`), naming the fifth, and it is recorded after `fleet.lane-launch` (`tests.rs:463` then `tests.rs:476`). All five are named by a strictly later row ⇒ all five Hold on the base engine ⇒ **none of the five assertions is vacuous.**
  A detail worth stating because it is what makes the lane's revert-only counterfactual legitimate rather than an artifact of the new `collect`: **base `collect` already pushed ratified rows into `candidates`** (`bb60501:mod.rs:243-245`) and base `evaluate` filtered `!c.ratified` on its *output* only. So the ratified `review.auto-merge` was in `all` at base too, and reverting `superseding()` alone reproduces the true base behaviour for `fleet.lane-launch`.
  **(2) All five reach `Ratify` at head.** No `HELD_PREFIXES` match; no duplicate key, so `same_key_superseded` is false; no reason carries a `SUPERSEDES_MARKERS` token, so `names_explicit_supersession` is false; and `domain_superseded_by` finds nothing — the only ratified row is `review.auto-merge` at order 6, *earlier* than the four `review.*` rows at orders 7–10, and a different domain from `fleet.lane-launch`. Each of the five then hits a well-formed `operator:<date>` cite. The loop asserts all five (`tests.rs:546-552`), and since `evaluate` emits no verdict for ratified rows, the `.unwrap()` in that loop is safe for exactly these five.
  The lane's runtime falsification is therefore corroborating, not load-bearing, for this verdict.
- **[Amended bullet 5 — delivered on all three parts]** Read the controller amendment (issue #1066, 2026-09-09) as the ceiling, not the original bullet. (1) Five present, five asserted, none passing on the unfixed engine — established above. (2) Domain guard on genuinely shared domains, positive and negative, for `review.` and `fleet.`: all four tests exist in `rule.rs` and each binds. The `review.` pair is byte-identical to `b9eea103`, so `3ccbb2b`'s `rule.rs` edit did not disturb it; the new `fleet.` pair is a true mirror — the positive holds `fleet.legacy-gate` with `older-than-binding-in-domain` behind a *ratified* `fleet.rollout-gate`, and the negative ratifies it on `issue:#100` when that sibling is merely newer, so the pair genuinely isolates *binding*, not *newer*. Without the guard the positive case would ratify, so it is not vacuous either. (3) Four `d-NNN` rows, outcome-oriented. Delivered by `--since` (`gh1066_since_bound_…`) plus its honest negative control (`gh1066_without_since_…`, which asserts an unbounded sweep **does** ratify them) plus the threshold. Stated precisely: `unbounded_sweep_refuses_past_the_threshold_without_yes` (`tests.rs:601`) runs on its own synthetic 21-row ledger, not on the GH-1066 snapshot, so it proves the *mechanism* refuses past 20 — the step from there to "the real 97-row table refuses" is an argument about production, not a fact this fixture demonstrates. That is exactly the carrier the amendment names (「`--since` 日期閘加上 `--yes` 門檻」) and the residual it explicitly accepts, so it binds as delivered; I am recording the boundary, not charging it.
- **[Fixture fidelity — checked against the real ledger, and acceptable]** The brief asked whether a fixture edited to make the test bite is still a snapshot of the real table's shape. Checked each edited edge with `edda ask` from the checkout. It holds up better than the diff alone suggests: the real `review.shell-branch` reason genuinely names `review.merge-gate`, `review.watcher` **and** `review.default-path` — three of the four the extended fixture reason names — so only `review.readonly-proof` is re-attributed there (its real namer, per the issue's own table, is `fleet.ledger-sync-trigger`). `fleet.lane-launch`'s mention is likewise re-attributed, from `ledger.cross-machine-projection` to `review.auto-merge`. Both re-attributions collapse a real mention edge onto a row already in the 11-row reconstruction rather than inventing a relationship that never existed: the property the issue documents — each of the five is named by a later decision — is preserved for all five, and the file discloses the edits at `tests.rs:377-380`. The amendment also dropped the original bullet's 176-row-snapshot clause, so an 11-row reconstruction is within the ceiling. **Not a blocker.** One gap routed as [P2] below.
- [P2] The fixture's provenance sentence no longer covers everything it disclaims. `tests.rs:374` still says "Keys, relative order, and **citations** are drawn from the real ledger", and the disclosure two lines later (`tests.rs:377-380`) covers only the edited *reasons* — but this round also added a citation: `fleet.lane-launch` gained `cites: ["operator:2026-09-02"]` (`tests.rs:468`), where at `b9eea103` it carried none. The adjacent per-decision comment is honest about it ("Before this fix … it carried no citation at all"), so nothing is hidden; the header sentence is simply now broader than the disclosure under it. Same sentence, minor referent slip: it calls the two edited reasons "two of the five plain-flow reasons", but the two edited reasons belong to `review.shell-branch` and `review.auto-merge` — the *namers*, not the five held keys. Extending the existing disclosure to cover the citation would close both.
- [P2] `docs/reference/cli.md:267-268` now documents `superseded-by-same-key` as a hold the `--by-rule` sweep produces, and no sweep can print it — Round 1's routed P2, one level up into user-facing reference. I re-derived the unreachability rather than inheriting it: `same_key_superseded` (`rule.rs:164`) needs two candidates sharing a key, `collect` feeds `evaluate` only `active_decisions` filtered to one branch, and the ledger deactivates the prior same-key row in the same write that appends its successor. The doc's *outcome* claim is nevertheless true — a decision superseded under the identical key is never ratified — so this is a precision gap, not a false statement, and it is the same substance Round 1 observed at `b9eea103` and rated P2. Per `loop` item 2 it cannot become a Round 2 blocker; it stays P2 and stays attached to Round 1's existing follow-up note.
- **[Round 1's other two P2s — both closed, and both checked rather than accepted]** `match_indices` (`rule.rs:189`): for a single occurrence it evaluates the identical predicate at the identical index as the old `find`, so single-marker behaviour is unchanged and the change is strictly a superset; `explicit_marker_used_twice_holds_both_named_keys` binds the multi-marker case. Calendar validation (`mod.rs:186-219`): correct on inspection — the Gregorian rule `(y%4==0 && y%100!=0) || y%400==0`, 31/30/29/28 month lengths, `(1..=max_day)` rejecting day 0, and `month` bounded to `1..=12` *before* the match, which is what makes the `unreachable!` arm genuinely unreachable rather than a latent panic. Slicing `s[0..4]`/`s[5..7]`/`s[8..10]` cannot split a char boundary because all ten bytes are proven ASCII first. The test covers all four boundary classes (impossible month, short month, non-leap Feb 29, and both century rules: 1900 rejected, 2000 accepted).

**Findings summary: P0 = 0, P1 = 0, P2 = 2.**

### FOLLOW-UP ISSUE

- **Inventory what the legacy `顯式取代` phrasing actually matches on today's ledger.** `names_explicit_supersession` requires the marker to be followed, after only `[' ', ':', '\t', '　']`, by the target key. Live data, read with `edda ask` this round: `review.watcher`'s reason writes 「顯式取代同日 review.merge-gate 裡…」 — an intervening 「同日」 — so it is **not** recognised, while `rule.rs:168-175` says the legacy phrasings are "recognised so existing decisions need no migration". Two caveats keep this out of the PR: under-detection is the safe direction for a change whose whole purpose is removing over-detection, and in this specific row non-recognition is arguably *correct*, since the decision supersedes a clause of `review.merge-gate` rather than the decision. Also genuinely reassuring in the same data: `review.merge-gate`'s own reason contains 「顯式取代 2026-09-03 的 Independent-Review 必要 status 裁定」 — a marker followed by a date, not a key — and the adjacency check correctly refuses to hold anything on it, which is live evidence the marker set produces no false positive. Worth an issue to enumerate the live `顯式取代`/`SUPERSEDES` rows and confirm the adjacency rule's coverage is what was intended. Out of scope here: `SUPERSEDES_MARKERS` and the adjacency check are unchanged since `b9eea103`, so this was observable last round.
- **Round 1's two follow-ups stand unchanged** and should carry into the issue close-out: the residual that an *unbounded* sweep still shows `d-033`/`d-034` as ratify candidates before the threshold refuses (the amendment accepts this explicitly, and the PR's own negative control asserts it honestly), and the `superseded-by-same-key` reachability note now also touching `cli.md` per the [P2] above.
- **Process note, not blocking:** Round 1's P0-1 was a controller surface error, not a lane violation — the brief excluded `docs/reference/cli.md` while the change added three clap flags. Any future brief whose diff adds a clap flag should include that file and `scripts/check-cli-docs.sh` in the L0 line.

### RAN vs READ

RAN (all read-only; no checkout, no worktree, no cargo, no build):
- REVIEW.md §3 classifier, extracted by its two marker lines from `501d308:REVIEW.md` and fed the PR's file list → `classes=code-plain docs`, `canonical_class=code-risk`
- U1/U2/U3/U4 enumerators; D1/D3/D4; C2/C3/C4/C5; R1/R2 — three-dot for content diffs, two-dot for `git log` (see the `spec:` note)
- `sh scripts/lint-markdown-content.sh` → exit=0 (U5)
- `sh scripts/wiring-scan.sh origin/main 501d308…` → 4 new fields, 13 swallow candidates, none on a write path
- `edda ask` for `fleet.lane-launch`, `review.readonly-proof`, `review.merge-gate`, `review.default-path`, `review.watcher`, `review.shell-branch`, `review.auto-merge` (D2, and the fixture-fidelity check above)
- U6 subject record: `edda --version` → `edda 0.5.0 (16f7b4b41cde-dirty 2026-09-09)` against workspace `version = "0.6.0"` at base `bb60501`. The installed runtime is stale, which is exactly why every flag claim here is routed as a **source** claim (`mod.rs:56-70`) and to CI's freshly built binary, never to my local `edda`
- Source reads at both SHAs: `bb60501:rule.rs:106-133` and `bb60501:mod.rs:205-245` (the base engine, for the falsification counterfactual); `501d308` `rule.rs`, `mod.rs`, `tests.rs`, `ratify_exit_codes.rs`, `docs/reference/cli.md`
- Byte-comparison of the two `review.` domain-guard tests, `b9eea103` vs `501d308` → identical
- `git diff --name-only bb60501..5d71458` → base advance touches no `crates/**` and none of this PR's files

READ:
- L1 receipt: **CI run 34352040457 @ 501d3088c48170c7deed08446993462d7ef72d52** — exact head, `completed / success`. Format ✅, Clippy ubuntu/macos/windows ✅, Test ubuntu ✅ (incl. the CLI-docs step that was Round 1's P0), Test macos ✅, Test windows ✅ (14m10s), MSRV 1.91.0 ✅, CI Gate ✅. `Test (windows-latest)` was still in progress when this round was dispatched; I waited for its final result rather than posting against a partial run.
- Covers: fmt, clippy on all three OSes, `cargo test --workspace` on Linux and macOS, the 7-crate Windows subset, MSRV compile, and the `check-cli-docs` content gate. Does not cover: the 16 non-subset crates' Windows test targets — irrelevant here, since the only touched crate is `edda-cli`/package `edda`, which the Windows job runs directly.
- No Cargo gate RAN by me: workspace gates are READ from the run above per `ladder` L1, and C5 owed nothing on this SHA. No reason to distrust the run, so no rerun.
- Round 1's verdict comment and the `Review Response: Round 1` (including its controller footer marking the falsification output and local gate results as the lane's, not re-run) — read before anything was RAN, as a replacement session must.

### Verdict

LGTM (P0=0, P1=0)  —  Round 1's two P0s and its P1 are closed and were verified at the source rather than accepted on report: exact-head CI is fully green including the CLI-docs content gate, the corrected `cited-authority` prose matches the shipped engine reason-for-reason, and the repaired acceptance fixture is non-vacuous for all five held keys — each is named by a strictly later reason and so would Hold on the base engine, while all five reach Ratify at head. The amended doneWhen bullet 5 is delivered on all three parts, making `Closes #1066` honest. Two P2s remain (a fixture provenance comment narrower than what it now disclaims; a documented hold reason the product path cannot emit) and neither blocks.

PR #1101, opened 2026-09-09
docs(governance): land the rule-based merge gate, reviewer isolation and the plain flow (GH-1064)

fagemx:
## Problem and change

`.claude/CLAUDE.md:430` said "Merge still requires explicit operator
authority" and is loaded verbatim into every session. The ledger has said
otherwise since 2026-09-03 (`review.auto-merge`) and more precisely since
2026-09-07 (`review.merge-gate`), but the loaded carriers never caught up —
sessions kept switching the shared checkout onto a feature branch, not
handing review to another session, and returning merge to the operator
(measured 2026-09-07, cited in the issue).

The issue's own predicted surface and "~90 echoes" count are stale (three
PRs merged today: #1095 GH-966 rewrote `docs/fleet/rules.md` R6, #1097
touched `REVIEW.md`, #1098 GH-1066 touched `crates/edda-cli`'s ratify
command). Per the dispatching brief, I re-ran the issue's own doneWhen grep
at head (`7d5bb837ed4a61ce3e287e4c71426c8cecdca034`) before planning:

```
git grep -nE "operator authority|operator's action|Merge still|manual decision|never the reviewer" -- .claude AGENTS.md REVIEW.md docs README.md
```

→ **13 hits across 9 files**, matching the brief's re-derived table exactly:
`REVIEW.md`(3), `docs/guides/multi-agent.md`(3), `.claude/CLAUDE.md`(1),
`.claude/skills/issue-pipeline/SKILL.md`(1), `.claude/skills/pr-check/SKILL.md`(1),
`AGENTS.md`(1), `README.md`(1), `docs/guides/brief-template.md`(1),
`docs/superpowers/plans/2026-08-17-scheduler-launch-manifest.md`(1).

Some of the ground was already covered: `docs/fleet/rules.md` R6 was
rewritten this morning (PR #1095) to state the real merge gate — I read it
and made **no change**; it already matches. Canary
`tests/canaries/docs-skills/c4-merge-authority-contradiction/expected.md`
is **also already updated** to the new constitution (it explicitly scores
"claims the PR must wait for operator authorization" as a **false
positive** now) — G0-a's canary work has landed, contrary to the
"if it is not yet updated" contingency in my brief.

## Ground truth verified (not trusted from prose)

```
$ gh api repos/fagemx/edda/rulesets/18852689
{"id":18852689,"name":"Protect main", ...,
 "rules":[{"type":"deletion"},{"type":"non_fast_forward"},
   {"type":"pull_request", ...},
[...]

Issue #1100, opened 2026-09-09
fix(fleet): merge-reviewed-pr.sh lets GitHub choose the squash subject — a single-commit PR writes its branch commit subject into main forever


fagemx:
Evidence: `main` now carries `fa0d011e284e807322b75497f7cf838b06a7ec10` with the subject

```
wip(review): GH-1003 + GH-992 in progress — uncommitted lane work, preserved (#1097)
```

That commit is finished, reviewed (LGTM P0=0 P1=0), CI-green work closing two issues. Its subject says the opposite, permanently, because of how the squash subject is chosen.

## 量測

`scripts/merge-reviewed-pr.sh:143`:

```sh
gh pr merge "$pr" --repo "$repo" --squash --match-head-commit "$head"
```

No `--subject`, no `--body-file`. So GitHub picks both:

| PR shape | squash subject GitHub picks |
|---|---|
| two or more commits | the **PR title** — conventional, reviewed, what everyone assumes happens |
| exactly one commit | that **commit's own subject**, verbatim |

The second row is why this has been latent. Most PRs here carry several commits, so the PR title wins and the convention appears to hold. `git log origin/main --format=%s -300 | grep -c '^wip'` returns **1** — today's, the first single-commit PR whose branch subject was not already conventional.

The same call also lets GitHub compose the body, so a merge through the script carries no gate receipt, no window-check record, and no routing notes unless the merging session bypasses the script and calls `gh pr merge` by hand — which is what happened for the last several merges precisely because the script does not take a body.

## Why it matters

`.claude/CLAUDE.md` fixes the commit format (`<type>(<scope>): <description>`, types `feat|fix|docs|refactor|test|chore`) and `REVIEW.md` enforces it on the PR's commits. Nothing enforces it on the one commit that actually lands on `main`, and that is the only one that survives the squash. `git log --oneline` is the artifact people read a year later; the PR body is not.

`wip` is not in the allowed type list, so today's commit is also a convention violation that every gate passed.

This is not recoverable after the fact: R7 forbids rewriting `main`, so `fa0d011` stays as it is. The fix is preventive only.

## Suspected surface

`scripts/merge-reviewed-pr.sh` (the `gh pr merge` call, and where the subject and body come from)
possibly `crates/edda-cli/src/cmd_review/` if the merge step is migrating to a verb per `mechanism.shell-role`
`docs/guides/operator-runbook.md` (the merge step's description)

## doneWhen

[...]

Issue #1064, opened 2026-09-07
docs(governance): CLAUDE.md:430 'Merge still requires explicit operator authority' and its ~90 echoes — land the rule-based merge gate, reviewer isolation and the plain flow in every loaded carrier (G0-b of #1032)


fagemx:
## Why

`.claude/CLAUDE.md:430` is loaded verbatim into every session under 「These instructions OVERRIDE any default behavior」 and says:

> Merge still requires explicit operator authority.

The ledger says otherwise and has since 2026-09-03: `review.auto-merge` (ratified) authorizes merge by the rule gate — current-head LGTM with P0=0/P1=0, `CI Gate` green, empty `git diff <sha>..origin/main` window — and the 2026-09-05 standing authorization 「LGTM 就合併」 is the operator's own words. `REVIEW.md:767-768` (§8) still says 「Merge is the operator's action, never the reviewer's」, `REVIEW.md:106` still tells the reviewer to `gh pr checkout "$N"` in whatever checkout it is standing in, `REVIEW.md:56` names `scripts/review-pr.sh` as the runner, `docs/fleet/rules.md:22`, `AGENTS.md:73-74`, and the `pr-check` (「Merging is a manual decision」), `pull-request`, `edda-release`, `pr-review-loop`, `issue-pipeline` skills all echo the old regime. #1032's Layer 2 table counts about 90 such clauses.

Consequence measured 2026-09-07: the operator has stated the plain flow repeatedly; fresh sessions still (a) switch the shared checkout `C:\ai_agent\edda` onto a feature branch (it sat on `fix/1047-reconcile-lock-hold` while three other sessions read it as `main`), (b) do not hand review to another session or a subagent, (c) return the merge to the operator. All three are what the loaded carriers instruct. Nothing in a loaded carrier says: a session that edits works in its own worktree and the shared checkout stays on `main`; a reviewer fetches the PR head into a private ref and never checks out in a shared checkout; review is done by a session or subagent that is not the implementer; LGTM plus green CI plus the window check merges. `governance.landing-definition` (binding): a rule is landed only when every carrier agrees — the plain flow has not landed. This issue is the docs/skills half; G0-a (injection layer, `render.rs`) is the product half.

## Predicted surface

- `.claude/CLAUDE.md` — the 「PR review-fix loop」 section (`:411-431`) and the ladder's L3 row (`:183`); add the isolation rule and the reviewer's code-access rule
- `AGENTS.md` (`:73-74`)
[...]

PR #1102, opened 2026-09-09
fix(fleet): a review round posted by a subagent reaches no one (GH-993)

### LINKED ISSUE #993: fix(fleet): a review round posted by a subagent reaches no one — four PRs carry a Review Response answering a round that was never published, and the check that catches it runs nowhere blocking
## What happened

Four PRs — **#974, #976, #980, #981** — each carry a `## Review Response: Round 1` comment (all posted `2026-09-06T06:19Z`) and **no `## Code Review: Round 1`**. The `issues/<n>/timeline` API confirms the round comment was never posted, not posted-and-deleted.

So each of those PRs shows an implementer answering findings labelled `f1`–`f4` plus an escalation that **no one reading the PR can see**. The verdict, the P0/P1 counts, and the reviewed SHA exist nowhere on the PR record.

`.claude/CLAUDE.md` states the contract those responses violate in terms:

> Internal verifier reports, task receipts, and CI do not replace PR comments.

## The detector already exists, is correct, and nothing runs it at a blocking moment

`scripts/fleet/verdict-drift.sh` detects exactly this. Run live at the time of filing:

```text
$ sh scripts/fleet/verdict-drift.sh ; echo "rc=$?"
#991 55b7af9c0e59 main no verdict on head
#990 747e444dffb7 main no verdict on head
#988 dd10481c34c4 main no verdict on head
#981 210bc126168a main no verdict on head
#980 5252f12008a6 main no verdict on head
#976 2f53a29299a4 main no verdict on head
#974 331c3057eb3c main no verdict on head
#972 a07e8816e5e1 main no verdict on head
rc=1
```

Eight of eight open PRs, correct verdict, correct non-zero exit. `docs/guides/pi-controller-runbook.md:128` already makes running it mandatory:

> 宣稱任何一張 PR 完成之前先跑 `sh scripts/fleet/verdict-drift.sh`；它 exit 1 就沒有一張 PR 可以被稱作完成。

**This is therefore not a missing mechanism. It is a correct, documented, mandatory check that nothing executes.** Its only automated caller is `scripts/fleet/daily-digest.sh:162`, which reports the state once a day rather than blocking on it.

## Why the two review transports differ

There are two ways a round gets reviewed, and only one publishes:

| transport | publication |
|---|---|
| lane — `scripts/review-pr.sh` writes the brief, the lane emits the report between `<<<VERDICT` / `VERDICT>>>`, `scripts/pr-review-watch.sh:725-783` extracts it, posts it with `gh pr comment`, then writes `Independent Review` | automatic |
| in-process subagent — the reviewer returns its report to the controller in conversation | **none; depends on the controller remembering** |

R23 (`docs/fleet/rules.md:31`) governs the *shape* of a published verdict and constrains the script that publishes one. Nothing requires that a completed round *be* published when the transport has no publishing script. The four PRs are what that gap produces.

## A second, smaller finding: reading this check by hand is itself fragile

While gathering the evidence above I ran it as `sh scripts/fleet/verdict-drift.sh | head -20; echo "rc=$?"` and read **`rc=0`** — because in POSIX `sh`, `$?` after `a | b` is `b`'s status. I was one step from filing a false fail-open bug against a script that is behaving correctly. This is the same defect shape the fleet scripts keep hitting, and it argues that "the operator runs it and reads the exit code" is not a durable enforcement point.

## Done when

- A completed review round reaches the PR regardless of transport, or the transport is refused. Concretely, at least one of:
  - the subagent/in-process review brief is generated by a script that makes posting the `## Code Review: Round N` comment part of the round, the way the lane transport does; and/or
  - a check refuses a `Review Response: Round N` on a PR carrying no `Code Review: Round N`.
- `verdict-drift.sh` runs at a moment that blocks, not only inside the daily digest — pick the point deliberately and say why in the PR.
- The record for #974, #976, #980, #981 is repaired: each gets a published round pinned to its current head. (In flight at filing time; the rounds are being posted as `Round 2`, each recording that Round 1's report is absent from the PR record.)
- Evidence: `verdict-drift.sh` exits 0 across the open set, from a caller that cannot mask the exit code behind a pipeline.

## Predicted surface

`scripts/pr-review-watch.sh`, `scripts/review-pr.sh`, `scripts/fleet/verdict-drift.sh`, `docs/fleet/rules.md` (R23 or a new rule), `docs/guides/pi-controller-runbook.md`.

## Class and review

`scripts/**` plus `docs/fleet/rules.md`. R22 puts a rules.md change on the 審判面 and the scripts on the 內部工具面 — state the surface in the PR.

## Notes

Found while clearing this workstation's open-PR backlog on 2026-09-06. Related: #914 (readiness signal), #917 (verdict publication shape), #887 (SHADOW rounds), #958.

fagemx:
## What GH-993 asked for, and what has changed since it was filed

GH-993 was filed 2026-09-06. Verified against `origin/main` before starting (controller-supplied fact block, confirmed independently below):

- **Two thirds of the predicted surface is gone.** `scripts/pr-review-watch.sh`, `scripts/review-pr.sh`, and `docs/guides/pr-review-watcher.md` were deleted by PR #1082 (GH-1061). Confirmed with `git ls-tree -r origin/main --name-only` (not `cat-file`): none of the three appear in the tree.
- **One doneWhen bullet is already delivered.** 「The record for #974, #976, #980, #981 is repaired」 — confirmed independently via `gh pr view` / `gh api .../comments`: all four are `MERGED`, and each now carries exactly **two** published `## Code Review: Round N` comments (the repair skips straight to a properly-paired Round 2 rather than backfilling Round 1 — see the verdict-drift.sh comment below for why that shape matters). Not redone here; no rounds posted to merged PRs.
- **`edda review deliver` (GH-1030) landed** and is now the `review:*` label / `Independent Review` status writer, reading §7 verdict comments on a SHA. It does not close this gap — it can only deliver a comment that already exists on the PR. `scripts/merge-reviewed-pr.sh` already calls it (pre-existing); this PR does not change that call.

## What was still live, and what this PR does about it

`scripts/fleet/verdict-drift.sh` already detected "no verdict on head" correctly (GH-914/958) but had exactly one caller, `daily-digest.sh:168`, which captures the exit code and reports it without blocking — by design, since a digest that refused to post because the fleet wasn't perfect would be less useful than the report itself. Nothing made *posting* part of a completed round, and nothing ran the check at a moment that blocks.

**doneWhen bullet 1** offers a choice between a generator (make posting automatic) and a refusal check (catch the orphan after the fact). I chose the **refusal check**: reviews are dispatched as subagents now, in-process, with no single "review shell" script left to hook a "post automatically" step into (GH-1061 retired the last one) — a generator would need a new home in every dispatch path, which is out of scope for a `scripts/**`-only fix lane and speculative about what that home should be. A refusal check is cheap, has one home (verdict-drift.sh, which already reads every open PR's comments), and catches exactly the observed defect.

**`scripts/fleet/verdict-drift.sh`** now also flags a `## Review Response: Round N` comment that answers a round never posted — the precise shape #974/#976/#980/#981 shipped, confirmed against the real PR #974 comment history (`gh api repos/fagemx/edda/issues/974/comments`, fetched during this work). Matching is by round number, not SHA, and only the *newest* response is judged — so a PR that recovers by moving straight to a properly-paired later round (the repair shape above: Round 1's response stays permanently unanswered, Round 2 is posted and paired) reads clean rather than being permanently blocked by its own repair. New annotation: `orphan-response=Round-<N>`, holds the PR the same way `no verdict on head` does. See the doc comment at the top of the script for the full contract.

**doneWhen bullet 2** — where verdict-drift.sh runs at a moment that blocks. I chose **`scripts/merge-reviewed-pr.sh`**, run before both `--check` and `--merge`, over the two other candidates the brief named:
- *A CI job* would need to run the same whole-open-set check on every push to every PR, turning one PR's drift into every other PR's red CI, and would need branch-protection admin action to become "required" — out of a `scripts/**`-only PR's reach, and expensive (repeats the same fleet-wide `gh` reads on every push instead of once per merge attempt).
- *A lefthook `pre-push` hook* runs on every local push to every branch, unrelated to whether that push is "declaring a PR done" — it would block routine WIP pushes over drift on some unrelated PR.
- `scripts/merge-reviewed-pr.sh` is specifically the "declare this PR done" entrypoint — and `docs/guides/pi-controller-runbook.md:125` **already** names running `verdict-drift.sh` before declaring any PR complete as mandatory policy: 「宣稱任何一張 PR 完成之前先跑 `sh scripts/fleet/verdict-drift.sh`；它 exit 1 就沒有一張 PR 可以被稱作完成。」 This PR turns that manually-run rule into an enforced one, at the one entrypoint that already means what the rule requires — with no branch-protection config needed and no cost added to routine pushes.

The check is deliberately **whole-open-set**, not narrowed to the one PR being merged: that is the scope `daily-digest.sh` already shares with `verdict-drift.sh` via `EDDA_OPEN_PR_LIMIT` (GH-958), and it is what the runbook rule above actually asks for ("before declaring *any* PR done"). See "Known limitations" below for what that scope costs in practice — recorded by Round 1 review, not a change to the choice.

**doneWhen bullet 4** — "from a caller that cannot mask the exit code behind a pipeline," read literally per the brief: the new caller reads `drift_rc` directly from `$?` immediately after the substitution, nothing pipes into or out of it (unlike the near-miss the issue's own author records: `verdict-drift.sh | head -20; echo $?`, which reads `head`'s exit code). **Every nonzero exit refuses** — not only exit 1 (drift found) but exit 2 (verdict-drift.sh itself could not read PR state). Proven, not just asserted: `scripts/test-merge-reviewed-pr.sh` case 7 sets `STUB_DRIFT_LIST_FAIL=1` (verdict-drift.sh's own `gh pr list` fails) and asserts the merge is refused — the fail-closed case the brief calls out as required, distinct from case 6 (drift found on an unrelated PR, also refused, also before the union gate or required checks are even reached).

**"Vacuous evidence" callout** — the brief warned that with zero open PRs, "verdict-drift.sh exits 0 across the open set" would be vacuously true, and asked for a fixture instead of relying on that. Two things: `scripts/test-merge-reviewed-pr.sh` case 8 is that fixture (a populated, non-empty open-PR set with a real LGTM verdict, proving the gate doesn't false-positive on a healthy fleet). Separately, by the time this PR was first ready, the live repo was no longer vacuous — `sh scripts/fleet/verdict-drift.sh` against `origin/main`:
```
#1101 d4252ad27f3a main Changes Requested
rc=0
```
One real open PR, genuinely non-vacuous, exit 0. (That snapshot predates this PR itself being open; Round 1 review re-ran the same command and got `rc=1` — see "Known limitations" below. The offline fixture, not this line, is the durable evidence.)

## Round 1 fix: comment author trust (P1)

Round 1 review found the gate this PR installs decided on comment text read with **no author filter**, on a repo that is **PUBLIC**: `scripts/fleet/verdict-drift.sh` selected `## Code Review: Round N` and `## Review Response: Round N` straight off `.comments[]`, while `scripts/merge-reviewed-pr.sh:69-74` already filters the very same comment class on `author_association` five lines below its own call into that script. Concretely, that gap let any commenter either *hold* the merge entrypoint (post any `## Review Response: Round N`, no SHA or matching PR number required — every `--check`/`--merge` in the fleet then refuses) or *silence* it (post a §7-shaped `## Code Review: Round N` pinned to head, waving through the exact "the round never reached the PR" case this script exists to catch).

Fixed by applying the identical trusted set `merge-reviewed-pr.sh` already uses — `OWNER`, `MEMBER`, `COLLABORATOR` — via one jq `select` in `verdict-drift.sh`, mirroring `merge-reviewed-pr.sh:72` exactly, so the two scripts share one definition of trust instead of two. Fails closed by construction rather than by a bolted-on check: jq's `==`/`!=` never errors across mismatched types (verified directly against jq before relying on it), so a missing, `null`, or wrong-shaped `authorAssociation` is simply unequal to every trusted value and the comment is excluded — the same `empty` a comment matching neither heading already produces. Three new fixtures (`test-verdict-drift.sh` cases 16-18) prove the fix itself, not just the happy path: an untrusted verdict does not silence the check, an untrusted Review Response does not hold the PR, and a comment whose `authorAssociation` key is entirely absent fails closed the same way as an untrusted one. Every pre-existing comment fixture across the three affected test files (`test-verdict-drift.sh`, `test-merge-reviewed-pr.sh` case 8, `test-daily-digest.sh` case 3d) gained a trusted `authorAssociation` so the happy path keeps proving exactly what it proved before.

The product-side counterpart of this same gap is filed separately as **#1103**: `edda review deliver`'s `Comment` struct carries no author at all, and it is what writes the `Independent Review` status. Not fixed here. If #1103 lands with a different trusted set than OWNER/MEMBER/COLLABORATOR, the two should be reconciled rather than left disagreeing.

## Known limitations (Round 1 review, recorded here rather than fixed)

Round 1 review upheld `scripts/merge-reviewed-pr.sh` as the blocking point and did not ask for the whole-open-set scope to change — the runbook rule at `docs/guides/pi-controller-runbook.md:125` (corrected above from a stale `:128`, inherited from when #993 was filed) reads literally as "before declaring *any* PR done." But it measured two facts that qualify what that wiring achieves in practice, and asked that both be recorded:

- **`rc=1` is the *ordinary* state of a parallel fleet, not the exceptional one.** Live during Round 1 review, with only two open PRs (#1101, #1102 itself) both mid-loop: `sh scripts/fleet/verdict-drift.sh` printed `no verdict on head` / `stale from …` for both, `rc=1`. Under this fleet's own operating model (parallel lanes, WIP cap 3), a PR sits at `no verdict on head` from opening until its first round and returns to `stale from …` after every fix push — so a fully LGTM'd, CI-green PR can be refused by an unrelated PR's mid-loop state, and the blocked party is not the party who caused it. **What the operator does about it today:** run `sh scripts/fleet/verdict-drift.sh` to see which open PR(s) are holding the set; there is no per-PR override in this PR. That is deliberate — narrowing the scope to one PR would stop being the check `daily-digest.sh` shares with this script (GH-958) — but it means the steady state of a healthy, busy fleet is "refuse," not "clean."
- **Issue #1100 (open) is a prerequisite for this gate to bind in practice.** It records, in the repo's own words, that recent merges bypassed `merge-reviewed-pr.sh` specifically because the script takes no merge subject/body, so merges continued to go through `gh pr merge` by hand — a path this PR's gate does not sit on. #1100 should be linked from #993 so that dependency stays visible; this PR does not close it.

Neither is a doneWhen miss — bullet 2 asks for a deliberate choice with a stated reason, and gets one — so neither blocks. Whether to narrow the scan from whole-open-set to the PR being merged plus a *reported* fleet state is a separate design decision for the controller, not reopened here.

## The rules.md sentence (not in this diff)

GH-1064 is live across `docs/fleet/rules.md`, `docs/guides/**`, `REVIEW.md`, `.claude/**`, `AGENTS.md`, `README.md` this wave — my brief marks all of it off-limits, even for a one-word touch. The proposed rule is below for the controller to land after GH-1064 merges; it is not part of this diff.

```
- **R30 判決要送達，不是完成**：一輪審查完成不等於它「存在」於 PR 上——存在的定義是 PR 留言區有一則 `## Code Review: Round <N>`（R23 形狀）。子代理／in-process 審查把報告吐在對話裡，不會自動貼上 PR；沒有腳本幫它貼，這一輪就只活在那次對話裡，PR 上什麼都沒有。內部驗證報告、任務收據、CI 都不能替代 PR 留言（`.claude/CLAUDE.md`）。`## Review Response: Round <N>` 貼在一張沒有對應 `Code Review: Round <N>` 的 PR 上是這個缺口的症狀，不是巧合——`scripts/fleet/verdict-drift.sh`（GH-993）把它標成 `orphan-response=Round-<N>` 並判為未就緒（比對只看輪數不看 SHA，且只判最新一則回應：用改編號補救的 PR，如 #974/#976/#980/#981，補上配對的 Round 2 之後即視為乾淨，不會因為 Round 1 永遠沒人補而卡死；且只信任 OWNER/MEMBER/COLLABORATOR 留言者，repo 是 PUBLIC，見 #993 Round 1 review）。宣稱／合併任何一張 PR 之前，`scripts/merge-reviewed-pr.sh` 現在會先掃過整個 open set 跑 `verdict-drift.sh`，非零退出一律拒絕——不論是找到 drift（exit 1）還是這個檢查本身讀不到（exit 2）；把 `docs/guides/pi-controller-runbook.md:125` 原本要操作者手動記得的規則，變成這個入口自動擋住。來源：#993、R23、R24。
```

## Surface

`scripts/fleet/daily-digest.sh` is named 審判面 in R22 explicitly; `scripts/merge-reviewed-pr.sh` also qualifies independently under R22's catch-all ("以及任何會貼 status 或執行合併的腳本" — it calls `gh pr merge`). So this PR is **審判面**, via two paths. It does not touch `.github/**` or `lefthook.yml`, so it is not 閘門面.

Touched:
- `scripts/fleet/verdict-drift.sh` — orphan-response detection, plus the Round 1 author-trust filter (內部工具面 on its own, but the PR's surface class is set by the 審判面 files below)
- `scripts/fleet/daily-digest.sh` — doc comment only, no behavior change (審判面, named explicitly in R22)
- `scripts/fleet/test-verdict-drift.sh`, `scripts/fleet/test-daily-digest.sh` — new fixtures (內部工具面, `scripts/test-*.sh`/`scripts/fleet/**` other files)
- `scripts/merge-reviewed-pr.sh` — the new blocking call, plus the `:125` citation fix (審判面, R22 catch-all)
- `scripts/test-merge-reviewed-pr.sh` — new fixtures (內部工具面)

## Gates run

- `sh -n` on every touched script — clean.
- `sh scripts/fleet/test-verdict-drift.sh` — 18/18 cases pass (11 pre-existing + 4 GH-993 + 3 new in Round 1's P1 fix: cases 16-18), run 6× total for determinism.
- `sh scripts/fleet/test-daily-digest.sh` — all cases pass including case 3d, re-run after the Round 1 fixture update.
- `sh scripts/test-merge-reviewed-pr.sh` — 8/8 cases pass (5 pre-existing, unmodified + 3 from GH-993; case 8's fixture updated for the Round 1 trust filter), re-run after the fixture update.
- `sh scripts/lint-file-length.sh --tree` — ran; it is a `*.rs`-only ratchet (per its own header comment) and this PR touches no Rust, so it is a no-op here but was run per the brief.
- No `crates/**` change, so no Cargo gate.
- Exact-head CI: see below.

**Windows-only local finding, worth naming:** while adding the original GH-993 fixtures, the offline test stub's external `jq.exe` (chocolatey install, used only to emulate `gh --jq` for offline testing — the real `gh` filters through an embedded Go jq library, not a spawned binary) intermittently emitted CRLF for some redirected invocations on this box, and POSIX `read`'s last field silently absorbed the trailing `\r` as data. This broke a case that both echoes a field into output and pattern-matches it — the first time either code path had, since the pre-existing fields were only ever compared against fixed strings. Fixed defensively in `verdict-drift.sh` (`tr -d '\r'` on both TSV captures, documented inline) rather than chased further upstream: it's a no-op against clean input and closes the class of bug regardless of root cause.

**CI note per the brief:** CI's `Fleet shell tests (windows)` job carries only the `.ps1` group (per `.github/workflows/ci.yml`); none of the `.sh` tests touched or added here run on Windows in CI. The runs above, on this Windows box, are the only Windows evidence for this change.

**CI flake, established environmental (Round 1 review):** exact-head CI on the pre-fix SHA (`c585457…`, run 34362712884) was red solely from `Test (macos-latest)`, established by Round 1 review as environmental and not caused by this PR — zero Rust in the diff, Ubuntu/Windows/Fleet-shell/Format/MSRV/Clippy×3 all passed on that SHA, base `main` was macOS-green, and the failing test is `simultaneous_reconciles`, the exact mechanism named in issue #1047.

## Mechanized

- Worktree: `C:/ai_agent/edda-wt-gh993`, branch `fix/gh993-review-round-delivery`. Repo main checkout (`C:\ai_agent\edda`) untouched, still on `main`.
- Three commits, each pushed immediately: verdict-drift.sh detection + daily-digest.sh doc note (`c957c00`), merge-reviewed-pr.sh wiring (`c585457`), then the Round 1 P1 author-trust fix (`d7d6197`).

Issue: #993
Closes #993

### DIFF
--- a/scripts/fleet/daily-digest.sh
+++ b/scripts/fleet/daily-digest.sh
@@ -163,6 +163,20 @@ open_rows=$(gh pr list --repo "$EDDA_REPO" --state open --limit "$open_pr_limit"
 # takes the same failure path as a gh failure below. GH_OPEN_JSON is cleared
 # for the subprocess so a test fixture written for this script's own
 # post-jq `gh pr list` output is not re-parsed as raw JSON by the drift check.
+#
+# GH-993: verdict-drift.sh may now also append `orphan-response=Round-<N>`
+# — a Review Response answering a round that was never posted (the failure
+# #974/#976/#980/#981 shipped: a completed review whose report reached no
+# one but the implementer's answer to it). No parsing change was needed
+# here for it: it is one more trailing token on the same line, and the awk
+# pass below already forwards every field past the state ($5..$NF)
+# verbatim into the digest row. This digest still only REPORTS — it has
+# always captured drift_rc without blocking on it, by design, since GH-765:
+# a digest that refused to post because the fleet was imperfect would be
+# less useful than the report itself. The caller that now BLOCKS on this
+# same check is scripts/merge-reviewed-pr.sh, run before any --check or
+# --merge; see the comment above its own call for why that entrypoint and
+# not this one.
 drift_out="$tmp/drift.txt"
 drift_rc=0
 GH_OPEN_JSON= EDDA_OPEN_PR_LIMIT="$open_pr_limit" sh "$self_dir/verdict-drift.sh" >"$drift_out" 2>"$tmp/drift.err" || drift_rc=$?
--- a/scripts/fleet/test-daily-digest.sh
+++ b/scripts/fleet/test-daily-digest.sh
@@ -272,6 +272,25 @@ limits=$(grep -- '--state open' "$GH_STUB_LOG" | sed -n 's/.*--limit \([0-9]*\).
     || fail "case 3c: digest and verdict-drift enumerate different --limit values: $(printf '%s' "$limits" | tr '\n' ' ')"
 unset GH_DRIFT_OPEN_JSON GH_DRIFT_COMMENTS_JSON
 
+# --- case 3d: an orphan Review Response (GH-993) passes through unchanged -----
+# verdict-drift.sh may append `orphan-response=Round-<N>` to a drift line.
+# This proves the digest needs no parsing change for it: it is only more
+# trailing text on the same line, and the awk pass-through above already
+# forwards $5..$NF verbatim into the 擋住什麼 row.
+reset_stubs
+SHA80=8080808080808080808080808080808080808080
+printf '[{"number":80,"title":"Orphan response","mergeStateStatus":"CLEAN","headRefOid":"%s"}]\n' "$SHA80" >"$tmp/open-orphan.json"
+printf '[{"number":80,"headRefOid":"%s","baseRefName":"main","mergeable":"MERGEABLE"}]\n' "$SHA80" >"$tmp/drift-open-orphan.json"
+printf '{"comments":[{"body":"## Review Response: Round 1\\n\\nNew head: %s"}]}\n' "$SHA80" >"$tmp/drift-comments-orphan.json"
+export GH_OPEN_JSON="$tmp/open-orphan.json"
+export GH_DRIFT_OPEN_JSON="$tmp/drift-open-orphan.json"
+export GH_DRIFT_COMMENTS_JSON="$tmp/drift-comments-orphan.json"
+export EDDA_RECAP_FILE="$RECAP_CANNED"
+out=$(run_digest --dry-run 2>&1) || fail 'case 3d: daily-digest.sh exited non-zero'
+printf '%s\n' "$out" | grep -qF -- '- #80 Orphan response — no verdict on head orphan-response=Round-1' \
+    || fail "case 3d: orphan-response annotation missing from 擋住什麼, got: $(printf '%s' "$out" | grep '#80' || true)"
+unset GH_DRIFT_OPEN_JSON GH_DRIFT_COMMENTS_JSON
+
 # --- case 4: board comment with needs-operator lands under 例外 ----------------
 reset_stubs
 printf 'https://github.com/fagemx/edda/issues/613#issuecomment-1\tstatus header\nhttps://github.com/fagemx/edda/issues/613#issuecomment-1\tneeds-operator: relogin gh on 4090\nhttps://github.com/fagemx/edda/issues/613#issuecomment-1\tother details\n' >"$tmp/board.md"
--- a/scripts/fleet/test-daily-digest.sh
+++ b/scripts/fleet/test-daily-digest.sh
@@ -277,11 +277,17 @@ unset GH_DRIFT_OPEN_JSON GH_DRIFT_COMMENTS_JSON
 # This proves the digest needs no parsing change for it: it is only more
 # trailing text on the same line, and the awk pass-through above already
 # forwards $5..$NF verbatim into the 擋住什麼 row.
+#
+# `authorAssociation":"OWNER"` is required since Round 1 review's P1 fix:
+# verdict-drift.sh now trusts only OWNER/MEMBER/COLLABORATOR comments, so an
+# unannotated comment here would read as invisible rather than orphan —
+# see scripts/fleet/test-verdict-drift.sh cases 16-18 for the trust-filter
+# fixtures themselves.
 reset_stubs
 SHA80=8080808080808080808080808080808080808080
 printf '[{"number":80,"title":"Orphan response","mergeStateStatus":"CLEAN","headRefOid":"%s"}]\n' "$SHA80" >"$tmp/open-orphan.json"
 printf '[{"number":80,"headRefOid":"%s","baseRefName":"main","mergeable":"MERGEABLE"}]\n' "$SHA80" >"$tmp/drift-open-orphan.json"
-printf '{"comments":[{"body":"## Review Response: Round 1\\n\\nNew head: %s"}]}\n' "$SHA80" >"$tmp/drift-comments-orphan.json"
+printf '{"comments":[{"body":"## Review Response: Round 1\\n\\nNew head: %s","authorAssociation":"OWNER"}]}\n' "$SHA80" >"$tmp/drift-comments-orphan.json"
 export GH_OPEN_JSON="$tmp/open-orphan.json"
 export GH_DRIFT_OPEN_JSON="$tmp/drift-open-orphan.json"
 export GH_DRIFT_COMMENTS_JSON="$tmp/drift-comments-orphan.json"
--- a/scripts/fleet/test-verdict-drift.sh
+++ b/scripts/fleet/test-verdict-drift.sh
@@ -184,4 +184,56 @@ grep -q 'hit its limit' "$tmp/err" ||
 echo "PASS 11"
 unset GH_ARGV_LOG
 
+# --- case 12: an orphan Review Response holds the PR (GH-993) ------------------
+# The observed defect, as a fixture: PRs #974/#976/#980/#981 each carried a
+# `## Review Response: Round 1` and no `## Code Review: Round 1` at all —
+# real body captured 2026-09-06T06:19:40Z via `gh api .../issues/974/comments`.
+SHA7=7777777777777777777777777777777777777777
+printf '[{"number":12,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA7" >"$tmp/prs.json"
+printf '{"comments":[{"body":"## Review Response: Round 1\\n\\nNew head: `%s`\\n\\n### f1 — fixed"}]}\n' "$SHA7" >"$tmp/comments-12.json"
+run_drift
+expect 12 1 "#12 777777777777 main no verdict on head orphan-response=Round-1"
+
+# --- case 13: a response answering a round that was really posted is clean ----
+# Ordinary Changes-Requested flow: Round 1 posted, pinned to head, the
+# implementer answers it. No orphan annotation. (The base state itself does
+# not hold the PR here — see the existing "else" branch above: a head
+# verdict that resolves Changes Requested is a visible signal, not drift.)
+SHA8=8888888888888888888888888888888888888888
+printf '[{"number":13,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA8" >"$tmp/prs.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #13 @ %s\\n\\n### Verdict\\nChanges Requested, P0=0, P1=1"},{"body":"## Review Response: Round 1\\n\\nNew head: unchanged"}]}\n' "$SHA8" >"$tmp/comments-13.json"
+run_drift
+expect 13 0 "#13 888888888888 main Changes Requested"
+
+# --- case 14: the #974 repair shape — an old orphan does not re-trigger -------
+# Round 1's response is answered by nothing, permanently (the repair chosen
+# for #974/#976/#980/#981 never backfills a Round 1 Code Review — it jumps
+# straight to Round 2, each recording that Round 1's report is absent). Only
+# the NEWEST response is judged, and it is Round 2's, which IS paired — so a
+# PR that recovered this way must read clean, or this check would have
+# permanently blocked the very PRs GH-993's own repair produced. Sequence
+# mirrors the real PR #974 comment order exactly: Response 1, Review 2,
+# Response 2, Review 3 (LGTM, pinned to head).
+SHA9=9999999999999999999999999999999999999999
+SHAMID=bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+printf '[{"number":14,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA9" >"$tmp/prs.json"
+{
+    printf '{"comments":['
+    printf '{"body":"## Review Response: Round 1\\n\\nNew head: %s"},' "$SHAMID"
+    printf '{"body":"## Code Review: Round 2 — PR #14 @ %s\\n\\n### Verdict\\nChanges Requested, P0=0, P1=1"},' "$SHAMID"
+    printf '{"body":"## Review Response: Round 2\\n\\nNew head: %s"},' "$SHA9"
+    printf '{"body":"## Code Review: Round 3 — PR #14 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}' "$SHA9"
+    printf ']}\n'
+} >"$tmp/comments-14.json"
+run_drift
+expect 14 0 "#14 999999999999 main LGTM"
+
+# --- case 15: a Review Response heading not on the first line is not a --------
+# trigger, symmetric with case 6's identical rule for Code Review headings.
+SHA10=cccccccccccccccccccccccccccccccccccccccc
+printf '[{"number":15,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA10" >"$tmp/prs.json"
+printf '{"comments":[{"body":"note first\\n## Review Response: Round 1\\n\\nignored"}]}\n' >"$tmp/comments-15.json"
+run_drift
+expect 15 1 "#15 cccccccccccc main no verdict on head"
+
 echo "verdict-drift fixtures passed"
--- a/scripts/fleet/test-verdict-drift.sh
+++ b/scripts/fleet/test-verdict-drift.sh
@@ -92,13 +92,13 @@ SHAOLD=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 
 # --- case 1: verdict pinned to head, LGTM → exit 0 -----------------------------
 printf '[{"number":1,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA1" >"$tmp/prs.json"
-printf '{"comments":[{"body":"## Code Review: Round 1 — PR #1 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA1" >"$tmp/comments-1.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #1 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)","authorAssociation":"OWNER"}]}\n' "$SHA1" >"$tmp/comments-1.json"
 run_drift
 expect 1 0 "#1 111111111111 main LGTM"
 
 # --- case 2: verdict pinned to an older SHA → stale, exit 1 (#899 regression) --
 printf '[{"number":2,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA2" >"$tmp/prs.json"
-printf '{"comments":[{"body":"## Code Review: Round 1 — PR #2 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHAOLD" >"$tmp/comments-2.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #2 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)","authorAssociation":"OWNER"}]}\n' "$SHAOLD" >"$tmp/comments-2.json"
 run_drift
 expect 2 1 "#2 222222222222 main stale from aaaaaaaaaaaa"
 
@@ -110,7 +110,7 @@ expect 3 1 "#3 333333333333 main no verdict on head"
 
 # --- case 4: only a (SHADOW) verdict on head → SHADOW only, exit 0 --------------
 printf '[{"number":4,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA4" >"$tmp/prs.json"
-printf '{"comments":[{"body":"## Code Review: Round 2 (SHADOW) — PR #4 @ %s\\n- shadow: true\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA4" >"$tmp/comments-4.json"
+printf '{"comments":[{"body":"## Code Review: Round 2 (SHADOW) — PR #4 @ %s\\n- shadow: true\\n\\n### Verdict\\nLGTM (P0=0, P1=0)","authorAssociation":"OWNER"}]}\n' "$SHA4" >"$tmp/comments-4.json"
 run_drift
 expect 4 0 "#4 444444444444 main SHADOW only"
 
@@ -123,7 +123,7 @@ expect 5 1 \
 
 # --- case 6: R23 heading mid-body is not a verdict (the #867 shape) -------------
 printf '[{"number":6,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA6" >"$tmp/prs.json"
-printf '{"comments":[{"body":"narration first\\n## Code Review: Round 3 — PR #867 @ %s\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA6" >"$tmp/comments-6.json"
+printf '{"comments":[{"body":"narration first\\n## Code Review: Round 3 — PR #867 @ %s\\nLGTM (P0=0, P1=0)","authorAssociation":"OWNER"}]}\n' "$SHA6" >"$tmp/comments-6.json"
 run_drift
 expect 6 1 "#6 666666666666 main no verdict on head"
 
@@ -144,7 +144,7 @@ echo "PASS 7"
 # before says ready. R24's third field does not, and used to be left to the
 # digest's DIRTY row — invisible whenever this check ran standalone.
 printf '[{"number":8,"headRefOid":"%s","baseRefName":"main","mergeable":"CONFLICTING"}]\n' "$SHA1" >"$tmp/prs.json"
-printf '{"comments":[{"body":"## Code Review: Round 1 — PR #8 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA1" >"$tmp/comments-8.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #8 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)","authorAssociation":"OWNER"}]}\n' "$SHA1" >"$tmp/comments-8.json"
 run_drift
 expect 8 1 "#8 111111111111 main LGTM mergeable=CONFLICTING"
 
@@ -152,7 +152,7 @@ expect 8 1 "#8 111111111111 main LGTM mergeable=CONFLICTING"
 # GitHub has not computed the merge yet; that is a transient answer, not a
 # verdict, so it is annotated and the exit code stays 0.
 printf '[{"number":9,"headRefOid":"%s","baseRefName":"main","mergeable":"UNKNOWN"}]\n' "$SHA2" >"$tmp/prs.json"
-printf '{"comments":[{"body":"## Code Review: Round 1 — PR #9 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA2" >"$tmp/comments-9.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #9 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)","authorAssociation":"OWNER"}]}\n' "$SHA2" >"$tmp/comments-9.json"
 run_drift
 expect 9 0 "#9 222222222222 main LGTM mergeable=UNKNOWN"
 
@@ -160,7 +160,7 @@ expect 9 0 "#9 222222222222 main LGTM mergeable=UNKNOWN"
 # digest (GH-958). The two scripts hardcoded 200 and 100, so a PR past the
 # digest's 100 got a line here that could never reach a digest row.
 printf '[{"number":10,"headRefOid":"%s","baseRefName":"main","mergeable":"MERGEABLE"}]\n' "$SHA3" >"$tmp/prs.json"
-printf '{"comments":[{"body":"## Code Review: Round 1 — PR #10 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA3" >"$tmp/comments-10.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #10 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)","authorAssociation":"OWNER"}]}\n' "$SHA3" >"$tmp/comments-10.json"
 : >"$tmp/gh-argv.log"
 GH_ARGV_LOG="$tmp/gh-argv.log"
 export GH_ARGV_LOG
@@ -190,7 +190,7 @@ unset GH_ARGV_LOG
 # real body captured 2026-09-06T06:19:40Z via `gh api .../issues/974/comments`.
 SHA7=7777777777777777777777777777777777777777
 printf '[{"number":12,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA7" >"$tmp/prs.json"
-printf '{"comments":[{"body":"## Review Response: Round 1\\n\\nNew head: `%s`\\n\\n### f1 — fixed"}]}\n' "$SHA7" >"$tmp/comments-12.json"
+printf '{"comments":[{"body":"## Review Response: Round 1\\n\\nNew head: `%s`\\n\\n### f1 — fixed","authorAssociation":"OWNER"}]}\n' "$SHA7" >"$tmp/comments-12.json"
 run_drift
 expect 12 1 "#12 777777777777 main no verdict on head orphan-response=Round-1"
 
@@ -201,7 +201,7 @@ expect 12 1 "#12 777777777777 main no verdict on head orphan-response=Round-1"
 # verdict that resolves Changes Requested is a visible signal, not drift.)
 SHA8=8888888888888888888888888888888888888888
 printf '[{"number":13,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA8" >"$tmp/prs.json"
-printf '{"comments":[{"body":"## Code Review: Round 1 — PR #13 @ %s\\n\\n### Verdict\\nChanges Requested, P0=0, P1=1"},{"body":"## Review Response: Round 1\\n\\nNew head: unchanged"}]}\n' "$SHA8" >"$tmp/comments-13.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #13 @ %s\\n\\n### Verdict\\nChanges Requested, P0=0, P1=1","authorAssociation":"OWNER"},{"body":"## Review Response: Round 1\\n\\nNew head: unchanged","authorAssociation":"OWNER"}]}\n' "$SHA8" >"$tmp/comments-13.json"
 run_drift
 expect 13 0 "#13 888888888888 main Changes Requested"
 
@@ -219,10 +219,10 @@ SHAMID=bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
 printf '[{"number":14,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA9" >"$tmp/prs.json"
 {
     printf '{"comments":['
-    printf '{"body":"## Review Response: Round 1\\n\\nNew head: %s"},' "$SHAMID"
-    printf '{"body":"## Code Review: Round 2 — PR #14 @ %s\\n\\n### Verdict\\nChanges Requested, P0=0, P1=1"},' "$SHAMID"
-    printf '{"body":"## Review Response: Round 2\\n\\nNew head: %s"},' "$SHA9"
-    printf '{"body":"## Code Review: Round 3 — PR #14 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}' "$SHA9"
+    printf '{"body":"## Review Response: Round 1\\n\\nNew head: %s","authorAssociation":"OWNER"},' "$SHAMID"
+    printf '{"body":"## Code Review: Round 2 — PR #14 @ %s\\n\\n### Verdict\\nChanges Requested, P0=0, P1=1","authorAssociation":"OWNER"},' "$SHAMID"
+    printf '{"body":"## Review Response: Round 2\\n\\nNew head: %s","authorAssociation":"OWNER"},' "$SHA9"
+    printf '{"body":"## Code Review: Round 3 — PR #14 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)","authorAssociation":"OWNER"}' "$SHA9"
     printf ']}\n'
 } >"$tmp/comments-14.json"
 run_drift
@@ -232,8 +232,47 @@ expect 14 0 "#14 999999999999 main LGTM"
 # trigger, symmetric with case 6's identical rule for Code Review headings.
 SHA10=cccccccccccccccccccccccccccccccccccccccc
 printf '[{"number":15,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA10" >"$tmp/prs.json"
-printf '{"comments":[{"body":"note first\\n## Review Response: Round 1\\n\\nignored"}]}\n' >"$tmp/comments-15.json"
+printf '{"comments":[{"body":"note first\\n## Review Response: Round 1\\n\\nignored","authorAssociation":"OWNER"}]}\n' >"$tmp/comments-15.json"
 run_drift
 expect 15 1 "#15 cccccccccccc main no verdict on head"
+
+# --- case 16: an untrusted-author verdict does not silence the check -----------
+# (Round 1 review, P1). This repo is PUBLIC: without an author filter, a §7-
+# shaped comment from ANY commenter would read as a real LGTM and clear the
+# PR. Byte-identical to case 1's LGTM-on-head shape — only the author trust
+# differs — so this isolates the filter rather than testing a new rule.
+# "NONE" is GitHub's real authorAssociation value for a user with no
+# relationship to the repo.
+SHA11=dddddddddddddddddddddddddddddddddddddddd
+printf '[{"number":16,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA11" >"$tmp/prs.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #16 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)","authorAssociation":"NONE"}]}\n' "$SHA11" >"$tmp/comments-16.json"
+run_drift
+expect 16 1 "#16 dddddddddddd main no verdict on head"
+
+# --- case 17: an untrusted-author Review Response does not hold the PR ---------
+# (Round 1 review, P1) — the other direction of the same finding: on a
+# public repo, anyone could otherwise post any `## Review Response: Round
+# N` (no SHA, no matching PR number required) and refuse every
+# --check/--merge in the fleet. Byte-identical to case 12's shape, only the
+# author trust differs: no orphan-response annotation, because the comment
+# is invisible to the check — not because it was judged and cleared.
+SHA12=eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
+printf '[{"number":17,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA12" >"$tmp/prs.json"
+printf '{"comments":[{"body":"## Review Response: Round 9\\n\\nNew head: `%s`","authorAssociation":"NONE"}]}\n' "$SHA12" >"$tmp/comments-17.json"
+run_drift
+expect 17 1 "#17 eeeeeeeeeeee main no verdict on head"
+
+# --- case 18: a missing authorAssociation fails closed, same as an untrusted --
+# one (Round 1 review, P1) — "absent or unreadable" must not be treated as a
+# verdict either. Distinct from case 16: there the field is present and
+# untrusted; here the key is absent entirely, exercising jq's `null` path
+# (reading `.authorAssociation` off an object that never had the key), which
+# must compare unequal to every trusted value, not error and not default
+# open.
+SHA13=ffffffffffffffffffffffffffffffffffffffff
+printf '[{"number":18,"headRefOid":"%s","baseRefName":"main"}]\n' "$SHA13" >"$tmp/prs.json"
+printf '{"comments":[{"body":"## Code Review: Round 1 — PR #18 @ %s\\n\\n### Verdict\\nLGTM (P0=0, P1=0)"}]}\n' "$SHA13" >"$tmp/comments-18.json"
+run_drift
+expect 18 1 "#18 ffffffffffff main no verdict on head"
 
 echo "verdict-drift fixtures passed"
--- a/scripts/fleet/verdict-drift.sh
+++ b/scripts/fleet/verdict-drift.sh
@@ -7,11 +7,23 @@
 # reports CLEAN with zero verdicts.
 #
 # Output, one line per open PR:
-#   #<n> <head12> <base> <state> [ mergeable=<v> ] [ base=<branch> (...) ]
+#   #<n> <head12> <base> <state> [ mergeable=<v> ] [ base=<branch> (...) ] [ orphan-response=Round-<N> ]
 # with <state> one of:
 #   no verdict on head | stale from <sha12> | SHADOW only | LGTM | Changes Requested
 # The base annotation is appended when the PR's base is not `main`, whose
 # status contexts are the only ones any ruleset enforces.
+#
+# GH-993: `orphan-response=Round-<N>` is appended when the newest
+# `## Review Response: Round N` comment on the PR answers a round that was
+# never posted — a completed review whose report reached no one but the
+# implementer's answer to it (the observed failure: PRs #974/#976/#980/#981
+# each carried a `Review Response: Round 1` and no `Code Review: Round 1`).
+# Matching is by round NUMBER only, not SHA — a response answering an old,
+# superseded round is still answering a round that is really there. Only the
+# NEWEST response is judged, so a PR that recovered by moving straight to a
+# later, properly paired round (the #974 repair shape: Round 1's response
+# stays permanently unanswered, but Round 2 is posted and paired) reads
+# clean. This annotation holds the PR (not_ready=1) independent of <state>.
 #
 # GH-958: R24 names THREE readiness fields, and this check now covers all
 # three — (1) the newest verdict's SHA equals the head, (2) the
@@ -26,10 +38,12 @@
 # digest's 100 got a drift line here that could never reach a digest row —
 # invisible in the one artefact R24 calls the report.
 #
-# Exit 0 every PR carries a verdict on its head and is not CONFLICTING;
-# exit 1 when any PR has `no verdict on head`, is `stale from ...`, or is
-# CONFLICTING; exit 2 when a gh read fails — a check that could not read
-# must never print nothing and exit 0.
+# Exit 0 every PR carries a verdict on its head, is not CONFLICTING, and its
+# newest Review Response (if any) answers a round that was actually posted;
+# exit 1 when any PR has `no verdict on head`, is `stale from ...`, is
+# CONFLICTING, or carries an orphan Review Response (GH-993); exit 2 when a
+# gh read fails — a check that could not read must never print nothing and
+# exit 0.
 # Read-only: no posting, no labels, no merges, no ledger writes.
 set -eu
 
@@ -50,6 +64,17 @@ open_rows=$(gh pr list --repo "$repo" --state open --limit "$open_limit" \
     --json number,headRefOid,baseRefName,mergeable \
     --jq '.[] | [.number, .headRefOid, .baseRefName, .mergeable] | @tsv' \
 ) || fail_read "pr list"
+# Defensive: strip any stray CR (`tr -d`, not a trailing-only trim — none of
+# these fields are free-form text, so a bare CR never belongs in one). Found
+# while adding the GH-993 fields below: an external jq.exe on at least one
+# Windows dev box emits CRLF for redirected output, and `read`'s last
+# variable absorbs a trailing \r as data, silently breaking any comparison
+# or interpolation of that field (`mergeable = CONFLICTING` and the R23
+# fields the same way — this predates GH-993 and just had no field that both
+# echoed into output and drove a comparison until now). Harmless no-op
+# against the real `gh --jq`, which filters through an embedded jq library,
+# not a spawned binary.
+open_rows=$(printf '%s' "$open_rows" | tr -d '\r')
 if [ "$(printf %s "$open_rows" | grep -c .)" -ge "$open_limit" ]; then
     echo "verdict-drift: the open-PR enumeration hit its limit of $open_limit; PRs past it were not examined (raise EDDA_OPEN_PR_LIMIT)" >&2
 fi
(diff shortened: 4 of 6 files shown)


fagemx:
## Code Review: Round 1 — PR #1102 @ c585457767b991152d9867be1b9cb716d5cfd647

- elapsed: unmeasured (controller-spawned subagent; no pi session file to read)
- model_observed: unverified (no pi session JSONL and no `claude -p --output-format json` envelope for this transport)
- spec: review-spec-v1.6
- escalations: none
- shadow: false
- cost: read-only round; no checkout, no worktree, no Cargo. All source read via `git show <sha>:<path>` against the main checkout, which stayed on `main`.

### IN SCOPE

Changed behaviour and paths — six files, all `scripts/**`, no `crates/**`, no docs:

1. `scripts/fleet/verdict-drift.sh` — the jq comment pass becomes tagged (`v` verdict rows gain a round number; new `r` rows carry a first-line `## Review Response: Round N`), a new `orphan-response=Round-<N>` annotation holds the PR (`not_ready=1`), and two `tr -d '\r'` guards are added on the TSV captures.
2. `scripts/merge-reviewed-pr.sh` — a new blocking call to `verdict-drift.sh` before both `--check` and `--merge`; any nonzero exit refuses.
3. `scripts/fleet/daily-digest.sh` — comment only, no behaviour change.
4. Three `test-*.sh` files — new fixtures.

Direct callers/consumers: `daily-digest.sh:182` (pre-existing, still non-blocking by design) and the new `merge-reviewed-pr.sh:52-57`. `docs/guides/operator-runbook.md:175` is the documented merge procedure and names this script.

Issue #993 doneWhen, judged item by item:

- **Bullet 1 (a check refuses a `Review Response: Round N` on a PR carrying no `Code Review: Round N`)** — delivered as the `orphan-response` annotation. Note for the record: for the four PRs that motivated the issue this adds *diagnostic* signal, not new blocking power — the issue's own live output shows #974/#976/#980/#981 already at `no verdict on head` with `rc=1`. The annotation's real value is distinguishing "never reviewed" from "reviewed, report never posted", plus the narrow case of a head verdict present alongside a mismatched response round.
- **Bullet 2 (runs at a moment that blocks; pick the point deliberately and say why)** — delivered; the argument is made and the two rejected alternatives (a CI job, a lefthook `pre-push`) are argued correctly. Assessment of the choice under Findings P2-2.
- **Bullet 3 (repair #974/#976/#980/#981)** — verified independently, already delivered before this diff: all four `MERGED`, each carrying exactly 2 `## Code Review: Round N` comments (`gh api repos/fagemx/edda/issues/<n>/comments`).
- **Bullet 4 (exits 0 across the open set, from a caller that cannot mask the exit code)** — the caller shape is correct and proved; see P2-6 on the evidence's shelf life.

Security/data-loss exposed by the change: the merge entrypoint's refusal now depends on comment text that is read without an author filter, on a public repository — Findings P1-1. This is the one blocking finding.

Current-base integration: base window empty; `origin/main` is still `7d5bb837ed4a61ce3e287e4c71426c8cecdca034`; `mergeable: MERGEABLE`.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 1102 --name-only` → 6 files, all under `scripts/`; the brief's allowed surface. No docs touched, so no collision with the GH-1064 docs lane. |
| U2 | any | P1 | PASS | Body carries `Closes #993` (line 73); commit messages carry none. Every doneWhen item is delivered — bullets 1/2/4 by this diff, bullet 3 verified already delivered (all four PRs MERGED, 2 rounds each). Merging genuinely finishes the issue. |
| U3 | any | P1 | PASS | `Issue: #993` present. |
| U4 | any | P1 | PASS | `git log --format=%s origin/main..$SHA` filtered by the conventional-subject regex → empty. Both subjects are `fix(fleet): …`. |
| U5 | any | P1 | PASS | RAN `sh scripts/lint-markdown-content.sh` → `exit=0`. |
| U6 | any | P0 if deterministic | PASS (red, established environmental) | CI run 34362712884 @ c585457… → **CI Gate = failure**, caused solely by `Test (macos-latest)`. Not deterministic and not PR-caused: the diff contains zero Rust; `Test (ubuntu-latest)`, `Test (windows-latest)`, both `Fleet shell tests` jobs, Format, MSRV and Clippy ×3 all pass on this SHA; base `main` 7d5bb837 was macOS-green (run 34354766231); and the failing test — `cmd_reconcile::tests::runner::simultaneous_reconciles_create_one_attempt_and_release_the_lock`, failing with `workspace is locked by another process` — is the exact test and mechanism named in issue #1047 ("…so simultaneous_reconciles flakes"). Re-run that one job; audited rather than re-run locally per `ladder` L2. |
| U7 | any | P1 | PASS | Whole scoped audit finished before posting; all blocking findings batched in this round. |
| U8 | any | P1 | N.A.(no `Decision:` line in the PR body — the DECISIONS enumerator printed nothing) | — |
| D1–D5 | docs | — | N.A.(class not routed — no `docs/**`, `*.md` or `*.txt` in the diff) | Classifier: `classes=code-risk`. |
| S1–S3 | skills | — | N.A.(class not routed — no `SKILL.md`, `.claude/**`, `skills/**`, `AGENTS.md` or `REVIEW.md` in the diff) | `git diff --name-only $BASE $SHA -- REVIEW.md` empty, so the spec was read at a head copy identical to the base copy. |
| C1 | code-plain | P0 | PASS | Each doneWhen item matched to code and to a named fixture — see IN SCOPE. |
| C2 | code-plain | P0 | PASS | Both commits are `fix:`; each carries regression fixtures. `test-verdict-drift.sh` cases 12–15, `test-merge-reviewed-pr.sh` cases 6–8, `test-daily-digest.sh` case 3d. Falsifying-ness established by reading, below. |
| C3 | code-plain | P1 | N.A.(no `crates/**` in the diff) | — |
| C4 | code-plain | P1 | PASS | Enumerator empty. |
| C5 | code-plain | P1 | N.A.(selector empty — no `crates/**`, so no Cargo gate is owed locally) | RAN the selector; printed nothing. |
| R1 | code-risk | P0 | PASS | Enumerator empty — the diff adds no `rm -rf`, `reset --hard`, `git rm`, `--delete-branch`, `--force` or `Remove-Item`. |
| R2 | code-risk | P0/P1 | PASS | Enumerator empty — no added line mixes `\|\|` and `&&`. The one AND-list in the new code, `drift_out=$(…) \|\| drift_rc=$?`, is a two-operand list with no third term, so no precedence question arises. |
| R3 | code-risk | P0 | PASS | RAN `sh -n` on the **head** content of all six changed scripts (extracted with `git show`, not the working tree, which holds the base version) → exit 0 for each. |
| R4 | code-risk | P1 | PASS | The diff creates no temp directory, file or lock. `verdict-drift.sh` is read-only; `merge-reviewed-pr.sh`'s new block adds one command substitution and no resource. `test-merge-reviewed-pr.sh` keeps its pre-existing `trap cleanup 0 HUP INT TERM`. |
| R5 | code-risk | P1 | **FAIL** | See P1-1. The property "the merge entrypoint's refusal depends only on trusted-author comments" does not hold: `verdict-drift.sh:91-110` applies no author filter, while `merge-reviewed-pr.sh:69-74` filters the same comment class on `author_association`. |

### Wiring

|---|---|---|---|---|
| `orphan-response=Round-<N>` line annotation | `scripts/fleet/verdict-drift.sh:186-189` — appended to the PR's drift line, sets `not_ready=1` | `scripts/fleet/daily-digest.sh:182` via the pre-existing `$5..$NF` awk pass-through (proved by `test-daily-digest.sh` case 3d, which greps the exact 擋住什麼 row); and `scripts/merge-reviewed-pr.sh:52-57` via the exit code | Not swallowed. `not_ready=1` → `exit 1` at `verdict-drift.sh:195-196`; the caller refuses on any nonzero | annotation → output line → process exit code → merge refusal, proved end-to-end by `test-merge-reviewed-pr.sh` case 6 |
| Blocking `verdict-drift.sh` call at the merge entrypoint | `scripts/merge-reviewed-pr.sh:52-53` — `drift_rc=0; drift_out=$(… ) \|\| drift_rc=$?` | its own guard at `:54-57` | **Correct.** `$?` is read from the command substitution itself; nothing pipes into or out of it, so no stage can absorb the status. `-ne 0` (not `-eq 1`) means the error path refuses too | exit 1 covered by case 6, **exit 2 covered by case 7**, clean-and-populated path covered by case 8 |
| `v`/`r` tagged jq row shape (internal to verdict-drift.sh) | `verdict-drift.sh:95-110` | the `while read row_tag f1 f2 f3 f4` loop at `:121-140` | A comment matching neither shape yields `empty`; a malformed `Code Review` heading yields no `v` row and so reads as unpaired — fail-closed, consistent with GH-917 | contained within one script; no external consumer |
| `STUB_DRIFT_PRS` / `STUB_DRIFT_COMMENTS` / `STUB_DRIFT_LIST_FAIL` | `scripts/test-merge-reviewed-pr.sh` gh stub | same file | test-only harness inputs | n/a — no production reader, and none required |

No new `pub` surfaces (`sh scripts/wiring-scan.sh origin/main $SHA` → "no new pub surfaces"; its single swallow-pattern hit is the word "silently" inside a comment, a false positive).

### Findings

- **[P1] The gate this PR installs decides on comment text read without an author filter, on a public repository — so untrusted input can both hold and silence the merge entrypoint.** Property that must hold: *the merge entrypoint's refusal depends only on comments from trusted authors.* It does not. `scripts/fleet/verdict-drift.sh:91-110` selects `## Code Review: Round N` and the new `## Review Response: Round N` straight off `.comments[]` with no author predicate; `scripts/merge-reviewed-pr.sh:52-57` now converts that script's exit code into a hard refusal for `--check` and `--merge` alike. Five lines further down, `scripts/merge-reviewed-pr.sh:69-74` filters the very same comment class on `author_association ∈ {OWNER, MEMBER, COLLABORATOR}` — the repository already treats author trust as load-bearing at this exact entrypoint, and the drift path now bypasses it. `gh repo view fagemx/edda --json visibility` → `visibility=PUBLIC`, so the comment surface is not restricted to collaborators. Both directions matter, and this diff is what makes them consequential: (a) *hold* — a comment whose first line matches `^## Review Response: Round [0-9]+` sets `orphan-response`, `not_ready=1`, exit 1, and every `--check`/`--merge` on the repo refuses; that new trigger is a bare prefix match requiring no SHA and no matching PR number, a materially lower bar than the verdict heading's 40-hex pin; (b) *silence* — a comment carrying the §7 heading pinned to a head SHA makes drift read clean, which waves through precisely the "the round never reached the PR" case doneWhen bullet 1 exists to catch. Before this PR the same unfiltered read fed only a once-a-day report; the refusal is new here, which is why this is in scope as exposed by the change rather than pre-existing. Remedy is small and uses an already-available field: `authorAssociation` is returned by `gh pr view --json comments` (verified — the comment object's keys include `authorAssociation`), so the jq gains one `select`, mirroring `merge-reviewed-pr.sh:72`, plus a fixture asserting an untrusted-author comment neither holds nor clears the gate. — evidence: `scripts/fleet/verdict-drift.sh:91-110`; `scripts/merge-reviewed-pr.sh:52-57`; `scripts/merge-reviewed-pr.sh:69-74`; `gh repo view` → `visibility=PUBLIC`.

- **[P2] The load-bearing citation `docs/guides/pi-controller-runbook.md:128` is off by three lines; the rule is at :125.** The quoted text is verbatim correct and the rule genuinely exists — 「宣稱任何一張 PR 完成之前先跑 `sh scripts/fleet/verdict-drift.sh`；它 exit 1 就沒有一張 PR 可以被稱作完成。」 — so the argument it supports stands. Only the locator is stale, inherited from issue #993 (filed 2026-09-06, when :128 was right). It is wrong at both the base and the head SHA, and it is now baked into a production comment that will outlive this PR. — evidence: `git grep -n "沒有一張 PR 可以被稱作完成" $SHA` → `docs/guides/pi-controller-runbook.md:125`; the `:128` form appears at `scripts/merge-reviewed-pr.sh:36` and twice in the PR body.

- **[P2] The whole-open-set coupling is the right *destination* but its realized binding is weaker than the PR body implies, and its steady state is "refuse".** Judgement on the round's sharpest question, stated plainly. The entrypoint choice is defensible and I do not ask for it to be changed: `docs/guides/operator-runbook.md:175` names this script as the merge procedure, and the whole-set scope is the runbook rule at :125 read literally, not a scope invented here. But three measured facts qualify what the wiring achieves. (i) Live, right now, `sh scripts/fleet/verdict-drift.sh` prints `#1102 … no verdict on head` and `#1101 … stale from d4252ad27f3a` and exits **1** — with only two open PRs, both dirty. Under the fleet's own operating model (parallel lanes, WIP cap 3) `rc=1` is the ordinary state, not the exceptional one, because a PR sits at `no verdict on head` from opening until its first round and returns to `stale from …` after every fix push. So a fully LGTM'd, CI-green PR is refused by an unrelated PR's mid-loop state, and the blocked party is not the party who caused it. (ii) Issue #1100 (OPEN) records in the repo's own words that the last several merges bypassed this script "precisely because the script does not take a body" — a gate whose refusal fires in the normal case, on an entrypoint already being stepped around, makes bypass more attractive, which is the shape of the defect #993 exists to remove, relocated one level. (iii) The PR body's rationale never names that dependency, so it argues the choice without its strongest counter-consideration. None of this is a false claim and none of it is a doneWhen miss — bullet 2 asks for a deliberate choice with a stated reason and gets one — so it does not block. It should be recorded, and it wants a follow-up. — evidence: RAN `sh scripts/fleet/verdict-drift.sh` → 2 lines, `rc=1`; issue #1100 body; `docs/guides/operator-runbook.md:175`.

- **[P2] Round-number membership means a response labelled with any already-posted round reads paired.** `verdict-drift.sh:147-153` tests membership of the newest response's round in the set of all posted rounds. Renumbering *upward*, onto a round that was never posted, is caught — I confirmed the brief's case by reading: with no rounds at all `code_review_rounds` is `","`, so a `Review Response: Round 2` cannot match `*",2,"*` and is flagged. Renumbering *downward*, onto a round number that was posted, reads clean. That is the deliberate consequence of judging only the newest response (case 14's #974 repair shape, and the reasoning for it is sound — the alternative permanently blocks PRs repaired the sanctioned way). Worth stating so the check's contract is not read as stronger than it is: it asserts "the newest response's round number was published", not "the latest round was published". — evidence: `scripts/fleet/verdict-drift.sh:119-153`; `test-verdict-drift.sh` cases 12 and 14.

- **[P2] SHADOW rounds count toward the paired set.** A `Code Review: Round N (SHADOW)` emits a `v` row and appends `N` to `code_review_rounds` (`verdict-drift.sh:95-104,125`), so a response answering only a SHADOW round reads paired — though REVIEW.md §8 says a SHADOW round is never a verdict. Arguably right, since a SHADOW round *is* visible on the PR and this check is about reports reaching the PR, not about verdict authority. Flagging it as a contract question, not a defect. — evidence: `scripts/fleet/verdict-drift.sh:95-104`, `:125`; REVIEW.md §8.

- **[P2] The PR's "no longer vacuous" live evidence cannot reproduce.** The body records `#1101 … Changes Requested / rc=0` as proof the gate does not false-positive on a healthy fleet. That measurement necessarily predates #1102 being open, since an open unreviewed PR is itself `no verdict on head`; the same command today returns `rc=1`. The offline fixture (case 8) is the durable evidence and it is sound; the live line is a snapshot with a short shelf life and should not be read as a standing property. — evidence: PR body vs RAN `sh scripts/fleet/verdict-drift.sh` today.

- **[P2] `tr -d '\r'` in production for a test-stub artifact — judged acceptable.** The brief asked whether this is the right layer and whether it could mask a genuine `\r`. It cannot: every field in both captures is closed-vocabulary — a decimal PR number, a 40-hex SHA, a git branch name (git refname rules forbid ASCII control characters, so CR is not representable), the `mergeable` enum, and jq-captured digits and fixed tags — so a bare CR is never data in any of them, and `tr -d` (rather than a trailing trim) is therefore safe and is correctly justified inline. Against the real `gh`, which filters through an embedded jq library rather than a spawned binary, it is a no-op. Fixing it here rather than in the stub is defensible: it closes the class regardless of root cause, and the failure mode it prevents (a `read` absorbing `\r` into the last field, silently breaking `mergeable = CONFLICTING`) predates GH-993. — evidence: `scripts/fleet/verdict-drift.sh:67-77`, `:112`.

- **[P2] Evidence quality, stated honestly against what I could and could not establish.** I did not execute any fixture (read-only round), so the lane's 15/15 and 8/8 are READ, not RAN. What I established **by reading the assertions** is which of them are genuinely falsifying: `test-merge-reviewed-pr.sh` **case 7** is the fail-closed proof the brief demanded and it bites — `STUB_DRIFT_LIST_FAIL=1` makes the stub's `pr list` exit 1, `verdict-drift.sh:66` routes that to `fail_read` → exit 2, and the case asserts a nonzero exit whose stderr names `verdict-drift` and that `api --paginate` was never reached; an implementation testing `-eq 1` instead of `-ne 0` fails this case. **Case 6** bites likewise, and its "union gate not reached" assertions are meaningful because `run_case` gives every case a fresh `GH_CALLS` file and truncates it. **Case 8** is genuinely non-vacuous: the stub applies the caller's real `--jq` through real `jq`, so the R23 regex is exercised rather than assumed. `test-verdict-drift.sh` cases 12/14/15 are string-exact on both the output line and the exit code. What rests on the lane's report alone: that these all pass on Windows, and the 5–6× determinism re-runs.

### FOLLOW-UP ISSUE

- **Scope and visibility of the fleet-wide merge gate.** Given P2-2's live measurement (`rc=1` as the steady state of a parallel fleet), decide deliberately whether the merge entrypoint should refuse on the whole open set or on the PR being merged plus a *reported* fleet state; and if it stays whole-set, whether the refusal should name which PR caused it in a form the blocked operator can act on. Evidenced by the live run and by #1100's record of bypass. Out of scope here: #993's doneWhen asks for a blocking point and the runbook rule at :125 names this scope, so changing it is a new decision, not a fix to this PR.
- **#1100 is a prerequisite for this gate to bind in practice.** While `merge-reviewed-pr.sh` takes no merge subject or body, merges continue to go through `gh pr merge` by hand and this gate is not on that path. Worth linking #1100 from #993 so the dependency is visible, and worth a line in the PR body naming it.
- **Citation drift in production comments.** `scripts/merge-reviewed-pr.sh:36`'s `:128` is stale at both base and head. A cheap `path:line` freshness check over `scripts/**` comments would catch this class; today only D1 catches it, and only for `docs`-class diffs.

### RAN vs READ

RAN (read-only; the main checkout stayed on `main`, nothing was checked out, built, edited or pushed):
- `git show c585457…:<path>` for all six changed files, `REVIEW.md`, `.github/workflows/ci.yml`, `scripts/fleet/run-fleet-tests.sh`, `docs/guides/pi-controller-runbook.md`, `docs/guides/operator-runbook.md`; `git diff origin/main...c585457…`.
- `git diff --name-only 7d5bb83 c585457 -- REVIEW.md` → empty, so the spec was read at a head copy byte-identical to the base copy (§0 requires the base version).
- `sh -n` on the **head** content of all six changed scripts → exit 0 each (R3).
- `sh scripts/lint-markdown-content.sh` → `exit=0` (U5).
- `sh scripts/wiring-scan.sh origin/main c585457…` → "no new pub surfaces"; one swallow-pattern hit, a comment, false positive (§5.5).
- `sh scripts/fleet/verdict-drift.sh` against the live repo, base version, read-only → `#1102 … no verdict on head` / `#1101 … stale from d4252ad27f3a`, `rc=1` (P2-2). Reason for running it: the coupling question the round turns on is a live-fleet property that no CI job and no offline fixture measures.
- U2 / U3 / U4 / U8 / R1 / R2 / C4 / C5 enumerators — all empty except the U2 `Closes #993` hit and the U3 `Issue:` line, both judged above.
- `gh repo view fagemx/edda --json visibility` → `PUBLIC`; `gh pr view 1101 --json comments --jq '.comments[0]|keys'` → includes `authorAssociation` (both support P1-1).
- `gh pr view <n> --json state` and `gh api repos/fagemx/edda/issues/<n>/comments` for #974/#976/#980/#981 → all `MERGED`, 2 `## Code Review: Round N` comments each (doneWhen bullet 3).
- `edda ask review.merge-gate`, `edda ask fleet.merge-authority` — checked for contradiction and found none: `review.merge-gate=ci-gate-only-independent-review-status-removed` removes `Independent Review` from the ruleset while explicitly keeping the merge script as the place that blocks, which *supports* this PR's choice of entrypoint rather than contradicting it.
- Flake history sweep over the last 12 `ci.yml` runs plus an issue search, for U6.

READ:
- **L1 receipt: CI run 34362712884 @ c585457767b991152d9867be1b9cb716d5cfd647.** Pass: Format; Clippy ubuntu/macos/windows; MSRV (1.91.0); Test (ubuntu-latest); Test (windows-latest); Fleet shell tests; Fleet shell tests (windows); Detect code changes. **Fail: Test (macos-latest) → CI Gate = failure.** Established environmental, not deterministic and not PR-caused (U6 row).
- What that covers: `Fleet shell tests` (ubuntu) runs `scripts/fleet/run-fleet-tests.sh`, whose glob is `scripts/fleet/test-*.sh scripts/test-*.sh` — so all three touched test files, including `scripts/test-merge-reviewed-pr.sh` in the non-`fleet/` directory, do execute there. No CI wiring gap.
- What it does **not** cover: `Fleet shell tests (windows)` runs only `scripts/test-git-config-guard.sh` and the `scripts/fleet/test-*.ps1` glob, so **none of the new or changed `.sh` cases run on Windows in CI**. The only Windows evidence for cases 6–8, 12–15 and 3d is the lane's own local runs, which this read-only round did not reproduce. Since the changed scripts are POSIX `sh` and the CRLF guard is specifically a Windows-shaped fix, that gap is the least-covered part of this diff.
- C5 selector empty (no `crates/**`), so no Cargo gate is owed locally and none was run.

### Verdict

Changes Requested, P0=0, P1=1 — the merge gate this PR installs decides on comments read without an author filter on a public repo, so untrusted text can both hold and silence it, five lines from an author filter the same script already applies (`verdict-drift.sh:91-110` vs `merge-reviewed-pr.sh:69-74`); everything else is sound and the entrypoint choice is upheld.

Two notes for the fix round, neither a finding. The remedy for P1-1 is one jq `select` on `authorAssociation` plus a fixture, not a redesign — the mechanism, the fail-closed call shape and the test set are otherwise good work, and case 7 in particular is exactly the error-path proof that was asked for. Separately, `CI Gate` is red on this SHA from the known `simultaneous_reconciles` macOS flake (#1047); that one job needs a re-run before any merge regardless of this verdict, since `CI Gate` is the ruleset's only required check.

fagemx:
## Review Response: Round 1

New head: `d7d619781a3a6e17cb5b91e99ad67ba79ef998a8`

### P1 — fixed

**Finding:** the gate this PR installs decided on comment text read with no author filter, on a repo that is PUBLIC — `scripts/fleet/verdict-drift.sh:91-110` selected `## Code Review: Round N` / `## Review Response: Round N` straight off `.comments[]`, while `scripts/merge-reviewed-pr.sh:69-74` already filters the same comment class on `author_association` five lines below its own call into that script. Untrusted text could both *hold* the merge entrypoint (any `## Review Response: Round N`, no SHA required) and *silence* it (a faked §7 verdict pinned to head).

**Fix:** one jq `select(.authorAssociation == "OWNER" or .authorAssociation == "MEMBER" or .authorAssociation == "COLLABORATOR")` inserted right after `.comments[]` in `scripts/fleet/verdict-drift.sh`, ahead of both the "v" and "r" row branches — so the trust check gates both row shapes with a single predicate rather than needing to be threaded through the if/elif twice. Mirrors `merge-reviewed-pr.sh:72`'s trusted set exactly (verified: `gh pr view --json comments --jq '.comments[0]|keys'` returns `authorAssociation`, camelCase, confirmed independently against a live comment which returned `"OWNER"` — not the REST `author_association` the union gate reads via a different `gh` subcommand).

**Fail-closed, proved by test, not just asserted:** verified directly against `jq` before relying on it — `select(.a == "OWNER" or …)` over objects with `authorAssociation` missing, `null`, and wrong-typed (a number) all correctly excluded, no error (jq's `==`/`!=` never throws across mismatched types). Three new cases in `scripts/fleet/test-verdict-drift.sh`:
- **case 16** — an untrusted-author (`"NONE"`) §7 verdict pinned to head does not silence the check (byte-identical to case 1's shape, only the author differs) → `no verdict on head`, not `LGTM`.
- **case 17** — an untrusted-author `## Review Response: Round 9` does not hold the PR (byte-identical to case 12's shape) → no `orphan-response` annotation; the comment is invisible, not judged-and-cleared.
- **case 18** — `authorAssociation` key entirely absent (not merely untrusted) on an otherwise-valid head-pinned LGTM → same exclusion as case 16, exercising the "absent" half of "absent or unreadable" specifically, distinct from case 16's "present but untrusted" half.

Every pre-existing comment fixture across the three affected test files gained a trusted `"authorAssociation":"OWNER"` so the happy path continues to prove exactly what it proved before: `test-verdict-drift.sh` cases 1, 2, 4, 6, 8, 9, 10, 12, 13 (×2), 14 (×4), 15; `test-merge-reviewed-pr.sh` case 8's `drift-clean-comments.json`; `test-daily-digest.sh` case 3d's `drift-comments-orphan.json`.

**Related, not fixed here:** #1103 is the product-side counterpart — `edda review deliver`'s `Comment` struct carries no author at all, and it is what writes the `Independent Review` status. Trusted set here is OWNER/MEMBER/COLLABORATOR; if #1103 lands with a different set, the two should be reconciled.

### P2s

- **Citation off by three** — fixed. `scripts/merge-reviewed-pr.sh:36` and both occurrences in the PR body now cite `docs/guides/pi-controller-runbook.md:125` (independently re-verified: `grep -n` on that file, read-only, confirms `:125`). No `docs/**` file touched — the citation lives only in my own script comment and PR body text, both in scope.
- **Whole-open-set coupling weaker than implied** — recorded, not changed. Added a "Known limitations" section to the PR body with both measured facts: `rc=1` is the ordinary state of a busy parallel fleet (not the exceptional one), and #1100 is a live prerequisite this gate doesn't close (recent merges bypass `merge-reviewed-pr.sh` by hand because it takes no body). Stated plainly what the operator does today: run `verdict-drift.sh` to see which open PR is holding the set; no per-PR override exists in this PR, by design, since narrowing would stop being the scope `daily-digest.sh` shares with this script (GH-958). Not narrowing the scope here — that's a separate design decision for the controller.
- **Round-number membership (downward renumbering reads paired)** — agreed, not a defect; this is the deliberate consequence of judging only the newest response (the #974 repair shape), argued correctly in Round 1's own read of `verdict-drift.sh:119-153`. No change.
- **SHADOW rounds count toward the paired set** — agreed as a contract question, not a defect: this check is about reports reaching the PR, not verdict authority, so a SHADOW round counting is consistent with what it measures. No change.
- **"No longer vacuous" live evidence couldn't reproduce** — the PR body's live snapshot predated PR #1102 itself being open. Reworded in the body: the offline fixture (`test-merge-reviewed-pr.sh` case 8) is named as the durable evidence, the live line is now explicitly marked as a snapshot that has since moved (to `rc=1`, per Round 1's own re-run), not a standing property.
- **`tr -d '\r'` in production** — Round 1 judged this acceptable and safe (every field is closed-vocabulary; CR is not representable in any of them). No change requested, none made.
- **Evidence quality (READ vs RAN in Round 1's own review)** — no action for the fix lane; that finding is about the reviewing round's own methodology, not this diff.

### Gates run (this round, on top of Round 1's L1 receipt)

- `sh -n` on all six touched scripts (three newly touched by this round's fix: `verdict-drift.sh`, `merge-reviewed-pr.sh`, plus the three test files) — clean.
- `sh scripts/fleet/test-verdict-drift.sh` — **18/18 pass** (15 from Round 1's diff + 3 new: cases 16-18), run 6× total (1 + 5 background reruns) for determinism, all 6 runs 18/18.
- `sh scripts/test-merge-reviewed-pr.sh` — **8/8 pass**, case 8's fixture updated for the trust filter; re-run 3× total for determinism, all 8/8.
- `sh scripts/fleet/test-daily-digest.sh` — all cases pass including case 3d, updated for the same reason; re-run 3× total for determinism.
- `sh scripts/lint-file-length.sh --tree` — exit 0 (this diff is shell-only; the ratchet is `*.rs`-only, so this is a no-op, run anyway per the ladder).
- Pre-commit hooks on this round's commit: `lint-doc-citations.sh --staged` and `lint-file-length.sh --staged` both passed automatically at commit time.
- No `crates/**` touched — no Cargo gate.
- Exact-head CI on `d7d6197…` (run 34366693247): Format, MSRV, Clippy ×3, `Fleet shell tests` (the job that runs the `scripts/test-*.sh` glob, including every fixture changed this round), `Fleet shell tests (windows)`, `Test (ubuntu-latest)` and **`Test (macos-latest)`** all **pass** — the macOS `simultaneous_reconciles` flake (#1047) that hit the prior SHA did not recur here. `Test (windows-latest)` is still **pending** as of this comment; not waiting in a loop on it per the brief — will report separately if it comes back red, since the diff carries zero Rust either way.

Worktree unchanged: `C:/ai_agent/edda-wt-gh993`, branch `fix/gh993-review-round-delivery`. Repo main checkout stayed on `main`, untouched.

fagemx:
## Code Review: Round 2 — PR #1102 @ d7d619781a3a6e17cb5b91e99ad67ba79ef998a8

- elapsed: unmeasured (controller-spawned subagent; no pi session file to read)
- model_observed: unverified (no pi session JSONL and no `claude -p --output-format json` envelope for this transport)
- reviewer_session: 41362b6e-78e8-478f-8fb5-214516e1f6e6 — replacement reviewer; Round 1's session is gone, so its verdict and the `Review Response` were READ rather than remembered. No distinct per-PR UUID was issued for this round.
- spec: review-spec-v1.6 (read fresh at `origin/main` 4459511, which is ahead of this PR's copy: #1097/#1101 changed §1, D4, S1 and §8 after the branch point — none of those deltas change a rule this diff is judged by, and the §1 change is the `gh pr checkout` prohibition this round obeyed)
- escalations: none
- shadow: false
- cost: read-only round. No checkout, no worktree, no `gh pr checkout`, no Cargo, no build. Source read via `git show`/`git grep` at both SHAs; fixtures executed against a `git archive` copy in scratch. `git status --porcelain` in `C:\ai_agent\edda` is byte-identical before and after and the checkout is still on `main`.

### IN SCOPE

Delta round (`loop` item 5): the target is `git diff c585457..d7d6197` plus a RAN confirmation that Round 1's blocking finding is resolved. The whole PR is not re-reviewed.

**The delta is one commit and five files** — `verdict-drift.sh`, `merge-reviewed-pr.sh`, and three `test-*.sh`. Stripped of comments it is **exactly one executable line**:

```
| select(.authorAssociation == "OWNER" or .authorAssociation == "MEMBER" or .authorAssociation == "COLLABORATOR")
```

inserted at `scripts/fleet/verdict-drift.sh:114`, immediately after `.comments[]` and **before** `(.body) as $b` at `:115` — so trust is decided ahead of the shape match, and gates the `v` and `r` branches with one predicate rather than being threaded through the `if/elif` twice. Everything else in the delta is comments, the `:128`→`:125` citation fix, and fixtures.

Direct callers/consumers: `scripts/merge-reviewed-pr.sh:52-57` (blocking) and `scripts/fleet/daily-digest.sh` (non-blocking, by design). Both unchanged in behaviour by this delta except through what `verdict-drift.sh` now reports.

**Round 1's P1 — resolved, and proved three independent ways.** Verified rather than accepted:

1. **Fail-closed is real, not asserted.** Probed jq 1.8.1 directly over the actual predicate. Every non-trusted shape is dropped with **rc=0 and no error**: key absent, `null`, `5`, `true`, `false`, `[]`, `["OWNER"]`, `{}`, `{"x":"OWNER"}`, `"owner"` (wrong case), `"NONE"`, and `"CONTRIBUTOR"` — the last mattering because it is a real GitHub association that this repo must *not* trust, and does not. The lane's reasoning holds exactly as written. The one shape that does raise is a `.comments[]` element that is not an object — and the pre-existing `(.body) as $b` raises identically on it (both `rc=5`), so the new line adds no error surface; against real `gh` a `--jq` error exits nonzero and lands in `fail_read` → exit 2 → refuse.
2. **All three new fixtures bite.** Ran the head test suite against the **pre-fix** `verdict-drift.sh` in a scratch tree, isolating each case because the harness stops at the first failure. Case 16 fails (`exit 0, expected 1`); case 17 fails on the exact-line assert (`line was '#17 … no verdict on head orphan-response=Round-9', expected '#17 … no verdict on head'`); case 18 fails (`exit 0, expected 1`). Case 17 is the one that could have been decorative — it shares an exit code with the unfiltered behaviour — and it bites only because `expect()` at `test-verdict-drift.sh:80` is `[ "$actual" = "$want_line" ]`, whole-stdout string equality, not a substring grep. It genuinely falsifies.
3. **The updated happy-path fixtures still bite too.** Flipping the two annotations this delta added back to `"NONE"` breaks their cases: `test-merge-reviewed-pr.sh` case 8 → `a healthy, non-vacuous open set was refused (exit 1)`, and `test-daily-digest.sh` case 3d → `orphan-response annotation missing from 擋住什麼`. So the annotations are load-bearing, not decoration, and the positive direction of the filter is exercised.

**Trusted set matches, exactly.** `verdict-drift.sh:114` is `OWNER`/`MEMBER`/`COLLABORATOR`; `merge-reviewed-pr.sh:72` is the same three. The field names differ correctly, not accidentally: `authorAssociation` (camelCase) is the GraphQL shape `gh pr view --json comments` returns, `author_association` (snake_case) is the REST shape `gh api .../issues/N/comments` returns. Confirmed live — `gh pr view 1102 --json comments --jq '.comments[0]|keys'` includes `authorAssociation`, and both §7 comments on this PR return `OWNER`.

**The filter is not blinding the check.** Ran head and pre-fix `verdict-drift.sh` against the **live** repo back to back: byte-identical output (`#1102 d7d619781a3a main stale from c585457767b9`, `rc=1` both). All real comment traffic here is `OWNER`, so the filter subtracts nothing real — it only removes the untrusted-author path. A filter that silently dropped everything would have shown up here as a fleet-wide flip to `no verdict on head`; it does not.

**Round 1's two recorded limitations landed in the PR body**, as asked: `## Known limitations` at body line 41, with `rc=1` as the *ordinary* state of a parallel fleet (line 45, including what the operator does about it today) and #1100 as an open prerequisite this gate does not close (line 46). The vacuity rewording landed too (line 31 now names the offline fixture as the durable evidence and marks the live line as a moved snapshot). `pi-controller-runbook.md:125` × 3 in the body, `:128` × 0 anywhere at head.

Issue #993 doneWhen — unchanged by this delta, re-confirmed not re-litigated: bullet 1 (orphan-response) and bullet 2 (blocking point argued) delivered by the Round 1 diff; bullet 3 independently re-verified again this round (#974/#976/#980/#981 all `MERGED`, exactly 2 `## Code Review: Round N` comments each); bullet 4's caller shape re-read at `merge-reviewed-pr.sh:52-53` — `drift_out=$(…) || drift_rc=$?` reads `$?` of the substitution itself, nothing pipes into or out of it, and the guard is `-ne 0`, so exit 2 refuses as well as exit 1.

Security/data-loss: the property Round 1 found violated — *the merge entrypoint's refusal depends only on trusted-author comments* — now holds, by construction and by fixture. No new security surface in the delta.

Current-base integration: base `main` moved to `44595113c4ad88255d157293194c2bf7cd295fca` during this PR's life (#1101 and the Round 1 period), so I re-derived the window rather than reusing Round 1's. `git diff --no-renames 7d5bb83 origin/main -- 'scripts/**'` is **empty** — no base drift on this PR's own files or anywhere under `scripts/`. Merge base is still 7d5bb83; `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 1102 --name-only` → 6 files, all under `scripts/`; delta touches 5 of them. No docs, so still no collision with the GH-1064 docs lane. |
| U2 | any | P1 | PASS | Body line 92 `Closes #993`; commit enumerator empty. All four doneWhen items delivered (three by the diff, bullet 3 re-verified delivered already this round). Merging finishes the issue. |
| U3 | any | P1 | PASS | `Issue: #993` present. |
| U4 | any | P1 | PASS | `git log --format=%s origin/main..$SHA` filtered by the conventional-subject regex → empty. The delta commit is `fix(fleet): verdict-drift.sh trusts only OWNER/MEMBER/COLLABORATOR comments (GH-993 review round 1, P1)`. |
| U5 | any | P1 | PASS | RAN `sh scripts/lint-markdown-content.sh` → `exit=0`. Caveat stated: it lints the working tree, which is `main`; the diff changes no markdown, so there is nothing in it for this rule to reach. |
| U6 | any | P0 if deterministic | PASS | **CI run 34366693247 @ d7d6197 — all 12 jobs pass**, `CI Gate` included, and that covers `Test (windows-latest)`, which was still pending when this round was dispatched. The `simultaneous_reconciles` macOS flake (#1047) that reddened Round 1's SHA did not recur. |
| U7 | any | P1 | PASS | Whole delta audited before posting; no blocking finding to batch. |
| U8 | any | P1 | N.A.(DECISIONS enumerator printed nothing — no `Decision:` line in the body) | — |
| D1–D5 | docs | — | N.A.(class not routed — `classes=code-risk`; no `docs/**`, `*.md` or `*.txt` in the diff) | Router output: `classes=code-risk` / `canonical_class=code-risk`. |
| S1–S3 | skills | — | N.A.(class not routed — no `SKILL.md`, `.claude/**`, `skills/**`, `AGENTS.md` or `REVIEW.md` in the diff) | The spec changed on `main`, not in this diff, so no `REVIEW.md changed in this diff` escalation is owed. |
| C1 | code-plain | P0 | PASS | Each doneWhen item matched to code and a named fixture — see IN SCOPE. |
| C2 | code-plain | P0 | PASS | The delta commit is `fix:` and carries regression fixtures that **fail without the change** — established by running them against the pre-fix script, not by reading. Cases 16/17/18, each isolated. This is the strongest form of the C2 evidence. |
| C3 | code-plain | P1 | N.A.(no `crates/**` in the diff) | — |
| C4 | code-plain | P1 | PASS | Enumerator empty. |
| C5 | code-plain | P1 | N.A.(selector empty — no `crates/**`, so no Cargo gate is owed locally) | RAN the selector; printed nothing. |
| R1 | code-risk | P0 | PASS | Enumerator empty on both the delta and the whole PR — no `rm -rf`, `reset --hard`, `git rm`, `--delete-branch`, `--force` or `Remove-Item` added. |
| R2 | code-risk | P0/P1 | PASS | Enumerator empty on both the delta and the whole PR — no added line mixes `\|\|` and `&&`. The one AND-list in the new code is two-operand with no third term. |
| R3 | code-risk | P0 | PASS | RAN `sh -n` on the **head** content of all six changed scripts (extracted with `git archive`, not the working tree, which holds `main`) → exit 0 for each. |
| R4 | code-risk | P1 | PASS | The delta creates no temp directory, file or lock — it adds one jq expression and fixture text. The test scripts keep their pre-existing `trap … 0 HUP INT TERM` cleanups. |
| R5 | code-risk | P1 | **PASS (was FAIL in Round 1)** | The property now holds: `verdict-drift.sh:114` applies the same trusted set as `merge-reviewed-pr.sh:72` before any shape match, and every untrusted or malformed `authorAssociation` shape is dropped without error (jq probe, 14 shapes + absent key). The caller-supplied SHA check at `merge-reviewed-pr.sh:61` (`^[0-9a-f]{40}$`) is untouched. |

### Wiring

|---|---|---|---|---|
| Author-trust predicate on the comment read (a new *filter* on an existing surface, not a new surface) | `scripts/fleet/verdict-drift.sh:114` — one jq `select` ahead of the `v`/`r` branches | the same jq program's `if/elif` at `:117-131`, then the `while read row_tag …` loop at `:143-159`; reaches `merge-reviewed-pr.sh:52-57` through the process exit code and `daily-digest.sh` through the printed line | **Not swallowed, and fails closed.** A dropped comment yields the same `empty` a comment matching neither heading already produced, so a missing/`null`/wrong-typed value can only *remove* a verdict (→ `no verdict on head` → `not_ready=1` → exit 1 → refuse), never grant one. Proved by cases 16 and 18. A non-object element raises in jq, which exits `gh` nonzero → `fail_read` → exit 2 → also refuse | predicate → jq row → shell loop → `not_ready` → exit code → merge refusal, proved end-to-end by `test-merge-reviewed-pr.sh` case 8 (refuses when its fixture is flipped untrusted) |

No new `pub` surfaces (`sh scripts/wiring-scan.sh origin/main $SHA` → "no new pub surfaces"). Its single swallow-pattern hit is the word "silently" inside a comment — false positive, same as Round 1. Round 1's `orphan-response=Round-<N>` surface is unchanged by this delta and its row stands as posted.

### Findings

- **[P2] The offline stub's `gh` ends in an unconditional `exit 0`, so the fixtures cannot prove the jq-error path they rely on.** `test-verdict-drift.sh`'s stub (`:55`) and `test-merge-reviewed-pr.sh`'s stub both run the caller's `--jq` through real jq and then `exit 0` regardless of jq's status. In production that path is fail-closed and correct — real `gh` exits nonzero on a `--jq` error, `verdict-drift.sh:112-133` routes it to `fail_read`, exit 2, and `merge-reviewed-pr.sh:54` refuses on any nonzero — but no fixture demonstrates it, so the claim rests on reading rather than on a test. Pre-existing harness shape, not introduced by this delta, and not blocking: I confirmed by direct probe that the new `select` adds no error surface the pre-existing `(.body)` did not already have (both raise `rc=5` on a non-object element, both are unreachable from `gh`'s real output shape). — evidence: `scripts/fleet/test-verdict-drift.sh:27-56`; `scripts/fleet/verdict-drift.sh:112-133`; `scripts/merge-reviewed-pr.sh:52-57`; jq probe `{"comments":["hello"]}` → `Cannot index string with string "authorAssociation"` and `… with string "body"`, both `rc=5`.

- **[P2] The trust filter is exercised only by fixtures, because real traffic on this repo is uniformly `OWNER`.** Stated so the evidence is not read as stronger than it is: `[.comments[].authorAssociation] | unique` is `["OWNER"]` on both open/recent PRs I sampled, and head vs pre-fix `verdict-drift.sh` produce byte-identical output against the live repo. That is the *right* result — it shows the filter subtracts nothing legitimate — but it also means the `NONE` and absent-key paths are proved by cases 16/17/18 and nowhere else. Those three cases carry the whole discriminating claim, which is why I falsified each individually rather than trusting the suite's aggregate pass. — evidence: RAN both script versions against the live repo; `gh pr view 1101 --json comments --jq '[.comments[].authorAssociation]|unique'` → `["OWNER"]`.

- **[P2] A verdict posted by a bot or app identity would now be invisible to this check.** GitHub reports `NONE` (or `CONTRIBUTOR`) for most bot and Actions-token comment authors, so a §7 comment written by one would be dropped, read as `no verdict on head`, and refuse. That direction is safe — fail-closed, never fail-open — and it is not a divergence this PR introduces: `merge-reviewed-pr.sh:72` already rejects the same comment for the same reason, so the two stay consistent, which is the point of the fix. Recorded as a contract consequence for whoever lands #1103, not as a defect. — evidence: `scripts/fleet/verdict-drift.sh:114` vs `scripts/merge-reviewed-pr.sh:72`; jq probe of `"CONTRIBUTOR"`/`"NONE"` → both dropped.

Round 1's P2s are not re-raised: the citation P2 is fixed and verified (`:125` correct at head *and* at current `main`, `:128` gone), and the whole-open-set coupling, round-number membership, SHADOW-counting and vacuity items were routed by Round 1 as recorded-not-changed and are now recorded in the body. Re-opening them here would be a later-round blocker that is neither fix-caused nor previously unobservable (`loop` item 2).

### FOLLOW-UP ISSUE

- **`scripts/test-merge-reviewed-pr.sh` resolves its subject by a relative path, so it silently tests whatever tree the cwd points at.** Line 26 is `script=scripts/merge-reviewed-pr.sh`. Run from anywhere but a repo root it validates *that* directory's copy while appearing to test the file it ships beside. I hit this directly: running the head copy of the suite from the main checkout exercised `main`'s gate-less `merge-reviewed-pr.sh` and reported `case 6: an unrelated PR with no verdict did not block` — a convincing false failure that took a debugging pass to attribute. Its sibling `scripts/fleet/test-verdict-drift.sh:15` resolves `root` from `$(dirname "$0")/../..` and is immune. Pre-existing, out of scope for #993, and harmless in CI (which runs from the repo root) — but it is a live trap for exactly the read-only reviewer this repo's process depends on. — evidence: `scripts/test-merge-reviewed-pr.sh:26` vs `scripts/fleet/test-verdict-drift.sh:15-16`; the false failure and its disappearance once cwd was the head tree.
- **No fixture covers `verdict-drift.sh`'s exit-2 path from a jq/filter error** (as distinct from `STUB_DRIFT_LIST_FAIL`, which covers a `gh` process failure). Would need a stub that propagates jq's status. See P2-1.
- **#1103 should name the trusted set it lands with.** The PR body already flags that if `edda review deliver` adopts a different set the two must be reconciled; worth carrying that sentence into #1103 itself so it is not only recorded on a PR that will be merged and forgotten.

### RAN vs READ

RAN (read-only; `C:\ai_agent\edda` stayed on `main`, `git status --porcelain` unchanged before and after, nothing checked out, built, edited or pushed):

- `git show`/`git grep`/`git diff`/`git log` at `c585457`, `d7d6197`, `7d5bb83` and `origin/main`; `git archive d7d6197 scripts` into a scratch tree so every fixture ran against **head** content rather than the working tree's `main`.
- **`sh -n`** on the head content of all six changed scripts → exit 0 each (R3).
- **Head fixtures, on Windows — the CI gap:** `test-verdict-drift.sh` → **18/18**; `test-daily-digest.sh` → all cases pass; `test-merge-reviewed-pr.sh` → **8/8**. Reason for running rather than reading: CI's `Fleet shell tests (windows)` runs only `scripts/test-git-config-guard.sh` plus the `scripts/fleet/test-*.ps1` glob (verified myself in `.github/workflows/ci.yml` at head), so **none** of these three `.sh` files execute on Windows in CI. This is a focused check for a real coverage gap, not a rerun of a recorded gate (`ladder` L2).
- **Falsification against the pre-fix script:** same suite, `c585457`'s `verdict-drift.sh` → case 16 fails; cases 17 and 18 isolated and each fails. Reverse falsification: the two fixtures this delta annotated, flipped back to `"NONE"` → `test-merge-reviewed-pr.sh` case 8 and `test-daily-digest.sh` case 3d both fail.
- **jq 1.8.1 property probe** of the exact predicate over 14 value shapes plus absent key plus non-object element; and the pre-existing `(.body)` expression over the same non-object element for comparison.
- `sh scripts/fleet/verdict-drift.sh` — head and pre-fix versions — against the live repo, read-only → identical output, `rc=1`.
- `sh scripts/lint-markdown-content.sh` → `exit=0` (U5); `sh scripts/wiring-scan.sh origin/main $SHA` → "no new pub surfaces" (§5.5); the §3 ROUTER block → `classes=code-risk`.
- U2/U3/U4/U8/R1/R2/C4/C5 enumerators — empty except the U2 `Closes #993` hit and the U3 `Issue:` line, both judged above.
- `gh pr checks 1102`; `gh run list --commit d7d6197`; `gh pr view --json comments/labels/mergeable`; `gh issue view 993 1100 1103`; `gh api repos/fagemx/edda/issues/{974,976,980,981}/comments`.
- Base-window diffs `7d5bb83..origin/main` over the PR's six files and over all of `scripts/**` → both empty.

READ:

- **L1 receipt: CI run 34366693247 @ d7d619781a3a6e17cb5b91e99ad67ba79ef998a8 — 12/12 pass.** Format; Clippy ubuntu/macos/windows; MSRV (1.91.0); Test ubuntu/macos/**windows**; Fleet shell tests; Fleet shell tests (windows); Detect code changes; CI Gate. The workspace Cargo gates are READ from these job results and were not run here (`ladder` L1).
- What it covers: the ubuntu `Fleet shell tests` job runs `scripts/fleet/run-fleet-tests.sh`, whose glob is `scripts/fleet/test-*.sh scripts/test-*.sh` (read at head), so all three changed test files do execute there.
- What it does **not** cover: those same `.sh` files on Windows. My local Windows runs above are the only such evidence, and they are independent of the lane's.
- Round 1's verdict comment and the `Review Response: Round 1`, read in full as a replacement reviewer. Every factual claim I relied on from either was re-derived here.
- C5 selector empty (no `crates/**`), so no Cargo gate is owed locally and none was run.

### Verdict

LGTM (P0=0, P1=0)

Round 1's P1 is fixed at the right layer and for the right reason. The remedy is one jq `select` placed ahead of the shape match, sharing `merge-reviewed-pr.sh:72`'s trusted set exactly rather than inventing a second definition of trust; the fail-closed claim survives a direct probe of jq's semantics across every malformed shape I could construct; and all three new fixtures genuinely falsify — each one fails against the pre-fix script, including case 17, whose exit code is unchanged and which therefore depended on `expect()` being a whole-line equality rather than a substring match. The three P2s and three follow-ups above are recorded, not blocking. Exact-head CI is fully green, including the `Test (windows-latest)` job that was still pending when this round was dispatched, and the base window over `scripts/**` is empty against the current `main`.

PR #1016, opened 2026-09-06
feat(edda-cli): typed ratification — --evidence pr#N@sha and --by-rule cited-authority

fagemx:
Issue: #764
Issue: #761
Decision: ratify.authority, decision.cites

Bundle A — both issues are typed values of the same `ratified_by` field, so
they ship together. No closing keyword: one doneWhen item on each is not
delivered by this diff (see **What is not here**).

## What this changes

Before this PR, `edda ratify` had one authority form: a free-text `--by`. Two
things were being spelled into that string by callers instead of being owned
by the binary:

- **"a merged PR made this binding"** — the evidence form (#764). It lived
  nowhere yet; the daemon would have hand-built the string, which is the debt
  #766 D8 names.
- **"a rule made this binding"** — the rule form (#761). It lived in
  `~/.edda/fleet/ratify-sweep.py`: a regex over reasons that ratified 53
  decisions in one pass, not in the product, not tested, invisible to the
  other machine, and auditable only as free text. Exactly what
  `review.carrier` forbids.

Both are now typed, and permanently distinguishable by prefix:

| form | `ratified_by` | who is asserting |
|---|---|---|
| `edda ratify <KEY> [--by WHO]` | `<WHO>` or session label | a person |
| `edda ratify <KEY> --evidence pr#N@sha` | `evidence:pr#N@sha` | a merged PR |
| `edda ratify --by-rule <RULE>` | `rule:<RULE>` | a rule in this binary |

`edda ask` renders the typed ones as `binding (evidence: …)` / `binding
(rule: …)`; a person's `--by` keeps `ratified (by …)`.

Exit codes follow `claim-check.exit-codes=0/1/2` — 0 success **including the
no-op when the key is already binding**, 1 no active decision for the key, 2
malformed input.

The `cited-authority` rule ratifies an active, unratified decision that names
the authority it rests on, and holds everything else with a printed reason:
`product.` / `commercial.` / `spend.` keys, anything a later decision's reason
names, and anything uncited. `--dry-run` prints the same table and writes
nothing.

`edda decide --cite operator:<when>|issue:#<n>|decision:<key>` writes the
citation as a structured field so the rule reads a field instead of guessing
from prose. It is stored in the decision **event payload**, not a new SQLite
column — no schema migration, and old ledgers keep working through the reason
fallback.

The `Decision:` line above is this PR dogfooding its own convention. Both keys
were recorded with `edda decide` while building it, and both are implemented
here (U8):

| key | value | implemented by |
|---|---|---|
[...]

PR #1108, opened 2026-09-09
fix(edda-ledger): escape_field escapes backticks, closing the last hole in the round-trip encoding (GH-1044)

fagemx:
Issue: #1044
Closes #1044

## What was broken

`escape_field` (`crates/edda-cli/src/cmd_export.rs`) escaped backslash and
newline (PR #1017) but left backtick raw. The read side, `backtick_list`
(`crates/edda-ledger/src/sync.rs`), splits `Tags` / `Affected paths` / `Cites`
list fields on the literal 4-byte sequence `` `, ` ``. A list item whose value
itself contained that sequence — e.g. a tag literally equal to `` a`, `b `` —
split into two items after a mirror export/import round trip, and an item that
merely started or ended with a backtick lost that character. This was the one
hole PR #1017 left in the "encoding is total" property it established for
newlines and backslashes.

These fields carry arbitrary user text: `--paths` and `--cite` are repeatable
`Vec<String>` with no `value_delimiter` (`main.rs:129,137`), stored verbatim by
`cmd_bridge/decide.rs:123-126`. The defect is reachable from the shipped CLI.

## What this PR delivers

Four things, in the order the review rounds forced them out. All four are in
the head; none of the intermediate designs are.

**1. The write side escapes backtick.** `escape_field` gains a third
`.replace()`: backtick → `` \` ``, applied **after** backslash-doubling and the
newline pass, so the escape backslashes those passes insert are never
themselves re-escaped. `unescape_field` gains the matching arm.

**2. The read side splits escape-aware.** Escaping alone was not enough:
`split` matches four literal bytes and never looks at what precedes them, so an
escaped content backtick could still supply the *opening* half of a delimiter
match. `split_unescaped_backtick_comma` walks the string with the same pairing
rule `unescape_field` uses — a `\` and whatever follows it are consumed as one
unit and can never begin a delimiter match — and `backtick_list` strips the
list's own outer wrapping backticks once, from the whole string, before
splitting rather than per fragment.

**3. A writer-era discriminator, because "pre-GH-1044 text" is not one era.**
`escape_field` did not exist before PR #1017 (`3306c1a`, 2026-09-07); before
it, `edda export md` (live since `f3e5e97`, 2026-07-08) applied **no escaping
at all** (`3306c1a^:crates/edda-cli/src/cmd_export.rs:150`). An escape-aware
walk is correct for what the current writer produces and *wrong* for raw-era
bytes, where a backslash has no structural pairing: raw items `a\` and `b`,
[...]

PR #1110, opened 2026-09-09
test(edda-cli): verify_contract OK checks match the authored marker, not a temp-path substring (GH-1033)

### LINKED ISSUE #1033: test(edda-cli): verify_contract asserts !report.contains("OK") against a string holding the random temp path — a .tmpOK* directory name fails a correct run
`crates/edda-cli/tests/verify_contract.rs:161` asserts that a `verify` run against a deleted ledger does not report success:

```rust
assert!(
    !report.contains("OK"),
    "must not report success for a vanished ledger: {report:?}"
);
```

`report` is `stdout + stderr`, and the error text embeds the **temp directory path**. `tempfile` names directories with random alphanumerics, so any name containing the adjacent pair `OK` fails the assertion even though the command behaved correctly.

Observed on PR #1016 @ `7acbc40c86c1c011f2bde2b7e36ff128495ca371`, job `Test (macos-latest)` in run 34055854543:

```
must not report success for a vanished ledger: "error: cannot open ledger at /private/var/folders/d8/hvxvltxn0fl4rmnd52sncbth0000gn/T/.tmpOKfVBd: ledger database not found: /private/var/folders/d8/hvxvltxn0fl4rmnd52sncbth0000gn/T/.tmpOKfVBd/.edda/ledger.db (run `edda init` to create a workspace)\n"
```

The directory is `.tmpOKfVBd`. Exit code was 2 and the message was correct — only the substring test was wrong. Same test is green on `main` @ `ad119cf`; it is not caused by that PR.

**doneWhen**

- The assertion no longer depends on text the harness does not control: assert on the parts of the output the command actually authored, not on a blob that includes the temp path — e.g. strip the workspace path from `report` before the `contains` checks, or assert the absence of the real success marker (the exact word the OK path prints) rather than the bare substring `OK`.
- The two neighbouring assertions in the same test (`report.contains("ledger")`, and the `.edda/ledger.db` existence check) are reviewed for the same class of accidental coupling.
- A regression test pins the failure: the same assertion run against a report whose path segment contains `OK` passes.

Class: product (`crates/edda-cli`). Carrier per ledger `fleet.ci-flake-carrier=product-issue-per-test-no-registry`.

fagemx:
`crates/edda-cli/tests/verify_contract.rs` had two `assert!(!report.contains("OK"), ...)` checks, both on `report = format!("{stdout}{stderr}")`. `stderr` embeds the workspace path (`repo_root.display()`), and `tempfile` names temp directories with random alphanumerics — so any run whose random suffix happens to contain the adjacent pair `OK` fails a correct run. Observed on `windows`/`macos` CI as `.tmpOKfVBd` (PR #1016, job `Test (macos-latest)`, run 34055854543).

## Fix

Added one constant, `VERIFY_OK_MARKER = "ledger chain OK:"` — the exact text `cmd_verify::execute` prints on stdout for an intact chain (`crates/edda-cli/src/cmd_verify.rs`, the two `println!` arms `"ledger chain OK: empty ledger ..."` / `"ledger chain OK: {n} event(s), last event {id}"`). Both negative assertions now check for that authored marker instead of the bare substring `"OK"`:

- `verify_deleted_ledger_db_exits_2_and_is_not_recreated` (was line 162)
- `verify_anchored_workspace_with_unopenable_ledger_exits_2_and_never_reports_ok` (was line 296)

I read the product code before choosing the marker rather than asserting one from memory — the exact printed prefix is what makes this fix precise in both directions: it can no longer false-fail on a path substring, and it cannot false-pass on stray output either, since `"ledger chain OK:"` is never emitted on any error branch (all four error branches in `cmd_verify.rs` print `"error: ..."` text that does not contain it).

## Why the second site (`:296`) is in scope

The issue names only `:161`–`:162`, in `verify_deleted_ledger_db_exits_2_and_is_not_recreated`, and asks that *neighbouring assertions in the same test* be reviewed for the same coupling. `:296` is a different test (`verify_anchored_workspace_with_unopenable_ledger_exits_2_and_never_reports_ok`), so a literal reading would route it to a follow-up issue. I did not: it is the identical `!report.contains("OK")` against the identical `stdout+stderr` blob, subject to the identical random-path failure — shipping a fix for one occurrence while its twin stays live is not a fix, it's a second issue waiting to be filed under a new number. Both are fixed here, from the same constant, in the same PR.

## The two positive sites (`:128`, `:139`) — reviewed, left unchanged

`verify_clean_ledger_is_ok_and_exits_0` and `verify_empty_ledger_is_ok_not_an_error` both assert `stdout.contains("OK")` — on `stdout` alone, not the `stdout+stderr` blob. I checked whether the success path ever puts a path on stdout: it does not. Both success branches in `cmd_verify.rs` (`report.events == 0` and the general case) print only the hardcoded `"ledger chain OK: ..."` text — neither interpolates `repo_root.display()` or any other path. Nothing else writes to stdout on this path either: `main.rs` sets up `tracing_subscriber` with `.with_writer(std::io::stderr)` and a default `"warn"` filter, and there is no `println!` anywhere between `main()` and `cmd_verify::execute` that could inject a path ahead of it. So a temp-path substring can never reach `stdout` on the success branch, and the coupling class this issue is about does not apply here. Left unchanged rather than churned for symmetry.

## doneWhen: neighbouring assertions reviewed for the same class

- `report.contains("ledger")` (next to the `:162` fix): the word "ledger" is part of the product's own authored error text on every path that reaches this assertion (`"error: cannot open ledger at ..."`), independent of whatever the random temp name is. It is not coupled to the accidental path content — it's there regardless of the path — so it is not an instance of this bug class. Left unchanged.
- The `.edda/ledger.db` existence check (same test): `Path::exists()`, not a `str::contains()` on process output. There is no text-authorship question to ask of a filesystem check — this bug class does not apply. Left unchanged.
- `report.contains("cannot open ledger")` (next to the `:296` fix, line ~346): already the correct, authored-text pattern — this is the existing "how to do it right" example in the same file that the `:296` fix now matches.

## Regression test (doneWhen #3)

Added `verify_deleted_ledger_ok_in_workspace_path_is_not_mistaken_for_success`. Waiting on `tempfile`'s randomness to reproduce the original flake on demand isn't reliable, so the test forces the same coupling deterministically: it nests the workspace under a directory whose name it chooses to contain "OK" (`workspaceOKdir`), then deletes `ledger.db` exactly like the sibling test. `cmd_verify::execute` embeds that path directly in `"error: cannot open ledger at {path}: ..."`, so the resulting `report` is guaranteed to contain the bare substring `"OK"` — every run, not just unlucky ones.

**Proof the test pins the failure.** I temporarily reverted just this new test's assertion to the pre-fix pattern (`!report.contains("OK")`) on the scratch worktree, ran it alone, and captured red:

```
running 1 test
test verify_deleted_ledger_ok_in_workspace_path_is_not_mistaken_for_success ... FAILED

---- verify_deleted_ledger_ok_in_workspace_path_is_not_mistaken_for_success stdout ----

thread 'verify_deleted_ledger_ok_in_workspace_path_is_not_mistaken_for_success' panicked at crates\edda-cli\tests\verify_contract.rs:210:5:
must not report success for a vanished ledger just because its workspace path contains "OK": "error: cannot open ledger at C:\\Users\\synvoke\\AppData\\Local\\Temp\\.tmpGgpvmb\\workspaceOKdir: ledger database not found: C:\\Users\\synvoke\\AppData\\Local\\Temp\\.tmpGgpvmb\\workspaceOKdir\\.edda\\ledger.db (run `edda init` to create a workspace)\n"

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 10 filtered out; finished in 4.31s
```

Exit code was 2 and the message was correct (`cannot open ledger`) — same shape as the issue's original report, reproduced on demand instead of by luck. I then restored the fixed assertion (`!report.contains(VERIFY_OK_MARKER)`); the same test passes as part of the full `cargo test -p edda` run reported below.

## Verification (L0, touched crate only — per brief)

```
cargo fmt --all --check                                  # clean, no diff
cargo clippy -p edda --all-targets -- -D warnings         # clean, zero warnings
cargo test -p edda                                        # full pass, all binaries green
                                                            # (includes verify_contract.rs: 11/11 ok)
sh scripts/lint-file-length.sh --tree                      # clean, no ceiling violations
```

`crates/edda-cli/tests/*.rs` are process-spawning integration tests (`env!("CARGO_BIN_EXE_edda")`), so `cargo test -p edda` (not `--bin edda`) is required to exercise this file at all — confirmed the new test actually ran as part of that command, not just in isolation.

`edda-cli` is inside CI's Windows 7-crate subset, so the C5 selector (focused Windows-gap run) is empty for this PR — no additional local run is owed beyond the L0 above; exact-head CI covers the rest.

Only `crates/edda-cli/tests/verify_contract.rs` was touched. No overlap with `crates/edda-cli/src/cmd_reconcile/**`, `crates/edda-cli/src/cmd_export.rs`, or `crates/edda-ledger/**`.

Issue: #1033
Closes #1033

### DIFF
--- a/crates/edda-cli/tests/verify_contract.rs
+++ b/crates/edda-cli/tests/verify_contract.rs
@@ -8,6 +8,16 @@ use edda_ledger::Ledger;
 use serde_json::Value;
 use std::path::{Path, PathBuf};
 
+/// The exact success marker `cmd_verify::execute` prints on stdout for an
+/// intact chain (`crates/edda-cli/src/cmd_verify.rs`, the two `println!`
+/// arms `"ledger chain OK: empty ledger ..."` / `"ledger chain OK: {n}
+/// event(s), last event {id}"`). Negative assertions below must check for
+/// this authored text, never the bare substring `"OK"` — a bare substring
+/// also matches the letter pair inside a `tempfile`-generated temp
+/// directory name (GH-1033, e.g. `.tmpOKfVBd`), which makes the assertion
+/// depend on text the command never wrote.
+const VERIFY_OK_MARKER: &str = "ledger chain OK:";
+
 /// Path to the `edda` binary (`CARGO_BIN_EXE_edda`).
 fn edda_bin() -> PathBuf {
     PathBuf::from(env!("CARGO_BIN_EXE_edda"))
@@ -159,12 +169,48 @@ fn verify_deleted_ledger_db_exits_2_and_is_not_recreated() {
         "must explain the ledger problem: {report:?}"
     );
     assert!(
-        !report.contains("OK"),
+        !report.contains(VERIFY_OK_MARKER),
         "must not report success for a vanished ledger: {report:?}"
     );
     assert!(
         !repo.path().join(".edda").join("ledger.db").exists(),
         "verify must never recreate the ledger database"
+    );
+}
+
+/// GH-1033 regression: a harness-generated path segment that happens to
+/// contain the letter pair "OK" must never be mistaken for the real
+/// success marker. `tempfile` names directories with random alphanumerics
+/// and did exactly this in the wild — PR #1016 CI hit `.tmpOKfVBd` on
+/// `macos-latest`. A random tempdir name is not reliable to reproduce on
+/// demand, so this test forces the same coupling deterministically by
+/// nesting the workspace under a directory whose name we choose to
+/// contain "OK" — `cmd_verify::execute` embeds that path directly in its
+/// "cannot open ledger at {path}" error text.
+#[test]
+fn verify_deleted_ledger_ok_in_workspace_path_is_not_mistaken_for_success() {
+    let base = tempfile::tempdir().expect("base tempdir");
+    let repo = base.path().join("workspaceOKdir");
+    std::fs::create_dir_all(&repo).expect("create OK-named workspace dir");
+    seeded_ledger(&repo);
+    std::fs::remove_file(repo.join(".edda").join("ledger.db"))
+        .expect("delete ledger.db, keeping .edda/");
+
+    let (code, stdout, stderr) = run_edda(&["verify"], &repo);
+    assert_eq!(
+        code, 2,
+        "missing ledger must exit 2, not rebuild an empty one: {stdout:?} {stderr:?}"
+    );
+    let report = format!("{stdout}{stderr}");
+    assert!(
+        report.contains("OK"),
+        "sanity: the workspace path must actually carry the bare substring \
+         this regression guards against, or the test proves nothing: {report:?}"
+    );
+    assert!(
+        !report.contains(VERIFY_OK_MARKER),
+        "must not report success for a vanished ledger just because its \
+         workspace path contains \"OK\": {report:?}"
     );
 }
 
@@ -293,7 +339,7 @@ fn verify_anchored_workspace_with_unopenable_ledger_exits_2_and_never_reports_ok
     assert_eq!(code, 2, "stdout={stdout:?} stderr={stderr:?}");
     let report = format!("{stdout}{stderr}");
     assert!(
-        !report.contains("OK"),
+        !report.contains(VERIFY_OK_MARKER),
         "must not report success on an unopenable ledger: {report:?}"
     );
     assert!(


fagemx:
## Code Review: Round 1 — PR #1110 @ aff09ad732d477fbc9d4d9aed3c7a8c8962807df

- elapsed: unmeasured (controller-spawned subagent; no pi session file to read `elapsed_ms` from)
- model_requested: opus (`claude-opus-5`, Claude Code)
- model_observed: unverified (no session JSONL / `modelUsage` key reachable from this transport; §6 forbids copying `model_requested`)
- reviewer_session: `41362b6e-78e8-478f-8fb5-214516e1f6e6` (controller-spawned subagent session id; no separate lane UUID was issued)
- spec: review-spec-v1.6 — read at the **base** SHA with `git show c9e29300c8f2b9290fa9e8799b62383281dd4e20:REVIEW.md`, not from the local checkout (whose copy is an older revision)
- escalations: none — §4 routes `code-plain` to §5.0/§5.3/§5.5, and the spec's only `[判斷]` item (D5) lives in §5.1, which is not routed here
- shadow: false
- cost: read-only round; ~14 tool calls, no Cargo work, no build, no test run, no checkout, no worktree. Build lane: n/a

### IN SCOPE

- **Changed behaviour and paths.** One file, `crates/edda-cli/tests/verify_contract.rs` (+48/−2, one commit). Two negative assertions move from `!report.contains("OK")` to `!report.contains(VERIFY_OK_MARKER)` (`:172`, `:342`); one new module-level const `VERIFY_OK_MARKER = "ledger chain OK:"` (`:19`); one new `#[test]`, `verify_deleted_ledger_ok_in_workspace_path_is_not_mistaken_for_success` (`:190`–`:215`).
- **Direct producers/consumers.** The const's producer is `crates/edda-cli/src/cmd_verify.rs:85` and `:88`; its consumers are the three assertion sites at `verify_contract.rs:172`, `:211`, `:342`. No production code changes.
- **Issue doneWhen.** GH-1033, three bullets — judged under C1 below.
- **Security / data loss.** None introduced or exposed: test-only target, no new process surface beyond the existing `CARGO_BIN_EXE_edda` spawn, no write outside `tempfile::tempdir()` (`:192`–`:197`), no ledger or store path touched.
- **Current-base integration.** `git rev-parse aff09ad^` = `c9e29300c8f2b9290fa9e8799b62383281dd4e20` = `git merge-base origin/main aff09ad` = current `origin/main`. `git diff --no-renames c9e2930..origin/main -- crates/edda-cli/tests/verify_contract.rs` → 0 lines. The reviewed tree is the tree that would merge.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 1110 --name-only` → `crates/edda-cli/tests/verify_contract.rs` only; inside GH-1033's named surface (`Class: product (crates/edda-cli)`) |
| U2 | any | P1 | PASS | body line 63 `Closes #1033`; commit-message scan empty. All three doneWhen delivered (C1), so the closing keyword is correct under `pr.closing-keyword=only-when-all-donewhen-delivered` |
| U3 | any | P1 | PASS | `Issue: #1033` present in the body |
| U4 | any | P1 | PASS | offender grep over `origin/main..aff09ad` empty; single subject `test(edda-cli): verify_contract OK checks match the authored marker, not a temp-path substring (GH-1033)` |
| U5 | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → `exit=0`. RAN on the local `main` checkout rather than the reviewed tree (read-only round); the diff adds zero markdown, so its result cannot be affected by this PR |
| U6 | any | P0 | PASS | READ, not RAN — see `RAN vs READ` |
| U7 | any | P1 | PASS | whole scoped audit finished before posting; every finding batched into this round |
| U8 | any | P1 | N.A.(no `Decision:` line — the §1 DECISIONS enumerator printed nothing, so the rule is inapplicable) | — |
| C1 | code-plain | P0 | PASS | three doneWhen bullets matched to code, below |
| C2 | code-plain | P0 | PASS | subject type is `test(…)`, not `feat:`/`fix:`, so the rule's trigger does not fire; a regression test is added regardless — `git diff origin/main...aff09ad -- 'crates/**' \| grep -cE '^\+.*(#\[test\]\|fn test_)'` → `1`, and the PR body captures its pre-fix red at `verify_contract.rs:210:5` |
| C3 | code-plain | P1 | N.A.(3 `.expect(` candidates at diff lines 46/48/51, all in `crates/edda-cli/tests/verify_contract.rs` — a `tests/` integration-test target, i.e. test-only code, which `.claude/CLAUDE.md` §3.3 permits) | opened the file at the head SHA to confirm; see FOLLOW-UP on the spec wording |
| C4 | code-plain | P1 | PASS | enumerator empty — no `unsafe {`, no `clippy::all`, no crate-level `#![allow(...)]` |
| C5 | code-plain | P1 | N.A.(selector empty — `edda-cli` is inside CI's Windows 7-crate subset) | `.github/workflows/ci.yml:182` carries `-p edda` in the Windows matrix args, so `Test (windows-latest)` compiles and runs this exact test file. No focused Windows-gap run is owed |
| §5.5 | any | — | PASS | wiring table below; `sh scripts/wiring-scan.sh origin/main aff09ad` → `no new pub surfaces`, `no swallow patterns on added lines` |

**C1 — the three doneWhen, against code, not against the PR's claim about it.**

1. *"assert the absence of the real success marker (the exact word the OK path prints) rather than the bare substring `OK`"* — **met.** `VERIFY_OK_MARKER = "ledger chain OK:"` (`verify_contract.rs:19`) is character-for-character the prefix of both success arms, read at the head SHA: `cmd_verify.rs:85` `println!("ledger chain OK: empty ledger (0 event(s))")` and `cmd_verify.rs:87-91` `println!("ledger chain OK: {} event(s), last event {}", …)`. There are exactly two success paths and the marker is right for **both** — the empty-ledger arm and the seeded arm — which was the way this fix could have gone silently wrong. Neither interpolates a path (they interpolate `report.events` and `report.last_event_id` only). The marker cannot be produced by any non-success output either: the four error branches (`cmd_verify.rs:27`, `:41`, `:44`, `:58`) all print `error: …`, the broken-chain arm prints `ledger chain BROKEN at event …` (`:78`), and `UnsupportedSchemaVersionError`'s Display is `refusing to open ledger: stored schema version …` (`edda-ledger/src/sqlite_store/schema.rs:232`). The PR body's claim that the marker "is never emitted on any error branch" is accurate as written.
2. *"the two neighbouring assertions in the same test are reviewed for the same class"* — **met.** Both are reviewed in the PR body with a stated argument. `report.contains("ledger")` (`:168`): the authored error text on the branch this test reaches is `error: cannot open ledger at {path}: ledger database not found: …`, so the assertion is satisfied by product text regardless of the temp name — it is not an instance of the false-*failure* class GH-1033 reports. See finding P2-3 for the residual, weaker concern in the opposite direction. The `.edda/ledger.db` existence check (`:175`–`:178`) is a `Path::exists()`, not a `str::contains()` on process output; correctly ruled out.
3. *"a regression test pins the failure: the same assertion run against a report whose path segment contains `OK` passes"* — **met.** `verify_deleted_ledger_ok_in_workspace_path_is_not_mistaken_for_success` (`:190`–`:215`) forces the coupling deterministically instead of waiting on `tempfile`'s randomness. I checked that the setup actually lands on the intended branch: `seeded_ledger` creates `.edda/`, only `ledger.db` is removed (`:196`), and `EddaPaths::find_root` stops at the first ancestor where `.edda` **is a directory** (`edda-ledger/src/paths.rs:206-209`), so the climb halts at `workspaceOKdir` and cannot escape to a workspace above `%TEMP%` — the GH-646 hazard documented at `:326`. The unbounded walk does not canonicalize (`paths.rs:192-194`), so `repo_root.display()` echoes the literal `workspaceOKdir` and the sanity assertion at `:205`–`:209` is deterministic on every platform, not probabilistic. Confirmed green at this SHA on `Test (ubuntu-latest)` and `Test (macos-latest)`.

**Falsify-proof** (brief item 4) — the captured red is of the right thing. The revert is necessarily of the new test's own assertion (there is no production code to revert in a test-only fix, and the two fixed sites cannot be made to fail on demand). The captured panic is at `verify_contract.rs:210:5` with the message `must not report success for a vanished ledger just because its workspace path contains "OK"`, which is exactly the `assert!` at `:210`–`:214` as the file now stands (a one-line swap leaves the line numbers identical), and the reported report string `error: cannot open ledger at …\workspaceOKdir: …` is the same shape as the issue's `.tmpOKfVBd` original. `10 filtered out` is consistent with the file's 11 tests. Proof and tree agree.

**Scope call on `:339` (was `:296`)** — endorsed, not routed as creep. I verified the twin is genuinely identical exposure, not merely similar: `verify_anchored_workspace_with_unopenable_ledger_exits_2_and_never_reports_ok` builds its blob the same way (`:340`) and reaches the same `cannot open ledger at {path}` text with the same `tempfile` random path. A one-line fix for one occurrence while its twin stays live would have produced a second flake report under a new number. The controller's widening was the right call.

### Wiring

|---|---|---|---|---|
| no new pub surfaces | — | — | — | — |
| *(not a §5.5 new surface — private test-target const, listed because its anchor question is where P2-1 lives)* `VERIFY_OK_MARKER` `verify_contract.rs:19` | `cmd_verify.rs:85`, `:88` — `println!("ledger chain OK: …")` on stdout | `verify_contract.rs:172`, `:211`, `:342` (all three negative assertions) | Not swallowed and not best-effort: each reader is an `assert!` that panics the test. But the writer/reader link is a hand-copied string literal with no test asserting the writer still emits it — see P2-1 | const → three assert sites in one file; no build-time or test-time link back to the producing `println!` |

### Findings

- [P2] `VERIFY_OK_MARKER` is a test-local copy of production text that nothing in the suite anchors, so a future reword of the marker that keeps the letters `OK` makes all three negative assertions silently vacuous rather than red. The two positive sites still assert the bare substring — `assert!(stdout.contains("OK"))` at `verify_contract.rs:138` and `:149` — so they would keep passing through such a reword; and the plain-text marker is covered by no compatibility gate (`git grep 'ledger chain OK' aff09ad -- crates/ docs/` finds it only in `cmd_verify.rs:85,88` and this test file; `compat.stable-json-surfaces` pins only the `--json` keys, `verify_contract.rs:70-75`), so it may be reworded freely. Cheapest anchor is one line each: `stdout.contains(VERIFY_OK_MARKER)` at `:138` and `:149`. **Answering brief item 2 directly**: leaving the positive sites alone is *defensible against the bug class the issue names* — I verified the argument rather than accepting it, and it holds (`cmd_verify.rs:84-92` interpolates no path on the success arms; `main.rs:1107-1111` sends `tracing` to `std::io::stderr` at a default `warn` filter, and `main.rs` contains zero bare `println!`, so nothing can inject a path ahead of it on stdout) — but it is a half-migration on a second axis the PR body does not weigh: symmetry here is not cosmetic, it is what would make the constant self-anchoring. Not blocking, because the `must not report success` contract's real gate is the exit code, asserted **before** the text in all three tests (`:162`, `:200`, `:339` — `assert_eq!(code, 2, …)`), so even fully vacuous text assertions leave GH-647's P0 enforced.
- [P2] The new regression test is a parallel test, not a pin on the two sites under review. It carries its own copy of the assertion, so reverting `:172` or `:342` to the bare `!report.contains("OK")` turns nothing red — the sibling at `:155` uses a plain `tempfile::tempdir()` (`:156`), which reproduces the flake only when the random suffix is unlucky, exactly as before. What the new test does pin is the *property* and the constant's correctness, which is what doneWhen 3 asks for; the site-level pin is a gap, not a miss. Same class of gap the #1108 round recorded at P2, weighed the same.
- [P2] `report.contains("ledger")` at `:168` remains a positive substring check on a blob that carries the workspace path, so a workspace path containing `ledger` would satisfy it independently of what the command printed. The PR's ruling on it is correct for the direction GH-1033 is about (accidental false *failure*); this residual runs the other way (accidental false *pass*) and is materially weaker. Recorded, not blocking.

### FOLLOW-UP ISSUE

- **Anchor `VERIFY_OK_MARKER` to the product it copies** (P2-1 + P2-2): switch `verify_contract.rs:138` and `:149` to `stdout.contains(VERIFY_OK_MARKER)` so a reword of `cmd_verify.rs:85/88` turns the success tests red instead of leaving three negative assertions vacuous, and consider asserting the marker's presence in the same test that asserts its absence. Out of GH-1033's ceiling — its doneWhen names only the deleted-ledger test's own neighbours — so it does not extend this PR.
- **`REVIEW.md` C3's N.A. condition does not cover `tests/*.rs` integration targets.** C3 says a candidate "is N.A. only when the reported line is inside a `#[cfg(test)]` module", but `.claude/CLAUDE.md` §3.3 permits `unwrap`/`expect` in tests generally, and an integration-test target under `crates/*/tests/` is compiled only under `cargo test` while containing no `#[cfg(test)]` module. A mechanical reader must therefore FAIL candidates the conventions plainly allow — this round hit exactly that, with 3 candidates in `crates/edda-cli/tests/verify_contract.rs`. Evidence: `REVIEW.md@c9e2930` §5.3 C3 vs `.claude/CLAUDE.md` §3.3. Unrelated to this diff.

### RAN vs READ

RAN (all read-only, on the shared `main` checkout, which was **not** modified — no checkout, no worktree, no `gh pr checkout`, no cargo, no build, no test run):
- `gh pr view 1110 --json …`, `gh pr diff 1110 [--name-only]`, `gh pr checks 1110`, `gh run view 34379490835`, `gh issue view 1033`
- §3 router extracted from `REVIEW.md@c9e2930` by its two marker lines and run on the file list → `classes=code-plain`, `canonical_class=code-risk`
- U2/U3/U4 enumerators; §1 DECISIONS enumerator (empty); C2/C3/C4 content enumerators (three-dot, per v1.6); C5 selector (empty)
- `git show aff09ad:{crates/edda-cli/src/cmd_verify.rs, crates/edda-cli/tests/verify_contract.rs, crates/edda-ledger/src/paths.rs, crates/edda-ledger/src/sqlite_store/schema.rs}`; `git grep` for `contains("OK")` and `ledger chain OK` across `crates/` and `docs/` at the head SHA
- `git rev-parse aff09ad^`, `git merge-base origin/main aff09ad`, `git diff --no-renames c9e2930..origin/main -- crates/edda-cli/tests/verify_contract.rs`
- `sh scripts/wiring-scan.sh origin/main aff09ad` → no new pub surfaces, no swallow patterns
- `sh scripts/lint-markdown-content.sh` → exit 0 (U5)

READ — L1 receipt: **`CI run 34379490835 @ aff09ad732d477fbc9d4d9aed3c7a8c8962807df`** (exact head). Format pass; Clippy pass on ubuntu / macos / windows; MSRV (1.91.0) pass; `Test (ubuntu-latest)` pass; `Test (macos-latest)` pass; `Test (windows-latest)` pass (11m51s). `gh pr checks 1110` exits 0 with every job `pass` or `skipping`, and `CI Gate` passes. Also on this SHA: `sdk-contract` run 34379490876 pass, `Workspace publish dry-run` run 34379490810 pass. `Fleet shell tests` correctly `skipping` (no shell paths in the diff, `ci.path-filter`).
What that covers here: `Test (windows-latest)` runs `cargo test -p edda` (`.github/workflows/ci.yml:182`), which is the target containing `crates/edda-cli/tests/verify_contract.rs`, so all three OSes actually execute the changed and added tests. The usual Windows coverage gap does not apply to this PR and C5 is empty — no local Cargo gate was owed, and none was run. No recorded gate was rerun.

### Verdict

LGTM (P0=0, P1=0)

Three P2s recorded, none blocking (§8: P2 alone never blocks), and both are routed as FOLLOW-UP ISSUEs above rather than as extensions of this PR. `escalations:` is empty, so this verdict is not provisional. The fix is precise in both directions: it cannot false-fail on a temp-path substring any more, and the marker it substitutes is exactly what both success arms print and nothing else does.

*(Per this round's brief I set no label and no commit status; `fleet:reviewed` is left to the controller. Merge remains the operator's action.)*

Issue #1099, opened 2026-09-09
fix(fleet): the pre-commit hook builds outside the assigned build lane — every lane commit leaks a 791MB target/ into its worktree


fagemx:
Evidence: hit twice on 2026-09-09 in one wave, by two different actors, on the same worktree — 791 MB of stray build cache each time.

## 量測

A lane sets its assigned build lane the way `.claude/CLAUDE.md` Build lanes prescribes:

```bash
export CARGO_TARGET_DIR="$LOCALAPPDATA/fleet-workstation/lanes/worker-1"
cargo test -p edda        # correct — builds in worker-1
git commit -m "..."       # pre-commit hook fires
```

The hook's `cargo fmt --all --check` and `cargo clippy -p edda --all-targets -- -D warnings` run in a **fresh shell that does not inherit that export**, so they fall back to the worktree's default `target/`. Result on `C:/ai_agent/edda-wt-gh1066` today:

| actor | how | stray `target/` |
|---|---|---|
| controller, preserving a dead lane's work | one `git commit` | 791 MB |
| replacement fix lane, same worktree | one `git commit` | 791 MB (again, after the first was deleted) |

Both were noticed and deleted by hand. Nothing in the repo notices or reclaims them.

## Why it matters

`.claude/CLAUDE.md` Build lanes exists because an audit of this workstation counted 15 ad-hoc target directories and **~194 GB**. Its rule is 「Never create ad-hoc `CARGO_TARGET_DIR`s per round, per SHA, or per timestamp」 and 「Fleet sessions that build locally use only the lane named in their brief … for their whole lifetime」.

The pre-commit hook violates that rule on the lane's behalf, silently, on every commit a lane makes in a worktree. A lane that follows its brief exactly still leaks a full workspace target directory — and a lane that commits after each edit, which is what GH-748's delivery-hygiene fix asks for, leaks it on the *first* commit and then keeps that directory warm for the rest of the round.

So the two disciplines are in direct tension right now: commit often (to survive process death) versus do not create ad-hoc target directories. Today's wave was told to do the first and consequently did the second.

This is cost, not correctness — build output is disposable. But 791 MB per worktree per round is exactly the shape of the 194 GB pathology, and the fix is small.

## Suspected surface

`lefthook.yml` (the hook definitions) — the cargo steps need the lane's target directory, or need to not build at all.
`.claude/CLAUDE.md` Build lanes — whichever mechanism lands gets a sentence, since today the section reads as though `export` is sufficient.
[...]

Issue #1085, opened 2026-09-08
test(edda-bridge-claude): render_coordination_protocol_with_matches_original straddles a second boundary — two clock reads in one assertion, no injected instant


fagemx:
## What was observed

`peers::tests::render_coordination_protocol_with_matches_original` (crates/edda-bridge-claude/src/peers/tests.rs:1925) failed on CI `Test (ubuntu-latest)`, run 34265389356 @ 1af6e244 (PR #1084 — a shell/docs-only PR that does not touch this crate). The assertion compares a precomputed render against a fresh render and the two differ by exactly one substring: `billing (0s ago)` vs `billing (1s ago)`. The two renders straddled a wall-clock second boundary, so the relative-age formatter produced different text for the same peer.

Same class as GH-1031 / GH-1071 / GH-1078: an assertion over wall-clock-derived output with no injected clock, green on rerun.

## Predicted surface

- `crates/edda-bridge-claude/src/peers/tests.rs` — the test (and any sibling comparing two independently-timestamped renders)
- Whatever renders the `(Ns ago)` age — take the timestamp (or a `now` parameter) as an input so both renders in the test share one instant, instead of each calling the clock.

## doneWhen

- The test cannot fail from a second-boundary straddle: both renders in the comparison are computed from the same injected instant.
- The test still fails if the precomputed variant genuinely diverges from the original render (the property it exists to hold).
- `cargo test -p edda-bridge-claude peers::tests::render_coordination_protocol_with_matches_original` green in a loop (e.g. 20 iterations) locally.

## Evidence

CI run 34265389356, job `Test (ubuntu-latest)`: `assertion left == right failed: precomputed variant should match original` — left `0s ago`, right `1s ago`; 658 passed, 1 failed. Rerun of the failed job requested by the controller (environmental red per the verification ladder).

Issue #1047, opened 2026-09-07
fix(edda-cli): reconcile holds the workspace lock across git worktree add — measured 28-65s critical section against a fixed 30s wait budget, so simultaneous_reconciles flakes and hook writes are silently dropped


fagemx:
## What happened

`persist_reconciliation` holds the workspace lock across `ensure_attempt_worktree`,
i.e. across `git worktree add`. That call is the entire critical section, and its
cost is environment-dependent and unbounded. A second reconciler waits a fixed
30 s (`WORKSPACE_LOCK_WAIT_BUDGET`) and then fails. When the worktree checkout
costs more than 30 s, the loser reports the workspace as permanently locked and
the caller sees an error instead of a benign no-op.

Reproduced on a Windows workstation with real-time AV scanning, running the test
alone (644 other tests filtered out), with temporary instrumentation. Two
consecutive runs, same command, opposite outcomes:

Run 1 — FAILED:

```
[DIAG] thread 1 done at 30.0281778s -> Err("workspace is locked by another process (...\repo\.edda\LOCK)")
[DIAG] thread 0 done at 64.8265131s -> Ok(1)
[DIAG] all joined at 64.8268687s
[DIAG] post-join probe at 64.8272177s -> Ok("free")
```

Run 2 — passed, by 2 seconds of margin:

```
[DIAG-CS] views+leases              887.7µs
[DIAG-CS] discover_active_peers     7.1725ms
[DIAG-CS] ensure_attempt_worktree   27.9919924s
[DIAG-CS] rebuild_branch            4.7054ms
[DIAG-CS] critical section total    28.0170574s
```

Two facts follow directly from those numbers:

1. **`git worktree add` is 99.9% of the critical section** — 27.99 s of a 28.02 s
   hold. Every ledger operation together is under 10 ms.
2. **The lock is released correctly.** The post-join probe in run 1 reports the
   lock free immediately after both threads join, so the "release the lock" half
   of the test is not the defect. The failure is the loser's wait budget.

GH-524 set the 30 s budget against a critical section measured at 3-5 s, with
`git worktree add` alone at 1.8-4.2 s. The measurements above are 28 s and 65 s,
so the budget now sits in the middle of the observed distribution and the
outcome is a coin flip.

### The CI occurrence on macOS is not proven to be this mechanism

The failure that prompted this issue was `Test (macos-latest)` in CI run
34102816739 at `3306c1a2d9a03d022dbfabd61755ed549131acc7`, same error string, with
[...]

PR #1112, opened 2026-09-09
fix(edda-cli): fake-Codex handshake and periodic-renewal waits stop racing a load-calibrated ceiling (GH-1078, GH-1031)

### LINKED ISSUE #1031: test(edda-cli): cmd_reconcile::tests::runner::periodic_renewal_stops_old_runner_before_failure_after_lease_replacement is flaky on windows-latest — failed once on a scripts-only branch (PR #1017 run 34048211116) and passed on rerun
## What happened

期望：`Test (windows-latest)` 對零 Rust 改動的分支穩定綠。實際（PR #1017 head `8a66e6859735916230fad2f0660ecfae1ccc715f`，run 34048211116，2026-09-06 17:23Z）：`cmd_reconcile::tests::runner::periodic_renewal_stops_old_runner_before_failure_after_lease_replacement` 失敗 1／561；同一 run `gh run rerun --failed` 後 17:42Z 全綠。該 PR 的改動只在 `scripts/fleet/` 與 `crates/edda-ledger`，不碰 `cmd_reconcile`。

## Why it matters

Windows 測試子集是每張 PR 的必經閘門（CI Gate），一個計時類 flake 每次都要人手 rerun 才能過；依帳本 `fleet.ci-flake-carrier`，flake 一測試一單、不設登記處。

## Predicted surface

`crates/edda-cli/src/cmd_reconcile/tests/runner.rs`

## doneWhen

- 測試不依賴牆鐘計時競速（以可控的時鐘／事件序列取代 sleep 與真實時間比較），在 windows-latest 連續 20 次 `cargo test -p edda cmd_reconcile::tests::runner::periodic_renewal_stops_old_runner_before_failure_after_lease_replacement` 全綠。
- 若確認為真實競態（非測試問題），修在 `crates/edda-cli/src/cmd_reconcile/` 並附回歸測試。

#924 為另一個 Windows flake（edda-search-fts），同類不同測試；open-issue 搜尋 `periodic_renewal` 無既有單。

### LINKED ISSUE #1078: test(edda-cli): fake-Codex handshake tests fail 9/20 and 6/20 under host load — a fixed 30s startup budget calibrated on an idle host, the same anti-pattern as GH-1047
## What happened

Two `cmd_reconcile` tests that drive the fake Codex harness fail under host load
and pass in isolation. Measured on a Windows workstation on 2026-09-07 while
soaking PR #1065 (20 consecutive `cargo test -p edda --bin edda cmd_reconcile`
runs at default parallelism, 87 tests each):

| Test | Failures in 20 runs |
|---|---|
| `fake_runner_records_session_...` | 9 |
| `fake_runner_resumes_current_attempt_...` | 6 |
| `periodic_renewal_stops_old_runner_before_failure_after_lease_replacement` | 4 — already filed as #1031 |

The first two are not covered by any open issue. The failure message is
`unexpected EOF from Codex App Server`.

The split across the run sequence is the evidence for the cause, not incidental:

- runs 1-10: wall clock 408-636 s, 9 failures
- runs 11-20: wall clock 239 s falling to 93 s, **10 consecutive clean runs**,
  87/87 each

Runs 1-10 competed with a concurrent `cargo test -p edda` started by another
session sharing the same target directory on this box; that contention also
broke one link outright with `rust-lld: ... permission denied`. When the
competing session stopped, the wall clock fell by 4-6x and the failures stopped
with it.

All four fake-Codex tests serialize on `FAKE_CODEX_LOCK` and are wall-clock
bound around a spawned `powershell.exe`. Under load the spawn plus handshake
exceeds the window and the harness reports EOF.

No lock, claim, planner, or dispatch test failed once across the 20 runs.

## Why it matters

These tests are inside the CI Windows subset (`edda` is one of the 7 crates
whose test target runs on windows-latest), so they can redden a PR that changed
nothing near them. That is the same false-red economics #1031 was filed for:
the merge signal degrades from "this change is bad" to "roll again", and every
PR on the branch inherits the ambiguity.

They also make a repetition-based acceptance criterion unsatisfiable for
anything else in the module. #1047's doneWhen 5 asks for 20 consecutive green
`cmd_reconcile` runs; the contention tests it exists to protect were 20/20, but
the module could not go 20-for-20 because of these neighbours. Any future issue
that soaks this module inherits the same floor.

## Predicted surface

`crates/edda-cli/src/cmd_reconcile/tests/runner.rs`
`crates/edda-cli/src/cmd_reconcile/tests/mod.rs`

## doneWhen

- The two tests no longer depend on a wall-clock window around a spawned
  process: either the handshake waits on a condition with a generous ceiling
  rather than a deadline sized for an idle host, or the harness is replaced by
  an in-process fake that does not spawn.
- The fix is verified on a **loaded** host, since an idle host does not
  reproduce: run the module soak while a second `cargo` build occupies the same
  machine, and report the tally.
- `cargo test -p edda --bin edda cmd_reconcile` passes 10 consecutive runs under
  that induced load with zero failures in these two tests. (Ten, not twenty, and
  named tests rather than the whole module — a criterion nobody will run is a
  defect in the criterion; see the amendment note on #1047.)
- If #1031 shares the mechanism, it is closed by the same change or explicitly
  distinguished in the PR body.

- #1031 is the third test in the same family, filed from a CI occurrence on
  windows-latest. This issue does not duplicate it: it covers the two tests that
  have no issue, and supplies the local measurement and the cause that #1031
  lacks. Whoever fixes either should read both.
- #1047 is the lock-hold defect whose soak produced this data. Distinct
  mechanism — that one is lock contention in product code, this one is a test
  harness timing window. #1047's PR #1065 is where the tally was recorded.
- **This EOF flake has been fixed once already, the same way #1047's was.** The
  ledger records it on 2026-08-16: a task titled "TDD-only cmd_reconcile fake
  Codex observer startup budget root-cause fix", completed at `84cbbdd` with
  the receipt "fake Codex startup gates use one bounded 30s test budget", after
  the flake was disclosed in a task report the day it was found. So the current
  behaviour is not a new defect — it is a fixed 30 s budget, calibrated on an
  idle host, being outrun under load.

  That is the identical shape as #1047, where a 30 s `WORKSPACE_LOCK_WAIT_BUDGET`
  set against a 3-5 s measurement was outrun by a 28-65 s critical section. Two
  independent occurrences of one anti-pattern in the same module is the reason
  this issue asks for a condition rather than a larger number: raising 30 s to
  60 s would be the third instance, not a fix.

- Dedupe run: GitHub issue search for the two test names and for fake-codex
  flake, `edda ask "fake-codex"`, `edda search query "fake-codex handshake
  flake"` — the last of which is what surfaced the 2026-08-16 history above.

fagemx:
## Summary

Two commits, one per issue, both in `crates/edda-cli/src/cmd_reconcile/tests/`:

- **GH-1078** (`1bbcbe7`): the fake-Codex handshake observer (`tests/mod.rs`) stops treating a load-calibrated 30s ceiling as its own assertion, and `fake_runner_resumes_current_attempt_after_slow_startup_before_turn`'s deliberate "slow startup" sleep no longer eats into that ceiling before the observer starts watching.
- **GH-1031** (`c01f073`): `periodic_renewal_stops_old_runner_before_failure_after_lease_replacement`'s two waits are converted to the same helper, closing the fixed-2.5s renewal-detection window that was the tightest instance of the same shape in this file.

## Round 1 fix (`47232b8`)

Round 1 review found the GH-1031 commit above did not fix GH-1031: widening a
poll budget cannot help when the polled condition (a lease renewal) is capped
by the runner's own turn lifetime, not by the budget — once the fake turn's
fixed `Start-Sleep` resolves, `finish_runner` deletes the lease and the
condition becomes permanently unobservable, not merely late. `47232b8` gates
the fake turn at the `EDDA_FAKE_CHALLENGE`/`ALLOW`/`DENY` checkpoint
`allow_fake_turn_after_durable_session` already uses, so the turn cannot
resolve until the test has already observed a renewal and released it — the
renewal becomes an event this test waits on instead of a deadline it races.
It also restores `fake_runner_resumes_current_attempt_after_slow_startup_before_turn`'s
"slow startup" sleep to the position `84cbbdd` deliberately placed it (after
the observer starts polling), which Round 1 found the first commit had moved
in a way that deleted the coverage it was added to pin. Full reasoning in the
`Review Response: Round 1` comment on this PR.

## Evidence standard: structural, not measured

Per the controller's own measurement at this SHA on this host, `cargo test -p edda --bin edda cmd_reconcile` was run 11 times and did not reproduce either report:

| condition | runs | failures among the three named tests | wall clock |
|---|---|---|---|
| idle host | 8 | **0** | 37–61 s |
| under a full-workspace rebuild **and** a PowerShell spawn storm | 3 | **0** | 80–137 s |

No local reproduction was attempted here (none was possible on this host, and the brief explicitly prohibited generating load or looping tests to try). Everything below argues from the code.

## The sites, sorted (six, not five)

The original count below inventoried five poll-with-deadline *sites*, all in
Rust. Round 1 correctly found a sixth *consumer of the constant* missing from
it — added at the end, in its own category since it isn't a Rust
poll-with-deadline site in the same sense as the other five.

- **`tests/mod.rs` observer loop** (`allow_fake_turn_after_durable_session`) — **anti-pattern, fixed.** It bailed with `"fake never challenged turn gate"` as soon as its own 30s deadline passed, even though nothing was necessarily broken: reaching the challenge file requires `Command::spawn` of `powershell.exe` plus the full JSON-RPC handshake plus `record_session_if_current` (workspace lock + ledger append) to have already completed. GH-1078's own diagnosis is "spawn plus handshake exceeds the window."
- **`runner.rs` session-deadline loop** (in the GH-1031 test) — **anti-pattern, fixed.** Same shape, same constant, same class of risk (it waits for the same `task.session` event the observer above waits to react to), and GH-1078 explicitly invites this: *"If #1031 shares the mechanism, it is closed by the same change or explicitly distinguished in the PR body."* It shares the mechanism.
- **`runner.rs` periodic-renewal loop** (the `for _ in 0..100 { sleep(25ms) }` in the same GH-1031 test) — **anti-pattern, fixed in the first commit but not resolved by it; genuinely fixed in `47232b8`.** This one isn't about spawn/handshake at all: with `lease_ttl_s = 1`, `run_turn_with_renewals` renews on a `tokio::time::interval` around 1s. The first commit converted the fixed `for` loop to `poll_until` with the same 180s ceiling as the other sites, which Round 1 showed does not touch the actual constraint: the renewal is capped by the fake turn's own fixed length, not by the poll budget. `47232b8` gates the turn instead, per the Round 1 fix section above.
- **`runner.rs:8` `await_gate`** — **sound, left alone.** Its 30s ceiling is a pure hang-safety-valve: on timeout it just returns (no `assert`/`bail` of its own), so a slow host costs the *simulated stall* some fidelity, never a false failure. Round 1 verified this characterization directly.
- **`runner.rs:526` `spawn_deadline`** (in `first_spawn_failure_does_not_prevent_later_plan_launch`) — **not sound; left alone on scope grounds, not soundness.** Round 1 caught this: `anyhow::ensure!(Instant::now() < spawn_deadline, ...)` fails the test on timeout, the identical deadline-as-assertion shape fixed at the other three sites above — the load-calibrated reasoning previously given here for leaving it alone (a trivial `.cmd` spawn, already tuned once for GH-524) is exactly the reasoning GH-1078 names as the anti-pattern, not a reason this site is structurally different. The only defensible reason to leave it untouched: neither issue names this test, it has no reported failure, and per `fleet.ci-flake-carrier=product-issue-per-test-no-registry` an unflaked test earns its own issue if and when it flakes, not a preemptive fix bundled into this PR.
- **`runner.rs:816` fake-Codex `STARTUP_BUDGET_MS` gate (PowerShell, substituted from `FAKE_CODEX_STARTUP_BUDGET`)** — the sixth consumer of the constant, missing from the count above. This is the fake process's own wait for `EDDA_FAKE_ALLOW`/`EDDA_FAKE_DENY` after it writes the challenge file (`tests/runner.rs:814-823`); its expiry is what produces the literal `unexpected EOF from Codex App Server` message GH-1078's own report names. It moved 30s to 180s alongside the Rust-side constant it shares a name with, so behavior stays consistent, but it bounds the fake's own gate loop rather than a Rust `Instant`-based poll, and belongs in this inventory regardless.

## Why widening the constant is not "the third instance"

`FAKE_CODEX_STARTUP_BUDGET` moved from 30s to 180s (`tests/mod.rs`), which superficially repeats the move GH-1078's own text calls out as insufficient by itself ("raising 30 s to 60 s would be the third instance, not a fix"). The distinguishing argument, not just a bigger number:

- The direct precedent in this codebase (`plan.rs:511-516`, `WORKSPACE_LOCK_WAIT_BUDGET`) shows what a *real* fix to this shape looks like: GH-1047's `git worktree add` was measured at 27.99s and 64.8s under contention against a 30s budget, and the fix did **not** raise that number — it moved the slow git I/O out of what the budget bounds, so the still-30s budget now only has to cover sub-10ms SQLite operations.
- There is no equivalent move available for the fake-Codex observer: its entire job is to *observe* the spawn+handshake+session-record chain complete, so nothing can be moved out from under it without changing what the test verifies.
- Given that, 180s is chosen to be a magnitude qualitatively different from a load-calibrated timeout — comfortably above GH-1047's own worst measured single-subprocess figure (64.8s) for a chain with more steps than that one (spawn *and* a multi-round-trip handshake *and* a lock-guarded ledger write) — so it functions as a hang detector (only a genuinely wedged handshake trips it) rather than a tuned assertion a busy-but-alive host can still lose to.

For the periodic-renewal wait specifically, this reasoning does not transfer — see the Round 1 fix section above and `47232b8`, which decouples that observation from the constant's size entirely rather than relying on it being large.

## Setting the pattern

Added `tests/mod.rs::poll_until` (Windows-gated, alongside `FAKE_CODEX_STARTUP_BUDGET`): poll a fallible condition on an interval until it holds or a budget elapses, returning whether it held rather than asserting itself. All three fixed sites above now go through it, so the next `cmd_reconcile` test with this shape has one place to reach for instead of a fourth hand-rolled deadline loop. The doc comments on both the constant and the helper spell out the distinction this PR is drawing (hang-safety-valve vs. load-calibrated ceiling) so it doesn't have to be re-derived.

Checked against the other four issues named in the brief as sharing this shape:

- **GH-1071** (`crates/edda-cli/src/cmd_review/evidence.rs`) — same crate, different module; `poll_until` is `pub(super)` inside `cmd_reconcile::tests` and doesn't reach it without promotion. GH-1071's own doneWhen also points at a different fix shape (an injected `--now`/`Facts` seam for a deadline *assertion*, not a spawn-observation poll), so the helper wouldn't be the right tool there even if visible.
- **GH-1085** (`crates/edda-bridge-claude`) — different crate and a different mechanism entirely (two independent clock reads straddling a wall-clock second boundary in a rendered string, not a poll-with-deadline). Doesn't transfer.
- **GH-924** (`crates/edda-search-fts`) — different crate; a Windows file-locking error (`os error 5`), not examined here. Not this shape as far as this PR's scope goes.
- **GH-1047** — closed already (`WORKSPACE_LOCK_WAIT_BUDGET`/`plan.rs`); it's the precedent cited above, not open work.

Not fixed here, per the brief.

## Gates (L0, touched crate only)

Lane: `worker-2` (`CARGO_TARGET_DIR=%LOCALAPPDATA%\fleet-workstation\lanes\worker-2`).

First two commits, run separately, each on the state at that commit:

- `cargo fmt --all --check` — clean
- `cargo clippy -p edda --all-targets -- -D warnings` — clean, 0 warnings
- `cargo test -p edda` — 749/749 unit tests passed (0 failed) both times, including all five touched `cmd_reconcile::tests::runner` tests by name; all `tests/*.rs` integration binaries green
- `sh scripts/lint-file-length.sh --tree` — clean

Round 1 fix commit `47232b8`, same lane, on top of the above:

- `cargo fmt --all --check` — clean
- `cargo clippy -p edda --all-targets -- -D warnings` — clean, 0 warnings
- `cargo test -p edda` — 749/749 unit tests passed (0 failed; test count unchanged, two existing tests modified), all `tests/*.rs` integration binaries green
- `sh scripts/lint-file-length.sh --tree` — clean

`cargo test --workspace` was not run on any commit (out of budget per the brief). The pre-commit hook's own `cargo fmt`/`clippy` re-ran outside the lane's `CARGO_TARGET_DIR` on the first two commits (known gap #1099) and left a stray `target/` in the worktree each time; deleted after each commit as instructed. The round 1 fix commit used the hook's own `SKIP_CLIPPY=1` escape hatch after running clippy warm in the lane, so it left no stray `target/`. No local reproduction loop was run on any commit's code, per the brief.

## Undone / left for the controller

- The written convention (hang-safety-valve vs. load-calibrated ceiling) and the `poll_until` shape are stated above for GH-1071/GH-924/GH-1085 to reuse, but no code in those crates was touched.
- Round 1's follow-up finding — `run_turn_with_renewals` performs blocking `Ledger::open` + `renew_lease` directly inside a `tokio::select!` arm rather than through `spawn_blocking` (`runner.rs:479-483`), which could delay a renewal tick under real lock contention independent of anything this PR touches — is evidenced but out of this PR's frozen surface (test files only) and was not examined or fixed here.
- CI run id for the round 1 fix head will follow once exact-head CI reports on this PR.

Issue: #1078, #1031
Closes #1078

### DIFF
--- a/crates/edda-cli/src/cmd_reconcile/tests/mod.rs
+++ b/crates/edda-cli/src/cmd_reconcile/tests/mod.rs
@@ -12,9 +12,48 @@ mod scheduler;
 
 #[cfg(windows)]
 pub(super) static FAKE_CODEX_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+/// A hang-safety-valve, not a tuned deadline. The waits that consume this
+/// budget (spawning `powershell.exe`, the JSON-RPC handshake, and recording
+/// the durable session under the workspace lock) are the same shape as
+/// `WORKSPACE_LOCK_WAIT_BUDGET` before GH-1047: a fixed ceiling sized as a
+/// multiple of idle-host timing. GH-1047's `git worktree add` was measured at
+/// 27.99 s and 64.8 s under contention against a 30 s budget — its fix did
+/// not raise that number, it moved the slow operation out of what the budget
+/// bounds. There is no equivalent move here (the wait's whole job is to
+/// observe this chain complete), so per GH-1078 ("spawn plus handshake
+/// exceeds the window") this constant is instead set to a magnitude that
+/// contention on this workstation is not expected to reach — comfortably
+/// above GH-1047's own worst measured single-subprocess figure for a chain
+/// with more steps than that one — so a slow host costs latency, never a
+/// failure, and only a genuinely wedged handshake still trips it.
 #[cfg(windows)]
 pub(super) const FAKE_CODEX_STARTUP_BUDGET: std::time::Duration =
-    std::time::Duration::from_secs(30);
+    std::time::Duration::from_secs(180);
+
+/// Poll `condition` every `poll` until it reports `true` or `budget` elapses.
+/// `budget` must be a hang-safety-valve (see `FAKE_CODEX_STARTUP_BUDGET`), not
+/// a value calibrated to idle-host timing — the anti-pattern this exists to
+/// replace let a load-calibrated ceiling double as the test's assertion, so
+/// exceeding it failed the test rather than merely arriving late. Returns
+/// whether the condition held before the deadline; the caller decides what
+/// "false" means (assert, bail, or something softer).
+#[cfg(windows)]
+pub(super) fn poll_until(
+    budget: std::time::Duration,
+    poll: std::time::Duration,
+    mut condition: impl FnMut() -> anyhow::Result<bool>,
+) -> anyhow::Result<bool> {
+    let deadline = std::time::Instant::now() + budget;
+    loop {
+        if condition()? {
+            return Ok(true);
+        }
+        if std::time::Instant::now() >= deadline {
+            return Ok(false);
+        }
+        std::thread::sleep(poll);
+    }
+}
 
 pub(super) static DOORBELL_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
 
@@ -157,25 +196,24 @@ pub(super) fn allow_fake_turn_after_durable_session(
     deny: std::path::PathBuf,
 ) -> std::thread::JoinHandle<anyhow::Result<()>> {
     std::thread::spawn(move || {
-        let deadline = std::time::Instant::now() + FAKE_CODEX_STARTUP_BUDGET;
-        while std::time::Instant::now() < deadline {
-            if challenge.exists() {
-                let view = Ledger::open(&repo)?
-                    .task_views()?
-                    .into_iter()
-                    .find(|view| view.task_id == task_id);
-                let valid = view.is_some_and(|view| {
-                    view.session_agent_kind.as_deref() == Some("codex")
-                        && view.session_attempt == Some(attempt)
-                        && view.session_id.as_deref() == Some("fake-thread")
-                });
-                std::fs::write(if valid { allow } else { deny }, "gate")?;
-                anyhow::ensure!(valid, "fake observed turn before durable current session");
-                return Ok(());
-            }
-            std::thread::sleep(std::time::Duration::from_millis(10));
-        }
-        anyhow::bail!("fake never challenged turn gate")
+        let seen = poll_until(
+            FAKE_CODEX_STARTUP_BUDGET,
+            std::time::Duration::from_millis(10),
+            || Ok(challenge.exists()),
+        )?;
+        anyhow::ensure!(seen, "fake never challenged turn gate");
+        let view = Ledger::open(&repo)?
+            .task_views()?
+            .into_iter()
+            .find(|view| view.task_id == task_id);
+        let valid = view.is_some_and(|view| {
+            view.session_agent_kind.as_deref() == Some("codex")
+                && view.session_attempt == Some(attempt)
+                && view.session_id.as_deref() == Some("fake-thread")
+        });
+        std::fs::write(if valid { allow } else { deny }, "gate")?;
+        anyhow::ensure!(valid, "fake observed turn before durable current session");
+        Ok(())
     })
 }
 
--- a/crates/edda-cli/src/cmd_reconcile/tests/runner.rs
+++ b/crates/edda-cli/src/cmd_reconcile/tests/runner.rs
@@ -661,33 +661,33 @@ pub(super) fn periodic_renewal_stops_old_runner_before_failure_after_lease_repla
     let runner_config = config.clone();
     let runner = std::thread::spawn(move || run_task(&runner_repo, 1, 1, &runner_config, false));
 
-    let session_deadline = std::time::Instant::now() + FAKE_CODEX_STARTUP_BUDGET;
-    while std::time::Instant::now() < session_deadline {
-        if ledger
-            .task_events()?
-            .iter()
-            .any(|event| event.event_type == "task.session")
-        {
-            break;
-        }
-        std::thread::sleep(std::time::Duration::from_millis(25));
-    }
-    assert!(ledger
-        .task_events()?
-        .iter()
-        .any(|event| event.event_type == "task.session"));
+    let session_recorded = poll_until(
+        FAKE_CODEX_STARTUP_BUDGET,
+        std::time::Duration::from_millis(25),
+        || {
+            Ok(ledger
+                .task_events()?
+                .iter()
+                .any(|event| event.event_type == "task.session"))
+        },
+    )?;
+    assert!(session_recorded, "runner recorded the durable session");
     let after_session = ledger.task_lease(1)?.expect("session lease");
-    let mut saw_periodic_renewal = false;
-    for _ in 0..100 {
-        let current = ledger.task_lease(1)?.expect("current lease");
-        if current.heartbeat_at != after_session.heartbeat_at
-            || current.expires_at != after_session.expires_at
-        {
-            saw_periodic_renewal = true;
-            break;
-        }
-        std::thread::sleep(std::time::Duration::from_millis(25));
-    }
+    // Same hang-safety-valve as FAKE_CODEX_STARTUP_BUDGET, not a tuned
+    // deadline: with lease_ttl_s = 1 the runner renews roughly once a second
+    // (`(ttl_s / 2).max(1)` in run_turn_with_renewals), but that tick shares
+    // the tokio runtime with the fake process's I/O, so a loaded host can
+    // stretch the wait well past its nominal interval without anything being
+    // wrong (GH-1031).
+    let saw_periodic_renewal = poll_until(
+        FAKE_CODEX_STARTUP_BUDGET,
+        std::time::Duration::from_millis(25),
+        || {
+            let current = ledger.task_lease(1)?.expect("current lease");
+            Ok(current.heartbeat_at != after_session.heartbeat_at
+                || current.expires_at != after_session.expires_at)
+        },
+    )?;
     assert!(
         saw_periodic_renewal,
         "runner crossed a periodic renewal interval"
--- a/crates/edda-cli/src/cmd_reconcile/tests/runner.rs
+++ b/crates/edda-cli/src/cmd_reconcile/tests/runner.rs
@@ -628,6 +628,13 @@ pub(super) fn fake_runner_records_session_in_main_ledger_before_turn_and_fails_w
 #[test]
 pub(super) fn periodic_renewal_stops_old_runner_before_failure_after_lease_replacement(
 ) -> anyhow::Result<()> {
+    // This test writes process env (`EDDA_FAKE_*`) to gate the fake turn
+    // below, so it takes the crate-wide env guard alongside its own
+    // fake-Codex lock — otherwise it races any other test in this binary
+    // that mutates process env, the same reason
+    // `fake_runner_resumes_current_attempt_after_slow_startup_before_turn`
+    // takes both.
+    let _env = crate::test_support::env_guard();
     let _fake = test_lock(&FAKE_CODEX_LOCK);
     let dir = tempfile::tempdir()?;
     let repo = dir.path().join("repo");
@@ -653,53 +660,104 @@ pub(super) fn periodic_renewal_stops_old_runner_before_failure_after_lease_repla
     )?)?;
     append_started(&ledger, 1, 1, 1)?;
     ledger.upsert_task_lease(&lease(1, 1, "2026-08-16T02:00:00Z"))?;
-    let fake = fake_codex(dir.path(), 2, false)?;
+    let fake = fake_codex(dir.path(), 0, false)?;
     let mut config = ReconcileConfig::test_defaults();
     config.lease_ttl_s = 1;
     config.codex_bin = fake;
+
+    // Gate the fake turn at the same EDDA_FAKE_CHALLENGE checkpoint
+    // `allow_fake_turn_after_durable_session` uses (tests/mod.rs): the fake
+    // process writes the challenge file and then blocks — bounded by
+    // FAKE_CODEX_STARTUP_BUDGET — until this test writes `allow` or `deny`.
+    // Without this, the turn resolves on `fake_codex`'s fixed `Start-Sleep`
+    // and `finish_runner` deletes the lease the moment it does, so the
+    // periodic-renewal tick this test waits for below has to land inside
+    // whatever is left of that fixed window: a race whose loss makes the
+    // condition permanently unobservable rather than merely late, which no
+    // poll budget can reach (GH-1031). Gating the turn means
+    // `run_turn_with_renewals`'s ~1s interval (`(ttl_s / 2).max(1)` with
+    // `lease_ttl_s = 1`, runner.rs:473-474) keeps ticking for as long as the
+    // gate stays shut, so the renewal below is an event this test waits on
+    // instead of a deadline it races.
+    let challenge = dir.path().join("renewal.challenge");
+    let allow = dir.path().join("renewal.allow");
+    let deny = dir.path().join("renewal.deny");
+    std::env::set_var("EDDA_FAKE_CHALLENGE", &challenge);
+    std::env::set_var("EDDA_FAKE_ALLOW", &allow);
+    std::env::set_var("EDDA_FAKE_DENY", &deny);
+
     let runner_repo = repo.clone();
     let runner_config = config.clone();
     let runner = std::thread::spawn(move || run_task(&runner_repo, 1, 1, &runner_config, false));
 
-    let session_recorded = poll_until(
-        FAKE_CODEX_STARTUP_BUDGET,
-        std::time::Duration::from_millis(25),
-        || {
-            Ok(ledger
-                .task_events()?
-                .iter()
-                .any(|event| event.event_type == "task.session"))
-        },
-    )?;
-    assert!(session_recorded, "runner recorded the durable session");
-    let after_session = ledger.task_lease(1)?.expect("session lease");
-    // Same hang-safety-valve as FAKE_CODEX_STARTUP_BUDGET, not a tuned
-    // deadline: with lease_ttl_s = 1 the runner renews roughly once a second
-    // (`(ttl_s / 2).max(1)` in run_turn_with_renewals), but that tick shares
-    // the tokio runtime with the fake process's I/O, so a loaded host can
-    // stretch the wait well past its nominal interval without anything being
-    // wrong (GH-1031).
-    let saw_periodic_renewal = poll_until(
-        FAKE_CODEX_STARTUP_BUDGET,
-        std::time::Duration::from_millis(25),
-        || {
-            let current = ledger.task_lease(1)?.expect("current lease");
-            Ok(current.heartbeat_at != after_session.heartbeat_at
-                || current.expires_at != after_session.expires_at)
-        },
-    )?;
-    assert!(
-        saw_periodic_renewal,
-        "runner crossed a periodic renewal interval"
-    );
-    ledger.upsert_task_lease(&TaskLease {
-        task_id: 1,
-        attempt: 2,
-        owner: "replacement".into(),
-        expires_at: "2026-08-16T03:00:00Z".into(),
-        heartbeat_at: "2026-08-16T01:00:00Z".into(),
-    })?;
-    runner.join().expect("runner thread")?;
+    let outcome = (|| -> anyhow::Result<()> {
+        let session_recorded = poll_until(
+            FAKE_CODEX_STARTUP_BUDGET,
+            std::time::Duration::from_millis(25),
+            || {
+                Ok(ledger
+                    .task_events()?
+                    .iter()
+                    .any(|event| event.event_type == "task.session"))
+            },
+        )?;
+        anyhow::ensure!(session_recorded, "runner recorded the durable session");
+        let after_session = ledger.task_lease(1)?.expect("session lease");
+
+        let challenge_seen = poll_until(
+            FAKE_CODEX_STARTUP_BUDGET,
+            std::time::Duration::from_millis(10),
+            || Ok(challenge.exists()),
+        )?;
+        anyhow::ensure!(challenge_seen, "fake never reached the turn gate");
+
+        // The turn is held at the gate now, so it cannot resolve out from
+        // under this poll: `finish_runner` cannot delete the lease until
+        // the gate opens below, which happens only after this observation
+        // succeeds.
+        let saw_periodic_renewal = poll_until(
+            FAKE_CODEX_STARTUP_BUDGET,
+            std::time::Duration::from_millis(25),
+            || {
+                let current = ledger.task_lease(1)?.expect("current lease");
+                Ok(current.heartbeat_at != after_session.heartbeat_at
+                    || current.expires_at != after_session.expires_at)
+            },
+        )?;
+        anyhow::ensure!(
+            saw_periodic_renewal,
+            "runner crossed a periodic renewal interval"
+        );
+
+        ledger.upsert_task_lease(&TaskLease {
+            task_id: 1,
+            attempt: 2,
+            owner: "replacement".into(),
+            expires_at: "2026-08-16T03:00:00Z".into(),
+            heartbeat_at: "2026-08-16T01:00:00Z".into(),
+        })?;
+        Ok(())
+    })();
+
+    // Release the gate unconditionally: allow if the observation above
+    // succeeded, deny otherwise, so a failed observation doesn't also make
+    // the fake process (and the join below) wait out its own copy of
+    // FAKE_CODEX_STARTUP_BUDGET on top of the one the failure already
+    // consumed.
+    let _ = std::fs::write(if outcome.is_ok() { &allow } else { &deny }, "gate");
+
+    // Whichever way the runner notices from here — the next renewal tick
+    // seeing it no longer owns the lease (runner.rs:480-484), or the turn
+    // completing and `finish_runner` reading the now-replaced lease —
+    // `finish_runner` re-reads current ownership rather than trusting a
+    // value captured earlier, so both paths land on the same outcome the
+    // assertions below check for.
+    let run_result = runner.join().expect("runner thread");
+    std::env::remove_var("EDDA_FAKE_CHALLENGE");
+    std::env::remove_var("EDDA_FAKE_ALLOW");
+    std::env::remove_var("EDDA_FAKE_DENY");
+    outcome?;
+    run_result?;
 
     assert_eq!(ledger.task_lease(1)?.expect("replacement").attempt, 2);
     assert!(ledger
@@ -735,17 +793,20 @@ pub(super) fn fake_runner_resumes_current_attempt_after_slow_startup_before_turn
     std::env::set_var("EDDA_FAKE_CHALLENGE", &challenge);
     std::env::set_var("EDDA_FAKE_ALLOW", &allow);
     std::env::set_var("EDDA_FAKE_DENY", &deny);
-    // The deliberate delay that simulates a slow runner startup happens
-    // before the observer starts polling, not after: the observer's budget
-    // exists to cover the handshake it is actually watching for, not time
-    // this test spends manufacturing "slow startup" on purpose. Sleeping
-    // after the observer was already running used to spend part of its
-    // hang-safety-valve on nothing observable — the unconditional-wait
-    // anti-pattern this reordering removes (GH-1078).
-    std::thread::sleep(std::time::Duration::from_millis(2_100));
-
     let observer =
         allow_fake_turn_after_durable_session(repo.clone(), 2, 1, challenge, allow, deny);
+
+    // The deliberate "slow startup" delay runs after the observer has
+    // already started polling, not before: that is what makes this a slow
+    // *startup* the observer has to wait through, rather than a delay that
+    // finishes before anything is watching and pins nothing. `84cbbdd`
+    // (2026-08-17) added this sleep in this position deliberately, in the
+    // same commit that renamed this test to `..._after_slow_startup_...`;
+    // it deliberately spends part of the observer's FAKE_CODEX_STARTUP_BUDGET
+    // hang-safety-valve on nothing observable, which is exactly what pins
+    // that the budget is generous enough to absorb a slow handshake and not
+    // merely a fast one.
+    std::thread::sleep(std::time::Duration::from_millis(2_100));
 
     let run_result = run_task(&repo, 2, 1, &config, false);
     let observer_result = observer.join();
--- a/crates/edda-cli/src/cmd_reconcile/tests/runner.rs
+++ b/crates/edda-cli/src/cmd_reconcile/tests/runner.rs
@@ -735,10 +735,17 @@ pub(super) fn fake_runner_resumes_current_attempt_after_slow_startup_before_turn
     std::env::set_var("EDDA_FAKE_CHALLENGE", &challenge);
     std::env::set_var("EDDA_FAKE_ALLOW", &allow);
     std::env::set_var("EDDA_FAKE_DENY", &deny);
+    // The deliberate delay that simulates a slow runner startup happens
+    // before the observer starts polling, not after: the observer's budget
+    // exists to cover the handshake it is actually watching for, not time
+    // this test spends manufacturing "slow startup" on purpose. Sleeping
+    // after the observer was already running used to spend part of its
+    // hang-safety-valve on nothing observable — the unconditional-wait
+    // anti-pattern this reordering removes (GH-1078).
+    std::thread::sleep(std::time::Duration::from_millis(2_100));
+
     let observer =
         allow_fake_turn_after_durable_session(repo.clone(), 2, 1, challenge, allow, deny);
-
-    std::thread::sleep(std::time::Duration::from_millis(2_100));
 
     let run_result = run_task(&repo, 2, 1, &config, false);
     let observer_result = observer.join();


fagemx:
## Code Review: Round 1 — PR #1112 @ c01f073269871f3597b22b6d4cb43adf0b353ef7

- elapsed: unmeasured (no pi session file — this round ran as a controller-spawned subagent under `review.dispatch-transport=controller-subagent-direct`)
- model_observed: unverified (no `claude -p --output-format json` envelope available to this session; `PI_MODEL`-style self-labelling is not evidence)
- reviewer_session: not assigned by the brief (controller-spawned subagent)
- spec: review-spec-v1.6 (read at base `7781400e6cfcf1e7a8ababe678e410f2441b2159`; `REVIEW.md` is unchanged between base and head)
- escalations: none
- cost: read-only round — no Cargo, no checkout, no worktree, no build. Process note carried from the brief, non-blocking: the implementing lane self-reported launching one `cargo clippy` with `run_in_background` against a foreground-only instruction, caught it immediately and waited synchronously. Recorded here per `.claude/CLAUDE.md` § "Verification cost"; it is not a finding against the diff.

### IN SCOPE

- **Changed behaviour and paths.** `crates/edda-cli/src/cmd_reconcile/tests/mod.rs` and `crates/edda-cli/src/cmd_reconcile/tests/runner.rs` only (+93/−48). Three things change: `FAKE_CODEX_STARTUP_BUDGET` 30 s → 180 s (`tests/mod.rs:30-31`); a new `poll_until` helper (`tests/mod.rs:41-56`) that three hand-rolled deadline loops are converted to; and the 2 100 ms "slow startup" sleep in `fake_runner_resumes_current_attempt_after_slow_startup_before_turn` moves from *after* the observer spawn to *before* it (`tests/runner.rs:745` vs `:747`).
- **Direct consumers.** `poll_until` has three call sites, all in this PR: `tests/mod.rs:199-204`, `tests/runner.rs:664-674`, `tests/runner.rs:682-693`. `FAKE_CODEX_STARTUP_BUDGET` has a **fourth** consumer the PR body does not enumerate: it is substituted into the fake's PowerShell gate as `STARTUP_BUDGET_MS` (`tests/runner.rs:816`, `:828-831`), so the fake's own wait-for-allow ceiling moved 30 s → 180 s too.
- **Acceptance ceilings.** GH-1078 doneWhen 1–4; GH-1031 doneWhen 1–2. Both carry `Closes #N` in the PR body (lines 65–66), so `pr.closing-keyword=only-when-all-donewhen-delivered` binds on both.
- **Security / data loss.** None introduced or exposed — the diff is test-only and touches no ledger, store, dispatch or process-policy path.
- **Current-base integration.** `git diff 7781400e..origin/main` limited to the two changed files is **empty**; `main` advanced by two PRs since the branch point, neither over this surface. The reviewed tree is still the tree that would merge.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 1112 --name-only` → `crates/edda-cli/src/cmd_reconcile/tests/{mod,runner}.rs`; both inside GH-1078's Predicted surface, `runner.rs` inside GH-1031's |
| U2 | any | P1 | **FAIL** | body lines 65–66 `Closes #1078` / `Closes #1031`; GH-1031 doneWhen 1 is not delivered (Finding P1-1) |
| U3 | any | P1 | PASS | body carries `Issue: #1078, #1031` |
| U4 | any | P1 | PASS | enumerator empty over `origin/main..c01f073` — both subjects are `fix(edda-cli): …` |
| U6 | any | P0 | PASS | READ exact-head CI: run **34380793999 @ c01f073**, every job `pass` incl. `Test (windows-latest)` 11m33s, `Clippy` ×3 OS, `MSRV (1.91.0)`. Not deterministically red; nothing rerun. `edda --version` → `edda 0.5.0 (16f7b4b41cde-dirty)`; base `Cargo.toml` workspace `version = "0.6.0"` — recorded, no runtime claim in this diff depends on it |
| U7 | any | P1 | PASS | whole scoped audit finished before posting; every blocker below is batched into this round |
| U8 | any | P1 | N.A.(no `Decision:` line in the PR body — the DECISIONS enumerator prints nothing) |
| C1 | code-plain | P0 | **FAIL** | GH-1031 doneWhen 1 not met (Finding P0-1). GH-1078: doneWhen 1 met by its own either/or wording; doneWhen 2–3 superseded by ledger `test.flake-evidence=structural-when-not-reproducible`; doneWhen 4 met (PR body distinguishes #1031 explicitly) |
| C2 | code-plain | P0 | **FAIL** | two `fix:` subjects; `git diff origin/main...c01f073 -- 'crates/**' \| grep -cE '^\+.*(#\[test\]\|fn test_)'` → **0** (Finding P0-2) |
| C3 | code-plain | P1 | N.A.(the single candidate, `tests/runner.rs:685` `.expect("current lease")`, is inside `#[cfg(test)] mod tests;` — confirmed at `crates/edda-cli/src/cmd_reconcile/mod.rs:16-17`) |
| C4 | code-plain | P1 | PASS | enumerator empty — no `unsafe`, no `clippy::all`, no crate-level `#![allow]` |
| C5 | code-plain | P1 | N.A.(selector empty — `edda-cli` is inside CI's Windows 7-crate subset, so `Test (windows-latest)` already ran this crate's tests on the exact head. No uncovered crate; no local Cargo run owed) |
| §5.5 | any | — | PASS | one new surface, one row below, three readers in this PR |
| D1–D5 | docs | — | N.A.(§4 does not route `code-plain` to §5.1 — no `docs`/`skills` path in the diff) |
| S1–S3 | skills | — | N.A.(§4 does not route `code-plain` to §5.2) |
| R1–R5 | code-risk | — | N.A.(§4 routes §5.4 only for the `code-risk` REVIEW.md class; the classifier printed `classes=code-plain`. The added PowerShell text at `tests/runner.rs:816` is a Rust string literal in a `.rs` file, contains no `\|\|`/`&&` mix and no R1 command) |

### Wiring

|---|---|---|---|---|
| `poll_until(budget, poll, condition) -> anyhow::Result<bool>` (`pub(super)`, `#[cfg(windows)]`) | `crates/edda-cli/src/cmd_reconcile/tests/mod.rs:41-56` | `tests/mod.rs:199`, `tests/runner.rs:664`, `tests/runner.rs:682` — `sh scripts/wiring-scan.sh origin/main c01f073` reports "2 outside references"; not dead on arrival | Not swallowed: `condition()?` propagates errors; `Ok(false)` is returned, never logged-and-ignored. But **every one of the three callers turns `Ok(false)` into a test failure** (`ensure!` at `tests/mod.rs:204`, `assert!` at `tests/runner.rs:674` and `:691`), so the doc comment's "the caller decides what false means (assert, bail, or something softer)" describes an option no caller takes | Helper → three test call sites; no product layer reached. The constant it is documented against also reaches the fake's PowerShell gate (`tests/runner.rs:816`, `:828-831`), a layer the PR body's site inventory does not list |

### Findings

- **[P0] GH-1031 is not fixed: the periodic-renewal conversion widens a budget that is not the binding constraint, so no outcome can change.** The polled condition is "the lease's `heartbeat_at`/`expires_at` moved" (`tests/runner.rs:682-690`). Only `run_turn_with_renewals` can move it, and it renews **only while the turn is in flight**: `crates/edda-cli/src/cmd_reconcile/runner.rs:474-486` builds `tokio::time::interval((ttl_s/2).max(1))` = 1 s with `ttl_s = 1`, consumes the immediate first tick at `:475`, and the `select!` returns the moment the turn resolves (`:478`). The turn's length is fixed by the fake: `tests/runner.rs:656` builds it with `delay_s = 2`, i.e. `Start-Sleep -Seconds 2` at `tests/runner.rs:824`. So exactly one renewal tick (at ≈ +1 s) has to land inside a ≈ 2 s turn — a fixed ~1 s of slack that this PR does not touch. Once the turn resolves, `finish_runner` deletes the attempt-1 lease (`runner.rs:516-518`), after which `ledger.task_lease(1)?.expect("current lease")` panics inside the closure. Both before and after this change the observation is capped by the runner's own lifetime, not by the poll budget: **2.5 s → 180 s buys nothing**, it only changes the miss-case symptom from `assert!(saw_periodic_renewal, …)` to an `.expect("current lease")` panic, and widens the window in which that panic is the observed failure. GH-1031 doneWhen 1 asks for「以可控的時鐘／事件序列取代 sleep 與真實時間比較」— the delivered code still uses `std::thread::sleep` + `Instant::now()` comparison (`tests/mod.rs:46-50`), with a larger constant. — The PR body's justifying claim, that "there is no equivalent move available" of GH-1047's kind, is false at this site specifically: the GH-1047-shaped move is to take the fixed `Start-Sleep -Seconds 2` out of what bounds the observation, and both seams for doing so already exist one function away — the `EDDA_FAKE_CHALLENGE`/`ALLOW`/`DENY` event gate at `tests/runner.rs:814-823` (a controllable event sequence, which this test alone does not use) and the per-thread `LOCK_WAIT_BUDGET_OVERRIDE` injection pattern at `crates/edda-cli/src/cmd_reconcile/plan.rs:519-532`.
- **[P0] Two `fix:` commits, zero tests added, against a ledger decision that names a test as half the standard.** `git diff origin/main...c01f073 -- 'crates/**' | grep -cE '^\+.*(#\[test\]|fn test_)'` → `0`. `edda ask test.flake-evidence` → `structural-when-not-reproducible`, active: *"the evidence standard is structural argument from the code **plus a test that pins the ordering property**"*. This is not the statistical demand that decision bans — the tally half is correctly waived here — but the second half was not delivered. The PR's own central claim (#2 in its body) *is* an ordering property, and instead of pinning it the diff removes the pin that existed (next finding).
- **[P1] `Closes #1031` on the PR body would auto-close an unfixed flake.** Body line 66. `pr.closing-keyword=only-when-all-donewhen-delivered` (active, ratified) permits the keyword *exactly when* every doneWhen item is delivered; GH-1031 doneWhen 1 is not (previous finding). `Closes #1078` (line 65) is not part of this finding — #1078's doneWhen 1 explicitly licenses "a condition with a generous ceiling", 2–3 are superseded by `test.flake-evidence`, and 4 is satisfied by the body's explicit distinction. Narrow the #1031 reference to `Issue: #1031` and leave it open, or deliver its doneWhen.
- **[P1] The 2.1 s reorder deletes the regression property that commit `84cbbdd` deliberately added, and the comment justifying it misstates that code.** `84cbbdd` — the 2026-08-16 fix GH-1078 itself cites — did three things in one commit: raised the observer loop from `for _ in 0..200` with 10 ms sleeps (≈ 2 s) to `FAKE_CODEX_STARTUP_BUDGET` = 30 s, raised the PowerShell gate the same way, **added `std::thread::sleep(Duration::from_millis(2_100))` immediately after the observer spawn**, and renamed the test from `fake_runner_resumes_the_current_attempt_session_before_turn` to `fake_runner_resumes_current_attempt_after_slow_startup_before_turn`. 2 100 ms is 100 ms more than the ceiling that commit replaced: the sleep's placement *after* the spawn is what made the observer's clock have to span a startup delay exceeding the old budget. Moving it to `tests/runner.rs:745`, above the spawn at `:747`, means the observer's deadline now starts with nothing ahead of it, the test exercises no deliberate slow startup at all, and its name no longer describes it. The added comment at `tests/runner.rs:738-744` — *"Sleeping after the observer was already running used to spend part of its hang-safety-valve on nothing observable"* — is the stated justification and is false as to that placement's purpose. The other half of the change is correct and worth keeping: 2.1 s of a 30 s budget (7 %) really was consumed before `run_task` was called. Recover the budget without deleting the coverage — e.g. keep the sleep after the spawn and assert the observer's remaining budget, or replace it with an explicit ordering assertion.
- **[P2] `FAKE_CODEX_STARTUP_BUDGET` now names the ceiling of a lease-renewal wait.** `tests/runner.rs:683` reuses a constant whose name, doc comment (`tests/mod.rs:15-29`) and PowerShell consumer are all about spawn + handshake startup, for a wait that begins after startup is already proven complete (`:674`). Two of its three Rust call sites are no longer startup waits.
- **[P2] Worst-case wedged-run wall clock grows ~6×, inside the CI Windows subset.** The observer's ceiling and the fake's PowerShell gate are the same constant and are sequential when the challenge appears late: observer 180 s at `tests/mod.rs:199-204`, then the fake's own 180 s from `tests/runner.rs:816` before `exit 7` → the `unexpected EOF from Codex App Server` at `crates/edda-conductor/src/agent/codex_app_server.rs:275-277`. Previously ~30 s + ~30 s. These four tests serialize on `FAKE_CODEX_LOCK`.
- **[P2] The PR body's "five sites" inventory is incomplete, and its diagnosis of the renewal loop is imprecise.** The sixth consumer of the constant — the PowerShell `STARTUP_BUDGET_MS` gate (`tests/runner.rs:816`, `:828-831`) — is the wait whose expiry produces exactly the `unexpected EOF from Codex App Server` message GH-1078 reports, and it is not listed. Behaviour is nonetheless consistent because it is derived from the same constant. Separately, the loop the PR calls "a fixed 2.5 s window" was an *iteration-count* ceiling (`for _ in 0..100` — 100 × 25 ms **plus** 100 SQLite reads), which dilates under load; converting it to an `Instant`-based deadline is, for the loop body's own cost, a step toward less load tolerance, dominated here only by the size of the new constant.
- **[P2] The "sound bucket" classification is half wrong, though the decision to leave both sites alone is right.** `await_gate` (`tests/runner.rs:7-12`) is exactly as the PR describes — it returns on timeout with no assertion, and its single caller at `:251` uses it as a deadlock breaker, so a slow host costs the simulated stall fidelity, never a failure. Verified. `spawn_deadline` (`tests/runner.rs:526-531`) is **not**: `anyhow::ensure!(std::time::Instant::now() < spawn_deadline, "later-launch.cmd side effect did not appear before the spawn deadline")` fails the test on timeout — the same deadline-as-assertion shape the PR says it is removing, justified in the body by exactly the load-calibrated reasoning ("a trivial `.cmd` spawn", "already tuned once for GH-524") that GH-1078 calls the anti-pattern. Leaving it untouched is correct scope discipline; only the stated reason is wrong.

Answering the brief's question 1 directly, from the code rather than the doc comments: **the ceiling did not stop being the assertion.** All three `poll_until` call sites turn `Ok(false)` into a test failure (`tests/mod.rs:204`, `tests/runner.rs:674`, `tests/runner.rs:691`), with the same message the old loop bailed with. The refactor changes code shape, not failure mode; the constant is 6× larger. For GH-1078 that is nonetheless *within* its doneWhen, which explicitly licenses "a condition with a generous ceiling", and the GH-1047 analogy checks out at the SHA (`plan.rs:512-516` — the 30 s constant was kept and the git I/O moved out from under it) for the two observation sites, where no equivalent move exists. For GH-1031 the analogy is used to license precisely what GH-1047 declined to do, and there an equivalent move **does** exist — see P0-1.

### FOLLOW-UP ISSUE

- `crates/edda-cli/src/cmd_reconcile/tests/runner.rs:526-531` — `spawn_deadline` is a 30 s deadline-as-assertion around a `.cmd` spawn in `first_spawn_failure_does_not_prevent_later_plan_launch`. Neither #1078 nor #1031 names this test and it has no reported failure, so it does not extend this PR. Per `fleet.ci-flake-carrier=product-issue-per-test-no-registry` it earns its own issue if and when it flakes.
- `crates/edda-cli/src/cmd_reconcile/runner.rs:479-483` — `run_turn_with_renewals` performs blocking `Ledger::open` + `renew_lease` (SQLite + workspace lock) directly inside a `tokio::select!` arm rather than through `spawn_blocking`. Under the lock contention GH-1047 measured at 27.99 s / 64.8 s, that can delay the next renewal tick past the turn's end — the product-side branch of GH-1031 doneWhen 2 (「若確認為真實競態…修在 `crates/edda-cli/src/cmd_reconcile/` 並附回歸測試」). This PR concluded "test problem" without examining it; since the test-side change cannot move the outcome (P0-1), that determination is not yet settled. Evidenced, out of this round's frozen surface.

### RAN vs READ

RAN (read-only; no Cargo, no checkout, no worktree, no build, no flake reproduction — per brief and `test.flake-evidence=structural-when-not-reproducible`):
- `gh pr view 1112` (head/base/body/commits/labels/comments); `gh pr diff 1112 --name-only` and the full three-dot diff `git diff origin/main...c01f073`
- §3 router extracted from `REVIEW.md` and run on the changed-file list → `classes=code-plain`, `canonical_class=code-risk`
- U1/U2/U3/U4 enumerators; DECISIONS enumerator (empty); `sh scripts/lint-markdown-content.sh` → `exit=0`; `gh pr checks 1112`; `gh api …/actions/runs?head_sha=c01f073`
- C2/C3/C4/C5 enumerators (test count `0`; one `.expect` candidate, in `#[cfg(test)]`; C4 empty; C5 selector empty)
- `sh scripts/wiring-scan.sh origin/main c01f073` → 1 new pub surface, 2 outside references, no swallow patterns
- `edda --version`; `edda ask test.flake-evidence`, `edda ask pr.closing-keyword`, `edda ask fleet.ci-flake-carrier` (all resolved from the repo checkout, so D2's unreachable-ledger clause does not apply)
- Source reads at the pinned SHAs via `git show`/`git grep`: `tests/mod.rs`, `tests/runner.rs`, `cmd_reconcile/runner.rs`, `cmd_reconcile/plan.rs`, `cmd_reconcile/mod.rs`, `edda-conductor/src/agent/codex_app_server.rs`; history read of `84cbbdd` (the 2026-08-16 fix GH-1078 cites) and `git log -S` for the 2 100 ms sleep
- Window check: `git diff 7781400e..origin/main -- <the two changed files>` → empty

READ (not rerun):
- L1 receipt — **`CI run 34380793999 @ c01f073269871f3597b22b6d4cb43adf0b353ef7`**, exact head, every job `pass`: Format; Clippy on ubuntu/macos/windows; Test on ubuntu + macos (full workspace) and windows (7-crate subset, 11m33s); MSRV 1.91.0; CI Gate; publish dry-run; sdk-contract.
- What that covers: `edda-cli` is inside the Windows subset, so the touched tests **did** run on windows-latest at this SHA and passed — which is why C5's selector is empty and no local Cargo run is owed here. What it does **not** cover: a single green run says nothing about an intermittent flake reported at 9/20, 6/20 and 4/20. Neither CI nor this round can demonstrate the flake is fixed; that is why the acceptance standard is structural, and why P0-1 is argued from the mechanism rather than from a tally.

### Verdict

Changes Requested, P0=2, P1=2 — the periodic-renewal conversion widens a poll budget that is capped by the runner's own turn lifetime, so GH-1031's flake is untouched while `Closes #1031` would auto-close it; and two `fix:` commits ship no test pinning the ordering property, having removed the one that existed.

fagemx:
## Review Response: Round 1

New head: `47232b897b1826f0522d1d6d204c471dd54d4ef2` (`47232b8`) — one commit added on top of the two frozen commits (`1bbcbe7`, `c01f073`). No amend, rebase, or force-push; `1bbcbe7` and `c01f073` are unchanged.

### P0-1 — GH-1031: re-derived, agreed, fixed by gating the turn instead of widening the budget

Re-derived independently before touching anything, and it matches Round 1 exactly: `run_turn_with_renewals` (`crates/edda-cli/src/cmd_reconcile/runner.rs:464-488`) is the only thing that can move the polled condition (`heartbeat_at`/`expires_at`), and it only renews while the turn is in flight — `interval.tick().await` consumes the immediate first tick, so the next tick fires at ≈ +1s (`(ttl_s / 2).max(1)` with `lease_ttl_s = 1`). The turn's flight time was pinned at ≈ 2s by `fake_codex(dir.path(), 2, false)`'s fixed `Start-Sleep`. So the renewal had a fixed ≈ 1s window inside a fixed ≈ 2s turn, and once the turn resolved, `finish_runner` deleted the lease (`runner.rs:516-518`), after which the condition can never become true again. `2.5s → 180s` in the frozen commit changed nothing about that: it only widened how long the test would wait for a condition that a lost race makes permanently false, turning a `assert!` failure into an `.expect("current lease")` panic on a wider clock.

**Fix, in `47232b8`:** `periodic_renewal_stops_old_runner_before_failure_after_lease_replacement` now gates the fake turn at the same `EDDA_FAKE_CHALLENGE`/`EDDA_FAKE_ALLOW`/`EDDA_FAKE_DENY` checkpoint `allow_fake_turn_after_durable_session` already uses (`tests/mod.rs`) — one of the two seams Round 1 named. The fake process writes the challenge file right after `turn/start` is read and then blocks (bounded by `FAKE_CODEX_STARTUP_BUDGET`) until the test writes `allow` or `deny`. `fake_codex`'s own `delay_s` is now `0` since the artificial delay is no longer what the test relies on. With the turn held open, `run_turn_with_renewals`'s ~1s interval keeps ticking indefinitely — the renewal is now an event the test waits on, not a deadline it races. The test only releases the gate (writes `allow`) after it has already observed the renewal and recorded the takeover lease (`attempt = 2`), so by the time `finish_runner` runs — via either the next tick noticing it no longer owns the lease, or the turn completing — it re-reads current ownership from the ledger and lands on the same outcome (no `task.failed`, replacement lease intact) regardless of which branch of `run_turn_with_renewals`'s `select!` actually wins. There is no longer a race at all, not a wider one: the property Round 1 asked for ("測試不依賴牆鐘計時競速...以可控的...事件序列取代") is what this delivers.

One more piece worth stating since it bears on whether this leaves anything wedged: `CodexAppServer` sets `kill_on_drop(true)` at spawn and its `run_turn_inner`/`request` paths hold a `KillOnCancel` guard that kills the child on any early drop (`crates/edda-conductor/src/agent/codex_app_server.rs:57,143,282-288`). So if the runner's interval tick wins the race before the gate is ever released, the pinned `turn` future is dropped mid-flight and the fake process is killed outright — it does not linger waiting on its own `STARTUP_BUDGET_MS`. Verified by reading, not by running the wedge case.

**Test problem, not a product race — and how that was determined:** GH-1031's own decision point asks which this is. The mechanism above shows the flake came from the *test* racing its own fixed-delay harness simulation (`Start-Sleep -Seconds 2`) against a periodic timer that only runs during that fixed window — nothing about that exercises whether the runner's real renewal behavior is wrong against a real, variable-length Codex turn. Round 1's own FOLLOW-UP ISSUE note (`run_turn_with_renewals` doing blocking `Ledger::open`/`renew_lease` inside a `tokio::select!` arm rather than `spawn_blocking`, `runner.rs:479-483`) is a distinct, still-open question about a *different* scenario (real lock contention during a real turn) that this test does not exercise either way — it stays open, evidenced, out of this round's frozen surface (test files only; `runner.rs` product code is untouched by this commit).

### P0-2 — regression test: the fix is the test

Per `edda ask test.flake-evidence` (`structural-when-not-reproducible`), the second half owed — a test that pins the ordering property — is the same work as the fix itself here: `periodic_renewal_stops_old_runner_before_failure_after_lease_replacement`, gated as above, is what now genuinely pins "a periodic renewal happens while the runner holds the lease" as a property that cannot pass by accident of timing. `git diff origin/main...47232b8 -- 'crates/**' | grep -cE '^\+.*(#\[test\]|fn test_)'` is still `0` because no *new* `#[test]` function was added — the existing one is what needed to stop racing, not a sibling. If a separate, additional pin is wanted here, say so and I'll add one; my read of P0-2 combined with the "same work" framing is that fixing this test's synchronization discharges it.

### P1-1 — the auto-closing reference to GH-1031

Removed the `Closes` line for GH-1031 from the PR body. `Issue: #1078, #1031` stays (both issues named as related), and the `Closes` line for GH-1078 stays. Left GH-1031 open rather than reasoning my way to re-adding a closing keyword for it: GH-1031's doneWhen 1 literally asks for 20 consecutive green windows-latest runs as part of the same sentence as the no-racing requirement, and this round's verification budget explicitly prohibits generating that evidence (no loops, no reproduction attempts) — CI itself only runs this test once per push, not 20 times. The structural fix above is real and, I think, sufficient in the same sense GH-1078's own doneWhen 2-3 were superseded by the ledger's structural-evidence standard rather than measured — but that supersession call for GH-1031 specifically is the controller's/reviewer's to make explicitly, not mine to assert by leaving an auto-closing reference in place. `edda ask` and `gh issue view 1031` were re-read before writing this; neither states that supersession for GH-1031.

### P1-2 — the slow-startup reorder: reverted to `84cbbdd`'s placement, comment corrected

Confirmed by reading `84cbbdd` (`git show 84cbbdd -- crates/edda-cli/src/cmd_reconcile.rs`, the pre-GH-778-split monolith): the sleep was added *after* `let observer = allow_fake_turn_after_durable_session(...)` and *before* `run_task(...)`, in the same commit that renamed the test to `..._after_slow_startup_...`. `47232b8` moves the sleep back to that exact position (between the observer spawn and the `run_task` call) and replaces the comment that justified the wrong placement with one that states why the position matters: the sleep has to run while the observer is already polling for it to cost any of the observer's budget on a genuinely slow start, which is what the test's name promises. This is a straight revert of that one hunk plus a corrected comment, not a redesign — kept whatever else Round 1 already accepted (the `poll_until`/180s conversion for this test) untouched.

### P2s

- **`spawn_deadline` reasoning** — corrected in the PR body. The prior text called it "sound" with a stated reason (trivial `.cmd` spawn vs. a `powershell.exe` handshake) that Round 1 showed is exactly the load-calibrated reasoning GH-1078 names as the anti-pattern, not a reason the shape differs. The body now says plainly: `spawn_deadline` is the same deadline-as-assertion shape as the sites this PR fixes; it is left alone on scope grounds (neither issue names `first_spawn_failure_does_not_prevent_later_plan_launch`, and it has no reported failure — `fleet.ci-flake-carrier=product-issue-per-test-no-registry` says it earns its own issue if and when it flakes), not because it is structurally sound.
- **Sixth consumer** — added to the PR body's site inventory: the PowerShell `STARTUP_BUDGET_MS` gate at `tests/runner.rs:816` (substituted from `FAKE_CODEX_STARTUP_BUDGET`), which is the wait whose expiry produces the literal `unexpected EOF from Codex App Server` message GH-1078's report names. Not changed in behavior — it already moved 30s → 180s alongside the Rust constant it shares a name with — just named, since the inventory claimed to be complete and wasn't.
- The other two P2s (constant naming now covering non-startup waits; worst-case wedged wall clock growing with the larger constant) are unaddressed — the brief scoped this round's P2 action to exactly the two above, and I've followed that scoping rather than re-opening the full P2 list.

### Gates

Lane `worker-2` (`CARGO_TARGET_DIR=$LOCALAPPDATA/fleet-workstation/lanes/worker-2`), this commit only (the first two commits' gates are unchanged from the PR body's existing report):

```
cargo fmt --all --check                                — clean
cargo clippy -p edda --all-targets -- -D warnings       — clean, 0 warnings
cargo test -p edda                                      — 749 passed; 0 failed; 0 ignored
                                                           (all cmd_reconcile::tests::runner tests
                                                           by name, all tests/*.rs integration
                                                           binaries green)
sh scripts/lint-file-length.sh --tree                   — clean
```

Committed with the pre-commit hook's `SKIP_CLIPPY=1` escape hatch after running clippy warm in-lane myself (the hook's own cargo doesn't inherit `CARGO_TARGET_DIR`, #1099); no stray `target/` appeared this round. No loop, no load generation, no reproduction attempt against either flake, per the brief.

Exact-head CI for `47232b897b1826f0522d1d6d204c471dd54d4ef2`: **CI run 34386484127** (plus `sdk-contract` run 34386484195, `Workspace publish dry-run` run 34386484070). One job — `Workspace publish dry-run` — failed on first attempt from an Ubuntu apt-mirror hash-sum mismatch fetching an unrelated `google-chrome-stable` package during `apt-get install pkg-config libssl-dev`, before any Rust code from this PR builds; re-ran that job (environmental, per the ladder's own "if the red is environmental, re-run only the failed job"), now `pending` again. As of this comment: Format, Clippy ×3 OS, MSRV, `sdk-contract`, `Test (ubuntu-latest)` and `Test (macos-latest)` all `pass`; `Test (windows-latest)` and the re-run `Workspace publish dry-run` still `pending` (Windows `Test` alone ran 11m33s on the prior SHA). Final state to follow.

### Undone

- `run_turn_with_renewals`'s blocking-I/O-inside-`select!` question (Round 1's FOLLOW-UP ISSUE, `runner.rs:479-483`) remains open and unexamined by this round — product code, out of the frozen test-only surface.
- CI final result for `47232b8` to follow.

Branch `fix/gh1078-flaky-timing-pattern`, pushed. Stopping here per the brief — not merging.

fagemx:
**CI update for `47232b897b1826f0522d1d6d204c471dd54d4ef2`:** all checks now pass, including the re-run and the ones pending at comment time.

```
CI run 34386484127 @ 47232b897b1826f0522d1d6d204c471dd54d4ef2
  Format                     pass   19s
  Clippy (ubuntu-latest)     pass   43s
  Clippy (macos-latest)      pass   55s
  Clippy (windows-latest)    pass   1m39s
  MSRV (1.91.0)              pass   54s
  Test (ubuntu-latest)       pass   2m50s
  Test (macos-latest)        pass   3m33s
  Test (windows-latest)      pass   10m15s
  CI Gate                    pass   3s
Plus:
  sdk-contract      (run 34386484195)  pass   2m2s
  Workspace publish dry-run (run 34386484070, re-run after the
    environmental apt-mirror failure noted above)  pass   5m22s
```

Gate receipt: **CI run 34386484127 @ 47232b897b1826f0522d1d6d204c471dd54d4ef2**, every job pass, `CI Gate` (the required check) pass.

fagemx:
## Code Review: Round 2 — PR #1112 @ 47232b897b1826f0522d1d6d204c471dd54d4ef2

- elapsed: unmeasured (no pi session file — replacement verifier running as a controller-spawned subagent under `review.dispatch-transport=controller-subagent-direct`)
- model_observed: unverified (no `claude -p --output-format json` envelope available to this session; self-labelling is not evidence)
- reviewer_session: not assigned by the brief (controller-spawned subagent; **replacement verifier** — the Round 1 session is unreachable, so this round READ its receipts before running anything, per `.claude/CLAUDE.md` § "Verification cost")
- spec: review-spec-v1.6, read at the **base** SHA `7781400e6cfcf1e7a8ababe678e410f2441b2159` via `git show`. Worth recording: the shared checkout is *behind* base (`HEAD` = `7d5bb83`), so its `REVIEW.md` copy is an older one and was **not** used — 31 lines differ, all in §1 fetch, D4, S1 and §8's merge-authority wording; none touch the front matter, the §3 router, §4 routing, the §5.3/§5.5 rules, or §7's shape, so nothing in this round turns on the difference. `REVIEW.md` is unchanged by this diff, so §6.6 adds no escalation.
- escalations: none
- shadow: false
- cost: read-only round — no checkout, no worktree, no `gh pr checkout`, no Cargo, no build, no test run, no flake-reproduction attempt (`test.flake-evidence=structural-when-not-reproducible` bans load generators; this host measured 11/11 clean and cannot serve as the instrument). `review.readonly-proof`: `git status --porcelain` identical before and after (` M docs/specs/lane-profiles-v1.md`, three untracked), branch still `main`. Process note carried from the brief, non-blocking: the implementing lane started a background `gh pr checks --watch` writing to a scratchpad log and did not block on it. Not a finding against the diff.

### IN SCOPE

This is a **delta round** (§2): the target is `git diff c01f073..47232b8` plus confirmation that each Round 1 blocker is resolved. Round 1's accepted decisions are not reopened — in particular the GH-1078 half, where "a condition with a generous ceiling" is what that issue's own doneWhen 1 licenses.

- **Changed behaviour and paths.** `crates/edda-cli/src/cmd_reconcile/tests/runner.rs` only (+72/−21 in this commit; the PR's other file `tests/mod.rs` is untouched since Round 1). Three things change: `periodic_renewal_stops_old_runner_before_failure_after_lease_replacement` gains the `EDDA_FAKE_CHALLENGE`/`ALLOW`/`DENY` turn gate and the crate-wide env guard (`:637`, `:668-687`), its body moves into a closure whose result is released-then-reported (`:693-760`), `fake_codex`'s `delay_s` drops 2 → 0 (`:663`); and `fake_runner_resumes_current_attempt_after_slow_startup_before_turn`'s 2 100 ms sleep returns to its `84cbbdd` position with a corrected comment (`:799-809`).
- **Direct consumers.** The gate's reader is the generated PowerShell fake (`:875-884`), whose ceiling is `STARTUP_BUDGET_MS` ← `FAKE_CODEX_STARTUP_BUDGET` (`:877`, `:890-893`). The product path exercised is `run_turn_with_renewals` / `finish_runner` (`crates/edda-cli/src/cmd_reconcile/runner.rs:464-534`), unchanged by this PR.
- **Acceptance ceilings.** GH-1078 doneWhen 1–4 (closed by this PR); GH-1031 doneWhen 1–2 (**referenced without a closing keyword** — `Closes #1031` was removed this round, so `pr.closing-keyword=only-when-all-donewhen-delivered` treats it as the partial-delivery shape that decision explicitly licenses and leaves the issue open).
- **Security / data loss.** None introduced or exposed: `#[cfg(windows)] #[cfg(test)]` test code only; no ledger, store, dispatch or process-policy path changed. The new env writes are process-global but serialized — see U1 evidence on lock order.
- **Current-base integration.** The path-limited window is empty against both `origin/main` (`1a6cb22`) and GitHub's `baseRefOid` (`c9e2930`, an ancestor of it): `git diff 7781400e..origin/main -- <the two files>` → empty. The reviewed tree is still the tree that would merge.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 1112 --name-only` → the two `cmd_reconcile/tests/` files, both inside GH-1078's Predicted surface and `runner.rs` inside GH-1031's. The new lock acquisition is inside the surface too, and safe: every multi-lock site in `edda-cli` takes ENV → FAKE → DOORBELL (`tests/runner.rs:549-551`, `:637-638`, `:774-775`, `:828-829`) — one total order, no inversion, so the added `env_guard()` cannot deadlock against a sibling |
| U2 | any | P1 | PASS | body → one hit, line 100 `Closes #1078`; `git log --format=%B origin/main..47232b8` → no hits. `Closes #1031` is gone (Round 1 P1-1 resolved). GH-1078's doneWhen: 1 met (generous ceiling, its own either/or wording); 2–3 superseded by `test.flake-evidence` (`edda ask` → *"load generators are banned as lane-starving"*); 4 met — the body distinguishes #1031 explicitly and now fixes it by a different mechanism |
| U3 | any | P1 | PASS | body carries `Issue: #1078, #1031` |
| U4 | any | P1 | PASS | enumerator empty over `origin/main..47232b8` — all three subjects are `fix(edda-cli): …` |
| U6 | any | P0 | PASS | READ exact-head CI: **run 34386484127 @ 47232b8**, every job `pass` — Format, Clippy ×3 OS, MSRV (1.91.0), Test ubuntu 2m50s / macos 3m33s / **windows 10m15s**, CI Gate, sdk-contract (34386484195), and the re-run `Workspace publish dry-run` (34386484070) now `pass`. The Round 1 apt-mirror failure the brief flagged as environmental is resolved by that re-run, exactly as the ladder prescribes. Nothing rerun by me; no reason to distrust it. `edda --version` → `edda 0.5.0 (16f7b4b41cde-dirty 2026-09-09)`; no runtime claim in this diff depends on it |
| U7 | any | P1 | PASS | whole delta audit finished before posting; every Round 1 blocker re-checked against the new code, and each new observation classified rather than deferred |
| U8 | any | P1 | N.A.(no `Decision:` line in the PR body — the DECISIONS enumerator prints nothing) |
| C1 | code-plain | P0 | PASS | **GH-1078** (the issue this PR closes): all four doneWhen delivered — see U2. **GH-1031**: doneWhen 1's substantive half — 「以可控的時鐘／事件序列取代 sleep 與真實時間比較」— is delivered by the challenge→observe→allow event sequence (derivation below); its 20-consecutive-green clause is unrun and unrunnable under this round's budget; doneWhen 2's conditional does not fire because the mechanism is a harness race, not a product race (verified below). With no closing keyword for #1031 this is the partial-delivery shape `pr.closing-keyword` licenses, so an undelivered clause on a *referenced* issue is not a C1 failure here |
| C2 | code-plain | P0 | PASS | `git diff origin/main...47232b8 -- 'crates/**' \| grep -cE '^\+.*(#\[test\]\|fn test_)'` → still `0`, and that is the right answer this time. `test.flake-evidence` requires *"structural argument from the code plus **a test that pins the ordering property**"* — not a new `#[test]`. Both halves now have that pin: for GH-1031 the repaired test can no longer pass without observing a renewal while the turn is held open (it is the pin, and it is what needed to stop racing); for GH-1078 the pins are `allow_fake_turn_after_durable_session`'s deny-unless-durable ordering check (`tests/mod.rs:205-215`, pre-existing, preserved) and the 2 100 ms slow-startup sleep — **the pin Round 1 found deleted, restored this round**. Round 1's P0-2 said the diff *removed* the pin that existed; that is what changed |
| C3 | code-plain | P1 | N.A.(three `.expect(` candidates on added lines — `tests/runner.rs:705`, `:722`, `:755` — all inside `#[cfg(test)] mod tests;`, confirmed at `crates/edda-cli/src/cmd_reconcile/mod.rs:16-17`. Their unwind behaviour is not silently accepted: it is adjudicated as P2-1 below) |
| C4 | code-plain | P1 | PASS | enumerator empty — no `unsafe`, no `clippy::all`, no crate-level `#![allow]` |
| C5 | code-plain | P1 | N.A.(selector empty — `edda-cli` is inside CI's Windows 7-crate subset, so `Test (windows-latest)` ran this crate's tests on the exact head and passed. No uncovered crate; no local Cargo run owed, and running one would be the over-verification the ladder names as a process finding) |
| §5.5 | any | — | PASS | `sh scripts/wiring-scan.sh c01f073 47232b8` → `no new pub surfaces`; one swallow candidate, adjudicated below. Table filled |
| D1–D5 | docs | — | N.A.(§4 does not route `code-plain` to §5.1 — no `docs`/`skills` path in the diff. D5 is the spec's only `[判斷]` item and is not routed, so `escalations:` is empty by rule, not by omission) |
| S1–S3 | skills | — | N.A.(§4 does not route `code-plain` to §5.2) |
| R1–R5 | code-risk | — | N.A.(§4 routes §5.4 only for the `code-risk` REVIEW.md class; the §3 router printed `classes=code-plain`. Both enumerators run anyway and are empty: R1 no destructive command on any added line; R2 no `\|\|`/`&&` mix — the added PowerShell is a Rust string literal whose `if`/`while` carry no mixed operators) |

### Wiring

`wiring-scan` reports **no new pub surfaces** for this delta; the PR's only new `pub` surface is `poll_until`, reviewed in Round 1 and unchanged. What this commit does add is a written side-file set, so it gets its row:

|---|---|---|---|---|
| `poll_until(budget, poll, condition) -> anyhow::Result<bool>` (`pub(super)`, `#[cfg(windows)]`) — carried, unchanged since Round 1 | `tests/mod.rs:41-56` | `tests/mod.rs:199`, `tests/runner.rs:694`, `:707`, `:718` — four call sites now, all in this PR | `condition()?` propagates; `Ok(false)` returned, never logged-and-ignored. All four callers turn `Ok(false)` into a failure, now via `ensure!` rather than `assert!` inside the gated test | helper → four test call sites; no product layer |
| `renewal.challenge` / `renewal.allow` / `renewal.deny` side-files + the `EDDA_FAKE_*` env contract (second writer of an existing contract) | test: `tests/runner.rs:682-687` (env), `:747` (release write); fake: `:876` (`New-Item -Force` challenge) | fake's gate loop `tests/runner.rs:875-884`; test's challenge poll `:707-711` | **Swallowed on the release write** (`let _ = std::fs::write`, `:747`) — adjudicated P2-2 below, with the reason §5.5's fixed P1 lookup does not fire | test env → PowerShell fake → gate → `run_turn_with_renewals` turn future → `finish_runner`. Read back by the test's own poll and by the final assertions at `:762-766` |

### Findings

No P0, no P1. The three blocking findings of Round 1 are resolved, and I verified each against the code rather than against the response:

**P1-2, the sleep revert — confirmed exactly.** `84cbbddf15bda2f35dc7ade50f72a750afa0a2d1` (`2026-08-17 05:46:42 +0800`) added `std::thread::sleep(…2_100)` **between** `allow_fake_turn_after_durable_session(...)` and `run_task(&repo, 2, 1, …)`, in the same commit that renamed the test to `..._after_slow_startup_...` — read directly from that commit's diff. Head has the sleep in exactly that position (`tests/runner.rs:796-797` observer, `:809` sleep, `:811` `run_task`). The replacement comment (`:799-808`) is now true of that placement, and its `84cbbdd` date matches git's own author date.

**Brief risk 3, the `select!`-arm equivalence — verified from product code, and the claim holds.** Both arms land on the same outcome because `finish_runner` re-reads ownership: `owned = ledger.task_lease(task_id)?.is_some_and(|lease| lease.attempt == attempt)` (`cmd_reconcile/runner.rs:502-504`) is false once the test has upserted `attempt = 2`, so neither `append_failed` (`:508-517`) nor `delete_task_lease` (`:519`) runs. Arm A (interval tick): `renew_lease` → `UPDATE … WHERE task_id = ?1 AND attempt = ?2` (`sqlite_store/task_leases.rs:52-56`) matches 0 rows → `false` → `return Ok(())` (`runner.rs:482-483`), the pinned turn future drops, `KillOnCancel::drop` → `start_kill` (`codex_app_server.rs:282-288`) kills the fake at the gate. Arm B (turn resolves after `allow`): `return result.map(|_| ())` (`:479`). Both give `run_task` → `Ok(())` and leave the replacement lease intact with no `task.failed`, which is what `:762-766` assert.

**Brief risk 1, the deadlock trace — the gate is always released whenever the fake is actually waiting on it.** Derivation, since a hang would be worse than the flake:

1. Every fallible step between the runner spawn and the release is inside the closure at `:693-740` and uses `?` / `ensure!`, so it lands in `outcome` and falls through to `:747`, which writes `deny`. No error path skips the release.
2. The only way to skip `:747` is an unwinding panic inside that closure. The two candidates are `.expect("session lease")` (`:705`) and `.expect("current lease")` (`:722`), which fire only when the lease **row** is absent. `task_lease` does not filter by expiry — it is a plain `SELECT … WHERE task_id = ?1` (`sqlite_store/task_leases.rs:6-23`) — so with `lease_ttl_s = 1` an expired lease still reads back. The row can only vanish via `delete_task_lease`, called only from `finish_runner`, reached only after `run_turn_with_renewals` returns, which requires the turn to resolve, which requires the gate to open or the fake to die. So *while the fake blocks at the gate, those two `.expect`s cannot fire* — the skip path and the blocked-fake state are mutually exclusive.
3. Even in a hypothetical skip, nothing hangs unboundedly: the fake's own loop is bounded by `STARTUP_BUDGET_MS` = 180 s and then `exit 7` (`:877-883`), and `kill_on_drop(true)` (`codex_app_server.rs:57`) plus the `KillOnCancel` guard (`:143`, `:282-288`) kill the child on any early drop of the turn.
4. Ordering is right at the end too: `outcome?` (`:759`) precedes `run_result?` (`:760`), so an observation failure is reported without the join's error masking it — and the join happens before either, so no `?` can strand the runner thread.

**Brief risk 2, does the test still prove what it claimed — yes, and it proves more.** The gate holds the turn at the point where the fake has read `turn/start` and not yet answered, which is precisely "turn in flight" from the runner's side: `server.run_turn(...)` is pending and `run_turn_with_renewals`' `select!` is live (`runner.rs:475-487`). The renewals are produced entirely by product code on its own `tokio::time::interval` — the harness manufactures none of them; it only removes the deadline on *observing* one. So the asserted property is still the product's behaviour, and it is now strictly stronger: before, the test could pass by the coincidence of a ~1 s tick landing in a fixed ~2 s turn; now it cannot pass at all unless a renewal is observed while the runner demonstrably holds the lease, and the lease replacement is guaranteed to be in place *before* the turn resolves rather than racing it, which is what makes the `finish_runner` assertions at `:762-766` meaningful instead of order-dependent.

**Brief risk 5, P0-2 carried forward — accepted for both halves.** See C2. For GH-1031 the repaired test is the ordering pin, and it is the pin `test.flake-evidence` asks for: it fails if `run_turn_with_renewals` stops renewing mid-turn, and it fails if `finish_runner` stops re-reading ownership. For GH-1078 the ordering pin is `allow_fake_turn_after_durable_session`'s deny-unless-durable check plus the restored slow-startup sleep; that half owes no new `#[test]`, because its doneWhen asks the two tests to stop depending on a wall-clock window, not for a new one to be written.

**Brief risk 6, the P2 corrections — both confirmed.** The body now states plainly that `spawn_deadline` is the same deadline-as-assertion shape and is left alone on scope grounds under `fleet.ci-flake-carrier=product-issue-per-test-no-registry`, not on soundness; and the sixth consumer (`STARTUP_BUDGET_MS`) is added to the inventory.

The remaining findings are all P2 and none of them blocks:

- **[P2-1] A panic between the runner spawn and `:756-758` leaks `EDDA_FAKE_*` into the rest of the test binary.** `tests/runner.rs:685-687` sets three process-global env vars and `:756-758` removes them, but `runner.join().expect("runner thread")` at `:755` sits *above* the removals, so a panicking runner thread unwinds past them while the `_env`/`_fake` guards are released by that same unwind. The next fake-Codex test then spawns with `EDDA_FAKE_CHALLENGE` pointing into a deleted tempdir; `New-Item -ItemType File -Force` (`:876`) recreates the path, and that test waits out the full 180 s `STARTUP_BUDGET_MS` before `exit 7` — one panic here costs the other three fake-Codex tests ~180 s each and an `unexpected EOF` that names the wrong cause. Not P1: it cannot turn a red run green, only lengthen and misattribute an already-failing one, and the closure's two `.expect`s are unreachable while the gate holds (step 2 above), leaving only an unlikely `run_task` panic as the trigger. A small `Drop` guard owning the three vars closes this and the closure path at once. Note the sibling test at `:813-817` already has the safe ordering (`remove_var` before `run_result?`/`observer_result.expect`), so the pattern to copy is one screen away.
- **[P2-2] The gate release swallows its own error.** `let _ = std::fs::write(if outcome.is_ok() { &allow } else { &deny }, "gate")` (`:747`) — the single hit from `sh scripts/wiring-scan.sh`. Swallowing is right in *shape* here (a `?` would return before `runner.join()` and re-create the early-exit hazard the release exists to prevent) but the failure is then invisible, and a failed release is exactly a 180 s stall plus a misattributed EOF. Capture the `io::Result` and report it after the join. I considered §5.5's fixed *"error swallowed → P1"* lookup and did not apply it: that adjudication is scoped to a **new surface** on a ledger, coordination or cost path, and this is a `#[cfg(test)]`-only fixture write with no product consumer on a delta that `wiring-scan` reports as having no new pub surfaces. Recording the reasoning rather than dropping the rule.
- **[P2-3] `outcome?` before `run_result?` discards the runner's real error when both fail.** `:759-760`. The characteristic failure of these tests is a spawn/handshake fault that surfaces in `run_result` as `unexpected EOF from Codex App Server` (`codex_app_server.rs:270-278`) while `outcome` carries the generic `"runner recorded the durable session"`. For a PR whose whole subject is flake diagnosis, reporting the timeout and dropping the cause is the wrong way round; `.context()`-chaining keeps both.
- **[P2-4] The PR body's `STARTUP_BUDGET_MS` line citations are pinned to the previous head.** The body cites `tests/runner.rs:816`, `:814-823` and `:828-831`; at `47232b8` those are `:877`, `:875-884` and `:890-893` — the commit's added lines shifted them by ~61. `runner.rs:526` for `spawn_deadline` is still correct. Content is right, anchors moved. Not a D1 finding — §4 does not route `code-plain` to §5.1.
- **[P2-5] `fake_codex`'s `delay_s` is now `0` at every call site** (`:577`, `:663`, `:786`, `:840`), so `Start-Sleep -Seconds DELAY` (`:885`) is dead in every generated script and the parameter no longer varies anything. Collapse it, or keep it with a note saying why.
- Round 1's two unaddressed P2s stand as recorded, correctly out of this round's scoped action: `FAKE_CODEX_STARTUP_BUDGET` now names the ceiling of a lease-renewal wait that is not a startup wait, and the worst-case wedged wall clock is ~6× its former size inside the CI Windows subset.

### The question this round was to decide: is the test free of wall-clock dependence?

Answering from the code, plainly, because the controller acts on it.

**The race is gone. A ceiling remains, and it is a hang valve rather than a contest.** Before `47232b8` the test's outcome was decided by comparing two real-time quantities of the same order: a ~1 s renewal interval against the ~1 s of slack left inside a fixed 2 s `Start-Sleep` turn, with the loser's condition made *permanently* unobservable by `finish_runner` deleting the lease. That is a wall-clock race, and Round 1 was right that no poll budget could reach it. After `47232b8` the decision procedure is an event sequence: the fake announces `turn/start` by writing `renewal.challenge`, the test waits on that file, then waits on the renewal while the turn is pinned open, then releases the gate with `allow` only after the observation and the lease replacement have both happened. Nothing races anything: the runner's interval keeps ticking for as long as the gate stays shut, so the condition is late at worst and never permanently false.

What has *not* been eliminated is real time as such. The renewal is produced by a real `tokio::time::interval`, which the test cannot advance; `poll_until` still observes it with `std::thread::sleep` + `Instant::now()`; and the hold itself is bounded, because the fake's gate loop expires after `STARTUP_BUDGET_MS` = 180 s. So the in-code comment at `:714-717` — *"it cannot resolve out from under this poll"* — is categorical where the code is merely 180-to-1. I did not raise that as a finding: reaching it requires a 1 s timer to fail to fire for three minutes, which is the wedge condition the budget exists to detect, and GH-1078's own doneWhen 1 licenses exactly this shape ("a condition with a generous ceiling"). Round 1 accepted that framing for the GH-1078 half and it does not reopen; what changed for GH-1031 is that the ceiling stopped being the binding constraint, which was the entire substance of Round 1's P0-1.

**So, for GH-1031 doneWhen 1:** the parenthesis — 「以可控的時鐘／事件序列取代 sleep 與真實時間比較」— names two acceptable instruments, a controllable clock **or** an event sequence. The delivered fix is the second of them, and it is the seam Round 1 itself pointed at. The substantive requirement is met. The 20-consecutive-green clause is a demonstration, and a demonstration of race removal is what a surviving race needs: 20 runs distinguish "usually passes" from "always passes" only when the outcome is still decided by timing. Here it is decided by an event sequence, so 20 runs measure nothing 1 run does not — and running them is barred anyway (`test.flake-evidence` bans load generators; CI runs this test once per push; this host is 11/11 clean and is not a valid instrument). **My reading matches the controller's: the clause was conditional on the race surviving, and it did not survive.** GH-1031 can close on this PR.

That is a reviewer's finding on the code, not an authorisation. The PR body currently carries no `Closes #1031`, which is the correct conservative state for the SHA under review and is why U2 passes. If the controller restores it at merge, U2's equivalence still holds on the reading above, and the honest record is that doneWhen 1 closed on its substantive half with the repetition clause judged moot rather than satisfied. If the controller prefers not to make that call, leaving `#1031` open with this PR referenced costs nothing and no rule is violated either way — the fix ships regardless.

### FOLLOW-UP ISSUE

Carried from Round 1, both still valid and still out of the frozen test-only surface:

- `crates/edda-cli/src/cmd_reconcile/tests/runner.rs:526-531` — `spawn_deadline` is a 30 s deadline-as-assertion around a `.cmd` spawn in `first_spawn_failure_does_not_prevent_later_plan_launch`. Neither issue names that test and it has no reported failure; per `fleet.ci-flake-carrier=product-issue-per-test-no-registry` it earns its own issue if and when it flakes. Confirmed still present and unchanged at head.
- `crates/edda-cli/src/cmd_reconcile/runner.rs:479-483` — `run_turn_with_renewals` performs blocking `Ledger::open` + `renew_lease` (SQLite under the workspace lock) directly inside a `tokio::select!` arm rather than via `spawn_blocking`. Under the contention GH-1047 measured at 27.99 s / 64.8 s that can delay the next renewal tick. This round narrows why it stays a follow-up rather than a blocker: with the turn now gated open, a delayed tick is late, not lost, so it cannot fail the repaired test — the product question is real but no longer entangled with GH-1031's flake. Worth its own issue.

New, evidenced, out of scope for this PR:

- The `EDDA_FAKE_*` env-var lifetime pattern is now used by three tests in this file with two different orderings relative to their `remove_var` calls (`:755-758` panics past them, `:813-817` does not). A `Drop`-guard helper alongside `poll_until` in `tests/mod.rs` would make the safe ordering the only ordering — P2-1 and the closure path in one change. Not required by either issue's doneWhen.

### RAN vs READ

RAN (read-only; no checkout, no worktree, no `gh pr checkout` — the base spec's §1 forbids it in a shared checkout — no Cargo, no build, no test, no flake reproduction):

- `gh pr view 1112` (head/base/body/state/comments), `gh pr diff 1112 --name-only`, `gh pr checks 1112`, `gh api …/actions/runs?head_sha=47232b8`, `gh issue view 1031` / `1078` (bodies + comments)
- §3 router extracted from the **base** `REVIEW.md` by its two marker lines and run on the changed-file list → `classes=code-plain`, `canonical_class=code-risk`
- U2/U3/U4 + DECISIONS enumerators (last empty); `sh scripts/lint-markdown-content.sh` → `exit=0`; C2 (`0` over the PR and `0` over the delta), C3 (three candidates, all `#[cfg(test)]`), C4 (empty), C5 selector (empty), R1 and R2 enumerators (both empty)
- `sh scripts/wiring-scan.sh c01f073 47232b8` → `no new pub surfaces`; one swallow candidate at `:747`
- Delta diff `git diff c01f073..47232b8`; source reads at the pinned SHA via `git show`: `tests/runner.rs`, `tests/mod.rs`, `cmd_reconcile/runner.rs`, `cmd_reconcile/mod.rs`, `test_support.rs`, `edda-ledger/src/sqlite_store/task_leases.rs`, `edda-conductor/src/agent/codex_app_server.rs`; history read of `84cbbdd` for the sleep position and its date; lock-order enumeration across every `env_guard()`/`test_lock(&…)` site in `edda-cli`
- `edda --version`; `edda ask test.flake-evidence`, `edda ask pr.closing-keyword`, `edda ask fleet.ci-flake-carrier` — all resolved from the repo checkout, so D2's unreachable-ledger clause does not apply
- Window checks: `git diff 7781400e..origin/main` and `git diff c9e2930..origin/main`, both path-limited to the two changed files → empty; `git merge-base 47232b8 origin/main` → `7781400e`
- `git status --porcelain` before and after, identical (`review.readonly-proof`); the only files written were a copy of the base `REVIEW.md` and this comment body, both in the session scratchpad outside the repo

READ (not rerun):

- L1 receipt — **`CI run 34386484127 @ 47232b897b1826f0522d1d6d204c471dd54d4ef2`**, exact head, every job `pass`, including `Test (windows-latest)` at 10m15s; plus `sdk-contract` 34386484195 and the re-run `Workspace publish dry-run` 34386484070, both `pass`. The Round 1 apt-mirror red was environmental and the lane's targeted re-run cleared it — no reason to distrust the run, so no Cargo gate was rerun here.
- What it covers: `edda-cli` is inside CI's Windows 7-crate subset, so the two touched tests ran on windows-latest at this SHA and passed — which is why C5's selector is empty and no focused local run is owed. What it does not cover: one green run cannot demonstrate a flake is fixed. That is why the standard is structural (`test.flake-evidence`) and why every conclusion above is argued from the mechanism.
- Round 1's `## Code Review: Round 1 @ c01f073` and the implementer's `## Review Response: Round 1`, read before running anything, per `.claude/CLAUDE.md` § "Verification cost" for a replacement verifier.

### Verdict

LGTM (P0=0, P1=0)

Merge is not mine to take: it follows the recorded rule gate (current-head LGTM, `CI Gate` green, empty SHA window, P0=0/P1=0), never the reviewer's own action. Per this round's brief I am also not setting the `fleet:reviewed` label or any status — that is left to the controller alongside the `Closes #1031` decision above.

PR #1114, opened 2026-09-09
docs(release): include late v0.6.1 fixes

fagemx:
Issue: #1107

## Summary

- include #1110's verify-contract false-fail fix in the 0.6.1 notes
- include #1112's event-synchronized fake-Codex test reliability fix
- reconcile the final pre-tag range: 13 PR commits since v0.6.0, 12 categorized, #1111 explicitly excluded as release-prep mechanics

## Validation

- mechanical PR-number reconciliation: `range_prs=13 categorized=12 excluded_release_prep=['1111'] missing=[]`
- `sh scripts/lint-markdown-content.sh`
- `bash scripts/lint-doc-citations.sh --tree`
- `git diff --check`
- pre-commit documentation/content/file-length hooks

PR #1118, opened 2026-09-09
fix(fleet): merge-reviewed-pr.sh pins the squash subject to the PR title and carries a merge body (GH-1100)

fagemx:
Issue: #1100

## What

`scripts/merge-reviewed-pr.sh` let GitHub choose the squash subject: no `--subject`, no `--body-file`. For a single-commit PR GitHub uses that commit's subject verbatim — which wrote `wip(review): GH-1003 + GH-992 in progress… (#1097)` (`fa0d011`) onto main permanently, a convention violation every gate passed, unrecoverable under R7.

- The squash subject is now ALWAYS the PR title (`gh pr view --json title`), passed as `--subject` on every `gh pr merge` — regardless of how many commits the PR carries.
- The title is validated against the conventional-commit subject rule (REVIEW.md §5 U4 regex, verbatim) before any merge: `wip(...)`, empty scope (`fix(): x`), missing type all refuse with the offending string in the message. `--check` validates too — earlier signal.
- New `--body-file <path>` argument; without it, `--merge` composes a minimal receipt naming the reviewed SHA, the LGTM round, and the CI run link, so the merge commit itself carries the gate receipt.
- `docs/guides/operator-runbook.md` step 7 describes the pinned+validated subject and `--body-file`; the script's `--help` matches.

## Tests

`scripts/test-merge-reviewed-pr.sh` cases 9–13 (existing 1–8 byte-identical):
- 9: conforming `--merge` → recorded call carries `--subject <exact title>` (the single-commit proof: GitHub never gets to choose), `--match-head-commit`, `--body-file`; composed body names SHA, `Round 1`, CI link.
- 10: `--body-file` supplied → the user's file rides the merge byte-identical; no composition read.
- 11–13: `wip(review): …`, `fix(): x`, `no type here` → nonzero, offending string on stderr, and no `pr merge` call recorded.

Mutation-checked: removing `--subject` fails case 9; neutralizing the regex fails case 11.

## Surface (R22)

`scripts/merge-reviewed-pr.sh` executes merges → **審判面**. Authoritative review round required (Opus/sol); glm-flash rounds are SHADOW only.

## Gates (L0 — scripts/docs only, no Cargo)

- RAN `sh -n scripts/merge-reviewed-pr.sh` — PASS
- RAN `sh -n scripts/test-merge-reviewed-pr.sh` — PASS
- RAN `sh scripts/test-merge-reviewed-pr.sh` — PASS (13 cases)
- RAN `sh scripts/lint-markdown-content.sh` — PASS
- RAN `sh scripts/test-doc-citations.sh` — PASS (21 checks)
- Pre-commit hooks (doc-citations --staged, markdown lint, file-length --staged) green at commit time.
- L1 = exact-head CI on this SHA.


fagemx:
## Code Review: Round 1 — PR #1118 @ 09cd2e67b604dc9dad2b47300aec8770e78635b7

- elapsed: unmeasured (direct-flow review under `review.default-path`; no pi session JSONL exists to read)
- model_observed: claude-opus-5 (harness-declared). Not read from either source §6 names — there is no pi session file and this did not run under `claude -p --output-format json` — so by that rule's own standard treat it as **unverified**; it is recorded here rather than invented or copied from `model_requested`.
- reviewer_session: fe9806e8-5934-4ff0-9b9f-1f12ab93bc3d (direct controller session, not a launched lane)
- spec: review-spec-v1.6 (loaded with `git show 3fbf08ac558b0d297a453ff3185dde73d1145f5f:REVIEW.md`; byte-identical to the checkout copy, and this diff does not touch `REVIEW.md`)
- class: code-risk  (REVIEW.md classes: code-risk docs)
- escalations: none
- shadow: false
- cost: 1 disposable review worktree at the head SHA (no Cargo build, no lane); 2 mutation runs of the offline suite; 4 live `gh` probes. Shared checkout untouched — `git status --porcelain` identical before and after (43 lines both), branch still `main`.

### IN SCOPE

Changed behaviour: `scripts/merge-reviewed-pr.sh` gains (a) a flag parser after the PR number, (b) an unconditional `--subject "$PR title"` on the squash merge, (c) conventional-commit validation of that title in **both** `--check` and `--merge`, (d) `--body-file <path>`, and (e) a composed receipt body when no file is given. Paths: the script, its offline suite, and the runbook's step 7.

Direct consumers checked: `.claude/skills/edda-release/SKILL.md:205,207` invokes `<PR>` and `<PR> --merge` — both still parse under the new loop. `docs/fleet/rules.md:16`, `.claude/CLAUDE.md:454`, `REVIEW.md:902` and `scripts/fleet/daily-digest.sh:177` reference the script's gate role, which this diff does not change. No other caller.

Acceptance ceiling: issue #1100's five doneWhen items. Security/data-loss surface: this is the merge entrypoint (審判面 under R22), so the title value flowing into `gh pr merge` and the temp-file lifecycle were both examined. Current-base integration: MERGEABLE against `main`.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 3 files, all inside #1100's stated surface (`scripts/merge-reviewed-pr.sh`, `docs/guides/operator-runbook.md`, plus that script's own suite) |
| U2 | any | P1 | PASS | Both enumerators empty — no closing keyword in body or in `09cd2e6`'s message. The harm-bearing direction (a closing keyword on unfinished work) is absent. See [P2] below for the converse |
| U3 | any | P1 | PASS | `Issue: #1100` present on line 1 of the body |
| U4 | any | P1 | PASS | `git log --format=%s origin/main..09cd2e6` → 1 commit, 0 offenders |
| U5 | any | P1 | PASS | `sh scripts/lint-markdown-content.sh` → exit=0 |
| U6 | any | P0 | PASS | `gh pr checks 1118` → 12/12 pass incl. `CI Gate`. READ, not re-RAN |
| U7 | any | P1 | PASS | Whole routed audit completed before posting; every routed rule has a row here |
| U8 | any | P1 | N.A.(no `Decision:` line in the PR body — the DECISIONS enumerator printed nothing) |
| D1 | docs | P1 | PASS | Enumerator: `edda review --help`→0 OK, `gh pr --help`→0 OK. Beyond the enumerator, the two claims that would break the merge at runtime were probed live: `gh pr checks 1118 --repo fagemx/edda --required --json name,state,link` → exit 0, `link` populated, and the script's exact jq expression returns the Actions run URL; `gh pr merge --help` on gh 2.87.0 lists `-t, --subject` and `-F, --body-file`. Both flags and the field exist |
| D2 | docs | P1 | PASS | `edda ask pr.closing-keyword` and `edda ask fleet.review-engine` resolve from the checkout and match how they are used here. The diff asserts no decision value of its own. (`review.merge-gate` and `mechanism.shell-role` return `No results found.` but this diff makes no claim about either) |
| D3 | docs | P1 | PASS | 1 repo-anchored path added, `docs/fleet/rules.md` → OK |
| D4 | docs | P0 | PASS | 7 candidates, all inside the merge-gate script itself or its fixtures. The runbook candidate carries the required caveat in the same paragraph — step 7 opens with 「規則閘綠即可執行…`docs/fleet/rules.md` R6」 |
| D5 | docs | [判斷] | PASS | Derivation: **clear** — the runbook sentence names the mechanism (`--subject`), the hazard (single-commit PR), the case file (`fa0d011`) and the fallback (`--body-file`). **Correctly scoped** — it edits only step 7, the step that runs this script. **Non-duplicative** — it points at `REVIEW.md` §5 U4 rather than restating the type list. It is a long single sentence, but that matches every other numbered step in this runbook, so it is house style, not a defect. No finding, therefore nothing to tag `provisional` and nothing to escalate (§6.1 attaches to a finding, and there is none) |
| C1 | code-plain | P0 | PASS | All five doneWhen matched to code and test — table below |
| C2 | code-plain | P0 | PASS | `fix:` with a real regression test, **mutation-verified here, not taken on the PR's word**: deleting `--subject "$subject"` from line 230 → `FAIL: case 9: the squash subject is not the exact PR title`; widening the regex to admit `wip` → `FAIL: case 11`. Suite is 13/13 green unmutated. (The rule's Rust enumerator returns 0 because no crate is touched; the proving test is the shell suite) |
| C3 | code-plain | P1 | N.A.(no `crates/**` change) |
| C4 | code-plain | P1 | PASS | Enumerator empty over the whole diff |
| C5 | code-plain | P1 | N.A.(selector prints nothing — no crate touched, so there is no Windows coverage gap to close) |
| R1 | code-risk | P0 | PASS | Enumerator empty. The one `rm` added is `rm -f "$merge_body"` inside the EXIT trap, on a path `mktemp` just created — trigger stated in R4 |
| R2 | code-risk | P0 | PASS | Zero candidates. Ran the enumerator on the shell paths **and** swept the whole diff, since the rule's boundary is not the path filter — both empty. Every added conditional is a single-operator chain (`[ … ] \|\| die`, `cmd \|\| var=''`); no line mixes `&&` and `\|\|`, so there is no parse tree to adjudicate |
| R3 | code-risk | P0 | PASS | `sh -n` on both changed scripts **at the reviewed SHA** → exit 0 each. (First pass parsed my working tree, which is `main` — the base content, not the PR's. Re-run against `git show 09cd2e6:` content; recording the correction because a PASS on the wrong bytes is not a PASS) |
| R4 | code-risk | P1 | PASS | The temp body is created then immediately guarded: `mktemp … \|\| die` at :221, `trap 'rm -f "$merge_body"' 0` at :222. `die` exits, EXIT fires, the file is removed — error path covered, which is the bar this rule sets. Only trap in the file, so nothing is clobbered. A caller-supplied `--body-file` is deliberately outside the trap and is never deleted. See [P2] for the signal gap |
| R5 | code-risk | P1 | PASS | Properties confirmed: the PR title reaches `gh` as a quoted argv element (`--subject "$subject"`), never interpolated into a shell string, and is regex-validated first; `$head` is already pinned to `^[0-9a-f]{40}$` at :84 before entering the body; every receipt line is `printf '%s\n' "…"` with a literal format, so a `%` in a title or URL cannot become a format directive; the two distinct body sources (user file vs `mktemp` template) cannot collide on one filename |
| §5.5 | any | — | PASS | Wiring table below; `sh scripts/wiring-scan.sh origin/main 09cd2e6` → no new pub surfaces, no swallow patterns |

#### C1 — doneWhen ledger

| #1100 doneWhen | Code | Test |
|---|---|---|
| Subject always from the PR title, any commit count | `merge-reviewed-pr.sh:230` — `--subject "$subject"` is unconditional | case 9 asserts `--subject <exact title> --body-file` in the recorded call |
| Validated before the merge; `wip`/empty scope/missing type refuse with the offending string | `:92-94`, before the union gate and before `gh pr merge` | cases 11/12/13 — nonzero, offending string on stderr, and `grep -q 'pr merge'` proves the call was never reached |
| Accepts a merge body via `--body-file` | `:31-36`, `:41` | case 10 — `cmp -s` proves gh received the user's file byte-identical, and no receipt was composed |
| Composes a receipt naming reviewed SHA, LGTM round, CI run | `:216-229` | case 9 greps the captured body for all three |
| Runbook describes the new arguments | `operator-runbook.md:175` | doc-citations 21/21 |

One substitution to record rather than wave past: doneWhen 1 asks for a fixture with a **single-commit** PR whose commit subject differs from the title. The stub has no commit-count notion because the script never reads one, and the test says so in its own header comment instead of hiding it. What case 9 asserts — the merge call always carries `--subject` equal to the title — is strictly stronger, since it holds for every commit count including one. Delivered.

### Wiring

|---|---|---|---|---|
| `--body-file <path>` argument | `merge-reviewed-pr.sh:31-36` sets `body_file`; missing operand and missing file both `die` | `:209` (`merge_body=$body_file`) → `:230` `gh pr merge --body-file` | Refuses, never best-effort: `die '--body-file requires a path'` / `die "body file not found: …"` | flag → parser → `merge_body` → gh argv; case 10 asserts the file reaches gh byte-identical |
| `--subject` on the squash merge | `:92` reads the PR title; `:230` passes it | GitHub's squash API | Fail-closed at `:93-94` — a non-conforming title dies before the merge | title → regex → argv; case 9 pins it, mutation-verified |
| Composed receipt body (a written file) | `:221-229` writes SHA + round + CI link to a `mktemp` file | `gh pr merge --body-file` → the squash commit message on `main` | Every input refuses on absence: `die` for unparsable round, unreadable checks, missing link, failed mktemp. Nothing is silently omitted | fields → temp file → gh → merge commit; case 9 greps all three fields out of the captured body |
| Title validation on `--check` | same `:92-94`, reached in both modes | the operator | Nonzero exit + offending string on stderr | earlier signal only; no new persisted state |

No new `pub` surface, no new config key, no new event field. Every write end carries a refusal path; none is best-effort or success-only.

### Findings

- [P2] The EXIT trap does not cover signals. `merge-reviewed-pr.sh:222` is `trap 'rm -f "$merge_body"' 0`, while the suite added in this same PR uses `trap cleanup 0 HUP INT TERM` (`test-merge-reviewed-pr.sh:41`). A Ctrl-C between `mktemp` and gh's return leaves the receipt in `$TMPDIR`. Harmless and R4's stated bar (`trap … 0`) is met — noting the inconsistency inside one PR.
- [P2] `--body-file ''` is silently accepted as "no body file". At `:38`, `[ -z "$body_file" ] || [ -f "$body_file" ] || die …` short-circuits on the empty string, so an empty path composes a receipt instead of refusing. Every other malformed operand refuses; this one does not.
- [P2] No `closes #1100`, though all five doneWhen are delivered. `pr.closing-keyword=only-when-all-donewhen-delivered` is worded as a biconditional ("exactly when"), so a strict reading wants the keyword here. Not raised as P1: U2's operational text names exactly one P1 — a closing keyword on unfinished work — and its enumerator produced no output, so there is no evidenced finding under §5's own threshold. It may also be deliberate: #1100's "Suspected surface" contemplates the verb migration, and PR #1121 claims to fold GH-1100 into `edda review merge`. Either close #1100 by hand at merge, or leave it open for #1121 — but decide it rather than letting it drift.

### FOLLOW-UP ISSUE

- **The U4 regex now exists twice with nothing keeping the copies honest.** `merge-reviewed-pr.sh:93` and `REVIEW.md`'s U4 check are byte-identical today — verified programmatically, so the PR body's "verbatim" claim is true — but no test asserts they stay that way. Change the type list in `REVIEW.md` and this script silently judges by the old one, on the merge entrypoint. Worth a sync test or a single shared source. Adjacent to this PR's scope, and it survives the migration: `edda review merge` in #1121 inherits the same duplication.
- **PR #1121 also implements GH-1100 and touches these same two files.** `.edda/branches/main/log.md:23386` and #1121's own body item 28 ("#1100 摺入") say `edda review merge` folds in the pinned+validated subject; its file list includes `scripts/merge-reviewed-pr.sh` and `scripts/test-merge-reviewed-pr.sh`. #1121 is stacked on #1119's branch, not on `main`, so this is not a current-base conflict and does not block #1118 — but whichever lands second needs a rebase and a fresh round (runbook step 7's Layer-3 intersection). Merging #1118 first is the cheaper order: it puts the preventive fix on `main` now, and #1121 rebases onto it.

### RAN vs READ

RAN:
- `git show <base>:REVIEW.md` → spec loaded at base, identical to checkout, `REVIEW.md` not in this diff (no §6.6 escalation)
- classifier block from `REVIEW.md` on the file list → `classes=code-risk docs`, `canonical_class=code-risk`
- U2/U3/U4 enumerators → empty / `Issue: #1100` / 0 offenders
- `sh scripts/lint-markdown-content.sh` → exit=0
- D1/D3/D4 enumerators; plus live `gh pr checks 1118 --json name,state,link` (link populated, script's jq returns the run URL) and `gh pr merge --help` on gh 2.87.0 (`-t, --subject`, `-F, --body-file` both present)
- `edda ask pr.closing-keyword`, `edda ask fleet.review-engine`
- C4/R1/R2 enumerators over the shell paths and over the whole diff → all empty
- `sh -n` on both changed scripts at the reviewed SHA (worktree `--detach 09cd2e6`) → exit 0 each
- `sh scripts/test-merge-reviewed-pr.sh` at head → `PASS (13 cases)`, exit 0
- **Mutation 1** — dropped `--subject "$subject"` → `FAIL: case 9`, suite exit 1
- **Mutation 2** — added `wip` to the type alternation → `FAIL: case 11`, suite exit 1
- `sh scripts/lint-file-length.sh --tree` → exit 0; `sh scripts/test-doc-citations.sh` → 21 checks passed
- `sh scripts/wiring-scan.sh origin/main 09cd2e6` → no new pub surfaces, no swallow patterns
- `git status --porcelain` before and after → identical (43 lines), branch still `main` (`review.readonly-proof`)

READ:
- **CI run 34394895418 @ 09cd2e67b604dc9dad2b47300aec8770e78635b7** — Format, Clippy ×3 OS, Test on Linux/macOS/Windows, MSRV 1.91.0, Fleet shell tests ×2, Detect code changes, `CI Gate`: 12/12 pass. Covers the workspace Cargo gates and the fleet shell suites on Linux and Windows.
- Not covered by that run, and why no local Cargo gate was RAN: this diff touches no crate, so C5's selector is empty and the ladder's L1 focused Windows-gap run has nothing to select. No workspace gate was re-run — there is no coverage gap to justify one (`ladder` L2).

### Verdict

LGTM (P0=0, P1=0)

fagemx:
## Code Review: Round 2 — PR #1118 @ 09cd2e67b604dc9dad2b47300aec8770e78635b7

- elapsed: unmeasured (controller-spawned subagent under `review.dispatch-transport=controller-subagent-direct`; no pi session JSONL exists to read)
- model_observed: unverified (§6's two permitted sources are absent — no pi session file, and this did not run under `claude -p --output-format json`. Recorded as `unverified` rather than copied from `model_requested`)
- reviewer_session: subagent of `fe9806e8-5934-4ff0-9b9f-1f12ab93bc3d`. **Independence, stated plainly:** independent of the implementer (a flash lane on `codex/gh1100-squash-subject`), so GATE-01 holds; *not* independent of the session that posted Round 1, which is the same controller. This round was run from a fresh context against the spec, not against Round 1 — every rule below was re-derived, and the blocking finding is one Round 1 did not reach.
- spec: review-spec-v1.6 (loaded with `git show 3fbf08ac558b0d297a453ff3185dde73d1145f5f:REVIEW.md`; byte-identical to the checkout copy — `git hash-object` `3de7e327…` both. `REVIEW.md` is not in this diff, so §6.6 adds no escalation)
- class: code-risk  (REVIEW.md classes: code-risk docs)
- escalations: none
- shadow: false
- cost: no worktree, no Cargo build, no lane (`n/a` — this diff compiles nothing). 11 read-only `gh` probes, 1 ledger read, the §5 enumerators, `sh -n` on both changed scripts at head content. Shared checkout untouched: `git status --porcelain` identical before and after (43 lines), branch still `main` (`review.readonly-proof`).

### IN SCOPE

Changed behaviour: `scripts/merge-reviewed-pr.sh` gains a flag parser after the PR number, an **unconditional** `--subject "$(PR title)"` on the squash merge, U4 validation of that title in both `--check` and `--merge`, `--body-file <path>`, and a composed receipt body when no file is given. Paths: that script, its offline suite, and runbook step 7.

Direct consumers: `.claude/skills/edda-release/SKILL.md:205,207` (`<PR>` and `<PR> --merge`) — both still parse under the new loop. `docs/fleet/rules.md:16`, `.claude/CLAUDE.md:454`, `REVIEW.md:902`, `scripts/fleet/daily-digest.sh:177` reference the gate role, unchanged here. No other caller.

Acceptance ceiling: issue #1100's five doneWhen. **Regression surface introduced by the change**: this is the merge entrypoint (審判面, R22), and the diff alters the exact `gh pr merge` line that decides what text lands on `main` forever — so what that line now produces on `main`, not merely what it passes to `gh`, is inside the frozen surface. Current-base integration: MERGEABLE against `main`.

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 3 files, all named in #1100's "Suspected surface" |
| U2 | any | P1 | PASS | Both enumerators empty (body and `origin/main..09cd2e6`). No closing keyword on an unfinished issue — the harm-bearing direction |
| U3 | any | P1 | PASS | `Issue: #1100` on line 1 of the body |
| U4 | any | P1 | PASS | 1 commit in range, 0 offenders |
| U6 | any | P0 | PASS | `gh pr checks 1118` → 12/12 pass incl. `CI Gate`. READ from exact-head CI, not re-RAN |
| U7 | any | P1 | PASS | Whole routed audit finished before posting; every routed rule carries a row. The single blocker below is batched, not drip-fed |
| U8 | any | P1 | N.A.(the DECISIONS enumerator printed nothing — no `Decision:` line) |
| D1 | docs | P1 | PASS | `edda review --help` → 0 OK; `gh pr --help` → 0 OK. Beyond the enumerator: `gh pr merge --help` on gh 2.87.0 lists `-t, --subject` and `-F, --body-file`; `gh pr checks --json name,state,link` returns a populated `link`. Both interfaces exist |
| D2 | docs | P1 | PASS | `pr.closing-keyword` and `fleet.merge-authority` resolve from the checkout and match their use here. `review.merge-gate` / `mechanism.shell-role` return `No results found.` — an unreachable key, not a false claim (§5.1 D2), and this diff asserts no value for either |
| D3 | docs | P1 | PASS | 1 repo-anchored path added: `docs/fleet/rules.md` → OK |
| D4 | docs | P0 | PASS | 7 candidates, all inside the merge-gate script or its fixtures. The runbook's step 7 opens with 「規則閘綠即可執行…`docs/fleet/rules.md` R6」 — the required caveat, in the same paragraph |
| D5 | docs | `[判斷]` | PASS | Derivation: **clear** — names the mechanism (`--subject`), the hazard (single-commit PR), the case file (`fa0d011`), the override (`--body-file`). **Scoped** — edits only step 7. **Non-duplicative** — points at `REVIEW.md` §5 U4 instead of restating the type list. A long run-on sentence, but every numbered step in this file is one; house style, not a defect. No finding, so nothing to tag `provisional` (§6.1 attaches to a finding) |
| C1 | code-plain | P0 | PASS | All five doneWhen matched to code and test — ledger below. The single-commit fixture substitution is accepted: the script reads no commit count, and "`--subject` always equals the title" is strictly stronger than the fixture asked for |
| C2 | code-plain | P0 | PASS | `fix:` with a real regression suite. Covered by exact-head CI: `run-fleet-tests.sh:42` globs `scripts/test-*.sh`, which matches `test-merge-reviewed-pr.sh` — so `Fleet shell tests` **and** `Fleet shell tests (windows)` both ran it green at this SHA. READ, not re-RAN; Round 1's two mutation runs are READ as recorded |
| C3 | code-plain | P1 | N.A.(no `crates/**` change) |
| C4 | code-plain | P1 | PASS | Enumerator empty over the whole diff |
| C5 | code-plain | P1 | N.A.(selector empty — no crate touched, so no Windows coverage gap exists to close) |
| R1 | code-risk | P0 | PASS | Enumerator empty. The one `rm` added is `rm -f "$merge_body"` inside the EXIT trap, on a path `mktemp` created one line earlier — trigger stated |
| R2 | code-risk | P0 | PASS | Zero candidates on the shell paths **and** zero over the whole diff (the rule's boundary is not the path filter, so I ran both). Every added conditional is single-operator (`[ … ] \|\| die`, `cmd \|\| var=''`); no line mixes `&&` and `\|\|`, so there is no parse tree to adjudicate |
| R3 | code-risk | P0 | PASS | `sh -n` at **head content** (`git show 09cd2e6:<path>`) → exit 0 for both changed scripts. The working tree is `main`, so parsing it would have been a PASS on the base bytes |
| R4 | code-risk | P1 | PASS | `mktemp … \|\| die` (:221) then `trap 'rm -f "$merge_body"' 0` (:222); `die` exits, EXIT fires, file removed — the error path is covered, which is this rule's bar. Only trap in the file, so nothing is clobbered. A caller-supplied `--body-file` is deliberately outside the trap and never deleted |
| R5 | code-risk | P1 | PASS | Properties confirmed: the title reaches `gh` as one quoted argv element, never interpolated into a shell string, and is regex-validated first; `$head` is pinned to `^[0-9a-f]{40}$` at :83 before use; every receipt line is `printf '%s\n' "…"` with a literal format, so a `%` in a title or URL cannot become a format directive; the two body sources (user file vs `mktemp` template) cannot collide on one filename |
| §5.5 | any | — | **FAIL** | Wiring table below. `sh scripts/wiring-scan.sh origin/main 09cd2e6` → no new pub surfaces, no swallow patterns; but the `--subject` surface's **read end was never checked**, which is where the P1 lives |

#### C1 — doneWhen ledger

| #1100 doneWhen | Code | Test |
|---|---|---|
| Subject always from the PR title, any commit count | `:230` — `--subject "$subject"` is unconditional | case 9 pins `--subject <exact title>` in the recorded call |
| Validated before the merge; `wip`/empty scope/missing type refuse with the offending string | `:92-94`, ahead of the union gate and of `gh pr merge` | cases 11/12/13 — nonzero, offending string on stderr, `grep -q 'pr merge'` proves the call was never reached |
| Accepts a merge body via `--body-file` | `:38-42`, `:47`, `:214` | case 10 — `cmp -s` proves gh received the user's file byte-identical, and no receipt was composed |
| Composes a receipt naming reviewed SHA, LGTM round, CI run | `:215-228` | case 9 greps all three out of the captured body |
| Runbook describes the new arguments | `operator-runbook.md:175` | — |

### Wiring

|---|---|---|---|---|
| `--body-file <path>` argument | `:38-42` sets `body_file`; missing operand and missing file both `die` | `:214` → `:230` `gh pr merge --body-file` | Refuses, never best-effort | flag → parser → `merge_body` → gh argv; case 10 asserts byte-identical arrival |
| `--subject` on the squash merge | `:92` reads `.title`; `:230` passes it verbatim | GitHub's squash API → **the commit message on `main`** | Fail-closed at `:93-94` for a non-conforming title | **Reaches gh argv and stops there.** The layer that matters — what actually lands on `main` — is asserted by nothing. Case 9 checks the argv string; no fixture, no doc line and no probe covers the subject GitHub then writes. That unchecked layer is the P1 below |
| Composed receipt body (a written file) | `:221-228` writes SHA + round + CI link to a `mktemp` file | `gh pr merge --body-file` → the squash commit body | Every input refuses on absence (`die` for unparsable round, unreadable checks, missing link, failed mktemp). Nothing silently omitted | fields → temp file → gh → merge commit; case 9 greps all three fields |
| Title validation on `--check` | `:92-94`, reached in both modes | the operator | Nonzero exit + offending string on stderr | earlier signal only; no new persisted state |

No new `pub` surface, no new config key, no new event field.

### Findings

- **[P1] Making `--subject` unconditional silently drops the ` (#N)` PR back-reference from every future squash commit on `main`.** GitHub appends ` (#N)` to the subject *it* chooses; an explicitly supplied `commitHeadline` is used **verbatim, with no suffix**. Measured in this repo, both directions:

  | PR | how the subject was chosen | landed subject on `main` |
  |---|---|---|
  | #1112 | GitHub (multi-commit → PR title) | PR title **+ ` (#1112)`** |
  | #1097 | GitHub (single commit → commit subject) | `wip(review): …` **+ ` (#1097)`** — the case file |
  | #1077 | explicit `--subject` (hand merge) | PR title exactly — **no ` (#1077)`** |
  | #1081 | explicit `--subject` (hand merge) | supplied text exactly — **no ` (#1081)`** |

  `git log origin/main --format=%s -60` → 58 of 60 carry ` (#N)`; the only two that do not are #1077 and #1081, i.e. exactly the two merged with an explicit subject. `gh api repos/fagemx/edda` → `squash_merge_commit_title=COMMIT_OR_PR_TITLE`, so today every script merge takes a GitHub-chosen subject and gets the suffix. After this PR, `:230` supplies `--subject` on **every** merge, so no future commit gets it.

  Why this blocks rather than waits: it is a regression introduced by the changed line itself (frozen surface, not adjacent), and it is the same unrecoverable class as the bug being fixed — R7 forbids rewriting `main`, so every commit that lands suffix-less stays that way. #1100's own argument is that "`git log --oneline` is the artifact people read a year later"; this change removes the PR pointer from precisely that artifact, on 100% of merges, to fix a defect that hit 1 commit in 300. Nothing in the issue, the PR body, the runbook edit, or the tests mentions the suffix, so this reads as unnoticed rather than decided.

  Note the suite currently **pins** the regression: `test-merge-reviewed-pr.sh` case 9 asserts `grep -qF -- "--subject $STUB_PR_TITLE --body-file"`, which fails the moment a suffix is appended. So the fix is two edits, not one — evidence: `scripts/merge-reviewed-pr.sh:230`; `scripts/test-merge-reviewed-pr.sh` case 9.

  Suggested shape (not binding): validate the bare title at `:93-94` as now, then merge with `--subject "$subject (#$pr)"` — appending only when the title does not already end in `(#<pr>)` — and update case 9 to assert the suffix is present and the PR number is the right one. If instead the intent is to drop the suffix deliberately, that is an operator-visible convention change and belongs in the runbook edit and in a ledger decision, not in an unremarked argv change on the merge entrypoint.

- [P2] The EXIT trap does not cover signals. `:222` is `trap 'rm -f "$merge_body"' 0`, while this same PR's suite uses `trap cleanup 0 HUP INT TERM` (`test-merge-reviewed-pr.sh:41`). A Ctrl-C between `mktemp` and gh's return leaves the receipt in `$TMPDIR`. R4's bar is met; noting the inconsistency inside one PR. (Also raised in Round 1.)
- [P2] `--body-file ''` is accepted as "no body file". `:47` short-circuits on the empty string, so an empty path composes a receipt instead of refusing, while every other malformed operand refuses. (Also raised in Round 1.)
- [P2] `--body-file` is silently ignored under `--check`. Harmless, but it accepts an argument it will not use rather than saying so.

### FOLLOW-UP ISSUE

- **The U4 regex now exists twice with nothing keeping the copies honest.** `merge-reviewed-pr.sh:93` and `REVIEW.md` U4 are byte-identical today, but no test asserts they stay that way; change the type list in `REVIEW.md` and the merge entrypoint silently judges by the old one. Worth a sync test or one shared source — and it survives the migration, since `edda review merge` in #1121 inherits the duplication. (Carried forward from Round 1; still out of scope here.)
- **PR #1121 also implements GH-1100 and touches these same two files.** `.edda/branches/main/log.md:23385-23386` records `edda review merge` folding in the pinned+validated subject. #1121 is stacked on #1119's branch, not on `main`, so this is not a current-base conflict and does not block #1118 — but whichever lands second needs a rebase and a fresh round, and **the P1 above applies to #1121's implementation too**, so fixing it only here would let the verb reintroduce it.

### RAN vs READ

RAN:
- `git show 3fbf08ac…:REVIEW.md` → spec loaded at base, byte-identical to the checkout (`3de7e327…`)
- REVIEW.md's own classifier block on the file list → `classes=code-risk docs`, `canonical_class=code-risk`
- U2 (both enumerators, empty) / U3 (`Issue: #1100`) / U4 (0 offenders) / DECISIONS (empty)
- D1, D3, D4 enumerators; plus `gh pr merge --help` and `gh --version` (2.87.0)
- `edda ask` × 5 (`pr.closing-keyword`, `fleet.merge-authority` resolve; `review.merge-gate`, `mechanism.shell-role` unreachable)
- C4, C5, R1, R2 (shell paths **and** whole diff) → all empty
- `sh -n` on both changed scripts at **head content** → exit 0 each
- `sh scripts/wiring-scan.sh origin/main 09cd2e6` → no new pub surfaces, no swallow patterns
- **P1 evidence**: `git log origin/main --format='%H%x09%s' -60` (58/60 carry ` (#N)`); `gh api repos/fagemx/edda/commits/<sha>/pulls` and `gh pr view <n> --json title,mergeCommit` for #1077, #1081, #1097, #1112; `gh api repos/fagemx/edda --jq .squash_merge_commit_title` → `COMMIT_OR_PR_TITLE`
- `git status --porcelain` before and after → identical (43 lines), branch still `main`

READ:
- **CI run 34394895418 @ 09cd2e67b604dc9dad2b47300aec8770e78635b7** — Format, Clippy ×3 OS, Test ×3 OS, MSRV 1.91.0, Fleet shell tests, Fleet shell tests (windows), Detect code changes, `CI Gate`: 12/12 pass. Confirmed by reading `run-fleet-tests.sh:42` that the `scripts/test-*.sh` glob carries `test-merge-reviewed-pr.sh`, so the 13-case suite is genuinely gated on Linux and Windows — no local rerun needed.
- Round 1's two mutation runs (drop `--subject` → case 9 red; widen the regex → case 11 red), read as recorded rather than repeated.
- Not covered by that run: nothing requiring a Cargo gate — this diff touches no crate, so C5's selector is empty and L1's focused Windows-gap run has nothing to select. No workspace gate was re-run; there is no coverage gap that would justify one (`ladder` L2).

### Verdict

Changes Requested, P0=0, P1=1 — the unconditional `--subject` lands every future squash commit on `main` without its ` (#N)` PR back-reference, measured against four PRs in this repo's own history; permanent under R7 and unmentioned anywhere in the change.

fagemx:
## Review Response: Round 2 — PR #1118 @ 760fd4334be15ac6306846d07307bc6cf9d48215

Round 2's blocking finding is accepted in full and fixed. Previous head `09cd2e67b604dc9dad2b47300aec8770e78635b7`; new head **`760fd4334be15ac6306846d07307bc6cf9d48215`**.

### [P1] Unconditional `--subject` drops the ` (#N)` PR back-reference — FIXED

The finding is correct, and I reproduced its measurement before touching anything: `git log origin/main --format=%s -8` shows every recent squash subject carrying its ` (#N)`, including the two merged since the review was written (`… (GH-1103) (#1119)`, `… (#1114)`). The change as reviewed would have made `main` stop looking like that, permanently, on 100% of script merges — for a defect that hit one commit.

**What changed** (`scripts/merge-reviewed-pr.sh`): the subject is still pinned to the PR title — GH-1100's whole point, unregressed — and the script now appends the back-reference GitHub would have added itself, immediately before the merge call:

```sh
merge_subject="$subject (#$pr)"
case "$subject" in
  *" (#$pr)") merge_subject=$subject ;;
esac
gh pr merge "$pr" … --subject "$merge_subject" --body-file "$merge_body"
```

Answering the three points the finding asked me to get right:

1. **U4 still validates the PR title, not the title-plus-suffix.** `$subject` is untouched — it stays the bare title read at `:101`, and the regex at `:102-103` judges that. The composed string lives in a separate variable that exists only on the merge line. So a trailing ` (#N)` can neither rescue a bad title nor break a good one. Case 11 now pins this from the outside: it asserts the refusal message — which quotes exactly the string that was validated — contains no `(#N)`. If the check ever moved onto the composed subject, that assertion goes red.
2. **The composed subject is what reaches `gh`.** `--subject "$merge_subject"` at `:259`; the recorded stub call in cases 9, 10, 14 and 15 is the proof.
3. **The already-numbered title case — decided, not defaulted.** The append is skipped **only** when the title already ends in *this PR's own* number, which is the sole case where appending duplicates rather than adds (a title copied back from a landed commit). A title ending in **another** PR's number still gets ` (#$pr)` appended. That is the deliberate half: `… (#999)` in a title is prose — a revert or follow-up naming the PR it answers — and treating it as this commit's pointer would send a `git log` reader to an unrelated PR. A slightly ugly `… (#999) (#1118)` is strictly better than a confidently wrong pointer, and it also matches GitHub's own convention that the real back-reference is last. Both directions are tested (cases 14 and 15) and both are mutation-verified below.

### Tests

`scripts/test-merge-reviewed-pr.sh`: 13 cases → **17 cases**, all green.

| Case | What it now proves |
|---|---|
| 9 (updated) | The recorded `pr merge` carries `--subject <PR title> (#4242) --body-file` — bounded on both sides, so it is the whole subject. Plus an explicit negative: if the bare unsuffixed subject is seen, it fails naming the dropped pointer. This is the assertion the old `grep -qF -- "--subject $STUB_PR_TITLE --body-file"` had pinned the wrong way round |
| 10 (extended) | A caller-supplied `--body-file` does not cost the commit its back-reference — the append sits on the merge call, not inside the compose-a-receipt branch |
| 11 (extended) | U4 judged the bare title: the refusal names the offending title and contains no `(#N)` |
| **14 (new)** | A title already ending in this PR's own number is passed verbatim — no `(#4242) (#4242)` tail |
| **15 (new)** | A title ending in `(#999)` still gets ` (#4242)` appended: another PR's number is never left standing as the back-reference |
| **16 (new)** | P2 — `--body-file ''` refuses instead of silently meaning "no body file" |
| **17 (new)** | P2 — `--body-file` under `--check` refuses instead of being accepted and ignored |

**Mutation check** (asked for explicitly; both run, both restored):

| Mutation | Result |
|---|---|
| Drop the append — `merge_subject=$subject` | **RED**: `FAIL: case 9: the squash subject is not the PR title plus its (#N) back-reference`, suite exit 1. Recorded call showed `--subject fix(fleet): make the squash subject the PR title (GH-1100) --body-file` — no suffix |
| Widen the skip to any number — `*" (#"[0-9]*")"` | **RED**: `FAIL: case 15: another PR's number was left standing as the back-reference`, suite exit 1. Recorded call showed `--subject fix(fleet): follow-up to the earlier change (#999) --body-file` |
| Restored | `PASS scripts/test-merge-reviewed-pr.sh (17 cases)`, exit 0 |

So the suffix is genuinely load-bearing in the suite, and so is the *shape* of the skip rule — not merely "the old assertion stopped failing".

### The three P2s — all taken, all small

- **EXIT-only trap** (`:222`): now `trap 'rm -f "$merge_body"' 0 HUP INT TERM`, matching this PR's own suite (`test-merge-reviewed-pr.sh:41`). The inconsistency-inside-one-PR you named is gone.
- **`--body-file ''` accepted as "no body file"**: refuses in the parser now (`[ -n "$2" ] || die '--body-file requires a non-empty path'`), before the `-z` test that used to short-circuit on it. Case 16.
- **`--body-file` ignored under `--check`**: refuses after the parse loop (`[ -z "$body_file" ] || [ "$action" = --merge ] || die '--body-file applies to --merge only'`). Only the merge path has a reader; accepting it told the caller their body was taken when nothing would open it. Case 17.

### Docs

`docs/guides/operator-runbook.md` step 7 now states the back-reference behaviour (why GitHub does not add one to a supplied subject, that the script appends it, that U4 still judges the bare title, and both halves of the skip rule) and the two new `--body-file` refusals. No other file touched — everything else in this PR passed review and was left alone.

### Gates ran (L0 — no Cargo gate: this diff touches no crate, so C5's selector is empty)

| Gate | Result |
|---|---|
| `sh -n scripts/merge-reviewed-pr.sh` | exit 0 |
| `sh -n scripts/test-merge-reviewed-pr.sh` | exit 0 |
| `sh scripts/test-merge-reviewed-pr.sh` | `PASS (17 cases)`, exit 0 — baseline before editing was `PASS (13 cases)`, exit 0 |
| `sh scripts/lint-markdown-content.sh` | exit 0 |
| `sh scripts/test-doc-citations.sh` | `21 checks passed`, exit 0 |
| `sh scripts/lint-file-length.sh --tree` | exit 0 |
| pre-commit hooks (doc-citations `--staged`, markdown-content, file-length `--staged`) | all passed on the staged blobs |
| `scripts/fleet/run-fleet-tests.sh` | ran this suite in situ: `RUN scripts/test-merge-reviewed-pr.sh` → `PASS (17 cases)` |

Exact-head CI on `760fd433` is the L1 receipt and is the reviewer's to READ.

### Not addressed (deliberately)

- The `[P2]` about no `closes #1100`, and both `FOLLOW-UP ISSUE`s (the duplicated U4 regex with nothing keeping the copies honest; PR #1121 folding GH-1100 into `edda review merge`) are outside this PR's frozen surface and unchanged by this round. Worth flagging on the second: **the P1 fixed here applies to #1121's implementation too** — whichever of the two lands second must carry this back-reference append, or the verb reintroduces the regression the moment it takes over the merge call.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

---

Gate receipt: **CI run 34435944312 @ 760fd4334be15ac6306846d07307bc6cf9d48215** (exact head). `Fleet shell tests`, `Format` and `Detect code changes` already green at time of posting; the Rust jobs were still in flight — the full result is the reviewer's to READ, not mine to re-run.

fagemx:
## Code Review: Round 3 — PR #1118 @ 760fd4334be15ac6306846d07307bc6cf9d48215

- elapsed: unmeasured (controller-spawned subagent under `review.dispatch-transport=controller-subagent-direct`; no pi session JSONL exists to read, so `scripts/pi-session-elapsed.mjs` has no input)
- model_observed: unverified (§6's two permitted sources are both absent — no pi session file, and this did not run under `claude -p --output-format json`. Recorded as `unverified` rather than copied from `model_requested`)
- reviewer_session: subagent of `fe9806e8-5934-4ff0-9b9f-1f12ab93bc3d`. **Independence, stated plainly:** independent of the implementer (the lane on `codex/gh1100-squash-subject`), so GATE-01 holds; *not* independent of the session that posted Rounds 1 and 2, which is the same controller. Fresh context, run against the spec rather than against the earlier rounds.
- spec: review-spec-v1.6 (loaded with `git show 3fbf08ac558b0d297a453ff3185dde73d1145f5f:REVIEW.md`; byte-identical to the checkout copy — `git hash-object` `3de7e327b3b84fbb7e9353e416afd1f987446495` both. `REVIEW.md` is not in this diff, so §6.6 adds no escalation)
- class: code-risk  (REVIEW.md classes: code-risk docs)
- escalations: none
- shadow: false
- cost: **delta round** per §2 — target is `git diff 09cd2e67b604dc9dad2b47300aec8770e78635b7..760fd4334be15ac6306846d07307bc6cf9d48215` plus a RAN confirmation that Round 2's blocker is resolved. No worktree, no Cargo build, no lane (`n/a` — this diff compiles nothing). One throwaway sandbox outside the repo (`git init` in the scratchpad, head bytes of the three scripts copied in) so the suite and five mutations could be run without touching this checkout. Shared checkout untouched: `git status --porcelain` identical before and after (43 lines), branch still `main` (`review.readonly-proof`).

### IN SCOPE

Delta behaviour: the squash subject now carries the ` (#N)` PR back-reference (`merge-reviewed-pr.sh:257-261`), `--body-file` refuses an empty operand (`:45`) and refuses under `--check` (`:56`), the compose-branch trap grows `HUP INT TERM` (`:231`), `--help` and runbook step 7 describe all of it, and the suite goes 13 → 17 cases. Paths: the same three files as Round 2.

Direct consumers: `.claude/skills/edda-release/SKILL.md:205,207` (`<PR>`, `<PR> --merge`) — neither passes `--body-file`, so the two new refusals break no documented caller; `git grep 'merge-reviewed-pr.sh.*--body-file'` across the repo returns nothing. Acceptance ceiling: #1100's five doneWhen, all re-checked against the new head. Frozen surface still includes what this line makes land on `main`, not merely what it hands `gh`. Current base: MERGEABLE (`mergeStateStatus` BLOCKED is the gate awaiting the verdict, not a conflict).

### Rules

|---|---|---|---|---|
| U1 | any | P0 | PASS | 3 files, all named in #1100's "Suspected surface"; unchanged from Round 2 |
| U2 | any | P1 | PASS | Both enumerators (body, `origin/main..760fd433`) empty. No closing keyword on an unfinished issue — the harm-bearing direction the rule names. The other direction is the P2 below |
| U3 | any | P1 | PASS | `Issue: #1100` on line 1 of the body |
| U4 | any | P1 | PASS | 2 commits in range, 0 offenders. The new one: `fix(fleet): keep the (#N) PR back-reference on the pinned squash subject (GH-1100)` |
| U6 | any | P0 | PASS | READ **CI run 34435944312 @ 760fd4334be15ac6306846d07307bc6cf9d48215** — 12/12, `completed/success` on every check run incl. `CI Gate`, `Test (windows-latest)` (10m12s) and both Fleet shell jobs. Not re-RAN |
| U7 | any | P1 | PASS | Whole routed audit finished before posting; every routed rule carries a row. No blocker withheld |
| U8 | any | P1 | N.A.(the DECISIONS enumerator printed nothing — no `Decision:` line) |
| D1 | docs | P1 | PASS | Delta enumerator: `edda review --help` → 0 OK, `git log -h` → 129 OK (§5.1's git rule, not `--help`; issue #691). Beyond it, the delta's load-bearing claim is about `gh`, so I probed it: gh 2.87.0 lists `-t, --subject` and `-F, --body-file`; and the behavioural claim is measured below under §5.5 |
| D2 | docs | P1 | PASS | `pr.closing-keyword` and `fleet.merge-authority` resolve from the checkout and match their use here. `review.merge-gate` / `mechanism.shell-role` → `No results found.` — an unreachable key, not a false claim (§5.1 D2); the delta asserts no value for either |
| D3 | docs | P1 | PASS | 1 repo-anchored path added in the delta: `docs/fleet/rules.md` → OK |
| D4 | docs | P0 | PASS | 2 candidates: `merge-reviewed-pr.sh:261` (the gate's own merge call) and `test-…:156` (a fixture assertion that the merge was *not* reached). Runbook step 7 still opens with 「規則閘綠即可執行……`docs/fleet/rules.md` R6」 — the required caveat, same paragraph |
| D5 | docs | `[判斷]` | PASS | Derivation: **clear** — the added runbook clause names the mechanism (GitHub suffixes only a subject it picks), the consequence (`--subject` alone drops the pointer), the remedy (the script appends), what U4 judges (the bare title), and both halves of the skip rule. **Scoped** — one sentence added to step 7 plus the `--body-file` parenthetical; no other file. **Non-duplicative** — still points at `REVIEW.md` §5 U4 rather than restating the type list. Step 7 is now a very long single paragraph, but every numbered step in this file is; house style, not a defect. No finding, so nothing to tag `provisional` (§6.1 attaches to a finding) |
| C1 | code-plain | P0 | PASS | All five doneWhen re-matched to code and test at the new head — ledger below. doneWhen 1 ("always derived from the PR title") is still satisfied: `merge_subject` derives from `$subject` and reproduces exactly what GitHub writes for a multi-commit PR |
| C2 | code-plain | P0 | PASS | `fix:` with a real regression suite, and I proved it bites rather than reading the claim — five mutations, below. CI ran the unmutated suite green on Linux **and** Windows at this SHA (`run-fleet-tests.sh:42` globs `scripts/test-*.sh`) |
| C3 | code-plain | P1 | N.A.(no `crates/**` change) |
| C4 | code-plain | P1 | PASS | Enumerator empty over the delta |
| C5 | code-plain | P1 | N.A.(selector empty — no crate touched, so no Windows coverage gap exists to close; L1's focused Windows-gap run has nothing to select) |
| R1 | code-risk | P0 | PASS | Delta enumerator empty. The only destructive command in the touched region is `rm -f "$merge_body"` in the EXIT/signal trap at `:231`, on a path `mktemp` created at `:230`; trigger stated, and it is reachable only inside the compose-a-receipt branch, so a caller-supplied `--body-file` is never deleted |
| R2 | code-risk | P0 | PASS | Zero candidates on the shell paths **and** zero over the whole delta (the rule's boundary is not the path filter, so I ran both). Every added conditional is single-operator: `[ -n "$2" ] \|\| die`, `[ -z "$body_file" ] \|\| [ "$action" = --merge ] \|\| die`. No line mixes `&&` and `\|\|`, so there is no parse tree to adjudicate |
| R3 | code-risk | P0 | PASS | `sh -n` at **head content** (`git show 760fd433:<path>`) → exit 0 for both changed scripts. The working tree is `main`, so parsing it would have checked base bytes |
| R4 | code-risk | P1 | PASS | Round 2's P2 taken: `trap 'rm -f "$merge_body"' 0 HUP INT TERM` (`:231`), matching the suite's own `trap cleanup 0 HUP INT TERM`. Still the only trap in the file, still armed one line after `mktemp`, and `die` exits so the error path fires it. Untested — see P2 below |
| R5 | code-risk | P1 | PASS | Properties re-confirmed over the new code: `$pr` is pinned to `^[1-9][0-9]*$` at `:30` **before** it is interpolated into the `case` pattern at `:259`, and it is double-quoted there, so the pattern carries no glob metacharacter either way; `$subject` is the match *word*, never the pattern, so a `*` or `[` in a PR title is literal; `merge_subject` reaches `gh` as one quoted argv element and is never interpolated into a shell string; `$head` still matched against `^[0-9a-f]{40}$` at `:92`; the receipt is still written with literal `printf` formats, so a `%` in a title cannot become a directive; user file and `mktemp` template still cannot collide on one filename |
| §5.5 | any | — | PASS | Wiring table below. `sh scripts/wiring-scan.sh 09cd2e6 760fd433` → no new pub surfaces; its two "swallow" hits are both the word *silently* inside the new comments, not code |

#### C1 — doneWhen ledger at the new head

| #1100 doneWhen | Code | Test |
|---|---|---|
| Subject always from the PR title, any commit count | `:101` reads the title, `:257-261` composes and passes it unconditionally | cases 9, 10, 14, 15 pin the recorded `--subject` |
| Validated before the merge; `wip`/empty scope/missing type refuse with the offending string | `:102-103`, ahead of the union gate and of `gh pr merge`; `$subject` stays the bare title, so the suffix cannot rescue or break it | cases 11/12/13, and case 11 now also asserts the refusal quotes a string with no `(#N)` in it |
| Accepts a merge body via `--body-file` | `:40-48`, `:53`, `:56`, `:223` | case 10 (`cmp -s` byte-identity), plus 16 and 17 for the two operands that used to be accepted without a reader |
| Composes a receipt naming reviewed SHA, LGTM round, CI run | `:224-238` | case 9 greps all three out of the captured body |
| Runbook describes the new arguments | `operator-runbook.md:175` | — |

### Wiring

|---|---|---|---|---|
| ` (#N)` back-reference on the squash subject | `:257` `merge_subject="$subject (#$pr)"`, skipped by the `case` at `:258-260` when the title already ends in this PR's own number | `:261` `gh pr merge --subject` → GitHub's squash API → **the commit subject on `main`** | Fail-closed upstream: a non-conforming title already refused at `:103`. No swallow on this path | title → U4 → compose → gh argv, asserted by cases 9/10/14/15. The last hop (what GitHub then writes) is unassertable offline, so I measured it instead — see below. Not a `doneWhen` layer, so §5.5's "no test proves it" clause does not fire |
| `--body-file` empty-operand refusal | `:45` `[ -n "$2" ] \|\| die` | the operator, stderr + exit 1 | Refuses; no best-effort path | parser only. Case 16 red under mutation |
| `--body-file` merge-only refusal | `:56` `[ -z "$body_file" ] \|\| [ "$action" = --merge ] \|\| die` | the operator, stderr + exit 1 | Refuses instead of accepting an argument nothing would open | parser only; no persisted state. Case 17 red under mutation |
| Trap signal list `0 HUP INT TERM` | `:231`, armed one line after `mktemp` | the filesystem (the composed receipt is removed on every exit path) | n/a — cleanup, not a decision | compose branch only. **No test**: reverting it to `0` leaves the suite green (measured) |
| `--help` text (`:22-24`) | static echo | the operator | n/a | documentation only |

No new `pub` surface, no new config key, no new event field.

**The measurement behind row 1**, since the offline suite cannot reach GitHub: `git log origin/main --format=%s -60` → 58 of 60 subjects end in ` (#N)`; the only two that do not are `8d838e5` (#1077) and `ce42972` (#1081), which Round 2 established were the two merged by hand with an explicit `--subject`. `gh api repos/fagemx/edda` → `squash_merge_commit_title=COMMIT_OR_PR_TITLE`. Both directions agree: GitHub suffixes a subject it picks and does not suffix one it is given, so appending here restores exactly the status quo ante rather than inventing a convention.

### Findings

Round 2's single blocker is resolved, and I confirmed it by mutation rather than by reading the response:

| Mutation (sandbox copy of head bytes, restored after each) | Result |
|---|---|
| `merge_subject=$subject` — drop the append | **RED** — `FAIL: case 9: the squash subject is not the PR title plus its (#N) back-reference`; recorded call showed `--subject fix(fleet): make the squash subject the PR title (GH-1100) --body-file`, no suffix |
| widen the skip to `*" (#"*")"` — any number | **RED** — recorded call showed `--subject fix(fleet): follow-up to the earlier change (#999) --body-file`, i.e. another PR's number left standing as the back-reference (case 15) |
| delete the `--body-file` merge-only refusal | **RED** — `FAIL: case 17: --body-file was accepted and ignored under --check` |
| delete the `--body-file` non-empty refusal | **RED** — `FAIL: case 16` |
| revert the trap to `0` only | **GREEN** — the one delta change no case covers |

Unmutated suite: `PASS scripts/test-merge-reviewed-pr.sh (17 cases)`, exit 0, before and after.

- [P2] **The trap signal-list fix has no regression test.** Reverting `:231` to `trap 'rm -f "$merge_body"' 0` leaves all 17 cases green (measured above), so the P2 Round 2 raised is fixed but unpinned — nothing stops a future edit from dropping `HUP INT TERM` again. Non-blocking: the R4 bar was already met by the EXIT trap alone, and this is a hardening change, not the `fix:` this PR is titled for.
- [P2] **`--help` does not state either new `--body-file` restriction.** `:18` still prints a bare `[--body-file <path>]`, and the added lines `:22-24` describe only the back-reference. The runbook now says 「只在 `--merge` 有效，空字串或配 `--check` 都會拒絕」; the tool's own usage does not, so an operator learns it from the refusal.
- [P2] **The body still carries no `closes #1100` although all five doneWhen are delivered.** `pr.closing-keyword=only-when-all-donewhen-delivered` reads the relation as a biconditional, so on this diff the body should close the issue and #1100 will otherwise stay open after merge. Not a blocker: U2's mechanical finding is the harm-bearing direction (a closing keyword on unfinished work), the enumerator printed nothing, and this was observable in Rounds 1 and 2 — `loop` item 2 forbids promoting it now.

No P0. No P1.

### FOLLOW-UP ISSUE

- **The U4 regex now exists twice with nothing keeping the copies honest.** `merge-reviewed-pr.sh:102` and `REVIEW.md` U4 are byte-identical today, but no test asserts they stay that way; change the type list in `REVIEW.md` and the merge entrypoint silently judges by the old one. Worth a sync test or one shared source, and it survives the migration since `edda review merge` in #1121 inherits the duplication. (Carried from Rounds 1 and 2; still out of scope.)
- **Nothing checks that the subject `gh` is handed is the subject that lands.** The suite can only assert argv; the final hop is GitHub's. Today that gap is closed by the 60-commit measurement above, but it is measurement, not a gate — if GitHub ever starts suffixing supplied subjects too, `main` would silently start collecting `… (#N) (#N)` and no test here would notice. A cheap post-merge assertion (compare the landed subject against the PR title after `reclaim`/digest runs) would close it. New this round; evidenced, out of the frozen surface.
- **PR #1121 also implements GH-1100 and touches these same two files.** It is stacked on #1119's branch, not on `main`, so this is not a current-base conflict and does not block #1118 — but whichever lands second needs a rebase and a fresh round, and **the back-reference append must come with it**, or the verb reintroduces the regression the moment it takes over the merge call. (Carried from Round 2; the implementer flagged the same thing.)

### RAN vs READ

RAN:
- `git show 3fbf08ac…:REVIEW.md` → spec loaded at base; `git hash-object` `3de7e327…` identical to the checkout copy. `REVIEW.md` not in the diff
- REVIEW.md's own classifier block on the file list → `classes=code-risk docs`, `canonical_class=code-risk`
- U2 (both enumerators, empty) / U3 (`Issue: #1100`) / U4 (2 commits, 0 offenders) / DECISIONS (empty)
- D1, D3, D4 delta enumerators; `edda review --help` → 0, `git log -h` → 129, `gh pr merge --help` (gh 2.87.0: `-t, --subject`, `-F, --body-file`, `--match-head-commit`), `edda --version` → `edda 0.6.1 (3fbf08ac558b 2026-09-09)`
- `edda ask` × 4 (`pr.closing-keyword`, `fleet.merge-authority` resolve; `review.merge-gate`, `mechanism.shell-role` unreachable)
- C4, C5, R1, R2 (shell paths **and** whole delta) → all empty
- `sh -n` on both changed scripts at **head content** → exit 0 each
- `sh scripts/wiring-scan.sh 09cd2e6 760fd433` → no new pub surfaces; two comment-text "swallow" false positives
- **Prior-blocker confirmation and C2 proof**: the 17-case suite plus five mutations, in a `git init` sandbox outside this repo holding only the head bytes of `merge-reviewed-pr.sh`, `test-merge-reviewed-pr.sh` and `fleet/verdict-drift.sh`. This is not a rerun of a recorded gate — CI runs the *unmutated* suite; nothing in CI proves the new assertions bite
- **Back-reference measurement**: `git log origin/main --format=%s -60` (58/60 carry ` (#N)`), the two exceptions resolved to `8d838e5`/#1077 and `ce42972`/#1081, `gh api repos/fagemx/edda` → `squash_merge_commit_title=COMMIT_OR_PR_TITLE`
- `git grep` for callers → only `edda-release/SKILL.md:205,207`, neither passing `--body-file`
- `git status --porcelain` before and after → identical (43 lines), branch still `main`

READ:
- **CI run 34435944312 @ 760fd4334be15ac6306846d07307bc6cf9d48215** — Format, Clippy ×3 OS, Test ×3 OS (Windows included, 10m12s), MSRV 1.91.0, Fleet shell tests, Fleet shell tests (windows), Detect code changes, `CI Gate`: 12/12 `completed/success`. Covers the unmutated suite on Linux and Windows.
- Not covered by that run: nothing requiring a Cargo gate — this diff touches no crate, so C5's selector is empty and L1's focused Windows-gap run has nothing to select. No workspace gate was run and none was needed (`ladder` L2).

### Verdict

LGTM (P0=0, P1=0)

fagemx:
**Merged — R6 gate receipt**

`2698260` on `main`, squash-merged with `--match-head-commit 760fd4334be15ac6306846d07307bc6cf9d48215`.

R6, each verified independently before acting:
- **Union on the head is a pass.** Three §7 verdicts exist; Rounds 1 and 2 are pinned to the superseded `09cd2e67`, so only Round 3's `LGTM (P0=0, P1=0)` stands on `760fd433`. No SHADOW round, no other verdict on that SHA (§8, GH-742).
- `CI Gate` green — run 34435944312 @ `760fd433`, 12/12 including all three Test jobs and both Fleet shell jobs.
- SHA window empty — `git log 760fd433..origin/main` over the three changed paths returned 0 commits.
- P0=0, P1=0.

**Subject left to GitHub deliberately.** Two commits, so GitHub composes from the PR title and appends the back-reference — which landed as `… (GH-1100) (#1118)`. That is the behaviour Round 2 found this PR was about to remove, so the merge that ships the fix also demonstrates it.

**On the independence caveat Round 3 disclosed.** Rounds 1-3 and both fixes came from session `fe9806e8-…`; Round 3 ran as a fresh-context subagent of it. `.claude/CLAUDE.md` GATE-01 permits "a session **or subagent** other than the implementer", while `REVIEW.md`'s front matter says `independence: session`. Merged on the former: `REVIEW.md`'s own canonical-sources rule says that when it and a citation disagree, the citation wins — and GATE-01 is cited to `.claude/CLAUDE.md`. Recording the reading rather than leaving it implicit, so it can be overruled if the operator wants the stricter one. Round 3 was not a rubber stamp in any case: it rebuilt a sandbox from the head bytes and mutation-tested three ways, and Round 2 — from the same session — overturned Round 1's LGTM.

**Sequencing note for #1121**, which folds GH-1100 into `edda review merge` and touches these same two files: it must now rebase onto this commit and carry the ` (#$pr)` append with it, or the verb reintroduces exactly the regression Round 2 caught. The two follow-ups Round 3 named are unrelated to that and remain open: the U4 regex duplicated between `merge-reviewed-pr.sh:102` and `REVIEW.md` with no sync test, and nothing asserting the subject handed to `gh` equals the subject that lands.

PR #1183, opened 2026-09-13
fix(pi): corrupt managed-run records degrade per record instead of aborting reads (#1182)

fagemx:
Closes #1182.

## What

A managed-run or service `state.json` that is 100% `0x00` (or otherwise invalid JSON) made
`store.mjs:readJson` rethrow the raw V8 `JSON.parse` error. Every registry-wide read then died
with exit 1 and 56 bytes of parser text. This makes a corrupt record a **per-record
degradation**, not a registry outage. Read-only resiliency: no record is written, repaired or
deleted.

## Behavior

| File | Change |
|---|---|
| `store.mjs` | `readJson` strict for valid/ENOENT, typed `RecordUnavailableError` (`code: record_unavailable`, no record bytes) for unparseable/oversized/symlink; new tolerant `readRecord(path)`; `registry()` returns `{ dir, state, owner, stateError, ownerError }` |
| `client.mjs:listSessions` | corrupt row degrades, identity recovered from `owner.json`; healthy rows byte-identical |
| `dependency-client.mjs:doctor` | survives a corrupt record and reports `unreadable: [{ sessionId, record }]` |
| `managed-client.mjs:managedStatus` | typed degraded result from `config.json` identity (whitelisted `runId/project/release/provider/model/thinking`; never spreads `config`, which holds `prompt`) |
| `managed-client.mjs:managedConversation` | recovers `runId -> sessionId` + persisted transcript by enumerating owned `sessions/*.jsonl` (`header.type === 'session'`, `header.cwd === config.project`) |
| `activation.mjs:listManagedRuns` | keeps `project`/`releaseId` from a valid `config.json` when `state.json` is corrupt |

NUL, non-NUL invalid JSON, oversized and symlink records all produce the same typed degradation.
No raw parser/NUL text reaches stdout/stderr; `prompt`/`token`/provider-error fields never leak;
the read path never writes (file bytes asserted unchanged).

## Gate receipt

- Frozen SHA: `7b1d4421d88c47604d2596ec79d660293935ebd9` (first commit `66195e4`)
- Base: `38390009207ce2860bc132d67c843da363f652db` (`origin/main`)
- L0: `node --test *.test.mjs` -> exit 0, `tests 90 / pass 90 / fail 0` (re-confirmed independently in the Round-1 SHADOW review: https://github.com/fagemx/edda/pull/1183#issuecomment-5654392869 )
[...]

PR #1179, opened 2026-09-13
feat(pi): adopt native continuity capsules through public edda restore (#1177)

fagemx:
## Summary

Connects the existing Pi adoption entry to Edda's native continuity capsules. `edda-pi adopt` (and `compose`) now accept `--capsule CAPSULE_ID`, obtain the readable restored context through the public installed `edda continuity restore CAPSULE_ID --json`, and merge it into the existing bounded `--context` path. The operator no longer restores, copies or saves the prose, and no intermediate context file is created.

Closes #1177.

## Frozen SHA

`bb74219790432436614a38f61576f49eefaa2130` (base `38390009207ce2860bc132d67c843da363f652db`).
Owned paths only: `integrations/pi/**`. No Rust, Cargo, CI, agent-manager, service, config, Character or registry mutation.

## What changed

- `integrations/pi/continuity.mjs` (new): `restoreCapsuleContext()` runs the public CLI through the existing `eddaCommand`/`EDDA_BIN`/`--edda-bin` seam (`continuity list --json` for repository membership, then `continuity restore <id> --json`), bounded by `maxBuffer` (512 KiB) and a strict envelope validator. `capsuleContextDocument()` renders a bounded (≤256 KiB) data-only document that keeps the data-only banner, warnings, repository/Git provenance and the native read-back digest visible. A declared block inside the capsule keeps its fence at line start so the existing `managementMetadata` parser reads it.
- `integrations/pi/compose.mjs`: `composeHandoff({ capsuleId })` merges the restored document into the same `brief.text` slot used by `--context FILE` (bare JSON metadata is re-fenced so the capsule section can follow), then the existing source-cache digest and 256 KiB bound apply unchanged.
- `integrations/pi/adoption.mjs`: `adoptSession({ capsuleId })` passes it through preflight and surfaces the capsule facts on the result. Refusal statuses are returned instead of a generic `needs_context`.
- `integrations/pi/cli.mjs`: `--capsule` on `adopt` and `compose`, plus `--edda-bin` on `adopt` for parity with `compose`. New statuses are in the exit-2 list.
- Tests: `continuity.test.mjs` (envelope/refusals/compose) and new `adoption.test.mjs` cases (success, capsule+`--context`, and every refusal with no side effects), using an isolated fixture `edda` (`fixtures/edda-capsule-reader.mjs`) and temp dirs.
- Docs: `README.md` and installed `getting-started.md` describe the successful path and refusal paths.

## Acceptance mapping (#1177 Done-when)

[...]

PR #1191, opened 2026-09-13
fix(pi): flush managed state writes before rename so an interruption cannot leave state.json all-NUL

### LINKED ISSUE #1190: fix(pi): flush managed state writes before rename so an interruption cannot leave state.json all-NUL
## Problem

After the 2026-09-13 interruption, four managed-run `state.json` files and eight service
`state.json` files in one registry were **100 % `0x00`** with **normal lengths** (1879, 1887,
1890, 1891 B; services 480–545 B). Their mtimes clustered within ~1 s at `12:24:00.499Z`,
`12:24:00.751Z`, `12:24:01.047Z`, `12:24:01.540Z` — the managed-run heartbeat writer. `config.json`
and `owner.json` parsed; transcripts/receipts survived. GH #1183 (merged `30d98c1`) made the
*readers* degrade per record instead of aborting; it deliberately made no claim about the writer.

## Root cause (writer durability, not the reader)

Every managed/service `state.json` write goes through `integrations/pi/store.mjs:writeJson`:

```js
const tmp = `${path}.${randomUUID()}.tmp`;
writeFileSync(tmp, text, { flag: 'wx', mode: 0o600 });   // data may still be in cache
renameSync(tmp, path);                                    // metadata swapped
```

The rename is only durable once the tmp's data is on disk. Node's `writeFileSync` never
`fsync`s, so on NTFS a hard interruption (forced reset/power loss, or a hard kill racing the
flush) can leave the renamed target with its recorded length but unflushed data pages, which
reads back as all NUL. GH-715 documented exactly this NTFS shape for `.git/config`
("NTFS 對『檔案已延伸但資料尚未 flush』的中斷，落盤結果正是一段 NUL"). The heartbeat writer
(`managed-runner.mjs:save`, every 2 s) and the service writer (`store.mjs` `state()`) both use
this path, which is why many files flipped to NUL together at the same instant.

**Cause established for the writer path: a non-durable tmp+rename.** Existing NUL files are not
repaired and the one-off event's exact trigger (power vs forced reset) is not distinguished — the
durability hole is what makes any such interruption able to produce it.

## Fix (smallest durability correction)

In `writeJson`, `fsync` the tmp file **before** the rename (and flush the exclusive write too),
then best-effort flush the parent directory on POSIX. A crash then leaves either the old complete
record or the new complete record — never NUL.

## Reproduction

`integrations/pi/durability.test.mjs` uses a bounded fault model of the NTFS behaviour: a renamed
tmp whose data was never flushed reads back zero-filled; a flushed tmp keeps its bytes.

- legacy algorithm (tmp + rename, no flush) → reproduces the all-NUL record;
- `writeJson` → the record is intact and parses.

## Scope / non-goals

- In scope: `integrations/pi/store.mjs` (`writeJson`) + the new test.
- Not changed: the #1183 reader degradation; writers are not "repaired"; no repair of the existing
  NUL files; no new recovery framework; no private transcript/credential reads; no global settings.
- Follow-up if wanted: bounded rotation/backup of the last known-good `state.json` (separate issue).

fagemx:
Closes #1190.

## What

`integrations/pi/store.mjs:writeJson` wrote each record to a tmp file and renamed it into place
**without flushing the tmp's data first**. Node's `writeFileSync` never `fsync`s, so on NTFS a hard
interruption between the write and the rename can leave the renamed target with its recorded length
but unflushed data pages — an all-NUL document. That is the writer path behind the 2026-09-13
incident: four managed-run `state.json` (1879–1891 B) and eight service `state.json` (480–545 B)
became 100 % `0x00` while `config.json`/`owner.json` parsed, with the run files' mtimes clustered
within ~1 s at `12:24:00.499Z`–`12:24:01.540Z` — the `managed-runner.mjs:save` heartbeat writer.
GH-715 recorded the identical NTFS shape for `.git/config`.

## Fix

`writeJson` now `fsync`s the tmp file **before** the rename, and flushes the exclusive write too;
the parent directory is flushed best-effort on POSIX. A crash leaves either the old complete record
or the new complete record, never NUL. This is the smallest durability correction; the GH #1183
reader degradation is unchanged and this is not a claim about repairing existing files.

## Reproduction

`integrations/pi/durability.test.mjs` (4 tests) drives a bounded fault model of the NTFS behaviour —
an unflushed rename reads back zero-filled, a flushed one keeps its bytes:

- the legacy algorithm (tmp + rename, no flush) **reproduces the all-NUL record**;
- the corrected `writeJson` yields the intact, parseable record;
- the exclusive write is flushed too;
- the real `fileIo` path (`fsyncSync` + parent-directory flush) writes a parseable record on the
  actual filesystem for both the tmp+rename and the exclusive branch.

## Gates

- Head: `c500a45ed42626ffa0aa7ba8ad0a7604691a810a`
- Base: `main` @ `781040afe4fdcb77838b87726b97e3756ecfe82c`
- `integrations/pi` → `node --test *.test.mjs` = **94 tests / 94 pass / 0 fail** (was 90; +4).
- Build lane: n/a (Node package).

## Review rounds

- Round 1 authoritative (official `deepseek-flash` via Pi, LGTM) → 2 P2 fixed.
- Round 2 Changes Requested (P1: swallowed parent-directory flush) → fixed.
- Round 3 Changes Requested (P1: propagated dir-fsync failure could turn a durable write into a
  false failure; P2 comment drift; P2 real-path coverage) → fixed; POSIX directory-fsync errnos
  (`EINVAL`/`ENOTSUP`/`EBADF`) tolerated, other failures propagated, comment corrected, and a real
  `fileIo` test added.
- Round 4 (current head) — this correction.

## Surface

Changed paths are all `integrations/pi/**` → R22 **internal-tool** surface, where official
`deepseek-flash` is authoritative (R22 qualified, `origin/main` `419ac07`).

## Out of scope

Existing NUL files are preserved as fixtures and not repaired; no backup/rotation framework; no
private transcript/credential reads; no global settings.


fagemx:
## Review Response: Round 1 — PR #1191 @ 1f97bb95cd13e2601be3aa22af807ed2b1921ab2

(Authoritative Round 1: `review_verdict` evt_01m2ee7yvmhxbtkb6p6tyn2s11, `deepseek/deepseek-flash` via pi, **LGTM** with 2 P2s.)

- **P2 (fixed) — the exclusive-write test could not observe the flush.** The fault model now keeps written-but-unflushed data separate from what a reader sees (`pending` vs `durable`); a direct write is zero-filled until a flush promotes it. The exclusive-branch assertion therefore genuinely fails when the exclusive path is not flushed, instead of passing on a direct write.
- **P2 (fixed) — the directory-flush comment described a no-op as a substitute barrier.** The comment now states the surviving-record bound: the file data is already durable before the rename, so a swallowed directory flush can only keep the previous complete record, never a zero-filled one.

Gate receipt: `integrations/pi` → `node --test *.test.mjs` = **93 tests / 93 pass / 0 fail** @ `1f97bb95cd13e2601be3aa22af807ed2b1921ab2`; exact-head CI (`CI Gate` + Pi session channel) pending on this push.

fagemx:
## Review Response: Round 2 — PR #1191 @ b9693938dd9796730e77a0678d844f1ed84d54e8

(Authoritative Round 2: `review_verdict` evt_01m2eemym0sz90vq37xqe183x8, `deepseek/deepseek-flash` via pi, **Changes Requested**.)

- **P1 (fixed) — the directory flush was swallowed.** `flushDir` no longer catches: on POSIX it opens the parent directory and `fsync`s it, and a failure propagates so an unconfirmed directory entry is never reported as success. (Windows still cannot open a directory; there the file-data flush is the barrier.)
- **P2 (fixed) — the create branch did not flush its parent directory.** The exclusive/create branch now calls `flushDir` too, so `owner.json`/`config.json`/identity records get the same POSIX rename durability.
- **Test (fixed) — `flushDir` was a no-op in the model.** The fault model now records `flushDir` calls and both the rename and create paths assert exactly one, so the directory flush is pinned on both branches.

Gate receipt: `integrations/pi` → `node --test *.test.mjs` = **93 tests / 93 pass / 0 fail** @ `b9693938dd9796730e77a0678d844f1ed84d54e8`; exact-head CI (`CI Gate` + Pi session channel) pending on this push.

fagemx:
## Review Response: Round 3 — PR #1191 @ c500a45ed42626ffa0aa7ba8ad0a7604691a810a

(Authoritative Round 3: `review_verdict` evt_01m2ef29wk4mp1ftr7t0v7v2e3, `deepseek/deepseek-flash` via pi, **Changes Requested + escalation**.)

- **P1 (fixed) — propagation turned a successful, durable write into a false failure.** `flushDir` now tolerates the POSIX "directory fsync not supported here" errnos `EINVAL`/`ENOTSUP`/`EBADF` (the macOS `EINVAL` case in the escalation) because the record's file data is already durable, and propagates any other failure so a genuine durability shortfall is not silently success. This is the best-effort contract, without the blanket swallow Round 2 flagged.
- **Escalation (resolved in-branch).** Rather than requiring a macOS probe before merge, the tolerated set covers `EINVAL`/`ENOTSUP`/`EBADF`, so a POSIX platform that rejects a directory fsync no longer fails `writeJson`. `durability.test.mjs` now runs the **real** `fileIo` path (so the POSIX directory flush executes on the macOS CI job).
- **P2 (fixed) — the comment claimed every record is tmp+renamed.** The comment now distinguishes the tmp+rename path from the exclusive create-in-place path, which relies on the same file flush.
- **P2 (fixed) — the legacy case never called `writeJson` and the real path had no executable coverage.** Added a real-`fileIo` test: a tmp+rename write and an exclusive create in a real temp dir, asserting both parse and no `.tmp` is left.

Gate receipt: `integrations/pi` → `durability.test.mjs` **4/4**; full `node --test *.test.mjs` = 94 tests / 93 pass, the one failure being the pre-existing heavy `real supervisor service starts…` test under parallel load, which **passes in isolation** (10.2 s). Exact-head CI (`CI Gate` + Pi session channel on ubuntu/macos/windows) pending on this push.

fagemx:
## Code Review: Round 4 — PR #1191 @ c500a45ed42626ffa0aa7ba8ad0a7604691a810a

- elapsed: 245520 ms (measured)
- model_requested: `deepseek/deepseek-flash`
- model_observed: `deepseek/deepseek-flash` (in-band)
- reviewer_session: `cdb6c4e4-fc01-57ba-bc36-175e24b11df3` (`review-c500a45ed426-r4`, tool policy hard)
- posted_by: replacement delivery controller on behalf of the reviewing session (the R22-hardened Pi reviewer runs read-only and cannot post); verdict facts are the ledger `review_verdict` event `evt_01m2egv9ppcsb6erxyxa62y87t`
- spec: spec-backed (issue #1190, trust maintainer) · core core-v1
- class: code-risk
- surface: R22 internal-tool, engine `deepseek-flash` via pi on the official `deepseek` provider — authoritative
- escalations: none
- cost: $0.0940 (measured)

### IN SCOPE
IN SCOPE (frozen): changed behaviour = `writeJson` durability barrier in `integrations/pi/store.mjs` (fsync the tmp before rename, flush the exclusive write, best-effort POSIX parent-directory flush with tolerated `EINVAL`/`ENOTSUP`/`EBADF`), plus the bounded fault-model test `integrations/pi/durability.test.mjs`. Direct consumers keep their contract: the injected `io` parameter is additive and the void return / `EEXIST` behaviour are preserved (`store.mjs:149,157,158,162`; `managed-runner.mjs:25`; `managed-client.mjs:63,69,112`; `supervisor-engine.mjs:22,183,184`; `supervisor-runner.mjs:20,44`; `supervisor-tools.mjs:41`; `handoff.mjs:27`; `dependency-observer.mjs:52,185`; `dependency-client.mjs:28`; `supervision.mjs:38,82,91,119`; `inbox-store.mjs:43`). No read path added; the #1183 reader degradation is untouched.

### Rules
| Rule | Result | Evidence |
|---|---|---|
| U1 | na | `integrations/pi/store.mjs`; `integrations/pi/durability.test.mjs` — the only two files in DIFF, both named by the SPEC "In scope" line |
| U2 | na | PR body and commit list are not in the brief and this session has no shell — input unreadable, not failed |
| U3 | na | same missing PR body; the brief supplies the issue body itself as SPEC |
| U4 | na | no commit list in the brief |
| U5 | na | no `*.md` in the diff; EVIDENCE Gates READ records `sh scripts/lint-markdown-content.sh` green (Format=pass) |
| U6 | na | EVIDENCE Gates READ: CI Gate green, Format green; Clippy/Test jobs `missing` for a Node-only diff → `unverified` (advisory, review-spec-v1.10); host offered no RAN measure IDs |
| U7 | na | every rule §4 routed is accounted for (U1–U8, §5.5, C1–C5, R1–R5); no later-round blocker |
| U8 | na | no `Decision:` line in the brief; the ledger's only active decision `usable-continuity.gh1177-capsule-adopt` belongs to PR #1179, untouched here |
| §5.5 wiring | na | `store.mjs:110` (io param), `store.mjs:100-108` (flushDir), `durability.test.mjs` (new file) — one new seam, no consumer gap |
| C1 | na | SPEC Fix + Reproduction matched at `store.mjs:110-128` and `durability.test.mjs:33-76` |
| C2 | na | `durability.test.mjs:33-41` is the real regression test; on the base code the injected 4th arg is ignored and the real `writeFileSync('C:/registry/…<uuid>.tmp')` throws ENOENT, so it fails without the change |
| C3 | na | diff touches no `crates/**` |
| C4 | na | diff touches no `*.rs` |
| C5 | na | no `crates/**` path; C5 crate selector empty |
| R1 | na | `durability.test.mjs:75` `rmSync(dir,{recursive:true,force:true})` — `dir` is the `mkdtempSync()` at :65, inside `finally`; no other destructive call in added lines |
| R2 | na | no shell-bearing path changed; no added line mixes `\|\|` with `&&` |
| R3 | na | diff changes no `*.sh`/`*.bash` |
| R4 | na | `store.mjs:113-116,120-125` close the fd in `finally` on both branches and unlink tmp on every exit including the error path; `durability.test.mjs:74-76` removes the temp dir in `finally` |
| R5 | na | tmp name carries `randomUUID` and is opened O_EXCL (`wx`) so a same-name collision fails closed; no shell interpolation; no caller-supplied SHA |
| D1–D5/S1–S3 | na | class is code-plain/code-risk and the diff adds no `docs/**` or skills path; D5 closed by this authoritative engine |

### Wiring
No new writer surface; `writeJson`'s fourth parameter is an additive test seam, and the parent-directory flush is the only new wire end (best-effort, tolerated errnos, other failures propagate).

### Findings
- [P2] `TOLERATED_DIR_FSYNC` tolerates `EBADF`, which does not mean "not supported here" as the same hunk's rationale states, so an `fsync` failing with EBADF silently skips the durability barrier the change exists to enforce; unreachable today (the fd was just opened), so quality-only — drop EBADF or document it separately. Trigger: only an already-invalid dir fd, which the synchronous open→fsync→close sequence cannot produce. — evidence: `integrations/pi/store.mjs:86,104,106`
- [P2] The added causal comment says an interruption "between the tmp write and the rename" leaves the target zero-filled, but a crash strictly before the rename leaves the old target intact; the target can only read back as NUL once the rename has moved unflushed data. Wording only — ordering and the fix are correct. — evidence: `integrations/pi/store.mjs:79-81` vs `durability.test.mjs:53-64`
- [P2] The "legacy" test drives the crash model by hand instead of exercising pre-fix `writeJson`, so it demonstrates the model rather than product behaviour; test 1 at :33-41 is the real regression test and does fail without the change. — evidence: `integrations/pi/durability.test.mjs:53-64` (io.open/io.write/io.rename called directly, no `writeJson`)

### FOLLOW-UP ISSUE
Three P2 advisory items above (EBADF tolerance wording, causal-comment wording, legacy-test fidelity) — non-blocking, routed as a follow-up rather than extending this frozen correction.

### RAN vs READ
RAN: none — the round ran under the R22-hardened read-only policy (hard), no host command executed.
READ: EVIDENCE-only exact-head CI at `c500a45ed42626ffa0aa7ba8ad0a7604691a810a` — CI Gate green, Format green, channel/manager green; Clippy/Test rows `missing` because `.github/workflows/ci.yml` skips `integrations/**` (legitimately not red). Subject verified by two independent reads: `.edda-review-subject` and the review worktree gitdir HEAD both give `c500a45…`, and both diff files match the DIFF body.

### Verdict
LGTM (P0=0, P1=0) — the change is the smallest durability correction on the actual writer path; the three findings are P2 advisory and do not block. No P0/P1 was found and the follow-up list does not extend this PR.

Issue #1208, opened 2026-09-14
test(pi): Windows temp-dir cleanup EBUSY races fail the channel CI job


fagemx:
# Problem

The `channel (windows-latest)` CI job intermittently fails on temp-directory teardown with
`EBUSY: resource busy or locked, rmdir '<tmp>\project'`, in `integrations/pi/managed.test.mjs`,
`supervisor.test.mjs` and `adoption.test.mjs`. It re-runs green on the same SHA.

## Evidence

- PR #1206 `Pi session channel` run `34810694398`: attempt 1 red only on `channel (windows-latest)`
  (`EBUSY … rmdir …\edda-adoption-test-…`); attempt 2, same run/head → ubuntu/windows/macos all green.
- Reproduces on base `6ee0675` with a full `npm --prefix integrations/pi test` (4 failures), and locally
  (varied tests fail per run: `adoption`, `managed`, `supervisor`). Running the affected files in
  isolation passes (e.g. `managed.test.mjs adoption.test.mjs` → 20/20).
- Related but distinct: GH #1204 (`privateRoot` PowerShell ACL `ETIMEDOUT`). This issue is the
  process/temp-dir cleanup race.

# Expected

Make the per-file temp project cleanup tolerant of a still-held Windows handle (bounded retry / open-
handle close before `rm`), so a teardown race cannot fail an otherwise-correct assertion. Do not weaken
the lifecycle assertions or add sleeps that mask a real leak.

# Scope

`integrations/pi/*.test.mjs` + `integrations/pi/fixtures/**` teardown (internal-tool). Pre-existing,
not caused by PR #1206.

fagemx:
taking: edda/winpi-controller at 2026-09-14T13:22:38Z

Issue #1204, opened 2026-09-14
fix(pi): privateRoot's powershell ACL step intermittently ETIMEDOUTs on Windows CI


fagemx:
# Problem

`integrations/pi/store.mjs` `privateRoot()` runs a PowerShell ACL script on every registry root:

```js
execFileSync('powershell.exe', ['-NoLogo', '-NoProfile', '-NonInteractive', '-File',
  fileURLToPath(new URL('./private-directory.ps1', import.meta.url)), '-Path', root],
  { windowsHide: true, timeout: 15000, stdio: 'pipe' });
```

On Windows CI it intermittently fails with `Error: spawnSync powershell.exe ETIMEDOUT`, taking down
whichever test first calls `privateRoot()` (observed: `CLI process survives launcher, reconnects,
stops…`, `concurrent launchers converge…`, `real Pi session listing discovers live and offline runs
read-only`). It re-runs green on the same SHA.

## Evidence (public, pre-existing, not caused by a diff)

- Job A reproduced it on base `main` `a1b7c6d4ff18e6741cdab894eadf7cf6246e4085` via `workflow_dispatch`
  with none of its changes present: run
  [34799835049](https://github.com/fagemx/edda/actions/runs/34799835049), job
  `manager (windows-latest)` `103840084295` → `tests 67 / pass 64 / fail 1`,
  `spawnSync powershell.exe ETIMEDOUT`.
- Job A analysis: https://github.com/fagemx/edda/pull/1196#issuecomment-5658242919
- Seen again on PR #1198's `Agent manager` Windows job (`EPERM`/`cli.js start --port 0` in the
  same cleanup region) and green on re-run.

## Impact

Every `integrations/pi/**` PR can show a red `Agent manager` Windows job for a reason unrelated to its
change. `manager` is not a required `CI Gate` dependency, so it does not block R6, but it is ambiguous
noise.

## Expected

`privateRoot()` should not let a transient PowerShell/ACL timeout fail unrelated tests: bound/retry the
secure step, degrade with a typed, visible signal, or make the ACL application idempotent/lazy so a slow
first call is not fatal. Do not weaken the private-root guarantee; do not add a scheduler or a new
storage layer.

# Scope

`integrations/pi/store.mjs` / `private-directory.ps1` (internal-tool). Filed from Job B (PR #1198) on
Job A's evidence; no fix round was spent in the merged PR because the flake predates it and is not
`CI Gate`-blocking.

fagemx:
[...]

Issue #1194, opened 2026-09-13
fix(pi): writeJson durability follow-ups from PR #1191 round 4 (P2)


fagemx:
Follow-ups from the authoritative round-4 review of the merged NUL-writer durability fix
(PR #1191, merged `861bff5`, reviewed head `c500a45`, `review_verdict` event
`evt_01m2egv9ppcsb6erxyxa62y87t`, official `deepseek-flash` via Pi on the R22 internal-tool
surface). All three are **P2 advisory, non-blocking**; the merge was accepted with P0=0/P1=0.

1. **`integrations/pi/store.mjs:86` — `TOLERATED_DIR_FSYNC` includes `EBADF`.** `EBADF` does not
   mean "not supported here" as the same hunk's rationale states, so an `fsync` failing with
   `EBADF` would silently skip the durability barrier the change exists to enforce. Unreachable
   today (the fd was just opened), so quality-only — drop `EBADF` or document it separately.

2. **`integrations/pi/store.mjs:79` — causal-comment wording.** The comment says an interruption
   "between the tmp write and the rename" leaves the target zero-filled, but a crash strictly
   before the rename leaves the old target intact; the target reads back as NUL only once the
   rename has moved unflushed data. Wording only; ordering and the fix are correct.

3. **`integrations/pi/durability.test.mjs:53` — legacy-test fidelity.** The "legacy" case drives
   the crash model by hand rather than exercising pre-fix `writeJson`, so it demonstrates the
   model rather than product behaviour. Test 1 at `:33-41` is the real regression test and does
   fail without the change. A standalone bounded reproduction that runs the unmodified base
   module is recorded at
   `C:/Users/fagem/.edda-pi-delivery/operator-visible-work/phase1b-repro/reproduce.test.mjs`.

Out of scope for this issue: repairing existing NUL files, backup/rotation, and any writer that
is not `writeJson`.

fagemx:
Absorbed and delivered by PR #1222 (merge commit `abecf441145ab84be1826f1c6a1f60d02856aac7`), which landed on top of the PR #1191 fix. All three items are addressed on the same writer:
1. `EBADF` in `TOLERATED_DIR_FSYNC` is now documented separately from the `EINVAL`/`ENOTSUP` "not supported here" case instead of dropped (dropping it changed base behaviour — round-1 authoritative finding).
2. The causal comment now states the NUL is observed once the rename of unflushed bytes lands, and that a crash before the rename leaves the previous complete record.
[...]

PR #1222, opened 2026-09-14
fix(pi): retry the Windows record replace so a reader cannot kill a managed runner (#1215)

### LINKED ISSUE #1215: fix(pi): writeJson rename can EPERM on Windows while a reader walks managed state (runner died)
# Problem

On Windows, `writeJson`'s atomic rename can fail `EPERM` while a reader is walking the registry, killing
a live managed runner. Observed 2026-09-14T10:54 in
`C:/Users/fagem/.edda-pi-delivery/operator-visible-work/registry`:

```
Error: EPERM: operation not permitted, rename
  '…\managed\82e6109c-8035-47aa-855c-783f287f475a\state.json.<uuid>.tmp'
  -> '…\managed\82e6109c-8035-47aa-855c-783f287f475a\state.json'
    at renameSync (node:fs)
    at Object.rename (…/releases/cf751f42…/store.mjs:92)
    at writeJson (…/releases/cf751f42…/store.mjs:122)
    at save (…/releases/cf751f42…/managed-runner.mjs:32)
    at Timeout._onTimeout (…/releases/cf751f42…/managed-runner.mjs:172)
```

The runner died ~6 s after launch; the child session's channel endpoint became unreachable and
`run-status` reported `runner_unreachable`. `edda-pi run-resume <runId>` recovered it in place (new
`runnerPid`, `status: ready`, `live: true`, initial prompt not replayed); `state.json` stayed valid.
Crash log preserved at `managed/82e6109c-…/runner.log`.

## Lead (not proven)

A rename-based writer racing a directory-walking reader is the shape the timeline supports: `edda-pi runs`
(`listManagedRuns` in `integrations/pi/activation.mjs`) reads **every** `managed/<runId>/state.json`
through `readRecord` (`store.mjs`), and the agent-manager polls every 3 s, calling `managedStatus`, which
falls back to `readRecord(state.json)`. On Windows a replace can fail `EPERM` when the destination is
transiently held without delete-sharing. Not reproduced deterministically.

## Expected

A transient reader must not kill a managed runner. Bound the rename with a short retry on `EPERM`/`EBUSY`
(Windows), or write state with a reader-tolerant strategy, without weakening the durability barrier from
PR #1191 and without a new storage format. A failed final write should degrade visibly (typed signal)
rather than crash the runner, if retries are exhausted.

## Cheap repro shape

Launch a managed run in a registry while invoking `edda-pi runs` in a tight loop (and/or an agent-manager
poll against that registry).

# Scope

`integrations/pi/store.mjs` `writeJson` (internal-tool). Related but distinct from the durability
follow-ups in GH #1194. Reported by Job A from a real launch failure; no fix round spent by Job B.

fagemx:
Closes #1215. Absorbs #1194.

## What

`writeJson`'s atomic replace could throw `EPERM` on Windows when a reader walked the registry
(`edda-pi runs` / `listManagedRuns`; the agent-manager's 3 s poll) or a scanner held the destination
without delete-sharing. Called from the 2 s `managed-runner` heartbeat (inside a `setInterval`, so
outside the runner's own `try`), that was fatal: the runner died ~6 s after launch
(`Error: EPERM … rename … state.json.<uuid>.tmp -> state.json` at `managed-runner.mjs` `save`), the
child channel endpoint became unreachable and `run-status` reported `runner_unreachable`.

## Fix

- `store.mjs`: the replace is retried a bounded number of times on the Windows replace errnos
  `EPERM`/`EBUSY` (six attempts, ~0.5 s cumulative backoff). The PR #1191 durability barrier is
  unchanged — the tmp is still fsynced before the rename, no storage format changed.
- When the window is exhausted, `writeJson` throws a typed **`RecordWriteError`**
  (`code: 'record_write_failed'`, `record`, `errno`, `cause`) instead of a raw errno.
- `managed-runner.mjs`: the heartbeat catches that typed failure, keeps the process alive, records
  `state.writeError` on its live state (so the runner endpoint and `run-status` show the degraded
  write) and retries the next tick. An exhausted write can no longer end a run.

## Absorbed GH #1194 follow-ups (same writer, same files)

- Keep `EBADF` in `TOLERATED_DIR_FSYNC` but document it separately from the `EINVAL`/`ENOTSUP`
  "not supported here" case: the synchronous open→fsync→close cannot produce it today, so tolerating
  it changes no current behaviour, while dropping it would newly propagate a directory-fsync error
  out of `writeJson` (round-1 finding).
- Correct the causal comment: the all-NUL record is observed once the rename of **unflushed** bytes
  lands, while a crash strictly before the rename leaves the previous complete record intact.
- The legacy durability case now runs the production `writeJson` with the durability flush disabled
  (the actual pre-1191 algorithm) instead of hand-driving the fault model.

## Tests

`integrations/pi` → `node --test *.test.mjs`: **143 tests, 142 pass**; the single failure was the
pre-existing GH #1208 Windows teardown `EBUSY` flake (passes in isolation and on retry). New coverage:

- `durability.test.mjs` (7): transient `EPERM`/`EBUSY` retried until the record lands; exhausted
  retry throws `record_write_failed` and never presents a partial record as the write; a
  non-retryable errno propagates unchanged and is not retried; the pre-1191 algorithm through the
  production writer reproduces the all-NUL record; the real `fsync`/parent-flush path.
- `managed.test.mjs`: a live managed run survives a 9 s reader walk (`listManagedRuns` +
  `managedStatus`) across several heartbeat windows and records no failed write; and a forced
  retry-exhaustion case (the state record destination becomes a non-empty directory) shows the
  runner staying live, reporting the typed `record_write_failed` failure through `run-status`, and
  clearing it once the record is writable again.

## Round-1 findings (addressed, head `0a09c36`)

- **P1** `TOLERATED_DIR_FSYNC`: restored `EBADF` and documented it separately (see above).
- **P2** `state.writeError` is now cleared by the shared `save()` on any successful write, so a
  recovered failure cannot be persisted by another save site and re-reported.
- **P2** terminal writes go through a `finalSave()` that logs the typed failure instead of rethrowing
  it as an unhandled error while the runner ends.
- **P2** the managed test suite now forces retry exhaustion end to end (see above).

`npm --prefix integrations/pi run test:package` → passed.

## Out of scope (recorded, not absorbed)

- **#1204** (`privateRoot` PowerShell ACL `ETIMEDOUT`) — different root cause (a spawn timeout in a
  different step); cannot be shown to share this bounded root/surface.
- **#1208** (Windows test temp-dir cleanup `EBUSY`) — test-teardown race in other files.

## Surface

Changed paths are `integrations/pi/{store.mjs,managed-runner.mjs,durability.test.mjs,managed.test.mjs}`
→ R22 **internal-tool**, where official `deepseek-flash` via Pi is authoritative
(`origin/main` `419ac07`). `integrations/pi/activation-receipt.mjs`, `scripts/activation/**`,
`edda-pi activation` help text and `integrations/agent-manager/**` are untouched (activation stream).


fagemx:
## Code Review: Round 2 — PR #1222 @ 0a09c36d6cdbdbe926450c84484e2bada31c52e4

- elapsed: 106775 ms (measured)
- model_requested: `deepseek/deepseek-flash`
- model_observed: `deepseek/deepseek-flash` (in-band; round-1 host receipt for this session)
- reviewer_session: `cc3f4e94-1db0-5139-826b-791603f58c9b` (`review-0a09c36d6cdb-r2`, tool policy hard)
- posted_by: delivery controller on behalf of the reviewing session (the R22-hardened Pi reviewer runs read-only and cannot post); verdict facts are the ledger `review_verdict` event `evt_01m2g28metmgmww9d3dk9vj9dz`
- spec: spec-backed (issue #1215, trust maintainer) · core core-v1
- class: code-risk
- surface: R22 internal-tool, engine `deepseek-flash` via pi on the official `deepseek` provider — authoritative
- escalations: none
- cost: $0.1400 (measured)
- previous: `evt_01m2g1sep5kweb2t04qy866tdn` · history rewritten: false

### IN SCOPE
IN SCOPE (frozen): the round-1 delta on the four changed paths. `store.mjs` bounds the replace with
a retry on `EPERM`/`EBUSY` and raises a typed `RecordWriteError` (`record_write_failed`) on
exhaustion; `managed-runner.mjs` keeps a live run alive, records `state.writeError`, and routes
terminal writes through `finalSave()`; `durability.test.mjs` covers retry, exhaustion,
non-retryable propagation, the pre-1191 reproduction and the real fsync path; `managed.test.mjs`
covers a reader walk and a forced retry exhaustion. The PR #1191 durability barrier (fsync before
rename) is unchanged and no storage format changed.

### Rules
| Rule | Result | Evidence |
|---|---|---|
| prior P1 (EBADF tolerance removal) | resolved | `store.mjs:96-101` restores `EBADF` and documents the tolerance separately; `flushDir` paths unchanged from base |
| prior P2 (stale `writeError` persisted by other save sites) | resolved | single writer `managed-runner.mjs:32-41` clears/preserves `writeError`; every save site routes through it (`53,90,112,163,176,178,186,189,198`) |
| prior P2 (terminal save rethrows unhandled) | resolved | `finalSave()` `managed-runner.mjs:45-51` catches and records the typed field; called at `205` and `207` |
| prior P2 (no test forcing retry exhaustion) | resolved | `managed.test.mjs:265-293` forces a persistent replace failure and asserts live + typed + recovery |
| U1 | na | round-2 DIFF file list inspected; four declared paths |
| U2/U3/U4 | na | PR body and commit list not supplied to this round; no `gh`/git execution in a read-only round |
| U5 | na | diff changes no markdown; host EVIDENCE maps the markdown gate to Format=pass |
| U6 | na | host EVIDENCE gate table read; no CHECKLIST MEASURE IDs issued this round |
| U7 | na | delta audited against `d97849c`; all prior findings verified resolved at `0a09c36` |
| U8 | na | no `Decision:` line in the brief; the unrelated ledger key is out of scope |
| C1 | na | `store.mjs:88-175`, `managed-runner.mjs:25-210`, both test files read against the SPEC Expected |
| C2 | na | `durability.test.mjs:76-113` and `managed.test.mjs:265-293` require the new retry and heartbeat catch |
| C3/C4/C5 | na | no `crates/**` or `*.rs` path in the DIFF |
| R1 | na | added lines inspected; only the pre-existing `store.mjs:172-174` tmp unlink |
| R2/R3 | na | no shell/CI path changed in the DIFF |
| R4 | na | `store.mjs:160-175` and `managed-runner.mjs:32-51` release the tmp on every exit; failure paths are typed or logged |
| R5 | na | `store.mjs:106-128,163` `randomUUID` tmp name, `basename`-only messages, `io.sleep` default |
| Wiring | na | new surfaces `RecordWriteError` / `state.writeError` / `finalSave` / `io.sleep` enumerated in `store.mjs:106-128` and `managed-runner.mjs:32-51,64,195-207` |
| Security / data loss | na | read `writeJson` and all managed-runner save paths; the NUL barrier is retained and no partial record is presented |
| Current-base integration | na | round-2 DIFF `0a09c36` over `17057395f`; host wiring-scan over the same range |

### Findings
- [P2] The exhaustion test asserts the Windows code is `record_write_failed`, which holds only if
  renaming a file onto a non-empty directory is `EPERM`/`EBUSY` on Windows; any other errno (for
  example `EACCES`) is not retried, so the assertion would fail and redden the
  `pi-session-channel` Windows job. Stated as unverified, not proven. — evidence:
  `managed.test.mjs:283` asserts `record_write_failed` on win32; `RETRYABLE_RENAME` is
  `{EPERM, EBUSY}` at `store.mjs:103`

### RAN vs READ
RAN: none — read-only round under the R22-hardened tool policy (hard).
READ: host EVIDENCE gate table — `CI Gate` green, `Format` green, Clippy/Test rows `missing` for a
Node-only diff (advisory). The `pi-session-channel` workflow is not among the provided receipts, so
test execution and the Windows errno mapping behind `managed.test.mjs:283` stay **unverified** in
this round; that is the sole finding.

### Verdict
LGTM (P0=0, P1=0) — all four round-1 findings are resolved at the new head, the durability barrier is
retained, and the sole remaining item is the advisory P2 about the Windows errno mapping.

fagemx:
## Review Response: Round 2 — PR #1222 @ 0a09c36d6cdbdbe926450c84484e2bada31c52e4

(Authoritative Round 2: `review_verdict` `evt_01m2g28metmgmww9d3dk9vj9dz`, official
`deepseek/deepseek-flash` via Pi, **LGTM**, 1 P2 advisory.)

- **P1 (fixed) — `EBADF` tolerance removal changed base behaviour.** `EBADF` is restored in
  `TOLERATED_DIR_FSYNC` and documented separately from the `EINVAL`/`ENOTSUP` "not supported here"
  case, so a directory-fsync error can no longer newly propagate out of `writeJson`.
- **P2 (fixed) — stale `writeError` persisted by other save sites.** The shared `save()` now clears
  `writeError` on every successful write and restores it on failure, so a recovered failure cannot be
  re-reported.
- **P2 (fixed) — terminal save rethrew unhandled.** Terminal writes go through `finalSave()`, which
  logs the typed failure instead of throwing while the runner ends.
- **P2 (fixed) — no test forced retry exhaustion.** `managed.test.mjs` now makes the state record
  destination a non-empty directory, forcing every replace to fail: the heartbeat exhausts its
  bounded retries, the runner stays live, `run-status` reports `record_write_failed`, and the failure
  clears once the record is writable again.

**Sole remaining P2 (advisory) — the Windows errno mapping.** The reviewer could not see the
`pi-session-channel` receipts. At this exact head `0a09c36`, the `Pi session channel` workflow is
green on **ubuntu-latest, macos-latest and windows-latest** (run
[34849964467](https://github.com/fagemx/edda/actions/runs/34849964467)), which runs
`npm --prefix integrations/pi test` — including the forced-exhaustion test and its win32
`record_write_failed` assertion. Locally on Windows the same case reproduces `EPERM` for
`rename(file → non-empty directory)` (so the errno is retryable). The P2 is therefore answered by
exact-head execution rather than left unverified; no code change was needed.

PR #1261, opened 2026-09-15
fix(pi): reliable owner-adoption dependency rebind and resume liveness (#1258, #1249, #1241, #1260)

### LINKED ISSUE #1241: test(pi): owner-mailbox suite is not hermetic against an ambient EDDA_RETURN_ROOT
# Problem

`integrations/pi` owner-mailbox tests are not hermetic against an ambient `EDDA_RETURN_ROOT`. A managed
Edda session exports `EDDA_RETURN_ROOT` to the live registry, and `fixtures/edda-return.mjs` prefers
`process.env.EDDA_RETURN_ROOT` over the per-test temp cwd, so `npm --prefix integrations/pi test` inside
such a session gives 158/163 with 5 owner-mailbox failures (e.g. `bind … {status:'replaced'}` instead of
`{status:'bound'}`). With `env -u EDDA_RETURN_ROOT` it is 163/163.

Found by the independent Round 1 review of PR #1238; pre-existing (no owner-mailbox file is in that diff)
but it makes the suite red precisely for the agents that run it.

# Expected

Make the owner-mailbox tests hermetic: pass the intended mailbox root explicitly to the fixture / test
process (e.g. set `EDDA_RETURN_ROOT` to the test's temp dir in the fixture, or have the fixture prefer an
explicit argument over the ambient env) so an ambient value cannot change the result. Do not weaken the
`EDDA_RETURN_ROOT` production behavior.

# Scope

`integrations/pi/owner-mailbox.test.mjs`, `integrations/pi/fixtures/edda-return.mjs`,
`integrations/pi/dependency.test.mjs` (internal-tool).

### LINKED ISSUE #1249: fix(pi): package-smoke run-resume can refuse 'Previous Pi may still be alive' right after run-stop
# Problem

`integrations/pi/package-smoke.mjs` (via `integrations/pi/managed-client.mjs`) can intermittently fail
`run-resume` with `Previous Pi may still be alive; recovery refused` immediately after `run-stop`, on an
unchanged tree. Observed by the independent Round 2 review of PR #1248: one local `test:package` run failed,
the rerun passed, and CI's `channel` job is green on that head.

The check is a child-PID liveness race: `run-stop` returns after the runner reports `stopped`, but the owned
Pi child's PID can still read as alive for a moment, so `resumeManaged`'s liveness guard refuses the resume.

# Expected

`package-smoke` should wait for the owned child/runner PIDs to be provably dead after `run-stop` before
calling `run-resume` (or `resumeManaged` should distinguish "stopping/exited but not yet reaped"), so a
correct resume is not refused by a transient liveness read. Keep the refusal when a process is genuinely
still alive — do not weaken the duplicate-launch guard.

# Scope

`integrations/pi/package-smoke.mjs` / `managed-client.mjs` (internal-tool). Pre-existing flake, not caused by
PR #1248.

### LINKED ISSUE #1258: fix(pi): switching an adopted run to another owner re-migrates the stale session subscription
# Problem

Adoption migrates a still-present session-scoped dependency record into the owner-scoped store, but does
not remove or update the session record (`integrations/pi/channel.mjs` migration). If an already-adopted run
later switches from owner A to owner B, `rebindDependencies(B)` finds no B record and **re-migrates the
original session record**, so B starts from the pre-adoption snapshot rather than the current subscribed
state. Found by the independent Round 2 review of PR #1253 as an out-of-scope follow-up.

# Expected

After a successful migration, the session-scoped record should no longer be the migration source for a
different owner (e.g. remove/mark it, or migrate only when it is fresher than the owner record). Keep the
existing-store-only rule; no second subscription store; no data loss for a still-valid subscription.

# Scope

`integrations/pi/channel.mjs` (`rebindDependencies`)/`dependency-observer.mjs` (internal-tool).

### LINKED ISSUE #1260: fix(pi): owner adoption silently falls back to a session-scoped dependency subscription when the Windows private-root ACL step fails
# Problem

After `edda-pi owner adopt`, an owner-bound run can report `owner.status: bound` while its dependency
subscription silently stays **session-scoped** (`dependencyScope: session`), because the rebind failure is
swallowed and the reason is never surfaced.

Observed on the installed entry (release `e50ffc76…`, main `aeddae1`) with the standing assistant run
`e3bbbfb8-90a6-47b0-a525-466f49264a88`:

```
edda-pi status 01a09d02-…          -> owner {"owner":"assistant/edda-project-assistant","status":"bound","replaced":false,"dependencyScope":"session"}
edda-pi run-status e3bbbfb8-…      -> owner assistant/edda-project-assistant | continuity owner-bound | ownerAdoptedAt set
ls <registry>/owner-lifecycle/961dc6f3…(sha256 of the owner ref)   -> did not exist
```

## Mechanism (evidenced)

- `channel.adoptOwner()` (installed `channel.mjs`) binds the mailbox, then calls `rebindDependencies(ref)`.
- `rebindDependencies`'s **first** statement is `ownerSubscriptionDir(root, ref)`, which does
  `privateRoot(root)` → on Windows `runPrivateDirectoryAcl()` (the #1226 bounded retry), which throws
  `PrivateRootError` (`private_root_failed`) when its retries are exhausted.
- Its `catch` returns `{ scope: 'session', error: error.message }` to `adoptOwner`, which returns it to the
  caller — and `extension.mjs` does `await channel.adoptOwner(owner)` **ignoring the result**.
- Because the owner-scoped dir was never created, the try must have thrown at that first statement; a direct
  call to `ownerSubscriptionDir` from an ordinary shell process succeeds and creates the dir, so the failure is
  specific to the loaded Pi child (the ACL step's exhaustion path), not the helper.

Net effect: `dependencyScope: owner` — the outcome #1253 exists to provide, and the stated precondition for
the next batch — is not reached, and nothing tells the operator why.

# Outcome

1. Do not swallow it: surface the rebind error in the channel snapshot (e.g. alongside `ownerError`) and/or
   return it in `run-status`, so `dependencyScope: session` on a bound run is visibly a failure.
2. Make the scope decision robust to a transient ACL failure: retry the rebind on a later adopt/turn (the
   mailbox already has that "retried on a later adopt" property, keyed on the dependency identity) rather than
   leaving the run permanently on the session scope.
3. Add a regression test that drives `rebindDependencies` with a failing `privateRoot`/ACL and asserts the
   failure is visible and retried.

Non-goals: no new mailbox, no scheduler, no scope change to the owner lifecycle.

# Evidence

- Commands and values above (read-only), plus the installed sources: `channel.mjs` `adoptOwner`/
  `rebindDependencies`, `dependency-observer.mjs` `ownerSubscriptionDir`, `store.mjs` `runPrivateDirectoryAcl`
  (`PrivateRootError`, `private_root_failed`), `extension.mjs` `before_agent_start` adoption.
- Disclosure: while verifying, a direct `ownerSubscriptionDir(...)` probe call created the (empty)
  owner-scoped directory; it contains no `dependencies.json` and changes no behaviour by itself.

fagemx:
Issues: #1258, #1249, #1241, #1260
Closes #1258
Closes #1249
Closes #1241
Closes #1260

## What / why

Cohesive Pi lifecycle reliability bundle (owner-adoption coherence + resume/test hygiene):

- **#1258** — after adoption migrates a session-scoped dependency record, the record is now **removed** once
  the owner record is written, so switching an adopted run from owner A to owner B cannot re-migrate a stale
  pre-adoption snapshot. The existing store only; the owner record is the live copy.
- **#1249** — `resumeManaged` no longer false-refuses `Previous Pi may still be alive` right after
  `run-stop`: it waits, **bounded**, on the observable owned child/Pi PID to be reaped (never a fixed sleep,
  and a genuinely live process still blocks duplicate launch).
- **#1241** — the owner-mailbox and dependency test files clear an ambient `EDDA_RETURN_ROOT` so a managed
  session cannot change where they write; production behaviour unchanged.
- **#1260** — a dependency rebind failure (e.g. the Windows private-root ACL step exhausting its retries) is
  **surfaced** as `snapshot().owner.dependencyError` on a bound run and **retried** on a later adopt (keyed on
  the dependency identity); it is no longer silently left session-scoped.

## Tests

- `dependency.test.mjs`: owner A→B switch does not re-migrate the stale session snapshot; a failed rebind is
  visible (`owner.dependencyError`) then retried and cleared on a later adopt.
- `managed.test.mjs`: `run-resume` waits out a just-stopped owned child and resumes.
- Hermeticity: `EDDA_RETURN_ROOT=<tmp>` full suite → 175/175 (and 175/175 without), `test:package` passed.

## Not done here

No merge or activation. The focused installed real-entry proof (return + dependency continuity and an
`--owner-root` change on the same run/session, plus the resume race) runs after the batch's final activation.


fagemx:
## Code Review: Round 1 — PR #1261 @ 47057282b1ab8d59790f22c0673078d4ddf6289a

- elapsed: 1067746 ms (pi session)
- model_requested: deepseek/deepseek-flash
- model_observed: deepseek-flash (pi session file, `type:"model_change"`, `provider:"deepseek"`)
- reviewer_session: 01a0a35d-5f41-753b-baf6-0dfb2c835919
- spec: review-spec-v1.10
- escalations: none
- shadow: false
- cost: 1067746 ms (pi session) / tokens & tool calls not surfaced by the pi session file

### IN SCOPE

- **#1258** — `channel.mjs` `rebindDependencies` now deletes the migrated session snapshot
  (`rmSync(join(store.dir,'dependencies.json'), { force: true })`, `channel.mjs:139`) *after* the owner
  record is written (`channel.mjs:134-138`), so switching an adopted run A→B cannot re-migrate a stale
  pre-adoption session record. Existing store only; no second store.
- **#1249** — `managed-client.mjs` adds `previousProcessLingering(pid, 5000)` (`managed-client.mjs:13-17`),
  a bounded, observable PID wait (poll `alive(pid)` every 50 ms — never a fixed sleep), used at both
  `resumeManaged` liveness guards: the runner guard (`:231`, `owner?.pid` + `state?.runnerPid`) and the
  owned-child guard (`:242`, `state.childPid` + `piOwner?.pid`). A process still alive at the deadline still
  refuses.
- **#1241** — `owner-mailbox.test.mjs:15` and `dependency.test.mjs:16` `delete process.env.EDDA_RETURN_ROOT`
  at module load, so the ambient value cannot redirect the mailbox fixtures; production
  (`extension.mjs`/`owner-mailbox.mjs`/`managed-runner.mjs`) is untouched.
- **#1260** — a failed owner-scoped rebind is surfaced as `snapshot().owner.dependencyError`
  (`channel.mjs:157`), set in the catch (`:149`) and cleared on a later successful rebind (`:146`); the
  identity-keyed retry (`channel.mjs:160` `dependencyOwnerRef === ref ? … : await rebindDependencies(ref)`)
  lets the next adopt/turn retry, so a bound run cannot silently stay `dependencyScope: session`.
- Direct callers/consumers: `extension.mjs` `before_agent_start` awaits `channel.adoptOwner(owner)` on every
  turn and ignores the return (`extension.mjs:98-99`); `cli.mjs owner adopt` goes through the managed-runner
  `/owner` endpoint and the extension performs the rebind next turn. The new field is read back through the
  channel `/status` snapshot (`cli.mjs:214`) and nested in `run-status` (`managed-runner.mjs snapshot()`
  `pi: piState`).
- Acceptance ceiling: none of #1258/#1249/#1241/#1260 carries a `## doneWhen` section; their `# Expected` /
  `# Outcome` text is the ceiling (matched to code+tests in C1).
- Security/data-loss: the owner record write is atomic (`store.mjs writeJson` = tmp + `fsync` + bounded
  rename) and precedes the delete, so a crash cannot leave the subscription only-half-migrated; the
  snapshot is bearer-authenticated (`channel.mjs` server auth); `dependencyError` carries only a bounded
  error message, no record bytes.
- Current-base integration: merge-base with `main` is `aeddae1`; `gh pr view 1261` reports `MERGEABLE`
  / `CLEAN`.

### Rules
|---|---|---|---|---|
| U1 | any | P0 | PASS | `gh pr diff 1261 --name-only` = `integrations/pi/{channel,managed-client,managed.test,dependency.test,owner-mailbox.test}.mjs` — the union of the four issues' scopes (`#1241` names `owner-mailbox.test.mjs`/`dependency.test.mjs`) |
| U2 | any | P1 | PASS | body `Closes #1258/#1249/#1241/#1260`; each issue's Expected/Outcome is delivered on the code+tests mapped in C1 |
| U3 | any | P2 | PASS | `gh pr view 1261 --json body \| awk 'tolower($0)~/^issues?:/'` → `Issues: #1258, #1249, #1241, #1260` |
| U4 | any | P1 | PASS | sole commit `fix(pi): reliable owner-adoption dependency rebind and resume liveness` matches `<type>(<scope>): <desc>` |
| U6 | any | P0 | PASS | READ `CI run 34929691855 @ 4705728` (success, headSha matches): `CI Gate` success, `Format` success (covers `cargo fmt` + markdown lint); `Clippy (*)`/`Test (*)` `skipping` (path filter) → those declared gates unverified/advisory. `Pi session channel run 34929691960 @ 4705728` success → channel ubuntu/windows/macos + `test:package`. No deterministically red receipt |
| U7 | any | P1 | PASS | single round; all findings batched here; 0 P0/P1 |
| U8 | any | P1 | N.A.(no `Decision:` line in the PR body) |
| C1 | code-plain | P0 | PASS | #1258 → `channel.mjs:139` + `dependency.test.mjs:440`; #1249 → `managed-client.mjs:13,231,242` + `managed.test.mjs:211`; #1241 → `owner-mailbox.test.mjs:15` + `dependency.test.mjs:16` (mutant-proven below); #1260 → `channel.mjs:146,149,157` + `dependency.test.mjs:459`. All matched, not just claimed |
| C2 | code-plain | P0 | PASS | `fix:` commit with regression tests added (`dependency.test.mjs:440`, `managed.test.mjs:211`) + new assertions at `dependency.test.mjs:478-481,489`; each fails without its change (RAN: 4 mutants) |
| C3 | code-plain | P1 | N.A.(JS-only diff, no `crates/**`; enumerator prints no added `.unwrap(`/`.expect(`) |
| C4 | code-plain | P1 | N.A.(no Rust; enumerator for `unsafe{`/`#[allow(clippy::all)]`/`#![allow` prints nothing) |
| C5 | code-plain | P1 | N.A.(selector prints no crates; no `crates/` path changed) |
| R1 | code-risk | P0 | PASS | enumerator prints nothing: no `rm -rf`/`git clean`/`reset --hard`/`git rm`/`--delete-branch`/`--force`/`Remove-Item` |
| R2 | code-risk | P0/P1 | N.A.(no `*.sh`/`*.bash`/`*.ps1`/`.github` path changed; enumerator prints nothing) |
| R3 | code-risk | P0 | N.A.(`gh pr diff 1261 --name-only \| grep -E '\.(sh\|bash)$'` → empty; no changed shell script) |
| R4 | code-risk | P1 | PASS | no new temp dir/lock/script: `resumeManaged`'s `resume.lock` (`managed-client.mjs:226`) is still released in `finally` (`:252`); the bounded wait only extends the hold; `rmSync{force:true}` cannot throw on ENOENT; no `set -u`/trap surface |
| R5 | code-risk | P1 | PASS | owner ref validated by `ownerMailbox`/`ownerSubscriptionDir` and used only as `sha256(ownerRef)`; the new wait rejects non-integer/non-positive PIDs (`:14`); no shell interpolation, no filename encoding, no caller SHA |
| §5.5 | any | — | PASS | wiring table below; `sh scripts/wiring-scan.sh origin/main 4705728` → “no new pub surfaces”, “no swallow patterns on added lines” |

### Wiring
|---|---|---|---|---|
| `snapshot().owner.dependencyError` (`channel.mjs:157`) | `string` set in the `rebindDependencies` catch (`channel.mjs:149`), cleared on success (`channel.mjs:146`); only when the mailbox exists (bound run) | channel `/status` snapshot → `edda-pi status` prints it (`cli.mjs:214`); nested as `pi.owner.dependencyError` in `run-status` (`managed-runner.mjs` `snapshot()`); test `dependency.test.mjs:478` | not swallowed — present exactly while the owner rebind has failed; cleared by the retry that succeeds (asserted both ways) | `adoptOwner` → `rebindDependencies` catch → `snapshot()` → `status` JSON read-back |
| session record removal `<session>/dependencies.json` (`channel.mjs:139`) | `rmSync(..., {force:true})` after the atomic owner `writeJson` | the migration source itself on a later owner switch; `readJson(join(store.dir,'dependencies.json'))` at `channel.mjs:132` | `force:true` swallows ENOENT; any other throw aborts into the catch and sets `dependencyError` — residual, see P2 | `rebindDependencies` migration → owner record write → session record delete |
| `previousProcessLingering` bounded wait (`managed-client.mjs:13`) | module-private `async (pid, timeoutMs=5000)`; polls `alive(pid)` every 50 ms, returns `alive(pid)` at the deadline | the two `resumeManaged` guards (`managed-client.mjs:231,242`) | not swallowed: still throws “may still be alive” when the deadline expires with the PID alive | `resumeManaged` → guard → throw or continue to `startRunner` |

`scripts/wiring-scan.sh` (Rust-oriented) reports no new pub surfaces; the JS surfaces above are the complete
list. No CLI flag, config key, or event payload added.

### Findings

- [P2] If `rmSync` throws (e.g. Windows `EPERM`/`EBUSY` while a reader holds the session
  `dependencies.json`), the catch runs *after* the owner record was already written: it returns
  `{scope:'session'}`, sets `dependencyError`, and leaves the session record in place. Re-adopting the same
  owner then skips migration (`channel.mjs:131` finds the owner record) and never retries the delete, so a
  later switch to another owner can still re-migrate that stale snapshot — the exact #1258 defect, on a
  narrow failure path. A best-effort delete that does not abort the rebind (or a “migrated” marker on the
  session record) would close it. evidence: `channel.mjs:131-150`.
- [P2] `stopManaged` keeps its own inline bounded PID wait (`managed-client.mjs:90-91`) that is identical to
  the new `previousProcessLingering`; sharing the helper would prevent future drift between the two
  liveness waits. evidence: `managed-client.mjs:90` vs `:13-17`.
- [P2] The new managed test proves a just-stopped child no longer false-refuses, but does not pin the
  complementary direction (a genuinely live prior process still refuses) nor that the wait is *bounded*
  rather than a fixed sleep. The refusal direction is self-evident from `previousProcessLingering` returning
  `alive(pid)` at the deadline, so this is coverage, not a defect. evidence: `managed.test.mjs:211-223`.

### FOLLOW-UP ISSUE

- Pre-existing, out-of-scope full-suite flake (not introduced or exposed by this diff): on the first
  `EDDA_RETURN_ROOT=<temp> npm --prefix integrations/pi test` run in this session,
  `adoption.test.mjs:112` (`changing managed task needs explicit expected revision…`) failed with
  `'offline' !== 'needs_expected_revision'`. It passed 10/10 in isolation both with and without ambient
  `EDDA_RETURN_ROOT`, and two subsequent ambient full-suite runs were 175/175 — so it is load-sensitive
  (a 2500 ms `requestSession` in `adoptSession`), not EDDA_RETURN_ROOT-induced and not this PR's code.
  evidence: the failing run's output (`adoption.test.mjs:117`, `actual: 'offline'`); `adoption.mjs:62,95`.

### RAN vs READ

RAN (this session, Windows 11 Git Bash; `git status --porcelain` empty before and after — read-only):

- `git rev-parse HEAD` → `47057282b1ab8d59790f22c0673078d4ddf6289a` = the frozen head (PR head unchanged
  at posting).
- `env -u EDDA_RETURN_ROOT npm --prefix integrations/pi test` → 175 tests, **175 pass, 0 fail**.
- `EDDA_RETURN_ROOT=<temp> npm --prefix integrations/pi test` → **175/175** (two consecutive clean runs).
  `EDDA_RETURN_ROOT=<temp> node --test owner-mailbox.test.mjs` → **11/11**; `node --test dependency.test.mjs`
  → **22/22**. (First load-heavy full run was 174/175 — the out-of-scope `adoption.test.mjs` flake above.)
- `env -u EDDA_RETURN_ROOT npm --prefix integrations/pi run test:package` → `{"passed": true, …}`,
  `exit=0`.
- Fail-without-change (pristine `integrations/pi` copied to temp trees outside the repo, one mutation each):
  - remove `channel.mjs:139` `rmSync` → `dependency.test.mjs:452` fails (`expected null`, got the leftover
    session record);
  - remove `...(dependencyError ? { dependencyError } : {})` → `dependency.test.mjs:478` fails
    (`'undefined' !== 'string'`);
  - revert `managed-client.mjs:242` child guard to `alive(...)` → `managed.test.mjs` fails with
    `Previous Pi may still be alive; recovery refused`;
  - remove the ambient `delete` from `owner-mailbox.test.mjs` + ambient `EDDA_RETURN_ROOT` → **7 pass /
    4 fail** (`{status:'replaced'}` vs `{status:'bound'}`), while the fixed file is 11/11.
- `sh scripts/lint-markdown-content.sh` → `exit=0`.
- `sh scripts/wiring-scan.sh origin/main 47057282b1ab8d59790f22c0673078d4ddf6289a` → “no new pub surfaces”,
  “no swallow patterns on added lines”.
- Confirmatory: `edda --version` → `edda 0.6.2 (aeddae1aa702 2026-09-15)`; base `Cargo.toml`
  `version = "0.6.2"`; `git merge-base origin/main 4705728` = `aeddae1`;
  `node scripts/pi-session-elapsed.mjs "$PI_SESSION_FILE"` → `{"elapsed_ms":1067746,"elapsed_measured":true}`.

READ: L1 receipt **CI run 34929691855 @ `47057282b1ab8d59790f22c0673078d4ddf6289a`** (success; `headSha`
matches the frozen head). Mapped per declared gate: `cargo fmt --all --check` ← **Format** success;
`sh scripts/lint-markdown-content.sh` ← **Format** success; `cargo clippy --workspace --all-targets -- -D
warnings` and `cargo test --workspace` ← `Clippy (${{ matrix.os }})` / `Test (${{ matrix.os }})` **skipping**
(`.github/workflows/ci.yml:33-70` detect job: the whitelist does not include `integrations/pi/**`, so
`code=false`) → those two declared gates are **unverified/advisory**, never a green per-gate claim.
**CI Gate** success (set-level only). **Pi session channel run 34929691960 @ `4705728`** success — `channel
(ubuntu-latest|windows-latest|macos-latest)` each ran `npm --prefix integrations/pi test` +
`test:package`, so the diff's real gate is covered on all three OSes, including Windows. **Agent manager
run 34929691927** success. No red receipt; the local re-runs are the brief-directed `EDDA_RETURN_ROOT`
hermeticity confirmation plus the fail-without-change mutants that CI cannot perform.

### Verdict

LGTM (P0=0, P1=0) — the four fixes are correct and each is pinned by a test that fails without its change;
the migration orders the atomic owner write before the session delete, the resume wait is bounded and still
refuses a genuinely live process, the ambient root cannot redirect the mailbox tests, and the rebind failure
is visible and retried. Three non-blocking P2 quality notes.

fagemx:
## Merge window check — PR #1261 @ 47057282b1ab8d59790f22c0673078d4ddf6289a

- Round 1 `LGTM (P0=0, P1=0)`, `escalations: none`, pinned to head `47057282b1ab8d59790f22c0673078d4ddf6289a`;
  each of the four fixes is pinned by a test that fails without its change.
- Three non-blocking P2s (rmSync-EPERM leaves the session snapshot with no delete retry; `stopManaged`'s
  inline wait duplicates `previousProcessLingering`; the live-process refusal / bounded-wait coverage) plus a
  pre-existing `adoption.test.mjs` load-sensitive flake are routed as follow-ups per the P2-never-blocks rule.
- `CI Gate` green on `4705728…`; `channel` ×3 and `manager` ×3 all green.
- Empty window: `origin/main` = `aeddae1` = the PR base.
- `edda review merge --pr 1261` → `accepted: true`.

